Issue #29, opened 2016-07-31
Offer to insert specific configuration variables


kaiyou:
Daemons embedded in Freeposte have many configuration options, most of which are not tunable through the usual Freeposte interface. It would be interesting to be able to tune them using specific ``freeposte.env`` entries or configuration files.

kaiyou:
As suggested by @jkarlobs: 

> Yes, the ability to set specific postfix options from the main freeposte.env file I think that is a great idea, even including a config folder with dovecot config files, postfix config files (main.cf and eventual lookup tables file for some entries) , etc. and when recreating the containers, this entries of these config files override the entries of main configuration. If this files are empty then we do nothing.

kaiyou:
PR !30 implemented this for Postix and provided a clear framework for other implementations. I will try and implement this for Dovecot, maybe using inclusions because the Dovecot configuration is not as linear as Postfix (cannot read line by line).

PR #31, opened 2016-08-01
Added config folder for Postfix

jkarlosb:
Feature #29

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ __pycache__
 /admin/include
 pip-selfcheck.json
 /data
+/docker-compose.mac.yml
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -47,7 +47,7 @@ services:
     volumes:
       - /freeposte/freeposte:/data
       - /freeposte/certs:/certs
-      - /freeposte/config/postfix:/config
+      - /freeposte/overrides:/overrides
 
   milter:
     build: rmilter
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -47,6 +47,7 @@ services:
     volumes:
       - /freeposte/freeposte:/data
       - /freeposte/certs:/certs
+      - /freeposte/config/postfix:/config
 
   milter:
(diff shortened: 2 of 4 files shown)


jkarlosb:
First, I created a volume with postfix config folder. Second, Into postifix container start.sh I added a script to verify optional main.cf and if it exist, override the entries through postconf e flag. Next, the script checks if some .map file exist (e.g. transport.map, sasl_passwd.map, etc.), it copies to /etc/postfix folder, executes postmap and configures its perms. Example config files:

/freeposte/config/postfix/main.cf:
```
relayhost = [email-smtp.eu-west-1.amazonaws.com]:25
smtp_sasl_auth_enable = yes
smtp_sasl_security_options = noanonymous
smtp_sasl_password_maps = hash:/etc/postfix/sasl_passwd.map
smtp_use_tls = yes
smtp_tls_security_level = encrypt
smtp_tls_note_starttls_offer = yes
smtp_tls_CAfile = /etc/ssl/certs/ca-certificates.crt
```

/freeposte/config/postfix/sasl_passwd.map:
```
[email-smtp.eu-west-1.amazonaws.com]:25 USER:PASSWORD
```

kaiyou:
Sounds very good.

Because we have very few configuration files for each service, I am thinking about gathering them all in something like ``/freeposte/overrides`` (or anything that clearly states that these are overrides, not the actual configuration files) and overriding on a filename basis with proper documentation to avoid having too many directories and different mount points.

The documentation could state something like "create a main.cf to override the postfix configuration", and maybe use some more meaningful names like ``postfix.conf``, ``dovecot.conf``, etc.

The only problem I see would be confidentiality between containers for password maps for instance. Does not sound like too much of a risk. What do you think?

jkarlosb:
Ok, It seems perfect for me, to avoid having too many mount points. About the confidentiality between containers I think this isn't a risk for now.

kaiyou:
Sounds perfect. Let's go for master, I will test it tonight. Thanks!

Issue #38, opened 2016-08-15
Catch-All Alias


XYunknown:
How to create a catchall alias from the admin interface? '*' doesn't work, and neither does an empty string. Is this even a feature here? If not, why, and if it is, how do I set it up?

kaiyou:
Catchall aliases or wildcard aliases are not supported yet.

Why? Mostly because we designed Freeposte.io with the strict minimum set of features in order to release early and we have been implementing missing features since.

This should be simple enough to implement. One way (option A) to go would be to leverage the power of SQL LIKE statements (https://www.sqlite.org/lang_expr.html#like). One could create an alias named ``%`` for catching all emails or ``admin-%`` for catching all emails sent to addresses that start with ``admin-``.

The only issue is that ``%`` is a valid character in an email address. Also, ``_`` (underscore) is wildly used in email addresses and would be interpreted as *any single character*. 

With the proper documentation, it would be possible to replace those two with ``*`` and ``?`` respectively, which are almost unused in local parts, would match your request exactly (an alias named ``*`` would indeed catch all emails) and would not bother users too much. We would then have to implement :

- an SQLAlchemy type adaptor for wildcard-enabled local parts (escape ``%`` and ``_``, replace ``*`` and ``?`` back and forth) ;
- ``LIKE`` statements in the Postfix configuration to handle aliases properly ;
- the proper Javascript snippet to warn users who are using ``*`` and ``?`` in their statement that they are enabling the wildcard feature, maybe with a link to documentation.

Another way (option B) to go is to offer simple catchall aliases with a separate implementation and stored directly in the domain table. I personnally do not like this solution.

The last way (option C) that I can imagine would be to offer a specific name for catchall aliases and simply match it in Postfix (like a single ``*`` for catching all emails). But as this would still require both documentation and modificiations in the Postfix SQL queries, I prefer to implement the full power version once and for all.

What do you think? I do not really like option B, but options A and C both sound fine. You are welcome to pull request if you have some time to implement one of them (or any other way you think of that seems fit), otherwise I will proabably implement option A by the end of the week (I also find myself in the need for wildcard aliases!)

kaiyou:
Also this is a standard enough feature to include in milestone 1.2 I believe, released as stable in a month or so.

XYunknown:
[...]

PR #41, opened 2016-08-15
Adding SQL Like format for aliases

XYunknown:
Resolving https://github.com/kaiyou/freeposte.io/issues/38

Needs review by @kaiyou


kaiyou:
I will add the Alembic migration script and merge this, thank you!

kaiyou:
After some testing, the union does not enforce ordering. this is not critical for now but I will mention it and leave the issue open until resolved.

XYunknown:
I already know a simply and straightforward fix, will push it tonight. Documentation on this feature is however still missing, so we should keep the issue open

kaiyou:
Have a look at the issue #38, I pushed two related commits.

Issue #40, opened 2016-08-15
Security concepts


XYunknown:
I strongly dislike the security measures/ practices which are implemented. In a default Email Server setup, the admin interface is reachable from the web, which wasn't audited for security and is prone to bugs. 

Admittedly, I am a proponent of security over usability when it comes to servers. Although freeposte is supposed to be a all-inclusive and easy-setup emailserver stack, The default setup should enforce stronger security requirements:

1. The admin interface should not be exposed to the web at all; Instead, it should bind to the localhost and be accessed by forwarding the server port to the client machine over SSH (changes/setup do not occur frequently). This ensures an encrypted channel and prevents any exploitability in the web interface.

2. Roundcube should be disabled by default. Any Webinterface introduces security risks, especially if there are no auto-updates and clients should prefer to connect using IMAPS.

3. Enforcing usage of certificates. Obtaining the necessary certificates using e.g. certbot is free and easy, and should be integrated in the default setup process. For now, disabling any unencrypted protocol by default is a best practice measure to ensure users know what they're doing.

kaiyou:
I see your point and I do agree about the lack of security tests and procedures so far. On the security vs. usuability scale, I wanted Freeposte.io to be simple to set-up and have easy defaults to ensure that setting up a mail server is simple enough for individuals and SME. Security should not be left apart however.

A quick answer to your three concerns:

1. I do not agree, most people who are not aware of the security issues related to TLS/HTTPS will not care about SSH security and are just as prone to interception over SSH as they are over HTTPS (maybe even more, as a lot of effort was put into educating them to pay attention to HTTPS-related issues). Additionally, the admin interface is intended for global admins, per-domain admins and mailbox owners, switching to a private administration interface would require a separate Web interface and increasing the complexity (thus reducing both usuability and maintainability, and possibly harming security in the process).

2. I still do not agree, it is a shame but people are used to having a Webmail available. Even if mail server administrators probably will never use it, Freeposte.io is not just intended for them but also for mailbox owners. Therefore I believe a default setup should include a Webmail. You make a great point about updates however, having an auto-update feature could be interesting if implemented properly.

3. Certbot is being implemented as part of #35 and will be part of the next release. A strong emphasis is put on using proper certificates, and I am wondering if disabling STARTTLS could increase security a bit.

Aditionnally : 

[...]

PR #43, opened 2016-08-19
Update postfix SSL/TLS configuration

d-fens:
Ensure that RC4 and SSLv3 is not used. This is based off mailinabox project settings, while not the most ideal settings this improves the configuration from what it is currently.


kaiyou:
I just opened #45 on the subject and there was some discussion on #40 about the matter. Thank you for the PR.

kaiyou:
This sounds like a reasonable first step towards hardening the TLS configuration, merging.

PR #55, opened 2016-08-29
Refactor the access control code

kaiyou:
The access control code was refactored to allow for easier maintenance and security audits.

- Move all the access control code in the ``access`` module.
- Implement a decorator based access control system.
- Implement a simple permissions audit.


kaiyou:
Only remaining comment: errors 403 and 404 should not be dinstinguishable. This will be addressed in a separate issue.

PR #69, opened 2016-09-24
Default message_size_limit to 50MB

aminb:
Add MESSAGE_SIZE_LIMIT variable in .env to allow setting the message size limit for postfix.

The current (default) message size limit of postfix is `10240000` bytes, or roughly 10MB.

### DIFF
--- a/.env
+++ b/.env
@@ -49,6 +49,10 @@ EXPOSE_ADMIN=no
 # Mail settings
 ###################################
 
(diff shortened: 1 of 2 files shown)


kaiyou:
Perfect and actually a great idea, I was thinking of implementing this in a couple weeks. Thanks!

aminb:
Cheers!

Issue #88, opened 2016-10-25
fetchmail not working (solved)


oschreibke:
Apologies for the formatting - Github's editor seems to do what it wants, not what I want it to.

I was having a problem logging in to my provider's server with a test account with both the 'Stable' and 'Latest' versions. Retrieving mails from the account works with my current but aging setup. 

Response to telnet mail@mysite.ch from the provider's mail system:
> 220 smtp-in-165.livemail.co.uk ESMTP Postfix

I solved the problem by modifying the Poll command in fetchmail.py from

```
    RC_LINE = """

    poll "{host}" proto {protocol} port {port}

        user "{username}" password "{password}"

        smtphost "smtp"

        smtpname "{user_email}"

        {options}

 """
```

to

```
    RC_LINE = """

    poll "{host}" proto {protocol}  port {port}

        user "{username}" password "{password}"

        is "{user_email}"

        smtphost "smtp"

        {options}

        sslproto 'AUTO'

    """
```

Seems connecting to my provider needs the sslproto clause.

Another issue was the logging. Messages with 'No mail...' were being logged as errors.
I changed fetchmail.py
from
```
        try:

            print(fetchmail(fetchmailrc))

            error_message = ""

        except subprocess.CalledProcessError as error:

            error_message = error.output.decode("utf8")

            print(error.output)
```

to

```
        try:

            print(fetchmail(fetchmailrc))

            error_message = ""

        except subprocess.CalledProcessError as error:

            error_message = error.output.decode("utf8")

            # No mail is not an error

            if not (error_message.startswith("fetchmail: No mail")): 

                print(fetchmailrc)

                print(error_message)

```

Finally, I received the following message in the log:
> fetchmail: WARNING: Running as root is discouraged.

This is resolved by adding

>  USER fetchmail

to the Dockerfile

My now working set up is as follows:
Modified fetchmail container:
Dockerfile: (~/fetchmail.Dockerfile)

```
 FROM ab77d96fd9ba 

 MAINTAINER Otto Schreibke <otto@mysite.ch>

 COPY fetchmail.py /fetchmail.py

 USER fetchmail

 CMD ["/fetchmail.py"]
```

fetchmail.py: (~/fetchmail.fetchmail.py)

```
 #!/usr/bin/env python

 import sqlite3

 import time

 import os

 import tempfile

 import shlex

 import subprocess


 FETCHMAIL = """

 fetchmail -N \

     --sslcertck --sslcertpath /etc/ssl/certs \

[...]

PR #89, opened 2016-10-26
Update to poll command; Upgrade logging - issue #88

oschreibke:
Think I got it right.
I'm not new to git, but have only been on github for about 4 hours.

### DIFF
--- a/fetchmail/fetchmail.py
+++ b/fetchmail/fetchmail.py
@@ -14,15 +14,14 @@ fetchmail -N \
     -f {}
 """
 
-
 RC_LINE = """
-poll "{host}" proto {protocol} port {port}
+poll "{host}" proto {protocol}  port {port}
     user "{username}" password "{password}"
+    is "{user_email}"
     smtphost "smtp"
-    smtpname "{user_email}"
     {options}
+    sslproto 'AUTO'
 """
-
 
 def escape_rc_string(arg):
     return arg.replace("\\", "\\\\").replace('"', '\\"')
@@ -45,7 +44,8 @@ def run(connection, cursor):
     for line in cursor.fetchall():
         fetchmailrc = ""
         user_email, protocol, host, port, tls, username, password = line
-        options = "options ssl" if tls else ""
+        options = "options fetchall antispam 501, 504, 550, 553, 554"
+        options += " ssl" if tls else ""
         fetchmailrc += RC_LINE.format(
             user_email=escape_rc_string(user_email),
             protocol=protocol,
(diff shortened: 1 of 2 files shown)


fetchmail/fetchmail.py
```diff
@@ -45,7 +44,8 @@ def run(connection, cursor):
     for line in cursor.fetchall():
         fetchmailrc = ""
         user_email, protocol, host, port, tls, username, password = line
-        options = "options ssl" if tls else ""
+        options = "options fetchall antispam 501, 504, 550, 553, 554"
```
kaiyou: I disagree with fetchall here. I suspect you will get duplicate messages, don't you?

fetchmail/fetchmail.py
```diff
@@ -60,7 +60,10 @@ def run(connection, cursor):
             error_message = ""
         except subprocess.CalledProcessError as error:
             error_message = error.output.decode("utf8")
-            print(error.output)
+            # No mail is not an error
+            if not (error_message.startswith("fetchmail: No mail")): 
+                print(fetchmailrc)
```
kaiyou: This looks like some debugging output.

kaiyou:
Thank you for the pull request, some comments above, perfect otherwise. I will include this in 1.3 as soon as Dovecot 2.2.26 is released.

fetchmail/fetchmail.py
oschreibke: Agree with your concern. fetchall can be dangerous if another application is relying on the messages being there on the remote server.
Being selfish: in my setup fetchmail is the only application pulling messages, and I don't want them to remain on the remote.

fetchmail/fetchmail.py
oschreibke: Yes, remember I was trying to debug the access issues.
There is a major flaw here. in case of an error the password is shown in the log.
Its probably wiser to remove the '''print(fetchmailrc)''' statement.

fetchmail/fetchmail.py
oschreibke: Some more thoughts on fetchall: 

Here's the poll command for my existing setup:
'''
poll mail.solwiz.ch proto imap
   user 'test@solwiz.ch' there with password 'aSecretNeverToBeTold'
   is 'test@solwiz.test' here
   options fetchall antispam 501, 504, 550, 553, 554
   sslproto 'AUTO'
'''

There's an option '''nokeep''', which seems to be the default, which means messages are deleted once fetched. I don't get duplicates with fetchmail, but that's the only process reading the mailbox.

kaiyou: Okay then, I will merge and test probably tonight.

PR #105, opened 2016-11-09
Add german translation

sanduhrs:


kaiyou:
That's an impressive job. Thank you!

PR #119, opened 2016-11-18
Adding fetchmail keep option to .env file

pe2mbs:
Here is the a simple patch to make fetchmail keep the e-mails on the server, its done through the .env file setting FETCHMAIL_KEEP=True. This patch until fetchmail options are supported in the web mail admin panel.


kaiyou:
Thank you for the nice work, that's a first step until these are stored in database.

kaiyou:
Why did you close? Do you plan on opening a separate merge request?

PR #132, opened 2016-12-14
Additional admin manager commands

sanduhrs:
Add some commands to be able to script an import:
* add user
* add alias


kaiyou:
Very nice, thank you!

Issue #167, opened 2017-02-22
TLS / SSL termination in service (container) apart from Nginx redirections logic


tiangolo:
# What

I would like to set up the TLS / SSL termination in a service / container apart from the current Nginx that handles the redirections and `proxy_pass` to the other components of the system.

For this I would keep the redirects logic in the current Nginx container, but I would create an additional Nginx container with the TLS / SSL part and the redirection of HTTP to HTTPS.

This would allow setting up a different TLS / SSL termination system by replacing only that part but keeping the rest of the Nginx logic.

# Why

As HTTPS needs to handle the certificates and establish the connection before giving the hostname requested, having a multi host (multi domain, "virtual hosts") set-up would not be possible with the current configuration, as the Nginx container has to handle the TLS connections, so it has to listen on the public 443 port.

If there was an additional container doing just the TLS / SSL termination, it would be easier to replace only that component by a custom system. 

# Example

For example, using Rancher, it would be possible to configure a load balancer that handles all the HTTPS connections and the certificates and does TLS / SSL termination for Mailu and for (possibly) other services running on the same host.

Also, it could integrate directly with the Let's Encrypt Rancher stack.

# Proposal

Is this something that you would like / accept?

I can make a PR with the changes if that is something that could be "desirable" or acceptable for the project.

kaiyou:
I don't know much about Rancher, but I have been looking for some time at possible solutions to solve that TLS nightmare. Of course, having a single TLS termination is one option that I would be glad someone could work on.

One problem however is terminating STARTTLS for IMAP and SMTP, which sounds quite difficult and would probably cause issues when it comes to only accepting authentication from encrypted connections for instance. One could argue that STARTTLS is terrible and Mailu should drop any support, but I am not ready to get down that road yet.

kaiyou:
I just had a look at rancher. If you can integrate this without dropping support to bare Docker, that would be fine, but as it requires some advanced setup in the first place, I would prefer to avoid making it the default option.

tiangolo:
So, what I would like to do is only separating the Nginx (`http`) service. I would let IMAP and SMTP continue as they are. I would only separate the HTTP internal redirection (from `/` to `/webmail`, `proxy_pass` from `/webmail` to the Roundcube service and `proxy_pass` from `/admin` to the Mailu admin service) and the HTTPS part (only the HTTPS handling, the first section of the Nginx config).

[...]

PR #168, opened 2017-02-23
Fix Nginx entrypoint

tiangolo:
Right now the Nginx entrypoint is checking if the string `ENABLE_CERTBOT` is not null, instead of checking if the environment variable `$ENABLE_CERTBOT` is not null.

This was making Nginx always start using the default config and never using the fallback config, and then it was breaking because it couldn't find the certificate with the Let's Encrypt names (instead of `snakeoil.pem`).

And as the Nginx server was breaking, the Let's Encrypt certificate was not being created.

---

**Note**: this is just to fix the current installation. I will make another PR to handle #167

### DIFF
--- a/nginx/start.sh
+++ b/nginx/start.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-if [[ -z ENABLE_CERTBOT || -f /certs/cert.pem ]]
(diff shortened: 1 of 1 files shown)


tiangolo:
Thanks!

PR #170, opened 2017-02-25
Create and use SSL dhparam file if not mounted, NGINX_SSL_DHPARAM_BITS variable in .env.dist file

andreasfaerber:
Will result in A+ rating on Qualys ssltest (https://www.ssllabs.com/ssltest/)

### DIFF
--- a/.env.dist
+++ b/.env.dist
@@ -70,6 +70,13 @@ RELAYHOST=
 FETCHMAIL_DELAY=600
 
 ###################################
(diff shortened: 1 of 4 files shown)


kaiyou:
Thank you for the contribution. I would like to generalize this to all services in the near future and maybe have the `admin` container generate the dhparams in the same fashin it generates Letsencrypt certificates.

PR #173, opened 2017-03-02
Add alternative frontend nginx-no-https

tiangolo:
As commented in #167, here's the alternative frontend without HTTPS.

If everything is OK, I will submit a PR to the Mailu/Rancher repo updating the template and another PR updating the documentation.

BTW, the `docker-compose.yml` won't work "as is" right now, as it assumes a new Docker image that doesn't exist yet. But to test it is possible to build the image locally tagging it with the corresponding name.

### DIFF
--- a/.env.dist
+++ b/.env.dist
@@ -36,7 +36,7 @@ COMPOSE_PROJECT_NAME=mailu
 # Optional features
 ###################################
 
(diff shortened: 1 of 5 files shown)


kaiyou:
Just finished testing locally, sounds perfect. I will create the image on Docker Hub and start building.

tiangolo:
Thanks! :tada:

PR #174, opened 2017-03-02
Add dulwich for git support

binor:
Radicale supports automatic versioning of calendar event changes if a git repository was initialized in the data directory. For this to work, Radicale needs dulwich as an interface to the git repository.

http://radicale.org/user_documentation/#idgit-support

### DIFF
--- a/radicale/Dockerfile
+++ b/radicale/Dockerfile
@@ -1,7 +1,7 @@
 FROM alpine:edge
 
 RUN echo "@testing http://nl.alpinelinux.org/alpine/edge/testing" >> /etc/apk/repositories \
- && apk add --no-cache radicale@testing
+ && apk add --no-cache radicale@testing py-dulwich@testing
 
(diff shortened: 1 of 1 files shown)


kaiyou:
This is pretty cool. Is installing Dulwith the only rquired step to enable git versionning of radicale repos? Also, what about migration from exisiting repos?

binor:
Yes, it's the only step to get radicale commiting it's changes to a repo. It's pretty straightforward. Radicale uses the name of the changed file as commit message. I got it working by installing git on the Docker host and issued a "cd /mailu/dav/; git init".

Migration of existing repos should be easy. Just copy the .git/ directory, too. Radicale itself doesn't depend on any historical data stored in the repo. It just logs it's actions there for back reference.

kaiyou:
Okay then, the fact that a manual ``git init`` has to be ran in the Radicale directory makes it opt-in which is just fine.

PR #180, opened 2017-03-14
Fix permission in nginx-no-https

tiangolo:
In my original PR in https://github.com/Mailu/Mailu/pull/173 I didn't set correctly the permissions of `start.sh` (it should be executable).

Then when running it in a proper Linux file system it throws an error. This just fixes that.


kaiyou:
I just merged #176 which is a duplicate. Thanks a lot for the original PR!

tiangolo:
Thanks @kaiyou ! :tada:

PR #182, opened 2017-03-27
Add portuguese translation

matiasvictor:
Created a portuguese (Brazil) translation with Poedit for the admin interface.


kaiyou:
Thank you so much. I will see about being more uniform across PO headers, this is just fine for now and the translation is more than welcome!

PR #210, opened 2017-06-21
Re-enable rmilter overrides

### LINKED ISSUE #209: rmilter overrides disabled
2a3f62fc65f36d9edc6f56460e11066d5212515d disabled rmilter overrides with no explanation in the commit message. Is there a reason why these overrides are disabled? Can we fix it to get the overrides enabled again?

Logan2211:
rmilter overrides were disabled without explanation in
2a3f62fc65f36d9edc6f56460e11066d5212515d. This will re-enable the
ability to override rmilter's configuration, which is necessary in
most environments where significant mail volumes occur on a single
inbox, since the default milter configuration only allows 20
mails/second per recipient.

Fixes #209


kaiyou:
Thank you for providing the PR :)

Issue #183, opened 2017-03-27
QUESTION: clamav disable


supersophie:
Hi, 

How can I disable Clamav from the workflow?
Thanks, Sophie

kaiyou:
There is no actual option to disable Clamav properly at the moment. I am not sure what is the ``rmilter`` behaviour if it cannot reach the antivirus. You could try and disable the ``antivirus`` container and check the behaviour.

supersophie:
Better would be to comment out the hook in rmilter  that calls it. Now I know where this is I shall look
--

kaiyou:
This will be possible once we merge #212.

PR #212, opened 2017-06-25
Added parameter for antivirus solution to be more flexible (see #183)

letzkus:
Simple new parameter to use images clamav or none as antivirus solution plus in- or exclusion of clamav configuration in rmilter.

### DIFF
--- a/rmilter/start.sh
+++ b/rmilter/start.sh
@@ -1,5 +1,9 @@
 #!/bin/sh
 
 rm -f /var/run/rsyslogd.pid
+if [ "$ANTIVIRUS" == "clamav" ];
+then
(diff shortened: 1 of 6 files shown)


rmilter/start.sh
```diff
@@ -1,5 +1,9 @@
 #!/bin/sh
 
 rm -f /var/run/rsyslogd.pid
+if [ "$ANTIVIRUS" == "clamav" ];
+then
+	echo "# .try_include /etc/rmilter-clamav.conf" >>  /etc/rmilter.conf
```
kaiyou: Configuration will not load if the include is commented out.

kaiyou:
That is pretty good, I added a small comment, looks good otherwise.

PR #222, opened 2017-07-19
Use exec to start rsyslogd

swoopla:
I use exec to start rsyslog in start.sh for postfix container. The rsyslog process optain PID 1, it's better when you stop the container.


kaiyou:
Thank you very much, I will apply this to other containers as well.

PR #226, opened 2017-07-29
The ACME spec says that http validation should be done via http, not …

jdrowell:
…https.

Plus, it seems that the "if" directive in nginx is quite unpredictable (https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/), so I replaced the redirect with a declarative version. We should consider removing the other "if" directives too.

My understanding is that the ACME validation was happening after the redirect to HTTPS, and thus using the self-signed certificate. This seems to contradict the ACME spec (which states that HTTP should be used, not HTTPS), and also relies on the remote library accepting invalid certificates, which is rarely the case (so might break in the future).

### DIFF
--- a/nginx/nginx.conf.default
+++ b/nginx/nginx.conf.default
@@ -26,8 +26,19 @@ http {
 
     server {
       listen 80;
-      listen 443 ssl;
       listen [::]:80;
+
+      location /.well-known/acme-challenge {
+        proxy_pass http://admin:8081;
+      }
+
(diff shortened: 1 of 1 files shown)


kaiyou:
Thank you for this! I do not have the time to check this on my test server right now, could anyone else confirm that certificate renewal still works after applying the patch?

kaiyou:
Finally had the time to check, thank you!

PR #235, opened 2017-08-24
User import routine (and selector for ecryption hashing algorithm)

droopy4096:
We have a number of users we need to migrate into Mailu setup and since we already have hashes for their passwords we needed some way of adding passwords as hashes vs plaintext like it is done in "user()" command. This adds hashing algorithm selection to old user() as well as new user_import() commands.


kaiyou:
Looks good to me, thank you for the contribution!
Also, an example about syncing configuration would be really welcome in the wiki :)

PR #238, opened 2017-08-26
Admin: Switch to alpine for smaller image size

nalt:
The admin image is very large. With a switch to alpine, its size can be reduced from 783MB to 186MB. The modified Dockerfile also removes build dependencies from the image. I have tested the resulting image - it work.


kaiyou:
Thank you, I was willing to switch recently as well!

PR #241, opened 2017-08-27
Fix #240

### LINKED ISSUE #240: Add customisation for dmarc values
It would be really nice to add custom values for rua field in dmarc entry as well as it would be really interesting to add ruf field.

Xumeiquer:
Adds two variables in .env that define the dmarc rua and ruf fields. Those fields are used in the domain view template.

### DIFF
--- a/.env.dist
+++ b/.env.dist
@@ -29,6 +29,10 @@ HOSTNAME=mail.mailu.io
 # Postmaster local part (will append the main mail domain)
 POSTMASTER=admin
 
+# DMARC rua and ruf email
+DMARC_RUA=admin
+DMARC_RUF=admin
+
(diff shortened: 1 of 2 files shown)


kaiyou:
I would rather implement this as conditional inserts in the DMARC, more like: 

```
<td><pre>_dmarc.{{ domain.name }}. 600 IN TXT "v=DMARC1; p=reject; {%if config["DMARC_RUA"] %}rua=mailto:{{ config["DMARC_RUA"] }}@{{ config["DOMAIN"] }}; {% endif %}{% if config["DMARC_RUC"] %}ruf=mailto:{{ config["DMARC_RUF"] }}@{{ config["DOMAIN"] }}; {% endif }adkim=s; aspf=s"</pre></td>
```

PR #246, opened 2017-09-02
Configurable default password scheme used for passwords

trnila:
Allow default password scheme to be configured from **.env**.


kaiyou:
Looks good to me, merging for further testing :)

PR #252, opened 2017-09-10
Implement alternative domains

kaiyou:
Alternative domains are domain aliases, where a full domain is aliased to another domain.


kaiyou:
I have been testing this for a week, with no apparent issue, merging to master.

Issue #250, opened 2017-09-10
Stop using Certbot manually and rely on a decent frontend implementation


kaiyou:
Multiple frontend exist out there, I am specifically thinking of traefik which works really well with Docker in many environments and handles Letsencrypt very well.

We would probably be a lot better off relying on such an implementation than trying to implement automated certificate generation for multiple services ourselves, which is where most popular reverse proxy implementations fail.

A first hint would be to replace the `nginx` frontend with `traefik` for instance, and enforce the use of a Web frontend when automated certificate generation is enabled. Then, the admin container could use the Docker socket to extract certificates from the `traefik` container.

If we rely on the latest Docker Compose features, we could also stop having these `none` nonsense images and dynamic routing would be enabled by `traefik` directly. This could also allow for multiple Webmails, etc.

kaiyou:
A first implementation is working as part of `feature-traefik`, I am testing on a couple servers. Redirection from `/` is still failing for now.

Also, the certbot related option was replaced with `TLS_FLAVOR`. The overall idea is to deprecate both `nginx` and `nginx-no-https` for the `1.5` release and rely on `traefik` for reverse proxying. A separate `traefik` instance could be used and the main one disabled; then one should take care of exposing the `acme.json` to the admin container (and accept the associated risk).

kaiyou:
Also, DH params are generated automatically by the admin container now, so that all services use properly generated params.

kaiyou:
Currently Dovecot does not support loading Diffie Hellman parameters directly. We will need to upgrade to Dovecot 2.3 before this becomes possible.

kaiyou:
I switched back to using a self-generated DH for Dovecot in the meantime. The redirection from the webroot is still missing. I am thinking about implementing it in the admin container, in order to provide a more complete interface (like configuration tips, links to various webmails and the admin)

kaiyou:
Okay, I am merging this into `master` for some more complete testing. It is currently running fine on test.mailu.io!

PR #254, opened 2017-09-10
Switch to traefik for proxying and querying letsencrypt

kaiyou:
Most of the discussion happens on #250.


pesimon:
Cool, was thinking to try the same. Did you manage to have traefik handle letsencrypt cert generation?

kaiyou:
As mentionned in #250, the current state of the PR works well with letsencrypt, copying the certificate for Postfix and Dovecot to use. It is not completely usuable yet due to a missing setting in the current Dovecot version, we will have to work with upstream to get 2.3 pushed to Alpine.

Issue #220, opened 2017-07-12
Mailman 3 integration?


stevelord:
Has anyone tried setting up a list manager like Mailman?

Mailman 3 apparently uses docker as a base, but it looks like it may have it's own postfix server. Before I experience the horror that is trying this first hand, I was wondering if anyone else has given it (or any other mailing list manager) a go successfully?

I suspect that any mail server that integrates with postfix is going to either conflict with mailu's postfix at some level or will need to carry a map file of some sort across.

Any thoughts/suggestions/experiences? Is it worth integrating a mailing list server into mailu?

kaiyou:
For this to work properly on the same public IP address, you must run both on the same Docker network (or with a Docker network in common), and have only one of the Postfix servers to listen publicly.

The expected behaviour would probably be to have Mailu listen on the public IP address and redirect mails to Mailman's Postfix. Unfortunately, Mailu does not yet have an option to redirect emails for a specific domain to a given MX. One would have to implement this first (should be quite simple though).

Archie1978:
stevelord use a BIND_ADDRESS variable in the .env.
If you are many IP, you must separate with coma.

I have a 20 dockers and it is a mess :)

kaiyou:
I will try and implement a dynamic domain routing table for this to work properly in the future. It will probably land in `1.5`.

kaiyou:
I started implementing this on branch `feature-relays`. Currently only the admin interface part is implemented.

kaiyou:
You can now integrate with Mailman using relays, as they were just merged. Closing this issue!

PR #257, opened 2017-09-11
Implement relaying emails to specific SMTP

kaiyou:
This is meant for multiple purposes : 
- relay domain from a backup server
- allow relay for a specific domain for technical reasons
- relay mails to mailing list servers (see #220)


kaiyou:
This has been tested for some days, with no apparent issue and should not interfere with normal delivery anyway. Merging to master.

PR #264, opened 2017-09-17
RainLoop: Allow two factor auth

agglrx:
This settings enables users to configure two-factor authentication for their account. As force_two_factor_auth is still off, users are not forced to use two-factor.

### DIFF
--- a/rainloop/config.ini
+++ b/rainloop/config.ini
@@ -5,6 +5,7 @@ attachment_size_limit = 25
 
 [security]
 allow_admin_panel = Off
+allow_two_factor_auth = On
 
 [labs]
 allow_gravatar = Off


kaiyou:
This is pretty cool, but I would definitely not advertise it as a security feature, as enabling 2FA on the webmail does nothing to protect the account against attacks over IMAP or SMTP. I will give it a try during the week hopefully.

Thanks for your PR! :)

agglrx:
I agree, it does not protect from attacks over IMAP or SMTP. Why do you mean _advertise as security feature_? RainLoop placed this setting into the security-section, if that's what you mean.

kaiyou:
No, I mean I am okay with enabling this, but I would avoid displaying it as a security feature, maybe even write a disclaimer somewhere explaining that it does not protect against bruteforce as is.

Waester:
@kaiyou Any update regarding this topic? 2FA seems sweet, especially if one only serve webmail to the public.

muhlemmer:
Thanks for submitting your pull request. We want to apologize for not reviewing your submission more promptly. Note that this is a [GitHub saved reply](https://help.github.com/articles/about-saved-replies/) and not all content might apply to you.

I would like to notify you, we had some significant changes on our master branch. If you are still looking to merge this PR, please merge the source branch with the upstream master. This will help us doing a functional review of this PR.

A possible workflow for this would be:
```
git remote add upstream https://github.com/Mailu/Mailu.git
git fetch --all
git checkout <your_branch>
git merge upstream/master
```

Git will tell you if there are any merge conflicts. See also: [how conflicts are presented](https://git-scm.com/docs/git-merge#_how_conflicts_are_presented). If the merge conflicts are extensive, you might want to use [`git mergetool`](https://stackoverflow.com/a/163659/1816774). After resolving any conflicts, you'll have to finish the merge by running:
```
git commit -a
```
Now you should push, to make the new changes available in the PR. We will be notified automatically about the updated PR.
```
git push
```

We will then start the review process. Thanks in advance!

muhlemmer:
No reply from original author. Re-submitted in PR #685 above with the merge conflict resolved. Closing this one.

PR #271, opened 2017-09-22
Added config_update verbosity and expanded YAML format a touch

droopy4096:
I've added verbosity to config_update as sometimes it's useful for debugging in prod. Alias destination now could be either comma-separated string or bona-fide list in YAML, code takes care of backward compatibility.

### DIFF
--- a/admin/manage.py
+++ b/admin/manage.py
@@ -98,7 +98,7 @@ def config_update(verbose=False, delete_objects=False):
         localpart=alias_config['localpart']
         domain_name=alias_config['domain']
         pre_destination=alias_config['destination']
-        if type(pre_destination) == type(""):
+        if type(pre_destination) is str:
             destination = pre_destination.split(',')
         else:
             destination = pre_destination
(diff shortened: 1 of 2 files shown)


kaiyou:
This looks good, though you included some changes to the Dovecot config that I do not believe are intended :)

admin/manage.py
```diff
@@ -90,9 +93,15 @@ def config_update(delete_objects=False):
     aliases=new_config['aliases']
     tracked_aliases=set()
     for alias_config in aliases:
+        if verbose:
+            print(str(alias_config))
         localpart=alias_config['localpart']
         domain_name=alias_config['domain']
-        destination=alias_config['destination']
+        pre_destination=alias_config['destination']
+        if type(pre_destination) == type(""):
```
kaiyou: The usual way to check a type would be `type(variable) is str`

droopy4096:
Dovecot changes are actually cipher changes that our security scanner picked since it considered "out-of-the-box" to be insecure... Do you want me to revert those to original?

kaiyou:
Well, ciphers were chosen after a long discussion about a compromise between security and compatibility with multiple email clients. They shall be reviewed if we move TLS terminations to the front container, which is still an ongoing work.

Anyway, this should not be part of a PR regarding the admin CLI :)

kaiyou:
Very nice addition :) nothing to add except for Dovecot of course.

droopy4096:
dovecot settings retracted

PR #276, opened 2017-09-28
multiple bugfixes for config_sync and full coverage of user/alias/domain attributes added

droopy4096:
I ended up refactoring quite a few items as I needed quick way of changing global admins etc. managed with external config file. This patch fixes typo from previous PR as well (163/180 ). Hope that logic for manager user maintenance is sound in line 183 - not sure if comparison is done in-depth or just pointer comparison... seems to work though.

### DIFF
--- a/admin/manage.py
+++ b/admin/manage.py
@@ -18,6 +18,7 @@ def admin(localpart, domain_name, password):
     db.session.add(user)
     db.session.commit()
 
+
 @manager.command
 def user(localpart, domain_name, password, hash_scheme=app.config['PASSWORD_SCHEME']):
     """ Create a user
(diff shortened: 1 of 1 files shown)


kaiyou:
Now that you are polishing the script, applying pep8 and flak8 would not hurt I believe :) not sure if this it's my original code or yours but indentation is not perfectly consistent, same goes for spaces and variable naming.

Looks goot to me otherwise!

droopy4096:
original code had some indentation inconsistencies I just adhered to existing ones :( will rehash and commit shortly

kaiyou:
Thank you, and sorry for the original inconsistencies!

PR #288, opened 2017-10-25
Update details.html

bcyrill:
- Fix assignment in template
- Use updated environment option


kaiyou:
Thank you for these catches! merging now.

PR #329, opened 2017-11-15
disable ssl_session_tickets

HorayNarea:
see https://github.com/mozilla/server-side-tls/issues/135 and [RFC 5077](https://tools.ietf.org/html/rfc5077)

TL;DR: Mozilla/IETF suggest disabling session-tickets¹ because the key to sign them is only changed when the webserver is restarted and that could lead to effectivly no perfect forward secrecy

¹ currently [Mozilla's recommendation](https://wiki.mozilla.org/Security/Server_Side_TLS#TLS_tickets_.28RFC_5077.29) is: schedule frequent restarts… not sure about that.

(Disclaimer: I'm no crypto-wizzard, my TL;DR might be not very accurate :D)


kaiyou:
I agree that we do not need top performance for reading mail and should care more about security, looks good to me otherwise :) thank you!

PR #338, opened 2017-12-01
This sets the max attachment size in roundcube

gregf:
I used the php.ini from the rainloop folder, so they are both set to
25MB.


kaiyou:
This looks good, except the file name once copied :)

HorayNarea:
Shouldn't the roundcube upload size be linked to the postfix maximum mail size?

kaiyou:
Not necessarily, it should not be any higher, but a lower limit does not bother me.

gregf:
Should be sorted now, sorry for the delay

kaiyou:
Perfect, no worries about the delay and thank you! Feel free to add your name to the AUTHORS file anytime.

Issue #322, opened 2017-11-12
pre-defined dhparam


HorayNarea:
[Mozilla and the IETF recommend](https://wiki.mozilla.org/Security/Server_Side_TLS#Pre-defined_DHE_groups) using pre-defined dhparam because it could be safer and is faster.

I don't know very much about crypto but if Mozilla/IETF say it's ok, I think Mailu can have an Option to choose these pre-defined dhparam instead of generating them on first run?

kaiyou:
You are right, we should simply use the predefined groups recommended by the IETF instead of generated new ones. I would basically remove the generation part.

HorayNarea:
I'm fine with that

gregf:
When you say remove the generation part do you mean to have the user fetch or generate the cert by hand before `docker-compose up -d` or do you mean you are just going to fetch ffdhe4096 for the user?

kaiyou:
We are just going to embed it inside the container.

kaiyou:
This was merged.

PR #339, opened 2017-12-01
Use a predefined dhparam.pem, This fixes issue #322

gregf:
Not positive this is the right solution, or if this even works. I don't currently have a environment setup to test this stuff. So documentation on the best way to do that would be great.


kaiyou:
This looks pretty good except you should not copy the cert in /certs because this will be mounted later.

I would copy the cert to /conf, move it to /etc/nginx in the startup script and adjust the configuration accordingly

kaiyou:
I merged this manually, closing the pull request :)

Issue #284, opened 2017-10-22
Template with Swarm support / environment variables documentation.


ghost:
Currently Mailu Compose template uses env file directive. It is not supportet by Docker Swarm - 
https://github.com/moby/moby/issues/29133 .

It would be nice to have a template with Swarm support. One of the possible ways - to use "environment directive" and supply directives for each service like:
```
  admin:
    image: mailu/admin:1.4
    environment:
      - DOMAIN=foo
      - POSTMASTER=baz
      - SECRET_KEY=bar
      - DEBUG=False
    expose:
      - "8000:80"
--------- etc...
```

And there is no documentation currently on relations between Mailu services and env variables form the env file. So it is extremely hard to rewrite current template to use with k8s or Swarm.

kaiyou:
Indeed, I could start mapping those if you are willing to write a Swarm or k8s template :)

dosimeter:
Yes, I have one for Swarm and Mailu v1.4 (most likely with partially wrong env-var setup, but still working). Will try to add it on Nov 5-7.

kaiyou:
If you write some documentation and i have not yet merged `refactor-repo` in master by then, feel free to pull request against that branch and add a specific compose file in `docs/compose` :)

kaiyou:
Also, in `docs/configuration.md` in that branch, I added some documentation about the main configuration settings.

dosimeter:
@kaiyou sorry for a delay. Here is my example config - https://github.com/Mailu/Mailu/pull/344 . (Has lot of copypasted env variables, would be nice to remove them ...)

kaiyou:
I would say this is outdated

PR #344, opened 2017-12-03
Add a Docker Swarm stack example. #284

dosimeter:


kaiyou:
This looks pretty good, just have a look at the documentation for Docker-Compose about adding proper links to configuration files and I think we can merge.

kaiyou:
Also, does Swarm support variables in Compose files? in that case, it would be interesting to keep version variables, root path and so on.

dosimeter:
I am not sure about vars in the stack file. But I know what there is no .env support (at least yet) - https://github.com/moby/moby/issues/29133#issuecomment-309429376

dosimeter:
And this ^ is why I needed env vars documentation - to at least somehow decrease variable duplication degree in the stack file.

kaiyou:
Okay then. So I would suggest a couple improvements:
- use the same defaults as for the Compose configuration
- use dynamic links in the document and so on
- add a warning in the documentation that this is for version ``1.4`` until you support ``1.5``.

kaiyou:
Also, could you merge against ``master``?

rgarrigue:
And that won't work as is I think

> Ignoring deprecated options:
>
> expose: Exposing ports is unnecessary - services on the same network can access each other's containers on any port.

muhlemmer:
This PR is pretty old, incomplete and the submitter seems to have deleted his Github account (Ghost). Furthermore, the docs seem already to be handled in PR #551.

@Kayiu, I would propose to Close this PR as rejected.

kaiyou:
Agrees since #551 does a pretty good job of this.

PR #353, opened 2017-12-04
Added adress verification before accepting mails for delivery

ripkens:
See https://www.endpoint.com/blog/2015/05/28/postfix-address-verification

Block client until address verifiction is completed and mail will not be rejected by relaying MTA or smarthost.
If verification fails, mail is rejected.
If verification takes too long, mail is temporaryly rejected and sending client will retry later.


kaiyou:
I finally had some time to test this and it is pretty good indeed. Thank you!

Issue #336, opened 2017-11-30
Check incoming mail against Blacklist


morbidick:
first of all mailu looks super nice im looking to use it in a production environment myself. Are you interested in using dnsbl to block known dynamic ips and known hosts sending spam?

I'ld have to look in which lists are free for everyone and post a PR. For example there is one provided by [iX a german computer newspaper](http://www.dnsbl.manitu.net/?language=en).

kaiyou:
It would be interesting to be able to configure blacklists indeed. Multiple options here:
- provide configurable variables
- provide a configuration UI in the admin website
- use static lists
- document how to override the list in the override section

morbidick:
I think as another environment variable would be easiest, ill have a look how to include it in the jinja config template.

PR #354, opened 2017-12-04
implement DNS Blacklists #336

morbidick:
This is a very basic implementation. I tested the template generation but couldn't test it "in the wild" with incoming  mails since i have no public setup yet. I'm not sure if postfix will check against the nginx proxy ip or the real client ip.

If the change itself is good i would add more documentation and will try to find some more public blacklists (most are only for private use).

### DIFF
--- a/core/postfix/conf/main.cf
+++ b/core/postfix/conf/main.cf
@@ -87,6 +87,11 @@ smtpd_recipient_restrictions =
   reject_non_fqdn_sender,
   reject_unknown_sender_domain,
   reject_unknown_recipient_domain,
+  {%- if BLACKLISTS -%}
+  {%- for BLACKLIST in BLACKLISTS.split(",") %}
+  reject_rbl_client {{ BLACKLIST|trim }},
+  {%- endfor -%}
+  {% endif %}
   permit
 
 ###############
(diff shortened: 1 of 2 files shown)


kaiyou:
The PR looks basically good :)

Just some thinking: does this need to be implemented in Postfix or Rspamd? The semantic would be different of course. Implementing it in Postfix means an actual blacklist while implementing it in Rspamd has more of a "spam" flavor.

I think I would feel better with it implemented in Rspamd, but then if the semantic is really that of a blacklist, is there a way to enforce email rejection?

For details: https://rspamd.com/doc/modules/rbl.html

ripkens:
Please make sure, if querying those blacklists, dont use public dns resolvers like google dns.
The lists use a fair use policy and blocks dns queries if queried to often.

Implement a local resolver to avoid query limits like dnsmasq or unbound dns.
At least mention in the wiki, to use the ISP dns or a docker dns server and not google dns.

kaiyou:
Adding something to the documentation is a nice first step. Then, at some point maybe we could embed a dnsmasq for this sort of thing but it sounds overkill at this point.

morbidick:
implementing the same functionality with rspamd is still on my todo list, but havent found the time yet. So i gonna close this PR for now.

PR #358, opened 2017-12-04
Adding options for when using reverse proxy (real ip header / from and mail-letsencrypt)

SunMar:
Added some options to allow for using LetsEncrypt for mail (IMAP/POP3/SMTP), but keep HTTP(S) behind a reverse proxy.

- TLS_FLAVOR: added `mail-letsencrypt` (uses LetsEncrypt, but still allows reverse proxy for HTTPS)
- REAL_IP_HEADER: env to set the `real_ip_header` in nginx when behind a reverse proxy
- REAL_IP_FROM: env to set the `set_real_ip_from` in nginx when behind reverse proxy
- HTTPS_REDIRECT: redirect traffic to HTTPS also for `mail` and `mail-letsencrypt`

### DIFF
--- a/core/nginx/conf/nginx.conf
+++ b/core/nginx/conf/nginx.conf
@@ -20,6 +20,14 @@ http {
     absolute_redirect off;
     resolver {{ RESOLVER }} valid=30s;
 
+    {% if REAL_IP_HEADER %}
+    real_ip_header {{ REAL_IP_HEADER }};
+    {% endif %}
+
+    {% if REAL_IP_FROM %}{% for from_ip in REAL_IP_FROM.split(',') %}
+    set_real_ip_from {{ from_ip }};
+    {% endfor %}{% endif %}
+
     # Header maps
     map $http_x_forwarded_proto $proxy_x_forwarded_proto {
       default $http_x_forwarded_proto;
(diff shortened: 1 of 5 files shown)


kaiyou:
I have multiple comments : 
- I like the real IP headers, and it would probably be perfect if they were commented by default in the .env, plus documented in the docs
- In the end, I agree about the use case for ``mail-letsencrypt``
- But I strongly disagree about ``HTTPS_REDIRECT``

Redirecting to https if you stand behind the proxy should be the responsibility of the frontend proxy and nothing else.

SunMar:
The reason for `HTTPS_REDIRECT` is because you can't configure that on an AWS ELB, the only way to do it (as far as I know) is to do it in the back-end. In that regard AWS ELB's are rather "dumb" :(. That's why I added it, but for me it doesn't matter because I plan to completely disable port 80 anyways.

SunMar:
Ok I removed `HTTPS_REDIRECT`, updated the documentation as I forgot to add `REAL_IP_FROM` and `REAL_IP_HEADER` there and I added some security headers in the Nginx configuration. I however omitted the CSP header because I don't know how that might influence the various applications, maybe it's better if they set it themselves to the correct values. If they for example pull in data from a CDN (jQuery or something similar) it needs to be allowed in the CSP.

kaiyou:
Thank you for updating the PR, merging now :)

PR #373, opened 2018-01-23
Parametrize hosts

mildred:
Allows to use mailu without docker-compose when hostnames are not set up
by docker itself but provided via a separate resolver.

Use case: use mailu using nomad scheduler and consul resolver instead of
docker-compose. Other servers are provided by the DNS resolver that
resolves names like admin.service.consul or webmail.service.consul.
These names needs to be configurable.

This is probaby a duplicate of https://github.com/Mailu/NGINX/pull/2

### DIFF
--- a/core/admin/mailu/internal/nginx.py
+++ b/core/admin/mailu/internal/nginx.py
@@ -1,5 +1,6 @@
-from mailu import db, models
+from mailu import db, models, app
 
+import re
 import socket
 import urllib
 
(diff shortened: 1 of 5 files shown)


curry684:
I closed the other one already and archived the repo to avoid further confusion.

kaiyou:
Thank you @curry684 this is indeed the proper repository for this PR. Ill review and make some comments, but it looks pretty good at first sight and addresses a real use case, thank you!

core/admin/mailu/internal/nginx.py
```diff
@@ -75,12 +76,16 @@ def get_status(protocol, status):
 
 
 def get_server(protocol, authenticated=False):
+    host_imap = os.environ["HOST_IMAP"] if "HOST_IMAP" in os.environ else "imap"
```
kaiyou: In the admin interface, the configuration is declared in ``core/admin/mailu/__init__.py``, is overloaded by environment variables at startup and is then available through ``app.config``. I would prefer if you set a default value in the configuration declaration, that can be overloaded by the environment, then simply use ``app.config["HOST_IMAP"]`` and such where necessary.

kaiyou:
Looks good now, merging :) thank you!

PR #376, opened 2018-01-30
Update configuration.rst

navossoc:
Fix environment variable.


kaiyou:
Thanks a lot for the catch, could you simply change it to ``master`` and I'll cherry pick in ``1.5`` :)

navossoc:
I think this should do the trick.

[]'s

PR #377, opened 2018-01-30
FIX : Error front container

tfournier:
Error : 
- Generating DH parameters, 4096 bit long safe prime, generator 2
- This is going to take a long time

Environnement:
Orchestrator : Rancher 1.6.14
Docker : 17.09.0
Mailu: 1.5.1

Solution: 
Add param : "-dsaparam" in line 8
https://security.stackexchange.com/questions/95178/diffie-hellman-parameters-still-calculating-after-24-hours

### DIFF
--- a/core/nginx/start.py
+++ b/core/nginx/start.py
@@ -6,7 +6,7 @@ import subprocess
 
 # Actual startup script
 if not os.path.exists("/certs/dhparam.pem") and os.environ["TLS_FLAVOR"] != "notls":
-    os.system("openssl dhparam -out /certs/dhparam.pem 4096")
+    os.system("openssl dhparam -dsaparam -out /certs/dhparam.pem 4096")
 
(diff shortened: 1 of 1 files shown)


kaiyou:
I'll review this as a temporary measure waiting for ``1.6`` and static dhparams. Thanks :)

Could you change it to merge into 1.5 instead?

tfournier:
For when is the output of the `1.6` ? :D
Yes of course, it's done for the `1.5`

kaiyou:
I am not a cryptography expert, but the openssl documentation states:

> Beware that with such DSA-style DH parameters, a fresh DH key should be created for each use to avoid small-subgroup attacks that may be possible otherwise. 

Do you have any litterature stating otherwise?

kaiyou:
This has been waiting for too long not being able to tell if it is worth it from a security pov.

PR #380, opened 2018-02-02
Add various environment variables to allow running outside of docker-compose

mildred:
This pull request follows https://github.com/Mailu/Mailu/pull/373

These changes allow to run the containers and parametrize the hostnames and IP addresses of the other containers. This is used in the end to run mailu using the nomad scheduler and consul delivery service.

With this, I could run most of the services without much problems. I could receive and send e-mails. I can follow up this pull request by providing the nomad configuration files too.


mildred:
Unfortunately I didn't plan for that PR to take so long, I'll instead open another PR based on another branch

PR #385, opened 2018-02-14
Adding unbound as dns resolver

obi12341:
Hello,

this should be discussed.
This allows to use unbound as dns-resolver, this is especially useful for blacklist checking, because some are blocking public DNS like 8.8.8.8 etc.
This helped me to increase spam blocking from 30% to 60% because rspamd was complaining about not able to resolve uribl.com checks.

### DIFF
--- a/core/unbound/Dockerfile
+++ b/core/unbound/Dockerfile
@@ -1,0 +1,14 @@
+FROM alpine:edge
+
+RUN apk add --no-cache unbound curl \
+	&& curl -o /etc/unbound/root.hints https://www.internic.net/domain/named.cache \
+	&& chown root:unbound /etc/unbound \
+	&& chmod 775 /etc/unbound \
+	&& apk del --no-cache curl \
+	&& /usr/sbin/unbound-anchor -a /etc/unbound/trusted-key.key | true
+
+COPY unbound.conf /etc/unbound/unbound.conf
+
+EXPOSE 53/udp 53/tcp
+
+CMD /usr/sbin/unbound
--- a/core/unbound/unbound.conf
+++ b/core/unbound/unbound.conf
@@ -1,0 +1,19 @@
+server:
+  verbosity: 1
+  interface: 0.0.0.0
+  interface: ::0
+  logfile: /dev/stdout
+  do-ip4: yes
+  do-ip6: yes
+  do-udp: yes
+  do-tcp: yes
+  do-daemonize: no
+  access-control: 0.0.0.0/0 allow
+  directory: "/etc/unbound"
+  username: unbound
+  auto-trust-anchor-file: trusted-key.key
+  root-hints: "/etc/unbound/root.hints"
+  hide-identity: yes
+  hide-version: yes
+  max-udp-size: 4096
+  msg-buffer-size: 65552
--- a/docs/compose/.env
+++ b/docs/compose/.env
@@ -21,6 +21,9 @@ SECRET_KEY=ChangeMeChangeMe
 BIND_ADDRESS4=127.0.0.1
 BIND_ADDRESS6=::1
 
+# Internal Docker network
+IPV4_NETWORK=172.22.1
+
 # Main mail domain
 DOMAIN=mailu.io
 
--- a/docs/compose/docker-compose.yml
+++ b/docs/compose/docker-compose.yml
@@ -1,4 +1,4 @@
-version: '2'
+version: '2.1'
 
 services:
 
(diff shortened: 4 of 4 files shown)


kaiyou:
After hating the idea of embedding a DNS server in Mailu, I am now convinced this is good. And your work looks really promising, thanks!

The Dockerfile and image build look perfect to me, but I really do not like the way the IP is hacked into the Compose file. Could this be resolved using the network declaration?
I am working on providing a config generator based on jinja templates instead of a static compose file, maybe this could help remove the ugly hacks?

obi12341:
Hey kaiyou,

I don't see a way to not "hack" it like that, because you can not deal with hostnames as dns servers. Of course this will not work, because you don't have a resolver yet.
I don't even think this is ugly. You just specify a subnet for your containers, which you should do most cases, because withouth specifing you will reach the limits of free docker networks quite fast (max. 32 networks).
If there is some better solution out there, this would be great.

core/unbound/start.py
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/python
+import os
+
+os.execv("/usr/sbin/unbound", [''])
```
HorayNarea: Couldn't this just be `CMD /usr/sbin/unbound` in the Dockerfile?

Then you could also remove python from the image

obi12341: yeah possible but maybe we want to change some settings via env vars later, so I included it exactly like in the other images.

obi12341: If I should remove it, let me know, I will change it quickly

HorayNarea: I think it would be better to include it when it's needed… without python the imagesize goes down to 10 MB (or 9 MB if you remove curl after using it :D)

obi12341: done :+1:

kaiyou:
Thanks @HorayNarea for the review. @obi12341 I started seriously working on the configuration generator yesterday, see the ``config`` directory in the master branch for a quick demo. I believe I can include a "use a dedicated (unbound) DNS server for Mailu" option, that will automatically include the server; I guess the local IP address hacking can be accomplished as well this way :)

ofthesun9:
> Thanks @HorayNarea for the review. @obi12341 I started seriously working on the configuration generator yesterday, see the `config` directory in the master branch for a quick demo. I believe I can include a "use a dedicated (unbound) DNS server for Mailu" option, that will automatically include the server; I guess the local IP address hacking can be accomplished as well this way :)

If we only want rspamd to use the (unbound) DNS server (not all the other services), it can be done by adding a file "options.inc" in $ROOT/overrides/rspamd/ containing:
dns {
  nameserver = ["unbound:53:1"];
}

Based on the review/changes proposed by @HorayNarea , I created an image at https://github.com/ofthesun9/unbound-mailu
As an improvement, we could have the "unbound-mailu" service creating/removing directly the options.inc file...

muhlemmer:
Seems interesting. I will have a look at this the coming week.

Anybody knows how this will work when resolving docker service names?

kaiyou:
> Seems interesting. I will have a look at this the coming week.
> 
> Anybody knows how this will work when resolving docker service names?

That would be the main issue I guess. Hopefully unbound itself is able to forward to the Docker host, but that will ruin any effort to be compatible with more complex setups like k8s, or stack. @ofthesun9 suggestion about specifying the server name for specific purposes (like in rspamd) sounds like the most reasonable one.

muhlemmer:
`antispam` also resolves against the docker hostname `front`.

Probably the most reliable way to do this, is running it on a seperate Docker network and let the host's `/etc/resolv.conf` point to it. This way the Docker resolver will first resolve docker names and pass along the lookups it can't resolve.

ofthesun9:
@muhlemmer , Is there any reason why you prefer to change the dns at container level, rather than change it at rspam application level ?
Specifying the dns only for rspam application is only a matter of specifying an option, which is much simpler to solve the op issue.
I am using this on my own setup and have described it at https://github.com/ofthesun9/unbound-mailu

muhlemmer:
@ofthesun9  No specific reasoning, I just noticed this PR became stale after the comment from @kaiyou. I just cleaned up a bit to see where we will go from there.

Basically, when including a DNS server, why not use it globally? Unbound resolves to root servers, it chashed properly. I can see performance benefits for Clam and SMTP as well.

For IMAP, webdav, admin and front I don't think we really need it.

muhlemmer:
Closing. We will integrate this in setup as an optional.

Issue #365, opened 2017-12-17
All attachments are zero sized on forward


ebrahim:
All attachments of an email are zero sized when I try to forward the email in RoundCube webmail. I get this error in /var/www/html/logs/errors on Mailu 1.5:

`PHP Error: abc@def.ghi can't read /tmp/rcmAttmnt0DXphm (not in temp_dir) in /var/www/html/plugins/filesystem_attachments/filesystem_attachments.php on line 216`

kaiyou:
This might be related to the zip extension, will try once that is fixed.

senilio:
Sending a new email with attachments from Roundcube is also broken (1.5.1). Probably related to this issue.

Email gets delivered OK, but attachments aren't sent.

kaiyou:
This  was fixed both on master and 1.5

PR #388, opened 2018-02-20
Fixes the inability to upload attachments from roundcube

Sharsie:
According to #365 this was already fixed in 1.5, but using the 1.5.1 version of roundcube webmail I could not upload the attachments due to permission error on `/var/www/html/temp` directory.

Please note that I have no idea whether `www-data` is the proper owner of this directory, I do not know if there are any other uses for the `temp` dir and if any other users are accessing the directory. Maybe chmod 777 would be more appropriate.

Nevertheless it fixes the problem for me regarding the attachments


kaiyou:
This is indeed the proper fix, thanks for the catch :)

PR #403, opened 2018-03-02
Added webmail-imap dependency in docker-compose.yml.

iBrotNano:
I noticed that the imap container was restarting on docker-compose up first because a python script depends on the webmail container.


kaiyou:
Nice catch, thank you!

Issue #411, opened 2018-03-14
SSL protocol error roundcube/imap


willydee:
## Steps to reproduce
* Set up a recent Debian 9.3 server
* Download `docker-compose.yml` and `.env` for `master`
* Configure `.env`
* Launch the compose project
* Create a user mail account
* Navigate to the webmail URL and try to log in

## Expected result
Login should succeed

## What happens instead
`⚠ Connection to storage server failed.`

## Log entries
```
imap_1       | Mar 14 09:15:02 imap-login: Fatal: Unknown ssl_protocols setting: Unrecognized protocol 'SSLv2'
imap_1       | Mar 14 09:15:02 master: Error: service(imap-login): command startup failed, throttling for 4 secs
```

According to `/webmails/roundcube/config.inc.php`, SSL should not be used at all for access to internal containers.

kaiyou:
Should now be fixed thanks to the fix by @adi90x

willydee:
Just pulled the master images, and problem is gone. Thanks @adi90x :+1:

Gaurav7005:
What kind of changes I need to do to solve this, as I am a beginner so it some complicated for me to understand this.
Please help me.
Thank you in advance.

PR #414, opened 2018-03-19
Add explicit ssl_protocols in conf

adi90x:
Hello, 
Quick pull request to avoid problem. 
Link to this : https://www.mail-archive.com/dovecot@dovecot.org/msg72812.html

### DIFF
--- a/core/dovecot/conf/dovecot.conf
+++ b/core/dovecot/conf/dovecot.conf
@@ -57,6 +57,7 @@ namespace inbox {
 ###############
 auth_mechanisms = plain login
 disable_plaintext_auth = no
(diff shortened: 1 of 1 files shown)


adi90x:
Should resolve #411

MFAshby:
This problem isn't restricted to webmail login, you can't log in to IMAP server using any mail clients either.
I can confirm the change in the above PR fixes the issue. This is caused by an upstream change to openssl, see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=844271

kaiyou:
Thank you for the fix, merging right now and just tested against 1.5, so backporting immediately.

Issue #415, opened 2018-03-19
Unable to send from alternative domains


calebj:
Steps to reproduce:
1. Add alternative-domain to an existing domain (primary-domain)
2. Verify that messages sent to both user@primary-domain and user@alternative-domain go to user's inbox (works)
3. Verify that user can send messages as both user@primary-domain and as alias@primary-domain (works)
4. Attempt to send as user@alternative-domain or alias@alternative-domain

Expected result: message can be sent.
Actual result: message doesn't send, with this error in the log:
```
NOQUEUE: reject: RCPT from unknown[172.19.0.3]: 553 5.7.1 <username@alternative-domain>: Sender address rejected: not owned by user username@primary-domain; from=<username@alternative-domain> to=<username@primary-domain> proto=ESMTP helo=<mail.primary-domain>
```

I modified the query in core/postfix/conf/sqlite-virtual_alias_maps.cf to work in some initial testing:
```sql
  SELECT destination
   FROM
     (SELECT destination, email, wildcard, localpart, localpart||'@'||alternative.name AS alt_email FROM alias LEFT JOIN alternative ON alias.domain_name = alternative.domain_name
      UNION
      SELECT (CASE WHEN forward_enabled=1 THEN (CASE WHEN forward_keep=1 THEN email||',' ELSE '' END)||forward_destination ELSE email END) AS destination, email, 0 as wildcard, localpart, localpart||'@'||alternative.name as alt_email FROM user LEFT JOIN alternative ON user.domain_name = alternative.domain_name
      UNION
      SELECT '@'||domain_name as destination, '@'||name as email, 0 as wildcard, '' as localpart, NULL AS alt_email FROM alternative)
   WHERE
     (
      wildcard = 0
      AND
      (email = '%s' OR alt_email = '%s')
     ) OR (
      wildcard = 1
      AND
      '%s' LIKE email
     )
   ORDER BY
     wildcard ASC,
     length(localpart) DESC
   LIMIT 1;
```
It works by using a left join to populate a fifth alt_email column with each possible instance of username@alternative-domain or alias@alternative-domain. If there's a match, the associated username is returned, just like it is for regular accounts and aliases.

kaiyou:
After some discussion, I agree this is a bug. I will review the PR this week end, as it is pretty complex and I would prefer to avoid any pitfall. Thanks!

PR #416, opened 2018-03-20
Add support for sending from alternative domains

calebj:
JOINs with alternative domains in the query for postfix's `virtual_alias_maps`.

Resolves Mailu/Mailu#415


kaiyou:
Finally had time to review and test this properly, thanks a lot!

Issue #420, opened 2018-03-21
Support multiple database backends, ie. move database stuff to admin


kaiyou:
The admin container should ultimately be responsible for everything related to the database (or, more specifically, the API container once we separate the ui from the API. This will also enable more complex behaviors with dynamic aliases, login maps, quotas, etc.

There are multiple valid ways we can remove a container dependency on the SQLite database, by decreasing priority:
- access an internal API from the admin container (see authentication for instance),
- have the admin container populate a short lived structure in a redis, memcached or any other intermediary structure,
- serve a socket with any specific protocol for loading maps, etc.

In case APIs are written in the admin container, that return sensitive information (e.g. a list of domains, or accounts), these APIs must be authenticated using at least the legitimate container source IP address, so a new authentication wrapper should be written for that purpose.

Here are the tasks at hand:
- [ ] write (if required) a new authentication decorator that checks the source ip among authorized internal containers
- [x] move Postfix alias maps to the admin container
- [x] move Postfix login maps to the admin container
- [x] move Dovecot quotas to the admin container
- [x] move current sive dictionaries used for redirect/vacation to the admin container
- [x] move fatchmail settings to the admin container
- [ ] move redis keys to the something else than files

kaiyou:
This is related to #109, #177, #320, and #412

adi90x:
Do we really need to move every map to the admin container ?
Can't we just add a mysql containers store everything in, use admin to make change to the DB container and get data for maps straight from the DB ?

adi90x:
Another question : wouldn't it be better to merge redis and rspamd in the same image ? As they can't really be separate no ?

fabiorauber:
Obrigado

Em 22 de mar de 2018 5:46 PM, "adi90x" <notifications@github.com> escreveu:

Another question : wouldn't it be better to merge redis and rspamd in the
same image ? As they can't really be separate no ?

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/Mailu/Mailu/issues/420#issuecomment-375451981>, or mute
the thread

kaiyou:
The reasoning behind moving things to the admin container is the following.

We need to support multiple databases for power users that want to split Mailu across multiple servers, use shared storage like NFS, or use external authentication sources like LDAP. In order to accomplish this, we need all Mailu components to stop depending so strongly on SQLite.

[...]

PR #425, opened 2018-03-25
Use MySQL DB

adi90x:
Hello, 

I am not really good in python, but I successfully setup everything to work with a MariaDB container ! 
Everything default to sqlite as it is now , so no incompatibility !
Add a little documentation and may do more if interested !
Last point I will work out is to create a SQL dump from the admin interface ( need to figure out how Flask work ! )

Do not hesitate if you have any problem with this to be merge !

Regards


kaiyou:
I am really not comfortable with merging this in master as I know I might use it as an excuse to not implement anything more flexible.

I might however, agree to maintain it as an officiallish separated branch that we would keep up to date with master. What do you think?

adi90x:
Could make sense ! 
Tbh , this can be use as a base for other DB , I can even do something for Postgres DB if anyone interested ... However, it would be really complicated to implement something incompatible with postfix & dovecot for user request !
Btw , do you have a quick sheets of how to add a new page to the admin interface ? Would love to be able to create a SQL dump from the admin page , and I am struggling to understand how it work ( python if definitely not for me :)

adi90x:
Ok , I have look a little bit more into the postfix/dovecot auth against the admin container.
One possible way could be : 
Postfix auth against Imap server ( Dovecot ) and Dovecot against a LUA script 
[This is part of the idea (in french sorry)](https://wiki.auf.org/wikiteki/Postfix/Authentification)

kaiyou:
Authentication is not the issue really, it is performed by the frontend and shall remain that way (for later TLS based authentication among other reasons).

What's more tricky is maps and dictionaries, as stated in the issue.

kaiyou:
The best way to help with this currently would be to help solving #420, which tackles down the dependency to any database in postfix or dovecot and rely on api calls instead.

Branch feat-abstract-db is doing just that and needs lots of testing. Currently IMAP should be working fine but anyone trying it out and reporting bugs would be more than welcome. I will try and publish the SMTP part next week.

rakshith-ravi:
I can get you a VPS hosted with AWS or DigitalOcean from my company for free, and give you the credentials to it. Would something like that help you?

kaiyou:
Not really, but thanks a lot :) the tedious part is actually running the tests, starting said branch and finding corner cases.

rakshith-ravi:
Alright then. I'll set something up tomorrow (it's late in the night for me here), try sending out some emails, run it across mail-tester.com and let you know how it goes. My only question now is, say something went wrong: what log files should I provide you with in order to give you a good idea of what went wrong?

kaiyou:
For now, sending messages will not work, maybe try accessing mailboxes, save drafts, move them around, play with quota, etc. Logs from the IMAP and front container should be helpful.

kaiyou:
Good news, based on #612 I was just able to run a working Mailu instance backed by MySQL. Once it is merged, I hope I will be able to publish this in a couple days.

PR #429, opened 2018-03-28
Add various environment variables to allow running outside of docker-compose

mildred:
This pull request follows #380

These changes allow to run the containers and parametrize the hostnames and IP addresses of the other containers. This is used in the end to run mailu using the nomad scheduler and consul delivery service.

With this, I could run most of the services without much problems. I could receive and send e-mails. I can follow up this pull request by providing the nomad configuration files too.

### DIFF
--- a/core/dovecot/conf/dovecot.conf
+++ b/core/dovecot/conf/dovecot.conf
@@ -5,7 +5,7 @@ log_path = /dev/stderr
 protocols = imap pop3 lmtp sieve
 postmaster_address = {{ POSTMASTER }}@{{ DOMAIN }}
 hostname = {{ HOSTNAMES.split(",")[0] }}
-submission_host = front
+submission_host = {{ FRONT_ADDRESS }}
 
 service dict {
   unix_listener dict {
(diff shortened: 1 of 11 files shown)


kaiyou:
Sorry about the never ending delays, this looks really good, I will merge it when I find the time to change the rainloop startup script and use Python to generate the configuration instead of bash, for the purpose of consistency.

Thanks!

mildred:
No problem (I have to use a patched version of Mailu in the meantime, but I can wait)

Following that, I'll have nomad job files to publish that allows running a complete mailu setup on a machine running conul + nomad (the hashicorp stack :) ).

I'm running this setup as my mail server now and it works well for the moment. On CoreOS so I get automatic updates of the system. The only thing that doesn't work is the admin interface monitoring of docker containers, but that's expected.

mildred:
Updated pull request to fix the conflict with redis that is now used in dovecot

kaiyou:
After much thinking (and honestly a mix between an overwhelming job and procrastination), I finally had time to test it all. I'd prefer that the rainloop start script is changed to a Python one in the future, instead of using complex bash constructs, but it is good enough at this point to merge.

Thank you so much for the large contribution.

flantel:
> No problem (I have to use a patched version of Mailu in the meantime, but I can wait)
> 
> Following that, I'll have nomad job files to publish that allows running a complete mailu setup on a machine running conul + nomad (the hashicorp stack :) ).

Hi, I know it is a long time, but do you have the nomad files? I want to run this under Nomad./Consul.

Thanks!
 
-Barry

Issue #446, opened 2018-04-12
IDNA support


sanduhrs:
Since a while lots of TLDs support IDNA [1] (non ASCII chracters in domain names).
I tried creating a new domain, containing non-ASCII characters, in this case some german Umlauts (äöü).

**Created domain using Unicode notation:**
- New domain: dömäin.org
- Database: dömäin.org
- Mailfolder: n/a
- Admin UI login: 
  - test@xn--dmin-moa0i.org, Failure
  - test@dömäin.org, Success
- Admin UI display: dömäin.org
- Webmail login: 
  - test@xn--dmin-moa0i.org, Failed
  - test@dömäin.org, Failed
- Webmail UI display: n/a
- Testmail to:
  - test@xn--dmin-moa0i.org, Failed
  - test@dömäin.org, Failed

**Created domain using Punicode [2] notation:**
- New domain: xn--dmin-moa0i.org
- Database: xn--dmin-moa0i.org
- Mailfolder: test@xn--dmin-moa0i.org
- Admin UI login:
  - test@xn--dmin-moa0i.org
- Admin UI display: test@xn--dmin-moa0i.org
- Webmail login:
  - test@xn--dmin-moa0i.org, Success
  - test@dömäin.org, Success
- Webmail UI display: test@dömäin.org
- Testmail to:
  - test@xn--dmin-moa0i.org, Success
  - test@dömäin.org, Success

What I observed is, that the notation of the email address in the email source
code always is the Punicode one.

So I assume, whithout knowing anything about the inner workings of IDNA, that it is merely a admin frontend and admin cli issue and we could probably support this by:
1. Storing the IDNA in Punicode notation in the database
2. Converting the Punicode notated domain name to Unicode notation when displayed

[1] https://en.wikipedia.org/wiki/Internationalized_domain_name
[2] https://en.wikipedia.org/wiki/Punycode

kaiyou:
I definitely agree, it is a matter of displaying and saving data in the proper format.

I guess that a simple salalchemy type that does on the fly conversion would suffice to get immediate and full idn support. This can surely be implemented very quickly.

kaiyou:
This was merged and seems to work pretty good :)

timoschwarzer:
@kaiyou 

I cannot send E-Mails from an account with an IDN domain.

```
[...]

PR #447, opened 2018-04-12
Add a sqlalchemy custom type for unicode to idna conversion of domain names

sanduhrs:
First try at implementing IDNA support at #446 

This is a fast shot to check whether it's the right path, I'm on.
Feedback would be very welcome.

### DIFF
--- a/core/admin/mailu/models.py
+++ b/core/admin/mailu/models.py
@@ -263,6 +263,9 @@ class User(Base, Email):
         default=scheme_dict[app.config['PASSWORD_SCHEME']],
     )
 
+    def is_enabled(self):
+        return self.enabled
+
     def check_password(self, password):
         reference = re.match('({[^}]+})?(.*)', self.password).group(2)
         return User.pw_context.verify(password, reference)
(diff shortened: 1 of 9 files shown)


core/admin/mailu/models.py
```diff
@@ -126,8 +139,8 @@ class Relay(Base):
 
     __tablename__ = "relay"
```
kaiyou: Not sure this is necessarily to be handled as a mail domain, as it is merely  hostname.

core/admin/mailu/models.py
```diff
@@ -38,6 +38,19 @@ def process_result_value(self, value, dialect):
         return filter(bool, value.split(","))
```
kaiyou: We should maybe parametrize the type with a length for the underlying string, so that alembic properly handles the table creation.

sanduhrs:
I currently get a "Internal Server Error" when trying to create a "Auto-forward"

sanduhrs:
Approaching it bit by bit, but might have reached a state where it would be worthwhile reviewing it.

core/admin/mailu/models.py
sanduhrs: I moved the parameters into the class IdnaDomain and IdnaEmail in 7f5bd98a2e5bd94f3ac128a05d838a9b98d5b9d6

core/admin/mailu/models.py
sanduhrs: Probably not, so reseting it in c40e255f3b513fb7846d4a90e5057c69c4b90c72

kaiyou:
This now looks pretty good, merging.

PR #448, opened 2018-04-14
Use POD_ADDRESS_RANGE for Dovecot if it exists

romracer:
This is required to override allow_nets in the Dovecot passdb query.  In a Kubernetes environment where pods are not recreated with the same IP address, restricting to FRONT_ADDRESS and WEBMAIL_ADDRESS only works as long as the front and webmail pod don't die.   Once they die, the query no longer is valid as their IP addresses have changed.

This solves the problem by allowing a POD_ADDRESS_RANGE environment variable in Dovecot which can be defined as a subnet (10.0.0.0/12 for example) that encompasses any possible IP address a pod may receive.  If POD_ADDRESS_RANGE is not defined, previous behavior to use FRONT_ADDRESS and WEBMAIL_ADDRESS continues.

If its helpful for testing, an image is pushed to Docker hub romracer/mailu-dovecot:master with this change.  I am currently using that version of this container in a Kubernetes setup.

### DIFF
--- a/core/dovecot/conf/dovecot-sql.conf.ext
+++ b/core/dovecot/conf/dovecot-sql.conf.ext
@@ -3,7 +3,7 @@ connect = /data/main.db
 
 # Return the user hashed password
 password_query = \
- SELECT NULL as password, 'Y' as nopassword, '{{ FRONT_ADDRESS }}{% if WEBMAIL_ADDRESS %},{{ WEBMAIL_ADDRESS }}{% endif %}' as allow_nets \
+ SELECT NULL as password, 'Y' as nopassword, '{% if POD_ADDRESS_RANGE %}{{ POD_ADDRESS_RANGE }}{% else %}{{ FRONT_ADDRESS }}{% if WEBMAIL_ADDRESS %},{{ WEBMAIL_ADDRESS }}{% endif %}{% endif %}' as allow_nets \
    FROM user \
   WHERE user.email = '%u'
 


romracer:
Is there anything I could do to improve this commit?

I've been running Mailu in Kubernetes with this change.  All of the other containers in the Mailu stack that might restrict to specific IPs can be overridden via settings or use service IPs in Kubernetes.  Dovecot is the only one I couldn't do that with.

And technically, I could, but its ugly and prone to future errors.  Example: provide a dovecot.conf and dovecot-sql.conf.ext in /overrides and use the override dovecot.conf to point to the override dovecot-sql.conf.ext for queries.  But once upstream dovecot-sql.conf.ext changes (ie. the iterate query), this breaks as we have to update our override to match.

kaiyou:
Sorry mostly for the delays, which are unacceptable but due to a pretty tight professional life :)

I think your way of implementing this is probably the best there is at this point. My only fear is obviously that other containers can of course bypass authentication. If this is opt-in only, I guess it is a good enough workaround for power user who understand the consequences.

kaiyou:
Thanks anyway, I will merge it at this point :)

frgomes:
I think that ``core/admin/mailu/configuration.py`` should define the default:

```python
    'SUBNET': '0.0.0.0/0',
    'POD_ADDRESS_RANGE': None
```

so that it does not matter if you are using Swarm or Kubernetes, the default is "it should work without any additional configuration".

PR #452, opened 2018-04-15
Add enabled flag to user model

### LINKED ISSUE #449: Add disable account action
An admin option to disable an email account in addition to deleting it would be nice.

I really do not like deleting accounts, because any deleted account could be recreated for another person with the same localpart, potentially allowing the other person to restore access to web services that have been signed on with the email address by the initial owner.

I suggest to add a status field to the user table with possible values 0 or 1.
And respecting it in the authentication and email delivery procedures.
An action on the user list of the admin interface would be nice, too.

sanduhrs:
Fixes #449

### DIFF
--- a/core/admin/mailu/internal/nginx.py
+++ b/core/admin/mailu/internal/nginx.py
@@ -51,7 +51,7 @@ def handle_authentication(headers):
                     status = False
                 elif protocol == "pop3" and not user.enable_pop:
                     status = False
-        if status:
+        if status and user.enabled:
             return {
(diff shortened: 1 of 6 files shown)


sanduhrs:
@kaiyou is that the right way to implement #449 ?

kaiyou:
It is and the PR looks really good. The only missing bit I can think of is to check the enabled field for /internal/admin.

kaiyou:
Looks perfect, merging now.

Issue #391, opened 2018-02-22
Roundcube Enigma


AdvancedStyle:
How would I go about enabling the "enigma" plugin for roundcube, in order to support PGP keys?

Should be able to just add "enigma" to the plugin list here i think:
https://github.com/Mailu/Mailu/blob/master/webmails/roundcube/config.inc.php#L13

But i'm not very familiar with docker, so not sure how to deploy with the edited config.

kaiyou:
It should be pretty straightforward to enable indeed, just have to test it as I am not sure how it integrates with Roundcube.

AdvancedStyle:
Enigma is working but I'm having one issue where sending a message it does not encrypt using your own pub key (so you can't read mail in the sent folder unless you CC the message to yourself).

The roundcube guys say Engima should already do this, but I'm not sure how to debug the issue seeing as Mailu is dockerize (not familiar with docker myself): https://github.com/roundcube/roundcubemail/issues/6210

Any tips on how to turned on enigma_debug inside the docker container?

PR #453, opened 2018-04-15
[Security] Update Roundcube to 1.3.6

romracer:
Fixes a security issue in roundcube.  May also fix the last comment in #391.


kaiyou:
Thanks, merging right now.

PR #458, opened 2018-04-18
Dovecot: Add SQL iterate_query to config file.

Farthen:
This allows to use doveadm -A to execute maintenance tasks for all users on the server.


kaiyou:
Looks good, merging

Issue #463, opened 2018-04-22
Remove the Service Status page


kaiyou:
Mailu was initially meant for Docker Compose only and expects the Docker socket to be mounted inside the admin container.

Now that the project has evolved to support other environments, and suspecting that the "Service Status" page is not actually useful, I would like to remove it. Any objection?

sanduhrs:
Using it in rancher it did not work anyway 😉

muhlemmer:
+1, it doesn't work in swarm mode.

kaiyou:
This was implemented and merged

PR #465, opened 2018-04-23
Remove services status page

sanduhrs:
As seen on #463


hoellen:
I think we don't need the docker socket in the admin container, if we remove the status page?
I found it in the following files:
- docs/compose/docker-compose.yml
- docs/contributors/environment.rst
- (docs/kubernetes/kubernetes-mailu.yaml)

kaiyou:
Indeed @hoellen I'll change that in a separate commit. Merging this now, thanks a lot @sanduhrs

PR #468, opened 2018-05-01
TLS using configurable filenames

dtwardow:
The current implementation uses fixed filenames for certificates and key pairs for TLS flavors `mail` and `cert`.

My currently used Docker environment provides multiple services on different sub-domains. Therefore, a certificate bot (based on Let's Encrypt) provides different certificates and keys for each service (as required) in **different filenames**. Those certificates are also **updated periodically** in the background prior to their expiration date.

I think most users can work with the default implementation, because I didn't find any information in the repo prior to my request.
Therefore, I keep the default filenames, if the user's configuration doesn't override them explicitly.

### DIFF
--- a/core/nginx/config.py
+++ b/core/nginx/config.py
@@ -22,11 +22,13 @@ if "HOST_ANTISPAM" not in args:
     args["HOST_ANTISPAM"] = "antispam:11334"
 
 # TLS configuration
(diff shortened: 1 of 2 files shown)


kaiyou:
This looks good and did not break anything when testing. Merging =

Issue #474, opened 2018-05-12
Moving mails from Junk to Trash flags them as ham


Farthen:
Using an IMAP client to move mails from the Junk to the trash directory will flag them all as ham. This is a problem as many clients are configured by default to delete Junk mails by moving them to Trash, therefore training all those spam messages as ham.

kaiyou:
This is pretty critical indeed, thanks for reporting.

PR #475, opened 2018-05-12
Roundcube: Add defaults.inc.php and remove mails from Junk immediately

Farthen:
Moving mails from Junk to Trash will mark them all as ham. See issue #474. To mitigate
this problem at least in roundcube, this config option is required.

### DIFF
--- a/webmails/roundcube/Dockerfile
+++ b/webmails/roundcube/Dockerfile
@@ -24,7 +24,7 @@ RUN rm -rf /var/www/html/ \
 
 COPY php.ini /usr/local/etc/php/conf.d/roundcube.ini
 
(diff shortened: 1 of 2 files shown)


kaiyou:
Mmmh, why can't this be configured in ``config.php``?

muhlemmer:
The right fix is already implemented in PR #510, which is merged. Issue #474 is also closed.

@Farthen (or @kaiyou), I my opinion this PR can be closed as well.

kaiyou:
I agree, just waiting for @Farthen to confirm if all aspects of the PR are indeed already handled by #510 :)

kaiyou:
I'll close, @Farthen, feel free to comment and/or reopen in any case.

PR #477, opened 2018-05-22
don't require BootstrapCDN for FontAwesome (GDPR-compliance)

HorayNarea:


kaiyou:
Thanks a lot, looks good to me

PR #493, opened 2018-06-10
Made User and UserSignup validation consistent for the local part of

MFAshby:
…the email address. This prevents SQL errors when you add invalid character (@) to the local part of the email.


kaiyou:
Looks good to me, merging :)

PR #501, opened 2018-06-25
Add posibilty to run webmail on root '/'

hoellen:


kaiyou:
Again this looks very good, thanks!

PR #502, opened 2018-06-25
Use message_size_limit variable from env for webmail client_max_body_size

hoellen:


kaiyou:
I am afraid this fails for some edge cases. Maybe we should add a constant, like 8M plus max message size?

hoellen:
I added the 8M tolerance. Good idea!

kaiyou:
Perfect, merging now!

PR #507, opened 2018-06-27
roundcube: fix host parametrization

mildred:
Roundcube can be parametrized so it can take a different hostname than
'front' or 'imap' to connect to the mail servers through environment
variables. Unfortunately, this was not correct and in PHP a `||`
operator always returns a boolean. It did not work as expected.

Instead use the ternary operator `:?` that works in all cases.


kaiyou:
Looks just fine to me, thank you!

PR #510, opened 2018-06-29
Dont flag spam as ham if moved to trash

hoellen:
This should resolve the issue described in title (issue #474).


kaiyou:
Thanks a lot, merging :)

PR #513, opened 2018-07-02
Return correct status codes from auth rate limiter failure.

mprihoda:
When failing on rate limiter, the nginx reports on SMTP port unhelpful "Internal server error". This patch returns error response that nginx will present to the SMTP client.


kaiyou:
Pretty neat, and working as well. Merging now.

PR #514, opened 2018-07-02
Fixed libpng12-dev dependency, called libpng-dev now.

mprihoda:
Upstream docker base image probably upgraded distro, libpng12-dev does not exist anymore, lipng-dev does.


kaiyou:
Thanks for the fix, just compiled, looks good

PR #524, opened 2018-07-13
Fix typo

jake-walker:
Just spotted a typo in the documentation. I don't think there's much to write here, so sorry if I missed something.


kaiyou:
Perfect, thanks :)

PR #527, opened 2018-07-14
Trying to enable fuzzy hashes for rspamd

ofthesun9:
The configuration files are a copy/paste from rspamd documentation.
I guess that we should have a sieve filter to learn automatically when a user moves a mail in his spam folder...
However learning is Ok from the rspamd admin site.


muhlemmer:
I want to start reviewing this. I'm unfamiliar with the internals of fuzy-hashing. So I read a bit of documentation. But I would like to ask @ofthesun9 for some info to make the test easier:

1. How does `rspamd` gets its "mark as spam" notifications? I mean, someone moves a mail to the spam folder either manually or with a dedicated button in a mail client or webmail. Through IMAP dovecot is informed about the move to the Spam folder. What then? How does `rspamd` get informed?
2. How do I check it works properly? Do you have some sample log output perhaps? Rspamd output can be quite overwhelming sometimes.

Thanks in advance!

hoellen:
For (1) I can help you. `Rspamd` get the spam notifications by sieve scripts ([report-ham.sieve](https://github.com/Mailu/Mailu/blob/master/core/dovecot/conf/report-ham.sieve) and [report-spam.sieve](https://github.com/Mailu/Mailu/blob/master/core/dovecot/conf/report-spam.sieve)) which trigger the bash script [mailtrain](https://github.com/Mailu/Mailu/blob/master/core/dovecot/conf/bin/mailtrain).

muhlemmer:
@hoellen Thanks! Apparently those files where already there. I was looking for something in this PR. :sweat_smile: 

Pulling it into me test server now to see whats up.

kaiyou:
Is there any specific action to be taken for learning fuzzy hashes on a mail, separate from the mailtrain commands?
If not, this looks pretty good on paper, I'll test crash it during the afternoon.

muhlemmer:
> I'll test crash it during the afternoon
You have a good method to crash test a mail server?

kaiyou:
I would say that in lack of automated tests, one starts a server and throws emails at it in various corner cases regarding the currently tested feature.
As I feared, the current PR enables fuzzy_hashes but does not learn any fuzzy hash, so these stay pretty useless as is. I am updating the learning scripts to learn fuzzy hashes automatically. Also, I plan on adding a way to create honeypot address that will autolearn fuzzy hashes and bayes coefficients.

hoellen:
I don't know if this is the right [source](https://rspamd.com/doc/modules/fuzzy_check.html#learning-fuzzy_check). But there are `rspamc` commands to add a mail as fuzzy hash:

`rspamc -w <weight> -f <flag> fuzzy_add ...`
`rspamc -f <flag> fuzzy_del ...`

Maybe the commands have to be added to the mailtrain script.

kaiyou:
Indeed, and that is pretty much the part I added. See https://github.com/Mailu/Mailu/compare/master...kaiyou:ofthesun9-feat-fuzzyhashes?expand=1

@ofthesun9 can you pull from there and update this PR maybe?

kaiyou:
> Is the fuzzy worker not listen on port 11335? Or is `rspamd` doing it automatically right?

It does, but I do prefer to listen explicitely, especially when unsure if the defaut behavior is to listen on a v4, v6 or hybrid socket.

muhlemmer:
@kaiyou, (off topic, sorry) with the term "crash test" I was hoping for a way of creating a big wave of mails. Unfortunately I'm quite familiar with the manual sending of messages to myself, after the last weeks. :stuck_out_tongue_closed_eyes:

Remember that the guys from `opensmptd` had an article on how they where generating thousands of mails an hour for "crash testing". Can't seem to find it anymore though. But I believe they used the community to bomb their test server after sending them a request to do so.

PR #541, opened 2018-07-29
modify www-data user to match UID of docker user

hypnoboutique:
I was having the same issue that is documented here: https://github.com/RainLoop/rainloop-webmail/issues/262


kaiyou:
Well, setting the user id to something different is probably not the most suitable solution. It would probably make much more sense to set the permissions properly at startup, before switching to www-data.

PR #544, opened 2018-07-31
Created new manifests for Kubernetes

hacor:
Hello,

I created new manifests for Kubernetes, because the containers of the `stable` version were quite old and I couldn't get the DKIM signing working.
A few things to note:
- I implemented the new Mailu structure using the `front` container, but this pushed me to use the `latest` version instead of the more stable `1.5` because of all the `HOST_` environment variables are only available on master. (May be an interesting moment to release 1.6 ;-), so we can change all the images to version 1.6 )
- All the `HOST_` environment variables are needed because the NGINX resolver doesn't use a search domain, so it wants the complete DNS name as it is registered in Kubernetes. `smtp.namespace.svc.cluster.local`.
- I chose to use the standardized `NGINX Ingress controller` in front of the `front` container so users can still publish other websites and domains on their cluster. This way the NGINX config file of the front container can be left untouched. I included the manifests.
- I chose to use a `cert-manager` to provide the Lets Encrypt certificates just because I already had a working `cert-manager` deployed and it seemed to integrate nicely with Mailu.
- I chose to use a `ConfigMap` to provide the pods with the necessary environment variables which makes it easy to maintain. The same idea as the `.env` file in docker compose.
- I added a `README.md` into the directory to provide more documentation on how to deploy.

A few extra notes:
- These manifests are tested, but not yet 100%. I created some mailboxes and all seemed to work just fine.
- Apple Mail and iPhone 7 behaved a little bit strange when adding my mailbox. I need to retest this. Thunderbird works (as it always does)

Special thanks goes to the complete Mailu team and especially to @kaiyou, @t3hmrman and @pbinks who already spent a lot of time to create the manifests which guided me to create the new ones.

Best regards!

Hacor


ishitatsuyuki:
Just by the way, have you considered creating a Helm chart which allows easier installation and templating?

hacor:
To be honest, no. I'm really not a Helm expert, I'm more old school in this stuff, but I understand it might be considered.

kaiyou:
This looks good (real good). I am not an expert enough to test this yet and would gladly get some feedback. But hey, there does not seem like there's anything malicious in there, and it is filling a pretty big void, so I am merging now. Thank you so much!

PR #552, opened 2018-08-05
add full-text search support

HorayNarea:


kaiyou:
Sounds nice, simple, I will give it a try regarding performance

Issue #548, opened 2018-08-02
Postfix 3.3.0 failing massive


muhlemmer:
I was playing around a bit to get Swarm to work. This involved removing earlier downloaded images and re-download them (today). I got all kinds of funky behaviour and decided to got back to the "classic" approach (docker-compose). However, after all the cleanup and running `docker-compose up -d` against my original `.env` (which worked yesterday) I still get all sorts of strange behaviour.

After going throug the logs, I found this:
```
smtp_1       | postconf: warning: sqlite:/etc/postfix/sqlite-virtual_alias_maps.cf: unused parameter: dbpath=/data/main.db
smtp_1       | postconf: warning: sqlite:/etc/postfix/sqlite-virtual_alias_maps.cf: unused parameter: query=SELECT destination   FROM     (SELECT destination, email, wildcard, localpart FROM alias      UNION      SELECT (CASE WHEN forward_enabled=1 THEN (CASE WHEN forward_keep=1 THEN email||',' ELSE '' END)||forward_destination ELSE email END) AS destination, email, 0 as wildcard, localpart FROM user      UNION      SELECT '@'||domain_name as destination, '@'||name as email, 0 as wildcard, '' as localpart FROM alternative)   WHERE     (      wildcard = 0      AND      email = '%s'     ) OR (      wildcard = 1      AND      '%s' LIKE email     )   ORDER BY     wildcard ASC,     length(localpart) DESC   LIMIT 1
smtp_1       | postconf: warning: sqlite:/etc/postfix/sqlite-transport.cf: unused parameter: dbpath=/data/main.db
smtp_1       | postconf: warning: sqlite:/etc/postfix/sqlite-transport.cf: unused parameter: query=SELECT 'smtp:['||smtp||']' FROM relay WHERE name='%s'
smtp_1       | postconf: warning: sqlite:/etc/postfix/sqlite-reject-spoofed.cf: unused parameter: dbpath=/data/main.db
smtp_1       | postconf: warning: sqlite:/etc/postfix/sqlite-reject-spoofed.cf: unused parameter: query=SELECT 'REJECT' FROM domain WHERE name='%s'  UNION  SELECT 'REJECT' FROM alternative WHERE name='%s'
smtp_1       | postconf: warning: sqlite:/etc/postfix/sqlite-virtual_mailbox_domains.cf: unused parameter: dbpath=/data/main.db
[...]

PR #557, opened 2018-08-07
Pin Alpine 3.7

muhlemmer:
For issue #548. Note, that for an actual succesfull rebuild, PR #556 needs to be merged as well.


kaiyou:
Thank you, merging now.

PR #559, opened 2018-08-08
Finalized the kubernetes release

hacor:
Hello,

As promised I converted the markdown doc to RST as used in the Mailu repository. As I may suggest, the pull request as provided now is ready to use a new branch like `1.6` or `1.5-k8s` so the users (including me :) ) can download stable Kubernetes images instead of the master. Although I think the branch `1.6` seems more logical otherwise Kubernetes will become a different branch on its own which I think is not necessary. Mailu fits in perfectly.
If done, I will update the manifests again including the correct stable image tags.

Thanks!

Signed-off-by: hacor <hacornelis@gmail.com>


hacor:
@muhlemmer I just updated the deployments of the Postfix server and the Dovecot server to 2 and all of my clients (Thunderbird, Apple Mail, Spark on iPhone) were still able to connect perfectly. I think this is the reasons why this works for me:
It's the NGINX Ingress controller which uses the public network and not the `front` containers, they are using the standard POD IP network, therefor all those containers persist in the `RELAYNETS` and `POD_ADDRESS_RANGE` environment variable CIDR provided in the current configmap

@kaiyou Any updates on the merging of this PR and/or the branch point of view?

Thanks!
Best regards,

Hacor

hacor:
@muhlemmer Uses Mailu `xclient` to be able to login?
because two days ago my Mailu deployment suddenly went failing:
- I'm not able to login anymore (this might be based on the problem you stated, but logging is not very transparent)
- It seems that the `RELAYNETS` set as pod IP net opens up the complete system as a public relay server which was of course unintended. Because the Nginx Ingress controller now forwards all mail traffic to the front container, the front container receives it as being from the pod IP net and therefor opening port 25 to the public.

This moment I'm adapting the front manifest in order to be able to disable the port 80/443 section based on an environment variable `KUBERNETES_INGRESS=true`. This way the Kubernetes Nginx Ingress controller can do what it's good in and I can use the `hostNet` parameters for the front (which will then open all other mail ports to the public). I will configure the correct ingresses for the webmail, admin and internal endpoints.

I will keep you posted!

Best regards

H

Issue #530, opened 2018-07-17
swarm mode (stack deploy) & authentification ?


ofthesun9:
I am trying to deploy mailu (1.5) under swarm mode (docker stack deploy -c ....)
I modified a little bit the .yml file (I am using nfs volume(s)): 
[docker-compose-stack-simple.txt](https://github.com/Mailu/Mailu/files/2200943/docker-compose-stack-simple.txt)
and if I run the docker-compose with this file, everything is ok, but If I run the file with docker stack deploy, I am unable to login (imap, webmail etc)
I get this in the logs:
mailu-swarm_front.1.weuol4erohw8@coreos-01    | 2018/07/17 08:55:55 [info] 8#8: *31 upstream sent invalid response: "NO [AUTHENTICATIONFAILED] Authentication failed." while reading response from upstream, client: 10.255.0.2 using starttls, server: 0.0.0.0:143, login: "xxxxx@vmaltair.ddns.net", upstream: 10.0.1.14:143

That would be really cool to get swarm mode working (I intend to use some raspberry as workers :-))

ofthesun9:
@Wolfboy573 ,  yes the .env file is not supported for substitution *in* the docker-compose.yml itself. That is why I removed the $VERSION and so on in the .yml
But the .env file can be used to pass parameter within the service. I checked with the portainer how my services are deployed and all the environnment file in the .env file are populated.
Basically every services seems to be working (checking the log, they are all running) except the admin or authentification part...
Thanks for pointing #375 ! I will have a look :-)

Wolfboy573:
@ofthesun9  Nice, I wasnt that far yet with the params in service. Just started with docker swarm. Im also trying to get mailu on docker swarm. 
This morning I found that request. 
Ill look a bit deeper in your solution.

ofthesun9:
@Wolfboy573 , thanks to the description in #375, I got it working ! I had to add "endpoint_mode: dnsrr" in each deploy section:
[docker-compose-stack-simple.txt](https://github.com/Mailu/Mailu/files/2206078/docker-compose-stack-simple.txt)
I guess that it has to do with the Virtual IP mechanism (which is by-passed with the "endpoint_mode: dnsrr")
I will check whether all the services need the statement "endpoint_mode: dnsrr" or only some of them.

Next step : try to deploy some service(s) on another node and check if this working well :-)

ofthesun9:
[...]

PR #563, opened 2018-08-09
Fix swarm smtp auth

muhlemmer:
XCLIENT check was done against `FRONT_ADDRESS`, which is resolved in `start.py` from the `front` hostname. In Swarm mode this would resolve to the single VIP of all instances inside the `front` service.

Consequently, instances of `front` would try to authenticate using their individual IP != VIP. This would fail.

This PR allows XCLIENT from anywhere inside the docker network, as defined as [`RELAYNETS`](https://github.com/Mailu/Mailu/blob/f506966abc5ec5452f7db599a11e16d7cc497e80/docs/compose/.env#L68) in `.env`.

Using the correct `docker-compose.yml` settings, authentication is now working in Swarm mode. Issue #530 could be closed.

I would also like to make reference to Kubernets PR #559, which could also make good use of this change. In `configmap.yaml`, the current setting would allow only for a single `font` instance. (Replication: 1)

Tested send/receive in both `docker-compose up` and `docker stack deploy`.

### DIFF
--- a/docs/Dockerfile
+++ b/docs/Dockerfile
@@ -1,0 +1,14 @@
+FROM python:3-alpine
+
+COPY requirements.txt /requirements.txt
+
+RUN pip install -r /requirements.txt \
+ && apk add --no-cache nginx \
+ && mkdir /run/nginx
+
+COPY ./nginx.conf /etc/nginx/conf.d/default.conf
+COPY . /docs
+
+RUN sphinx-build /docs /build
+
+CMD nginx -g "daemon off;"
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -7,7 +7,7 @@ templates_path = ['_templates']
 source_suffix = '.rst'
 master_doc = 'index'
 project = 'Mailu'
-copyright = '2017, Mailu authors'
+copyright = '2018, Mailu authors'
 author = 'Mailu authors'
 version = release = 'latest'
 language = None
@@ -23,7 +23,7 @@ htmlhelp_basename = 'Mailudoc'
 # to template names.
 html_sidebars = {
     '**': [
-        'relations.html',  # needs 'show_related': True theme option to display
+        'relations.html', 
         'searchbox.html',
     ]
 }
@@ -36,24 +36,3 @@ html_context = {
     'github_version': 'master',
     'conf_py_path': '/docs/'
 }
-
-
-# Upload function when the script is called directly
-if __name__ == "__main__":
-    import os, sys, paramiko
-    build_dir, hostname, username, password, dest_dir = sys.argv[1:]
-    transport = paramiko.Transport((hostname, 22))
-    transport.connect(username=username, password=password)
-    sftp = paramiko.SFTPClient.from_transport(transport)
-    os.chdir(build_dir)
-    for dirpath, dirnames, filenames in os.walk("."):
-        remote_path = os.path.join(dest_dir, dirpath)
-        try:
-            sftp.mkdir(remote_path)
-        except:
-            pass
-        for filename in filenames:
-            sftp.put(
-                os.path.join(dirpath, filename),
-                os.path.join(remote_path, filename)
-            )
--- a/docs/contributors/environment.rst
+++ b/docs/contributors/environment.rst
@@ -89,3 +89,20 @@ Any change to the files will automatically restart the Web server and reload the
 
 When using the development environment, a debugging toolbar is displayed on the right side
 of the screen, that you can open to access query details, internal variables, etc.
+
+Documentation
+-------------
+
+Documentation is maintained in the ``docs`` directory and are maintained as `reStructuredText`_ files. It is possible to run a local documentation server for reviewing purposes, using Docker:
+
+.. code-block:: bash
+
+  cd <Mailu repo>
+  docker build -t docs docs
+  docker run -p 127.0.0.1:8080:80 docs
+
+You can now read the local documentation by navigating to http://localhost:8080.
+
+.. note:: After modifying the documentation, the image needs to be rebuild and the container restarted for the changes to become visible.
+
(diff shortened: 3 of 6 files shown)


muhlemmer:
Note afterwards: if it is not preferred to trust all containers in this network, there is the possibility to set up separate network overlays for different tasks. It should be possible to set up a "trusted" network for authentication related authentication.

Nevertheless, XCLIENT should reflect a network range and not a single IP.

muhlemmer:
Personally, I would like to refrain from using `POD_ADDRESS_RANGE`. The name of the variable name does not make any sense in a regular Docker deployment. It is a Kubernetes variable, which I misused in a past solution to pass the `RELAYHOSTS` network to Dovecot.

So I think, in any standard code we should use `RELAYHOSTS` variable, including in the Dovecot config. Subsequently, Kubernetes `.yml` files can set `RELAYHOSTS=POD_ADDRESS_RANGE`.

Any comments are welcome!

muhlemmer:
Excuse my English. I really meant to say `RELAYNETS`. It is already used in Mailu as "the Docker network".

I don't think we should support all the different implementation's variables. Keep our own set of variables and the startup files from docker-compose, docker-compose for swarm and kubernetes have to make sure those variables are assigned accordingly.

So this pull request makes sure that `RELAYNETS` is used instead of only the `FRONT_ADDRESS`. Subsequently, `RELAYNETS` is already set in `.env` and it will work for both the classic docker-compose and Docker swarm. Al that needs to be done is update the Kubernetes.

Then, as a general comment and probably off-topic for this PR, I would like to suggest to use the `RELAYNETS` in a uniform way, instead of `POD_ADDRESS_RANGE` which name does not make much sense if one is not involved in Kubernetes. (What the hell does `POD` mean anyways? :stuck_out_tongue: ). At this moment `POD_ADDRESS_RANGE` is only used in the [Dovecot config](https://github.com/Mailu/Mailu/blob/f9ef126e037293438fd79b255e2f75277a0bfefe/core/dovecot/conf/dovecot-sql.conf.ext#L6) and is not yet integrated in any release version. I mean, look at this line of code:

```
 SELECT NULL as password, 'Y' as nopassword, '{% if POD_ADDRESS_RANGE %}{{ POD_ADDRESS_RANGE }}{% else %}{{ FRONT_ADDRESS }}{% if WEBMAIL_ADDRESS %},{{ WEBMAIL_ADDRESS }}{% endif %}{% endif %}' as allow_nets \
```
3 inline `if` statements to cover all the possible cases. Ughh. If we replace those 3 variables by the single variable of `RELAYNETS` it would cover all the cases in a uniform and clean way.

ofthesun9:
Hi,
RELAYNETS is used to declare nets that can use postfix as a relay server. This is not necessarily the same than what we want to do here... I think it is preferable to use another variable, and I guess that is why POD_ADDRESS_RANGE was introduced.

I suggest that we use the same POD_ADDRESS_RANGE variable in the other parts of Mailu code where it seems necessary for swarm deployment. I wouldn't be surprised that it is needed anyway for kubernetes..

On a side note, It seems that we might not need a range but rather a single IP.  At least on my setup, this is the case: the logs of smtp &imap show that the originator IP is always 10.0.1.4, which is not the front service neither the front containers... By inspecting the mailu_default network, I found this IP under the name "mailu_default-endpoint". But I don't know how to resolve this IP from within mailu code, otherwise we could have another approach and set the FRONT_ADDRESS to this mailu_default-endpoint (only whan under swarm mode of course)...

muhlemmer:
> RELAYNETS is used to declare nets that can use postfix as a relay server. This is not necessarily the same than what we want to do here... 
[docs/compose/.env](https://github.com/Mailu/Mailu/blob/f9ef126e037293438fd79b255e2f75277a0bfefe/docs/compose/.env#L66): 
> # Networks granted relay permissions, **make sure that you include your Docker**
> # **internal network** (default to 172.17.0.0/16)
`POD_ADDRESS_RANGE` will just be redundant as it uses the same network range.

> I think it is preferable to use another variable, and I guess that is why POD_ADDRESS_RANGE was introduced.
>
>I suggest that we use the same POD_ADDRESS_RANGE variable in the other parts of Mailu code where it seems necessary for swarm deployment. I wouldn't be surprised that it is needed anyway for kubernetes..

If you run a repository[ search](https://github.com/Mailu/Mailu/search?q=POD_ADDRESS_RANGE&unscoped_q=POD_ADDRESS_RANGE), `POD_ADDRESS_RANGE` is only used once and is introduced **only** for kubernetes. A "POD" is a kubernetes thingy I guess and has nothing to do with regular Docker.

> On a side note, It seems that we might not need a range but rather a single IP. At least on my setup, this is the case: the logs of smtp &imap show that the originator IP is always 10.0.1.4, which is not the front service neither the front containers... By inspecting the mailu_default network, I found this IP under the name "mailu_default-endpoint". But I don't know how to resolve this IP from within mailu code, otherwise we could have another approach and set the FRONT_ADDRESS to this mailu_default-endpoint (only whan under swarm mode of course)...

Are you still using `dnsrr` as workaround, as mentioned in #530? The purpose of what I'm trying to do is getting swarm mode to work without `dnsrr`. In such a case you will not find the mentioned endpoint and originating IP's are spread throughout the network range. See see comments in that issue for background info.

muhlemmer:
I will change it to POD_ADDRESS_RANGE as it seems accepted elsewhere. After the build errors on `master` are resolved, I'll rebase and re-submit.

@ofthesun9. What is you intention on using the `dnsrr` workaround?

ofthesun9:
dnsrr workaround will not be needed anymore, see https://github.com/Mailu/Mailu/pull/604

muhlemmer:
@ofthesun9. Please re-evaluate your review.

PR #576, opened 2018-08-22
Kubernetes fixed for production

hacor:
Hello,

I made some breaking changes to the whole kubernetes layout and ideas. This was because it seemed after using it for a month with success my Mailu setup became an open relay and I had to redesign the implementation completely.

**Major changes:**
- `Nginx Ingress Controller` Daemonset will now coexist with the `Mailu Front` Daemonset both using the `hostNetwork: true` mode. The reason was that I made an open relay because the Nginx Controller who forwarded the mail traffic automatically made it appearing from the pod network. That's why I had to make the front container public to be able to see the real IPs.
- Port `80` and `443` will be closed on the front container when environment variable `KUBERNETES_INGRESS=true` exists. Leaving this ports in favor of the standard Ingress Controller. This has no effect on the docker compose environment!
- All mappings of the port `80` and `443` of the front containers `nginx.conf` were replaced by `Ingress manifests` for the Ingress controller (including the auto-fetching of lets-encrypt certificates)
- I replaced all the `:master` tags by images in my docker account (`haor/mailu-*:1.6`). This was for testing purposes to be sure no changes on master branch would affect my setup. These can of course be replaced when some definitive tags were chosen by the Mailu team.

**Elements tested successfully:**
- Rouncube
- Webdav
- Antispam
- Admin UI
And all seem to work very good and stable now

**Minor change:**
- I added the ability to use the environment variable `RSPAMD_PASSWORD` to provide a more secure password for rspamd. If the variable doesn't exist, it still uses the older `mailu` value.

All remarks are welcome!
Sorry for the big error on my previous release, I hope this one will make things right.

Best regards,

Hacor


hacor:
Any update on this PR?

Thanks!

kaiyou:
Sorry for the delay, I basically have had zero time lately to work on Mailu and am looking for maintainers. I am thinking about introducing an automated review+merging system.

Your pr looks nice except for the part where you use images from your own Dockerhub namespace, could you switch back to using ``mailu/``?

hacor:
No problem! I used my own docker images as a temporary solution because of following reasons:
- The `master` branch seems not a good solution for its future changes
- The `1.5` branch doesn’t contain all necessary changes

What do you suggest to use?  I will make the necessary changes!
I am willing to do following ups as a maintainer on the kubernetes side

Best regards!

Hacor

hacor:
No problem! I used my own docker images as a temporary solution because of following reasons:
- The `master` branch seems not a good solution for its future changes
- The `1.5` branch doesn’t contain all necessary changes

What do you suggest to use?  I will make the necessary changes!
I am willing to do following ups as a maintainer on the kubernetes side

Best regards!

Hacor

kaiyou:
I would suggest the master branch, and then pin a stable branch when we reach one with good kubernetes support, e.g. 1.6.

hacor:
I reverted the images and tags back to the Mailu repository.
Sorry for the delay!

docs/kubernetes/mailu/index.rst
```diff
@@ -0,0 +1,159 @@
...
+   metadata:
+     name: letsencrypt-prod
+   spec:
+     acme:
+       email: something@example.com
+       http01: {}
+       privateKeySecretRef:
+         key: ""
+         name: letsencrypt-stage
+       server: https://acme-v02.api.letsencrypt.org/directory
```
consideRatio: I think a cluster issuer isn't needed unless you want to make cert-manager provide certificates across kubernetes namespaces. An `Issuer` of some kind is enough if it is created in the same namespace as the Certificate resource as far as I know.

consideRatio: But perhaps it is quite common practice to utilize one single ClusterIssuer, so not changing anything may be fine.

consideRatio: This was my knowledge background, I'm not yet experienced using cert-manager:
https://www.youtube.com/watch?v=etC5d0vpLZE&t=0s&list=PLShDm2AZYnK3cWZpOjV7nOpL7plH2Ztz0&index=5

hacor: @consideRatio You're completely right, a standard `Issuer` should do the trick as well for the `mailserver` namespace. There are two main reasons why I choose for the `clusterIssuer`
- Many people will run a website along their mailing services. Hopefully in a different namespace. Therefor they can use the same `clusterIssuer` in order to obtain their certificates.
- My setup runs many websites and it just has a single `clusterIssuer` present. Depending on the *correct* `RBAC` that is set up, there should be no problem using this.

hacor: It depends completely on everyone's personal setup, but can be easily changed by every user. That's why I provided the YAML in the docs. But this should work in most setups, and could be easily implemented if there's already a `clusterIssuer` present

consideRatio: Ah thanks I learn actively about cert-manager right.
What about the following code btw?

```yaml
       privateKeySecretRef:
         key: "" # why an empty string here?
         name: letsencrypt-stage # this is named staging, but the server below is not for staging (should be acme-staging-v02 i think)
       server: https://acme-v02.api.letsencrypt.org/directory
```

I

hacor: - `key`-field seems to be removed now indeed. It is an older field that could reference an ACME provider which shared a private key with you. According to their [docs](http://docs.cert-manager.io/en/latest/reference/issuers.html), this field indeed doesn't exist anymore. I will remove it!
- `name`: That's indeed an error when adapting all the fields for the PR. This field just slipped my attention ;-)

Thanks for pointing this out! Commit will follow soon

docs/kubernetes/mailu/admin-ingress.yaml
```diff
@@ -0,0 +1,89 @@
...
+          serviceName: admin
+          servicePort: 80
+---
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: mailu-admin-static-ingress
+  namespace: mailu-mailserver
+  annotations:
+    kubernetes.io/ingress.class: tectonic
```
consideRatio: `tectonic` - What does this imply for mailu, does this mailu kubernetes installation require something specific due to this?

docs/kubernetes/mailu/index.rst
```diff
@@ -0,0 +1,158 @@
...
+~~~~~~~~~
+
+There’s chosen to have a double NGINX stack for Mailu, this way the main
+ingress can still be used to access other websites/domains on your
+cluster. This is the current structure:
+
+-  ``NGINX Ingress controller``: Listens to the nodes ports 80 & 443. We decided to keep the Ingress Controller, because he is good in what he does.
+-  ``Cert manager``: Creates automatic Lets Encrypt certificates based on an ``Ingress``-objects domain name.
+-  ``Mailu NGINX Front daemonset``: This daemonset runs in parallel with the Nginx Ingress Controller and only listens on all E-mail specific ports (25, 110, 143, 587,...)
+-  ``Mailu components``: All Mailu components are split into separate files to make them more handy to use
```
consideRatio: - *We have* chosen to have a double NGINX stack for Mailu

Perhaps remove `We decided to keep the Ingress Controller, because he is good in what he does.`, it references some previous state and a later state. As a reader I have no clue about what was the state before.

- What are the mailu components? Those above?

docs/kubernetes/mailu/index.rst
```diff
@@ -0,0 +1,158 @@
...
+   kind: ClusterIssuer
+   metadata:
+     name: letsencrypt-prod
+   spec:
+     acme:
+       email: something@example.com
+       http01: {}
+       privateKeySecretRef:
+         name: letsencrypt-stage
+       server: https://acme-staging-v02.api.letsencrypt.org/directory
```
consideRatio: a mix between prod / stage - go with one systematically?

docs/kubernetes/mailu/index.rst
```diff
@@ -0,0 +1,158 @@
...
+    kubectl create -f imap.yaml
+    kubectl create -f security.yaml
+    kubectl create -f smtp.yaml
+    kubectl create -f fetchmail.yaml
+    kubectl create -f admin.yaml
+    kubectl create -f webdav.yaml
+    kubectl create -f admin-ingress.yaml
+    kubectl create -f webdav-ingress.yaml
+    kubectl create -f security-ingress.yaml
+    kubectl create -f webmail-ingress.yaml
```
consideRatio: Perhaps `kubectl apply -f name-of-folder-where-all-these-reside` instead?

docs/kubernetes/mailu/index.rst
```diff
@@ -0,0 +1,158 @@
...
+And in the pod run the following command. The command uses following entries:
+
+- ``admin`` Make it an admin user
+- ``root`` The first part of the e-mail adres (ROOT@example.com)
+- ``example.com`` the domain appendix
+- ``password`` the chosen password for the user
+
+.. code-block:: bash
+
+    python manage.py admin root example.com password
```
consideRatio: Move the command up over the description of the parameters passed

docs/kubernetes/mailu/admin-ingress.yaml
hacor: This class is deleted on the nginx controller and in all the ingresses. It is indeed not necessary, but can be handy when you have multiple `nginx ingress controllers`
This is a setting from my personal setup which is based on the CoreOS Tectonic bare metal cluster. Their an ingress-class name is standard implemented.

docs/kubernetes/mailu/index.rst
hacor: That won't work unfortunately. All things coming before the `rbac.yaml` file will fail because the namespace won't exist yet. The apply works if the namespace exists already or the namespace creation happens in the first file alphabetically. 
I personally prefer this way because we can add more webmails in the future for example. Now only roundcube is supported. I also don't use fetchmail, so I don't start it

docs/kubernetes/mailu/index.rst
hacor: Caught me on speed here. Sorry! This is fixed, I also added docs for `Issuer` so the user can change if he wants. I also added an example staging and production clusterIssuer

docs/kubernetes/mailu/index.rst
hacor: Done!

docs/kubernetes/mailu/index.rst
hacor: Done!

kaiyou:
Sorry but I am not experienced enough with k8s to judge of the quality of the PR thoroughly. I am ready to merge when you feel comments by @consideRatio have been answered :)

hacor:
@kaiyou :+1: Let's wait for @consideRatio to reply to my last commit. I found all his comments very helpful to make future use of my PR more easy for others

docs/kubernetes/nginx/nginx-ingress.yaml
```diff
@@ -94,7 +94,6 @@ spec:
           - --udp-services-configmap=$(POD_NAMESPACE)/udp-services
           - --annotations-prefix=ingress.kubernetes.io
           - --enable-ssl-passthrough
-          - --ingress-class=tectonic
```
consideRatio: There is still a reference to tectonic-custom-error above, should it be?

Please note that I don't know whats going on so well, I'm mainly using generic ideas to review this.

consideRatio:
@hacor thanks for the encouragement! I'm very happy about the work you have done to make Mailu deployed on k8s!

@kaiyou @hacor please understand that I currently lack knowledge of relevance to the networking domain and that I'm currently learning about it. I also lack a lot of knowledge about mailing servers and Mailu. I'm very eager to learn though :D

---

@hacor you wrote about using `hostNetwork: true`, not being well versed I tried reading up and learning about it. I read [this blog post](http://alesnosek.com/blog/2017/02/14/accessing-kubernetes-pods-from-outside-of-the-cluster/) and they wrote:

> [...] two applications [pods] requiring the same port cannot run on the same node. This can lead to port conflicts when the number of applications running on the cluster grows. On top of that, creating a pod with `hostNetwork: true` on OpenShift is a privileged operation. For these reasons, the host networking is not a good way to make your applications accessible from outside of the cluster.

So I wonder, is `hostNetwork: true` mostly a simple way to accomplish what we want, or is it required? Also, if it is not required, what would the alternative solution be to using it? Reading further in the blog, they discuss the usage of hostPort, as a more controlled option of accomplishing the same thing as hostNetwork, specifically for the ingress-nginx controller. Something to consider? Perhaps one could use a nodePort service + a LoadBalancer service ahead of it? Hmm... Just guessing around here.

I'll try to grasp the GitLab Helm charts implementation for reference, they deploy a ingress-nginx controller as well and utilize cert-manager also.

To summarize: No obvious changes requested to this PR, just some discussion based on an actively evolving understanding of k8s networking.

hacor:
@consideRatio about the `hostNetwork: true` setting: My personal vision is to try to have as little services as possible running on the `hostNetwork`. In my case, it's only the Nginx Ingress Controller, all other services use `in-cluster` ports, so they're not accessible by the outside world. Only through the Ingress Controller (which publishes a very limited amount of ports). This way if something happens, I know to look first to the Ingress Controller (which is developed by Google, so quite save on its own). The controller can directly forward TCP and UDP ports as well, but uses only the ports you specify.

If I have another service running on port 80, I don't use `hostNetwork: true` and add an `Ingress`. This way the request arrives on the Nginx Ingress Controller, which will pass it on to the correct underlying service.

Using external loadBalancers is more complex and won't address as many users as the Ingress Controller.
Can you please specify a little bit more what do you think _strange_ or so about the ingress controller or `hostNetwork: true`?
Thanks!

docs/kubernetes/nginx/nginx-ingress.yaml
consideRatio: Ah yepp but above this, I could not comment on it directly:

One commented line, and one configmap setting.

```
       containers:
         - name: nginx-ingress-lb
           image: quay.io/kubernetes-ingress-controller/nginx-ingress-controller:0.16.2
           args:
           - /nginx-ingress-controller
           - --configmap=$(POD_NAMESPACE)/tectonic-custom-error
           - --default-backend-service=$(POD_NAMESPACE)/default-http-backend
           #- --default-ssl-certificate=tectonic-system/tectonic-ingress-tls-secret
```

consideRatio:
@hacor My understanding is that using `hostNetwork: true` is like a wildcard allowing a lot, when it would perhaps be sufficient to specify hostPort on two specific ports, 80 and 443, but I'm out of my depth right now and I'm not confident about things here.

From [this blog post](http://alesnosek.com/blog/2017/02/14/accessing-kubernetes-pods-from-outside-of-the-cluster/).

> What is the `hostPort` used for? For example, the nginx based Ingress controller is deployed as a set of containers running on top of Kubernetes. These containers are configured to use hostPorts 80 and 443 to allow the inbound traffic on these ports from the outside of the Kubernetes cluster.

I figured one could use hostPort on those ports when needed instead of using the `hostNetwork:true`, but it may have the same result unless we add more ports that should not be opened on the host node? I don't know.

hacor:
@consideRatio Ok now I understand your opinion. I took some time to reread the complete Ingress Controller documentation and to be honest, my manifest using `hostNetwork: true` (which was the standard in this dark age 😄  ) is in use for about 1,5 years, so a review is interesting. Based on my findings I came to following conclusions:

- `hostNetwork: true`: Any program in the docker container can open up ports and they will be directly available on the hosts network interfaces. A firewall which only allows the correct ports in front of the Ingress Controller is probably a good thing. <= This is the setup I use for the moment
- `hostPort`: This setting opens up a single port on the host public network, but this means you need to specify every port in the `configMap` as well. So you need to specify all ports (25, 110, 193,...) we use for Mailu also in the Ingress Manifest.
- `securityContext`: This is the standard that seems to be used now on the [Ingress Controller examples](https://github.com/kubernetes/ingress-nginx/blob/master/deploy/with-rbac.yaml#L35). This means that only a single proces of the container can open up host ports. In this case user 33 which is the standard `www-data` user which Nginx uses. This way malicious processes in the container can not open ports on the host themselves. I'm planning of implementing this system but I need some time to test this myself before adapting the PR. @kaiyou : please give me some time on this to investigate.

Best regards!
It's becoming a very interesting walk!

Hacor

docs/kubernetes/nginx/nginx-ingress.yaml
hacor: Those lines indeed have no influence at all if the configmap doesn't exist. But I removed them! I only did a search on `tectonic` in the `mailu` subdirectory, not in the `nginx` directory. This should be solved now. All traces of tectonic are wiped out.

hacor: They will get removed in the following commit in which I will look into the Ingress Controller manifest as well. For this I need some time to test and make sure it works. I'll get beck here soon!

docs/kubernetes/nginx/nginx-ingress.yaml
```diff
@@ -87,9 +87,7 @@ spec:
           image: quay.io/kubernetes-ingress-controller/nginx-ingress-controller:0.16.2
           args:
           - /nginx-ingress-controller
-          - --configmap=$(POD_NAMESPACE)/tectonic-custom-error
```
consideRatio: Should this be set explicitly to something else perhaps? Like:
```
        - --configmap=$(POD_NAMESPACE)/nginx-configuration
```

**About the command line argument:**
https://github.com/kubernetes/ingress-nginx/blob/c3ce6b892e5f1cc1066f81c19482dded2901ad45/docs/user-guide/cli-arguments.md#command-line-arguments

**Examples of this being done:**
https://github.com/kubernetes/ingress-nginx/search?utf8=%E2%9C%93&q=--configmap&type=

consideRatio:
@hacor ah thanks for sharing your insights about `hostPort` and `hostNetwork` and `securityContext`, it helps me learn. You seem to have done a very thorough work with this.

I'll try utilize it myself soon I think - if I dare go the path of self-hosting a mail server at all. And if that would work out, I'll be interested in making it a Helm chart.

Thanks for all your work on this!

docs/kubernetes/nginx/nginx-ingress.yaml
hacor: That's true, but in fact there's no need to overwrite the configuration in a standard HTTP/HTTPS  + Mailu setup
So I removed it to make everything transparent. Advanced users will figure out themselves how to add extra configuration options

consideRatio:
@hacor @kaiyou this LGTM!

docs/kubernetes/mailu/security-ingress.yaml
```diff
@@ -0,0 +1,30 @@
...
+      proxy_set_header X-Forwarded-For "";
+  labels:
+    app: mailu
+    role: mail
+    tier: frontend
+spec:
+  tls:
+  - hosts:
+    - "mail.example.com"
+    secretName: letsencrypt-certs-all # If unsure how to generate these, check out https://github.com/ployst/docker-letsencrypt
```
consideRatio: @hacor this is taken care of by cert-manager, right?

docs/kubernetes/mailu/webdav-ingress.yaml
```diff
@@ -0,0 +1,46 @@
...
+      }
+  labels:
+    app: mailu
+    role: mail
+    tier: frontend
+spec:
+  tls:
+  - hosts:
+    - "mail.example.com"
+    secretName: letsencrypt-certs-all # If unsure how to generate these, check out https://github.com/ployst/docker-letsencrypt
```
consideRatio:
```
secretName: mail-example-com-tls # should be automatically created by cert-manager
```

docs/kubernetes/mailu/security-ingress.yaml
hacor: Ok thanks!
I added `mail-example-com-tls` to the next commit!

hacor:
@consideRatio Thanks for all the input! I really think you made this PR more transparent for others to use!
@kaiyou I think we're almost ready to pull here, but I was wondering whether you already have an idea about the container versioning? `1.6` or `1.5-k8s` or so. This way I can adapt all the image tags and merge all my commits into one for beter overview of the Mailu commit history

Thanks to you all for all this productive input!

hacor:
@muhlemmer It seems I added all files which weren't in my repository to this PR. I think I made some errors using git. Any advice please? Sorry for the troubles

muhlemmer:
ooh ooh, this looks nasty. I'll have a look.

hacor:
@muhlemmer I think we're there! Luckily I had my finished branch (before merging) still on my fixed PC. Sorry for the trouble, I don't use rebasing and merging that often and it always give me some troubles.

services/rspamd/conf/worker-controller.inc
```diff
@@ -1,3 +1,7 @@
 bind_socket = "*:11334";
+{% if RSPAMD_PASSWORD %}
+password = "{{ RSPAMD_PASSWORD }}";
+{% else %}
 password = "mailu";
+{% endif %}
```
muhlemmer: This seems to define new functionality, which  does not necessarily belong to Kubernetes. A change like this should be proposed in a separate pull request. Futhermore, I beleive (assumption) that the password is passed to the rspamd page by Admin. Changing overriding the password will break this behavior.

docs/kubernetes/mailu/configmap.yaml
```diff
@@ -40,11 +40,20 @@
     TLS_FLAVOR: "cert"
 
     # Authentication rate limit (per source IP address)
-    AUTH_RATELIMIT: "10/minute;1000/hour"
+    AUTH_RATELIMIT: "100/minute;10000/hour"
```
kaiyou: Why change this specifically for kubernetes?

kaiyou:
Not being a Kubernetes expert, this is looking rather good. Maybe now that feat-abstract-db is merged, a couple of the mount points are not needed anymore, even if this is not critical atm.

services/rspamd/conf/worker-controller.inc
hacor: OK I can separate those

docs/kubernetes/mailu/configmap.yaml
hacor: Long story, I created 2 e-mail addresses and connecting through IMAP on my laptop and mobile gave me login errors. It seemed that IMAP opens up multiple connections to the Dovecot server.
Although I have some strange behavior now: 
My laptop (apple mail, which is normally very stable) asks me for my password all the time (sometimes it works, sometimes it doesn't) and in the log files I see something like `[AUTHENTICATIONFAILED]`. This is rather bothering and I don't see a solution yet for this. I need some time to investigate more.

hacor: BTW, any hints on this are always welcome, I really now the entered passwords are 100% correct

core/postfix/conf/main.cf
```diff
@@ -32,7 +32,7 @@ relayhost = {{ RELAYHOST }}
 recipient_delimiter = {{ RECIPIENT_DELIMITER }}
 
 # Only the front server is allowed to perform xclient
-smtpd_authorized_xclient_hosts={{ FRONT_ADDRESS }}
+smtpd_authorized_xclient_hosts={{ RELAYNETS }}
```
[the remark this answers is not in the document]
hacor: I'll look into this

hacor:
I'm sorry this PR seems like such a long job, for the moment I'm very busy in my daily job and it will take some time to update this PR. But I make sure the asked adaptations will follow. Just give me some time.
I want to thank @kaiyou and @muhlemmer for the very nice and helpful comments to this PR

docs/kubernetes/mailu/front.yaml
```diff
@@ -1,23 +1,41 @@
-
-apiVersion: extensions/v1beta1
-kind: Deployment
+apiVersion: apps/v1beta2
+kind: DaemonSet
```
briantopping: Why a `DaemonSet`? I believed the use of `Deployment` before was good. Also see my notes below WRT affinity.

docs/kubernetes/mailu/front.yaml
```diff
@@ -1,23 +1,41 @@
...
+        component: ingress-controller
+        type: nginx
     spec:
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: node-role.kubernetes.io/master
+                operator: DoesNotExist
```
briantopping: My current main deployment has master taint turned off, specifically it's a two-node system where both nodes are masters. This affinity selection would mean it runs on no nodes.

Even if that were not true, I'm not sure I want the containers running on every worker node.

docs/kubernetes/mailu/webmail.yaml
```diff
@@ -15,7 +15,7 @@ spec:
     spec:
       containers:
       - name: roundcube
-        image: mailu/roundcube:1.5
+        image: mailu/roundcube:master
```
briantopping: I've seen this update to `master` in several places. I'm assuming this is just for development, but in a production environment, it can lead to problems if an untested change updates the image and the user's machine pulls it.

docs/kubernetes/mailu/front.yaml
hacor: The `Deployment` usage has it limitations, when the container starts it opens up node ports (all mail ports) and whenever a node fails or the pod fails it can be rescheduled to another node which means you need to adapt DNS settings every time your pod reschedules. The only solution to this is to use a cloud based LoadBalancer which connects to a kubernetes service (which limits usage to specific cloud providers only)
With the `DaemonSet` configuration, the front container is placed on the same level as the Nginx Ingress Controller, meaning every node is able to accept mail connections. This keeps the overview of the complete system transparent. You have the Ingress Controller DaemonSet opening up ports 80 and 443 (accepting all kinds of HTTP(S) services you have on your cluster) and the Mailu Front DaemonSet accepting all mail ports.
That's why I still go for the `DaemonSet` configuration.
Hopefully this clears things out

docs/kubernetes/mailu/front.yaml
hacor: > Even if that were not true, I'm not sure I want the containers running on every worker node.

Of course it's to the end user to decide whether he wants it or not. But to enable HA I think this is the most easy solution. 
The whole `nodeSelectorTerms` can of course be adapted to only apply to a few nodes only. This can be easily adapted by a user.

> My current main deployment has master taint turned off, specifically it's a two-node system where both nodes are masters. This affinity selection would mean it runs on no nodes.

This is true but can be easily changed to (If I remember correctly):
```
- matchExpressions: 
  - key: node-role.kubernetes.io/worker
     operator: Exists
```

> Ideally, the nodes would be autoscaling rather than pinned to specific nodes.

 I think you mean the pods would be autoscaling, but this is mentioned above and gives you no real certainty about which nodes will run the `Deployment`

docs/kubernetes/mailu/webmail.yaml
hacor: As you can read above during this PR, it's just for testing purposes now. When the PR is finished @kaiyou will help decide if we step into a new version `1.6` (or whatever is decided), build new containers and then we change the container images versions in all of the manifests.

services/rspamd/conf/worker-controller.inc
hacor: @kaiyou Is this true that the password is provided by Admin and not always `mailu`
- If the password is always `mailu` I will create another PR providing this functionality
- If the password is provided by Admin, any hints on how to configure are welcome and I will remove the lines above

docs/kubernetes/mailu/front.yaml
briantopping: I see, that's an innovative solution to the problem. Most of the solutions I have seen consider an internal load balancer (HAProxy-based solutions are good for this) such that there is a singular endpoint for the services as they are distributed within the cluster. This solves the ephemeral (portable) pod problem. The internal load balancer is then exposed externally using whatever facility is available, including MetalLB for bare metal (non-SaaS) clusters.

That said, there are a lot of moving parts with this common pattern when the choice of CNI is included and in the real world and the lack of a comprehensive TCK for kubernetes means that these moving parts often don't work together they way they should, which is very frustrating for users.

In my deployment, I have many addresses that are not associated directly with interfaces. This is facilitated by MetalLB announcing these addresses to upstream BGP and service routed through Cilium CNI. So `NodePort` and `hostPort` are not viable solutions.

briantopping: @hacor I also see that the conversation does not leave it's state of "resolved" after leaving another comment. Hopefully it is not hidden on your end.

briantopping:
@muhlemmer Indeed, we are all very excited for this, it's great to see so much excitement for the changes. It shows that the efforts will not languish, for sure. 

@hacor, sorry we're being annoying. As soon as you're checked in, it's going to be a love explosion. Just saying.. :)

docs/kubernetes/mailu/front.yaml
briantopping: Correct, apologies. I couldn't even see this response until digging around, I thought I didn't send it at all! :/

In any event, that does help, it puts the pod on every node. I'll experiment with setting up a HAProxy load balancer. I don't want to interrupt your flow with features you aren't interested in.

hacor:
@muhlemmer I think all remarks are resolved. We can move to the final review! 
Thanks to all for the patience and all the valuable input!

briantopping:
@muhlemmer Until your review is marked `resolved`, I don't think the merge will happen, but I can't figure out which one it is! It appears everything through the "See review" above is marked as `Outdated`. Hmm.

@kaiyou I would try testing in a free setup on Google Cloud at https://cloud.google.com/free/.

muhlemmer:
Hi. All it needs is two reviews from persons with write access. My review can be resolved or dismissed by any of them.

I'm mostly working on Mailu during business hours, or when I'm bored in my free time. Currently neither is the case ;).

As with all other reviews, I will do another iteration once I'm online. For now some patience is advised.


On October 20, 2018 6:55:49 PM GMT+03:00, Brian Topping <notifications@github.com> wrote:
>@muhlemmer Until your review is marked `resolved`, I don't think the
>merge will happen, but I can't figure out which one it is! It appears
>everything through the "See review" above is marked as `Outdated`. Hmm.
>
>@kaiyou I would try testing in a free setup on Google Cloud at
>https://cloud.google.com/free/.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/Mailu/Mailu/pull/576#issuecomment-431593822

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

briantopping:
@muhlemmer I'm very sorry for offending you with input. I certainly regret doing so and I deleted the input in question. I can go ahead and do the same with all my contributions in this thread and fork the project if that suits you better. 

Please edit your response and remove my comment. I do not approve of you using it in this manner.

muhlemmer:
Dude, no offense was taken. Just saying I'm not available right now for a review.

On October 20, 2018 7:46:22 PM GMT+03:00, Brian Topping <notifications@github.com> wrote:
>@muhlemmer I'm very sorry for offending you with input. I certainly
>regret doing so and I deleted the input in question. I can go ahead and
>do the same with all my contributions in this thread and fork the
>project if that suits you better. 
>
>Please edit your response and remove my comment. I do not approve of
>you using it in this manner.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/Mailu/Mailu/pull/576#issuecomment-431597876

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

briantopping:
* We’re on GitHub. You don’t have to get into a condescending display telling people how to use Git here. I let it slide the other day.
* Everyone is busy. Different people are busy for different reasons. “Some patience is advised” after the previous snipe sets a trend.

We all have things to do. Start by assuming good intent in others. It’s not your role to police others here, this is a community of contributors. 

If you feel the need to say that kind of thing, don’t.

hacor:
@consideRatio & @kaiyou Remember like mentioned in [this article](https://cloud.google.com/compute/docs/tutorials/sending-mail/) that Google blocks port 25.
Just to let you know all other mail ports can be used

hacor:
I do wanted to ask one more thing, which happens some times and I was wondering for people using Docker Swarm or Docker compose have the same problem.

As we all know we provided `POD_ADDRESS_RANGE` (in my case 10.2.0.0/16), my hosts have an interface with an address of 10.1.6.0/24. Because the `Front` container uses the `hostNet: true` option it listens on the 10.1.6.0/24 address and forwards the traffic through the IMAP pod (having a 10.2.0.0/16 address). In a very particular case: 
when the mail client asks access (based on round robin DNS) to the node which also hosts the IMAP pod, it seems to forward the 10.1.6.0/24 address through the front container and not the cluster address of the front container (10.2.0.0/16)
Therefor the Dovecot auth fails based on IP filtering (10.2.0.0/16 is allowed, 10.1.6.0/24 is blocked)
I still can not define what really causes this particular problem but I do think it's Kubernetes network related.

But the following questions do spring to mine:
- Is IP filtering on the Dovecot container really necessary? Because the Dovecot pod/container will always be running on a separate network (being the docker bridge or the K8s CNI, doesn't matter) and will never accept public connections directly. Only through the container network (Also here Docker Swarm, compose or K8s doesn't matter). Conclusion: Dovecot will almost always see the container network coming in, does this really adds an extra security layer to logging in to IMAP? Probably I just don't see it.
- Wouldn't be IP filtering more interesting on the `front` container? This container can filter out 10 failed logins from the same IP within 1 minute and block the public IP.
- Probably I could add 10.1.6.0/24 to accepted IPs, but may be I open up as a relay again (I'm not 100% sure)

I know this subject isn't really related to the K8s, but probably many people will experience some kind of logging in trouble based on the same behavior. May be I can still fix this before merging. Because now after some times (when accessing 1 specific node) the mail client re-asks the password for my e-mail.

Thanks for the suggestions

Best regards!!

muhlemmer:
@hacor, we had similar problems on Swarm. See issue #530. It was resolved by using POD_ADDRESS_RANGE in Dovecot config. See PR #604. This fix was only merged recently to master, so you might need to pull the new version of images to see this to work. Hope it does so for you as well.

PR #580, opened 2018-08-26
Fix binding in setup utility's gunicorn

muhlemmer:
`gunicorn` invocation explicitly wanted to bind to the global ipv6 address. Start up fails on hosts where Docker is not running with ipv6 support.

The updated command in this PR makes gunicorn to bind on port 80 of any available protocol. Now, it will not fail if ipv6 is not available inside docker.


kaiyou:
Very nice, thanks

PR #597, opened 2018-09-18
use safer cipher in roundcube

HorayNarea:
"Default is set for backward compatibility to DES-EDE3-CBC,
but you can choose e.g. AES-256-CBC which we consider a better choice."

https://github.com/roundcube/roundcubemail/blob/master/config/defaults.inc.php#L512


kaiyou:
Merging :)

Issue #601, opened 2018-09-23
Docs as docker


muhlemmer:
Like discussed on Matrix. I would like to create a Docker image for the Documentation server. This would eliminate the need of using travis to upload the webfiles and subsequently make life easier for developing a test environment.

A good side-effect is the ability for developers to run a private instance of the Documentation server when working on the documentation, for previewing.

Before implementing this, we need to have some things clear. Hence, I created this issue. I'm looking primarily to @kaiyou  to give some feedback to make this a bit more easy to implement

- [x] How does the current Documentation environment look like? What are the requirements? A copy-paste of the currently use `Dockerfile` and `requirements.txt` will help a great deal.
- [x] It was mentioned that the setup site would need to run in the same way. Does this mean the same Docker instance? Will they share the same Dockerfile?
- [x] Where should we put the Dockerfile? If it will be in `/docs` it would be included inside the web-server files?

kaiyou:
1. Currently the documentation is fully managed in the ``docs/`` folder, hence it uses ``requirements.txt`` there and the build code is actually in ``.travis.yml``
2. I did not mean the same instance when discussing this on Matrix: I see no point in trying to get both systems to work together while they are running on very different stacks
3. A Dockerfile in ``/docs`` would probably do the job pretty well, it should simply load the requirements, build the docs and have nginx or any other webserver run as the default command.

kaiyou:
I tried to implement a Dockerfike yesterday, there us just one catch: when building, sphinx needs to access the git folder in order to build the multiple docs versions, except that git folder is now outside the Docker build context.

I would suggest we set the build context for that specific image at the root of the repo. What do you think?

muhlemmer:
In the long run it was my intention to use built images from Travis to upload to docker hub. Using a deploy Section. So maybe it's better to reimplement the documentation build section in Travis and create the Docker image from there?

However, Travis deploy was a bit further down my roadmap, since it will push images for master and tags. I've raised the question about tagged releases before, but I'm not in front of the PC now to find that issue discussion.

If you want this functionality, I can prioritize the implementation and come with a proposal PR in a few days. It would be helpful if you can merge a Dockerfile + requirements.txt.

If I think better, a PR against the docs also needs to be checked for correct compilation on Travis.

On September 25, 2018 8:33:45 AM GMT+02:00, kaiyou <notifications@github.com> wrote:
>I tried to implement a Dockerfike yesterday, there us just one catch:
[...]

PR #602, opened 2018-09-23
Travis-CI automated test build

muhlemmer:
This PR is the first step into creating testing suite. At this stage, it instructs travis-ci to build all the Mailu images and the success of the test is only evaluated on this.

Note that the documentation deploy system has already been removed from `.travis.yml` , as it was giving complications. This functionality should be replaced as per #601.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,11 @@
-language: python
-python:
-  - "3.6"
-install:
-  - pip install -r docs/requirements.txt
+sudo: required
+services: docker
+addons:
+  apt:
+    packages:
+      - docker-ce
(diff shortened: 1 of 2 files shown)


muhlemmer:
I see that travis build will not be triggered in this PR, probably because it involves changes to the `.travis.yml` file. However, successful tests where done on my own fork.
https://travis-ci.org/usrpro/Mailu/builds/432272890

kaiyou:
More precisely, Travis integration is absolutely broken. I am setting up a new integration using Github apps at travis-ci.com, will get back once this is done.

PR #604, opened 2018-09-24
Enabling swarm deployment on master branch

ofthesun9:
-Extends the usage of POD_ADDRESS_RANGE
-Provides documentation

I guess this pull request could be usefull to other kind of deployment (like kubernetes), hence the choice to reuse POD_ADDRESS_RANGE which was already introduced in a commit

### DIFF
--- a/docs/swarm/master/README.md
+++ b/docs/swarm/master/README.md
@@ -56,12 +56,12 @@ core@coreos-01 ~ $ sudo umount /mnt/local/
 ```
 
 
-### Networking mode
+## Networking mode
 On this example, we are using:
 - the mesh routing mode (default mode). With this mode, each service is given a virtual IP adress and docker manages the routing between this virtual IP and the container(s) providing this service. 
 - the default ingress mode.
 
-## Allow authentification with the mesh routing
+### Allow authentification with the mesh routing
 In order to allow every (front & webmail) container to access the other services, we will use the variable POD_ADDRESS_RANGE.
 
 Let's create the mailu_default network:
@@ -73,7 +73,7 @@ core@coreos-01 ~ $ docker network inspect mailu_default | grep Subnet
 In the docker-compose.yml file, we will then use POD_ADDRESS_RANGE = 10.0.1.0/24 
 In fact, imap & smtp logs doesn't show the IPs from the front(s) container(s), but the IP of  "mailu_default-endpoint". So it is sufficient to set POD_ADDRESS_RANGE to this specific ip (which can be found by inspecting mailu_default network). The issue is that this endpoint is created while the stack is created, I did'nt figure a way to determine this IP before the stack creation...
 
-## Limitation with the ingress mode
+### Limitation with the ingress mode
 With the default ingress mode, the front(s) container(s) will see origin IP(s) all being 10.255.0.x (which is the ingress-endpoint, can be found by inspecting the ingress network)
 
 This issue is known and discussed here:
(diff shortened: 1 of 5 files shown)


ofthesun9:
Hi,

After a 2nd thought, I think the patch for dovecot.conf (submission_host=...) is not needed. Not doing any harm (I am using it since several weeks, but probably not needed)
I am currently doing some tests without it...

ofthesun9:
I reverted the modification done to dovecot.conf, as it is not needed indeed.
Sorry for the late notice

muhlemmer:
I started reviewing this. Some preliminary questions:
- In the documentation, is there a good reason why to use so many seperate NFS mounts? I use GlusterFS myself and use a single system mountpoint at `/mailu` without issues.
- Docs: In the NFS mount options you use `nolock`. I'm afraid this can lead to funky situations. Is it really neccesary in all deployments or something that just worked for you? A quick google gave me this:
> `nolock` — Disables file locking. This setting is occasionally required when connecting to older NFS servers. 

I would also like to discuss to which scope we need to provide documentation? Some examples:
- Setting up Docker swarm is very well step-by-step documented by Docker: https://docs.docker.com/get-started/
- Should we really provide all the NFS details? Maybe a warning box telling the reader something like "All mailu data needs to be on shared storage between the hosts". There are so many solutions to this out there, that it is impossible to cover them in our Documentation.

Anayway, thanks for the hard work putting this together :+1:

ofthesun9:
I agree it is more an Howto rather than a synthetic documentation focusing only on Mailu specifics.
Regarding the number of nfs volumes, I prefered to declare each volume needed directly from the stack declaration. I don't need to have every node to mount the "mailu data" at /mailu. But it could work as well.
Regarding the nolock option, I don't remember why I am using it.  I think this the default on my nfs-server but not 100% sure. I will have a look at it.

ofthesun9:
A general comment for those interested by a swarm deployment:
Due to the way the mesh routing is working, the source IP is "lost" and therefore not available to smtp, imap etc containers....
I guess this will apply to all of us see https://github.com/moby/moby/issues/25526

- We need to keep this in mind, as any rate limiter of fail2ban stuff will then be impossible (or even harmful)
- the other point is to be careful with RELAYNETS variable. In my case , if RELAYNETS encompasses  10.255.0.0/24, I create an smtp relay :-(

ofthesun9:
> I would also like to discuss to which scope we need to provide documentation? Some examples:
> 
>     * Setting up Docker swarm is very well step-by-step documented by Docker: https://docs.docker.com/get-started/
> 
>     * Should we really provide all the NFS details? Maybe a warning box telling the reader something like "All mailu data needs to be on shared storage between the hosts". There are so many solutions to this out there, that it is impossible to cover them in our Documentation.
> 
@muhlemmer ,
Yes the documentation can be simplified. But I would then rather keep (in a separate chapters) the working example(s). Personnaly, It helps me a lot when I found one of them :-). Would you be ok to add an example related to glusterfs ?

muhlemmer:
@kaiyou, I think this needs the final go-ahead from you. @ofthesun9 and myself are the only swarm users in the contributors tean.

kwek:
Not sure if this should be mentioned in the documentation, but currently I see it states:
> The docker stack deploy command doesn't support variable substitution in the .yml file itself

Another way could be to use the 'docker-compose config' command before the deploy. This also reads the .env file and outputs a parsed config (including variables). That file can then be deployed. Something like:

`docker-compose config > stack.yml && docker stack deploy stack.yml`

docs/swarm/master/README.md
```diff
@@ -0,0 +1,366 @@
...
+      replicas: 1
+
+  imap:
+    image: mailu/dovecot:master
+    restart: always
+    env_file: .env
+    environment:
+      - POD_ADDRESS_RANGE=10.0.1.0/24
+    volumes:
+#      - "$ROOT/data:/data"
```
kaiyou: This volume should not be needed anymore if you use the latest Dovecot container, since it does not query the database anymore.

docs/swarm/master/README.md
```diff
@@ -0,0 +1,366 @@
...
+      replicas: 2
+
+  smtp:
+    image: mailu/postfix:master
+    restart: always
+    env_file: .env
+    environment:
+      - POD_ADDRESS_RANGE=10.0.1.0/24
+    volumes:
+#      - "$ROOT/data:/data"
```
kaiyou: Same comment regarding the sqlite not being required anymore in the smtp container

docs/swarm/master/README.md
```diff
@@ -0,0 +1,366 @@
...
+      - imap
+    deploy:
+      replicas: 2
+
+  fetchmail:
+    image: mailu/fetchmail:master
+    restart: always
+    env_file: .env
+    volumes:
+#      - "$ROOT/data:/data"
```
kaiyou: Same comment about the sqlite not being required in the fetchmail container anymore.

services/rspamd/conf/worker-controller.inc
```diff
@@ -1,3 +1,3 @@
 bind_socket = "*:11334";
 password = "mailu";
-secure_ip = "{{ FRONT_ADDRESS }}";
+secure_ip = "{% if POD_ADDRESS_RANGE %}{{ POD_ADDRESS_RANGE }}{% else %}{{ FRONT_ADDRESS }}{% endif %}";
```
kaiyou: ``{{ POD_ADDRESS_RANGE or FRONT_ADDRESS}}`` sounds cleaner to me, but okay.

PR #609, opened 2018-09-25
admin: number of workers == number of CPU-cores

HorayNarea:


kaiyou:
If I am not mistaken, the number of workers is also the number of requests that can be answered simultaneously.

As opposed to CPU-intense scientific computation or compilation, serving a web request is 90% composed of io wait (waiting for the database, the socket, etc.), so having more processes than CPUs actually makes sense because multiple processes can live property through context changes and serve request as fast as before.

The question would be: how many processes do we need, and unfortunately I don't think that gunicorn has any adaptative worker manager.

hoellen:
> Gunicorn should only need 4-12 worker processes to handle hundreds or thousands of requests per second.

According to the [gunicorn docs](http://docs.gunicorn.org/en/stable/design.html#how-many-workers) they recommend (2 x $num_cores) +1 as the number of workers.

kaiyou:
Then I guess we could either set it to 8 or so, or blindly apply the formula, I just would not like gunicorn to get a single worker if the CPU is monocore.

HorayNarea:
I just found out: multi-threaded workers don't work with sqlite… so I suggest either @kaiyou closes this PR or leaves it open for future discussion about the worker-count

curry684:
Isn't it just easier to rebuild the admin in PHP so we can optimize it together with webmails and phpmyadmin?

muhlemmer:
@curry684, the bottle neck here is sqlite, not python. Using a database which can handle row-based locking like Postgresql will be the performance benefit in this subject.

@HorayNarea I will close this PR, as it pollutes the project overview. Also there is #670 for the admin refactoring.

PR #614, opened 2018-09-27
Attempt to fix the docs build context

muhlemmer:
The build context is solved by not referring twice to the parent directory.

However, sphinx is not able to find it's docs and the build still fails. I've tried a thing or two, but I don't have enough understanding on how sphinx is working with git. Maybe you have a suggestion, @kaiyou?

### DIFF
--- a/tests/build.yml
+++ b/tests/build.yml
@@ -50,4 +50,4 @@ services:
     image: mailu/docs:$VERSION
     build:
       context: ../
-      dockerfile: ../docs/Dockerfile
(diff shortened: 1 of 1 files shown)


kaiyou:
Okay, after spending >1h diving into sphinx-versioning, I conclude with a fair degree of certainty it is not the road I want to go down.

Let's think about something simpler, like what you were initially suggesting, build an image per branch of Mailu. If every image is an nginx running, I don't mind running a couple of them and having a reverse proxy on top of it. Builds will be far simpler that way.

PR #617, opened 2018-09-29
backport fix for #474 to 1.5

HorayNarea:
as @kaiyou said in the original issue #474: "this is pretty critical" …so we should fix it in the current release too.


kaiyou:
I am building to check it does not break the whole system before merging to 1.5 :)

kaiyou:
Merging this one manually since it runs just fine

PR #608, opened 2018-09-25
Docs as a Docker image

kaiyou:
This is a work in progress, not yet ready for review.

PR #620, opened 2018-09-30
Documentation on running a local docs container

muhlemmer:
As mentioned in #608, this provides the documentation on a local docs container.


kaiyou:
This looks really good, thanks :)

PR #621, opened 2018-09-30
1.5: Travis to always succeed

muhlemmer:
This will cause travis to never fail on PR's against 1.5. This is the quickest solution against build errors.

If you want to backport the testing suite, feel free to deny this PR and close it. However, the testing suite is still very new and will be extended in the coming weeks.


kaiyou:
Simple enough, thanks :)

PR #626, opened 2018-10-04
Add variable to choose whether to bounce or reject email when recipient is unknown. Fixes #583

### LINKED ISSUE #583: Unable to configure Mailu to reject instead of bounce
I'm having a spam issue in that I'm getting ~250.000 spam mails and mailu is bouncing them all. This got me on some spam lists so I'm looking into sending a reject to the server that's sending them to me instead of sending a bounce to the FROM address. As far as I can see this is not (yet) configurable. How would I do this?

At the moment my logs are littered with messages like `smtp_1       | 2018-08-22T07:30:18.130405545Z postfix/smtp[137]: 73EBC8B09: to=<some-random-address@yahoo.com.tw>, relay=mx-tw.mail.gm0.yahoodns.net[180.222.103.9]:25, delay=29, delays=17/0.03/12/0.24, dsn=4.7.0, status=deferred (host mx-tw.mail.gm0.yahoodns.net[180.222.103.9] said: 421 4.7.0 [TSS04] Messages from 94.142.240.35 temporarily deferred due to user complaints - 4.16.55.1; see https://help.yahoo.com/kb/postmaster/SLN3434.html (in reply to MAIL FROM command))` which lead me to believe this is a bounce message being sent. The domain yahoo.com.tw is not in my configuration and I'm certainly not trying to send mails there.

ofthesun9:
As described in #583, Mailu bounces email when recipient is unknown.
This PR add a variable (in .env) to choose whether to reject or to bounce.

If the variable is not set, the current Mailu behaviour is kept (equivalent to set it to "no")
To REJECT mail, the .env file has to set the variable to yes

### DIFF
--- a/core/postfix/conf/master.cf
+++ b/core/postfix/conf/master.cf
@@ -8,6 +8,7 @@ smtp      inet  n       -       n       -       -       smtpd
 10025     inet  n       -       n       -       -       smtpd
   -o smtpd_sasl_auth_enable=yes
   -o smtpd_recipient_restrictions=reject_unlisted_sender,reject_authenticated_sender_login_mismatch,permit
+  -o smtpd_reject_unlisted_recipient={{ REJECT_UNLISTED_RECIPIENT }}
   -o cleanup_service_name=outclean
 outclean  unix n       -       n       -       0       cleanup
   -o header_checks=pcre:/etc/postfix/outclean_header_filter.cf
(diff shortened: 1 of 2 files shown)


muhlemmer:
Perhaps it's wise to indicate in `.env` that `no` / bounce is the default?

Issue #627, opened 2018-10-05
[Test suite] - Rspamd fails occasionaly


muhlemmer:
cc: @ionutfilip

While developing the test suite, we had random failures on the `anitspam` service. It would usually end in:
```
2018-10-05 12:09:48 #7(controller) <3nxzfe>; monitored; rspamd_monitored_dns_cb: DNS query blocked on multi.uribl.com (127.0.0.1 returned), possibly due to high volume
```

This is most probably related to the freshclam issues I've been seeing (#625). It seems this is fixed in the Alpine git repository, only the `alpine:edge` image has not been rebuild yet. We've been able to run our test suite successful a number of times without errors on `rspamd`.

I'm opening this issue as other users that are using `mailu:master` might find themselves in this situation.

Links to the travis build logs:
- `rspamd` fail: https://travis-ci.org/usrpro/Mailu/builds/437588016#L3883
- `rspamd` success: https://travis-ci.org/usrpro/Mailu/builds/437594682#L4062
**Note:** Build still fails on fetchmail, which we are investigating.

The issues seems to originate from `musl` which is the standard C library used in Alpine and should be fixed in the following [commit](https://git.alpinelinux.org/cgit/aports/log/main/musl):
```
* | main/musl: backport fix for getaddrinfo regression | Natanael Copa | 2018-09-20 | 2 | -1/+55
```

Rebuild of `alpine:edge` had been requested in gliderlabs/docker-alpine#447

ofthesun9:
"DNS query blocked on multi.uribl.com" could be also related to:
https://rspamd.com/doc/faq.html#what-is-the-meaning-of-the-uriblblocked-symbol
see #385

muhlemmer:
We are also seeing segfaults and other kinds of failures when on `alpine:edge`. I don't want to bother too much in debugging, since on `alpine:3.8` everything seems fine. We've done a number of forced re-tests at Travis and it seems okay.

So hopefully a simple rebuild will help.

PR #612, opened 2018-09-27
Abstract db access from Postfix and Dovecot

kaiyou:
This is a work in progress but I guess the review can start already as 90% of it is testable.

Most of the original work is discussed in #420 and #445. The general idea is that  

- Podop (https://github.com/mailu/podop) is a proxy framework between Postfix/Dovecot and any Python available service, including notably a REST API
- A fairly simple (and still a bit too tightly dependent on Postfix/Dovecot structures) API was written as part of the internal REST API already leveraged for authentication, the API includes transparent authentication for Dovecot, domain, alias, mailbox and transport resolution for Postfix, quotas, sieve scripts.
- A simple sieve script as a Jinja template was written, removing the need for extdata in sieve.
- Dovecot and Postfix were bound to the admin container API using Podop, embedded as a side process in both start scripts.
- Fetchmail was bound to a simple internal API on the admin container.

No change should be required in a standard compose environment. There are still a lot of tests to run before declaring this equivalent to the current database queries.

Issue #412, opened 2018-03-14
Fetchmail error


adi90x:
Hello,
On a newly installation, fetchmail container fail because fetch table seems to be missing... 

Is it a known error ?

kaiyou:
It is not, and the fetch table should be created by the admin container on first launch. Did you check the sqlite?

adi90x:
Hello,
Yeah I guess it is link to the NFS share I am trying to use, seems that
sqlite3 doesn't like that at all and is doing random things !!

I will try to check whether I can move missing part to admin container or
maybe find a workaround !

kaiyou:
I read locking is pretty bad on NFS indeed, so it might explain some erratic behaviors. I'll create a meta issue to keep track of what's missing to switch to other databases.

adi90x:
Would be really nice to have a way to follow what's missing

Aksem:
I have the same problem(tried both 1.5 and master):
`fetchmail_1  | sqlite3.OperationalError: no such table: fetch`

Also tried reinstall and restart, no effect.
Any ideas?

kaiyou:
Aksem, does your Mailu instance run properly? Either there were issues when first initializing your database, or you do not have an initialized database yet.

kaiyou:
Closing as outdated

muhlemmer:
@ionutfilip

muhlemmer:
Hi, this problem also appears in our testing suite on travis:
https://travis-ci.org/usrpro/Mailu/builds/437588016#L3927

However, we've created `usrpro:debugging` branch which combines our testing suite with `feat-abstract-db` and the test finishes successfully. Most probably because fetchmail no longer tries to access the [sqlite database directly.](https://github.com/Mailu/Mailu/commit/f9c6c98180fb34a1628871610babe89ab8094b4a#diff-9fa615993b76409e9c6b2a13396fcbaa):
https://travis-ci.org/usrpro/Mailu/builds/437645741

I think we can safely conclude that the "old" setup from master imposes a race condition on the sqlite database. The race is exposed on systems where I/O throughput is limited. (So NFS, vitrual machines like travis). In my humble opinion this issue should be considered open until we are able to merge `feat-abstract-database`.

PR #628, opened 2018-10-05
Feat test suite v2

muhlemmer:
This PR has the next phase of the testing suite implemented. Features added:
1. Running of containers after build and testing for their actual existence after a 1 minute timeout
2. Auto deploy / upload of built images to docker hub upon successful tests. The deploy section is not executed on Pull Requests, only on commits to actual branches.

At this moment only the default / _core_ containers are included in the test. This scripts are written as such that new test configurations can be easily added by creating *.env files. We are publishing the PR now, so that it can be preliminary reviewed if this course of action is okay.

## Travis configuration
@kaiyou: For this to work, the following variables will need to be set at Travis (before merging to master):
 - DOCKER_ORG	first part of docker repo name (example: mailu)
 - DOCKER_UN	Docker-hub login username (private)
 - DOCKER_PW	Docker-hub login password (private)
**Note:** Docker hub autobuilds will have to be disabled.

## Issues
The following issues appeared in the process of testing our testing suite. These issues belong to other places, but prevents a successful Travis build for this PR/branch. These assumptions are tested and [confirmed](https://travis-ci.org/usrpro/Mailu/builds/437645741) in our [`debugging`](https://github.com/usrpro/Mailu/tree/debugging) branch.
- [ ] Rspamd issue #627 depends on issue gliderlabs/docker-alpine#447
- [ ] Fetchmail issue #412 , which will be solved by merging #612.

## ~Pending improvement~
- [x] ~For this phase to be completed, we will implement usrpro/Mailu#7 on Monday.~


muhlemmer:
Just for information, I've created a Github project which we are using internally at usrpro for ideas and management of the test suite:
https://github.com/usrpro/Mailu/projects/1

muhlemmer:
Did some alterations today.
- The Travis variables are now optional. `DOCKER_ORG` defaults to `mailu` and if `DOCKER_UN` is not set, deploy is skipped.
- Fetchmail is excluded in run verifications
- Cleanup with `docker-compose down`

muhlemmer:
In principle this is ready to merge. If there are failures on `rspamd`, just push `restart build` in Travis.

muhlemmer:
>  muhlemmer dismissed @kaiyou 's stale review via 1bbb86e 44 minutes ago
Not willingly. Maybe because of the new commits.

tests/compose/test-script.sh
```diff
@@ -0,0 +1,65 @@
...
+	antispam
+	admin
+	redis
+	antivirus
+	webdav
+#	fetchmail
+	front
+)
+
+# Verbose sleep, to prevent Travis to cancel the build
```
HorayNarea: Travis builds are only canceled after 10 minutes of log-silence¹, having a function that logs (at the current config) once _before_ sleeping for one minute seems unneccessary to me… am I missing something?

¹ from https://docs.travis-ci.com/user/common-build-problems/#my-builds-are-timing-out :
> Our builds have a global timeout and a timeout that’s based on the output. If no output is received from a build for 10 minutes, it’s assumed to have stalled for unknown reasons and is subsequently killed.

muhlemmer: At this moment the sleep time is only 1 minute, so it is easier to test the test script. Otherwise the wait would be too long. When we are are happy about the results, the sleep times will need to be increased to cover all sorts of cases:
1. ClamAV to be finished with pulling and updates the virus database
2. In case of Swarm, `docker stack deploy` returns immediately, but it can take significant time for the scheduler to pull in all the images and start containers.
3. Some crashes appear only after a few minutes after start. (ClamAV is doing this when it cannot download the virus database)

So this value can easily extent 10 minutes in the future.

`travis_wait sleep ..m` indeed is a cleaner solution. I will implement this.

muhlemmer: So the implemation is done with one note. We also use the script for local testing. This means I had to put a fallback sleep for the cases the script is not run on travis.

Issue #546, opened 2018-08-02
Performance Problem


t-t-yano:
I have tested a performance of Mailu system.

Following Python code is a performance test which measures how long does it takes on a POP3 over SSL connections. The test program connects a mail server with POP3 over SSL protocol, issues user authentication commands, and logouts simultaneously 300 times with threads.

It takes just 5-6 seconds to process this test on a Postfix server installed in a non-virtual CentOS server with a UNIX user account, but it takes over 30 seconds on a Mailu server.

Please give me some advice to make Mailu faster.

```
#!/usr/bin/python3

ACCOUNT   = 'testuser'
PASSWORD  = 'testpassword'
HOST      = 'test_host'

import poplib
from threading import Thread
import timeit 

def workload():
    p = poplib.POP3_SSL(HOST)
    p.user(ACCOUNT)
    p.pass_(PASSWORD)
    p.quit()

def threading(n):
    start = timeit.default_timer()
    thread_pool = []
    for _ in range(n):
        t = Thread(target=workload, args=())
        t.start()
        thread_pool.append(t)
    for t in thread_pool:
        t.join()
    return (timeit.default_timer() - start)

def main():
    print( threading(300) )

if __name__ == '__main__':
    main()

```

kolinger:
This can be caused because of hashing algorithm used for passwords. It is designed to be slow to make brute force attack against hash very hard or impossible. Unfortunately if your CPU is not fast enough then this takes too long and will make everything slow or even unusable. Hash is calculated on every new connection.

You can try change hash in configuration. But this will make old passwords invalid and also compromises security. You can use "PASSWORD_SCHEME=MD5-CRYPT" for testing. This will allow you to check of your CPU (or the hashing scheme, depends on your view) is the bottleneck. But MD5 is not secure. It can't be used for production. 

Configuration is not giving really good way to configure this. You don't want to change hashing algorithm but interation count instead - to make hashing faster but not insanely fast and insecure as md5. You will need to rewrite /core/admin/mailu/models.py for this and change configuration for passlib context.

jkaberg:
I agreed that this is a real issue, browsing the various IMAP folders remotely is very slow currently (I have validated that the amount of rounds is my main issue).

jkaberg:
Btw, https://pthree.org/2018/05/23/do-not-use-sha256crypt-sha512crypt-theyre-dangerous/

kaiyou:
[...]

PR #635, opened 2018-10-07
support PBKDF2 and use it as default

HorayNarea:
PBKDF2-SHA512 is as secure as SHA512-CRYPT but works in constant-time based on the rounds, not the length of the password - eliminating the leakage of password-length based on timing

this should also reduce the performance impact as reported by #546
(in my local tests with the script by @t-t-yano PBKDF2-SHA512 is 2-3x faster than SHA512-CRYPT)


HorayNarea:
I created an alternative PR #647 that just works so I will close this one now.

kaiyou:
The password is not passed to Dovecot. What is weird is this change worked pretty well for me. Let's switch to the alternative anyway.

PR #640, opened 2018-10-09
Adding unbound as dns resolver Pt. 2 [WIP]

muhlemmer:
This is subsequent work on PR #385, originally submitted by @obi12341. Mainly some cleanup, merge to latest master and DNS testing. This is still a work in progress and this PR should facilitate discussion.

## Docker compose ##
In a regular docker-compose setup, we cannot avoid using static IP. I don't know if this will break anything? Users will not be able to use `docker-compose up -d --scale unbound=x`. (Don't know if other Mailu components are capable of doing that, never tried).

## Docker swarm ##
It is not (yet) possible in Docker swarm to use a static VIP for a service. (moby/moby#24170) Replicated services rely on the service name to be resolved, which will not work as a DNS entry. This means also that `unbound` cannot be replicated, unless multiple serve entries are made.

I've tried exposing the DNS server to the host network and set the DNS entry to the host IP, but docker does not allow this kind of connections between containers, it has to by bridge or overlay network.

## Testing ##
The following things are tested:
1. Resolving of external hostnames
2. Resolving of Docker hostnames.

No mail testing has been conducted as of yet. First I want to discuss this approach.

### DIFF
--- a/core/unbound/Dockerfile
+++ b/core/unbound/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:edge
+FROM alpine:3.8
 
 RUN apk add --no-cache unbound curl \
 	&& curl -o /etc/unbound/root.hints https://www.internic.net/domain/named.cache \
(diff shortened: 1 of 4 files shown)


muhlemmer:
Closing. We will integrate this in setup as an optional.

Issue #625, opened 2018-10-03
Antivirus container crashes and restart all the time


muhlemmer:
Logs are flooded with entries like:
```
antivirus_1  | Wed Oct  3 15:16:22 2018 -> !TCP: getaddrinfo failed: System error
antivirus_1  | Wed Oct  3 15:16:53 2018 -> !TCP: getaddrinfo failed: System error
antivirus_1  | Wed Oct  3 15:17:23 2018 -> !TCP: getaddrinfo failed: System error
```
After a few minutes the container fails with exit code 1 and gets restarted by docker.

kaiyou:
At first I thought it would be related to recent changes but does not look like it. Does it output anything before that?

I will try and reproduce tonight.

muhlemmer:
So, this bug exposes when upgrading `1.5`->`master`. So it is not related to `feat-abstract-db`. In this situation this are the only entries in the log:

```
antivirus_1  | Thu Oct  4 08:33:05 2018 -> !TCP: getaddrinfo failed: System error
mailu_antivirus_1 exited with code 1
```

However, when doing a fresh install of `master` the logs become a bit more expressive. This seems to have to do whether or not the virus database is written to `/data`:

```
antivirus_1  | Thu Oct  4 08:39:02 2018 -> ClamAV update process started at Thu Oct  4 08:39:02 2018
antivirus_1  | Thu Oct  4 08:39:02 2018 -> ^Your ClamAV installation is OUTDATED!
antivirus_1  | Thu Oct  4 08:39:02 2018 -> ^Local version: 0.100.1 Recommended version: 0.100.2
antivirus_1  | Thu Oct  4 08:39:02 2018 -> DON'T PANIC! Read https://www.clamav.net/documents/upgrading-clamav
antivirus_1  | Thu Oct  4 08:39:02 2018 -> ^Can't get information about database.clamav.net: System error
antivirus_1  | Thu Oct  4 08:39:02 2018 -> ^Can't download main.cvd from database.clamav.net
antivirus_1  | Thu Oct  4 08:39:02 2018 -> Trying again in 5 secs...
antivirus_1  | Thu Oct  4 08:39:07 2018 -> ClamAV update process started at Thu Oct  4 08:39:07 2018
antivirus_1  | Thu Oct  4 08:39:07 2018 -> ^Your ClamAV installation is OUTDATED!
[...]

PR #645, opened 2018-10-10
Added some test cases to the test suite [BLOCKED]

muhlemmer:
This add the following _run_ tests to the test suite:
- antivirus
- webdav
- rainloop
- roundcube

Note that the tests are failing due to #625, #627. I'll update this branch once PR #644 is merged.


muhlemmer:
`tavis_wait` is only available when calling it directly from `travis.yml`. In the script environment [I get](https://travis-ci.org/Mailu/Mailu/builds/440081868#L6001):
```
tests/compose/test-script.sh: line 66: travis_wait: command not found
```
Hence that commit was reverted and `v_sleep` re-implemented. The time is now set to 10 minutes, as 5 minutes was too short for ClamAV to download the virus DB.

I will look into test-specific wait times as a future improvement to make things go a bit more fast.

PR #649, opened 2018-10-11
reduce webmail image-layers/sizes

HorayNarea:
it's logically the same as before, just less disk-space

### DIFF
--- a/webmails/rainloop/Dockerfile
+++ b/webmails/rainloop/Dockerfile
@@ -1,20 +1,21 @@
 FROM php:7.2-apache
 
-RUN apt-get update && apt-get install -y \
-      unzip python3 python3-jinja2
-
 ENV RAINLOOP_URL https://github.com/RainLoop/rainloop-webmail/releases/download/v1.12.1/rainloop-community-1.12.1.zip
 
-RUN rm -rf /var/www/html/ \
+RUN apt-get update && apt-get install -y \
+      unzip python3 python3-jinja2 \
+ && rm -rf /var/www/html/ \
  && mkdir /var/www/html \
  && cd /var/www/html \
  && curl -L -O ${RAINLOOP_URL} \
- && unzip *.zip \
+ && unzip -q *.zip \
  && rm -f *.zip \
  && rm -rf data/ \
  && find . -type d -exec chmod 755 {} \; \
  && find . -type f -exec chmod 644 {} \; \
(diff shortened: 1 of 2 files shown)


webmails/rainloop/Dockerfile
```diff
@@ -1,20 +1,21 @@
...
- && unzip *.zip \
+ && unzip -q *.zip \
  && rm -f *.zip \
  && rm -rf data/ \
  && find . -type d -exec chmod 755 {} \; \
  && find . -type f -exec chmod 644 {} \; \
- && chown -R www-data: *
+ && chown -R www-data: * \
+ && apt-get purge -y unzip \
+ && rm -rf /var/lib/apt/lists
```
curry684: If you want to make it as lean as possible:
`    && rm -rf /tmp/* /usr/share/doc/* /usr/share/man/* /var/lib/apt/lists/*`

webmails/roundcube/Dockerfile
```diff
@@ -17,7 +15,8 @@ RUN rm -rf /var/www/html/ \
  && cd html \
  && rm -rf CHANGELOG INSTALL LICENSE README.md UPGRADING composer.json-dist installer \
  && sed -i 's,mod_php5.c,mod_php7.c,g' .htaccess \
- && chown -R www-data: logs temp
+ && chown -R www-data: logs temp \
+ && rm -rf /var/lib/apt/lists
```
curry684: `    && rm -rf /tmp/* /usr/share/doc/* /usr/share/man/* /var/lib/apt/lists/*`

webmails/rainloop/Dockerfile
HorayNarea: In theory you are right but in practice man-pages don't get installed in debian-slim and the docs are only ~200KB

curry684: Like I said, no major difference, but if you're optimized for size why leave 200KB around?   😉 

Also in my experience `/tmp` can be a nasty one at times so I habitually clean all these folders up at the end of a `RUN`.

kaiyou:
It looks okay, ill try to build and approve later today

webmails/rainloop/Dockerfile
kaiyou: This could be further improved but currently none of these weighs more than a few kb.

Issue #584, opened 2018-09-04
Access logfile for Fail2ban


bilyboy785:
Can we mount a volume for logs to use Fail2Ban ?

v1ru535:
Same here, so I extract some lines from the docker-compose follow to be able to do the bans afterwards.
If one or more logs (front,smtp...) can be set up it would be great.
Thx

bilyboy785:
Nice idea @v1ru535  !! 
Waiting for this feature <3

oz123:
It turns out it's not only about `ufw` but some modifications on docker networking are needed:

https://askubuntu.com/questions/652556/uncomplicated-firewall-ufw-is-not-blocking-anything-when-using-docker

v1ru535:
Personally I extract some strings from the docker logs and put everything in a log.

After that I extract IPs and null routing using ip command like that "ip route add blackhole <IP>".

It's waiting to do better naturally, but with that I already have over 2000 banned IPs.

Hope this helps you

oz123:
I found a useful post about the issue:

http://blog.amigapallo.org/2016/04/14/configuring-fail2ban-and-iptables-to-get-along-with-docker/

muhlemmer:
See above PR. I was annoyed I couldn't write something useful for the FAQ, so wrote an actual solution.

I'm looking into suitable regex for documentation, but that is not my field of sport. So help would be appreciated. The standard Nginx regex supplied by Fail2Ban don't catch the login fails for Mailu.

muhlemmer:
Attempt 2, using Docker logging drivers. Please have a look.

muhlemmer:
Implemented in #661, we can close this.

v1ru535:
Thanks for this!

I went from version 1.5 to master to get these options, then added the necessary modifications on env and docker-compose.yml.
I added a jail and a rule to ban IPs on this pattern for fail2ban:
`client login failed: "Authentication credentials invalid" while in http auth state, client: [IP], server: 0.0.0.0:25, login: "someshit@domain.io"`
I filter on "client login failed" and get the IP, then all you have to do is adjust the number of failed login before the ban in fail2ban rule.
Now it works like a charm, thanks for the work.

_If you wish I can provide the jail for the faq._

hoellen:
It would be nice if you could share the complete fail2ban rules here, so we can add it to the FAQ for other users.

v1ru535:
For my environment, I put the log to syslog, so in env:
[...]

PR #654, opened 2018-10-14
Access to Nginx log for Fail2Ban

muhlemmer:
This implements Nginx logging to files. It uses logrotate to rotate weekly. Also, `docker-compose.yml` edited to mount the logs to the host in `/mailu/log`.

This should take care of the request in #584.

### DIFF
--- a/core/nginx/Dockerfile
+++ b/core/nginx/Dockerfile
@@ -7,6 +7,7 @@ RUN apk add --no-cache certbot logrotate nginx nginx-mod-mail openssl \
 
 COPY conf /conf
 COPY *.py /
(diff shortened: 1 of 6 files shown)


kaiyou:
I am still really uncomfortable with exporting the logs from the container. Isn't Docker supposed to handle logs properly, and in a fashion that fail2ban should be able to parse?

muhlemmer:
Option2: Use the docker [logging driver](https://docs.docker.com/config/containers/logging/configure/) for the export. If that is more preferred, I will close this and start investigating that solution. It will probably just end up being a compose file entry.

PR #655, opened 2018-10-14
Create README.md for kubernetes

briantopping:
I spent a lot of time working on `stable` before realizing it's, well, not stable...

### DIFF
--- a/docs/kubernetes/README.md
+++ b/docs/kubernetes/README.md
@@ -1,0 +1,3 @@
+# Kubernetes
+
+nb: The `stable` branch is deprecated, use 1.6!


briantopping:
I think it would be better to just delete the folder called `stable` and move the contents of the `1.6` folder to that root. It already has a very good `README.md`. WDYT?

muhlemmer:
Putting @hacor and @consideRatio I'm CC. They are involved in Kubernetes. It is probably better to discuss with them.

I just did the review of this PR ;)

briantopping:
Great, thanks! I'm not partial to a solution or doing the commit, just trying to reflect back with what I see and how I can help the next person be delighted with the experience!

hacor:
@briantopping Indeed the current `stable` version of kubernetes is not yet useable for production. I think it's better to wait for PR #576 to finish before using Mailu on Kubernetes.
Currently we're all reviewing/testing the current PR to be as stable as possible before publishing

briantopping:
Thank you @hacor, I will join you on #576!

PR #657, opened 2018-10-15
Added config XCLIENT_HOSTS

kwek:
With this option, we can run mailu in docker swarm (with the dnsrr workaround) with proper failover. 

When a docker host goes offline in docker swarm, the services get deployed on another server. With this change, the smtp service will still accept xclient commands from the new front host. 

This is a duplicate of https://github.com/Mailu/Mailu/pull/563 which has awaiting fixes. If this is pulled, that one can be closed.


muhlemmer:
#563 is not awaiting fixes, it is awaiting review and testing. Please note that after extensive discussion we have chosen to use `POD_ADDRESS_RANGE` for this purpose.

You can help doing actual tests / review on PR #563 and PR #604.

I will have to close this PS being a duplicate.

PR #661, opened 2018-10-16
Option to send front logs to journald or syslog

muhlemmer:
Supersedes PR #654. Implements request from issue #584.

This provides an option to let the Docker deamon send the logs from front to an alternative driver. In principle all [docker log drivers](https://docs.docker.com/config/containers/logging/configure/#supported-logging-drivers) can be set. However, `.env` only mentions the most applicable ones:
1. `json-file`: Docker's default driver, used for `docker-compose logs`
2. `journald`: Send to systemd
3. `syslog`: For non-systemd, breaks `docker-compose logs`.

When setting to `journald` logs can be checked using:
```
journalctl -b CONTAINER_NAME=mailu_front_1
```
On Fedora Fail2Ban is integrated with `journald`. So it would be a matter of finding the right regex to make this work. Make sure not to blacklist any 172.x.x.x addresses or whatever local range you are using.

I have not tested `syslog` myself, as I only use systemd on all my systems.

### DIFF
--- a/docs/compose/.env
+++ b/docs/compose/.env
@@ -120,6 +120,12 @@ WEBSITE=https://mailu.io
 # Advanced settings
 ###################################
 
+# Log driver for front service. Possible values:
+# json-file (default)
+# journald (On systemd platforms, useful for Fail2Ban integration)
+# syslog (Non systemd platforms, Fail2Ban integration. Disables `docker-compose log` for front!)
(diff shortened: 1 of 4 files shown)


kaiyou:
This looks a lot more docker friendly and I like it much more. Also, it just works, so approved :)

PR #664, opened 2018-10-16
Implement a multi-version documentation

kaiyou:
This is a follow up regarding #601. It builds documentation for various branches at different paths, either using a build arg or environment variables from the Docker hub.

It also implements the bottom-left widget for picking a version and the warning banner when browsing a non-stable version.

I still have to test with Docker hub, and maybe write some documentation for maintainers.


muhlemmer:
Yeah, go for it!

PR #667, opened 2018-10-17
Improve password checking performance

kaiyou:
After @HorayNarea closed #635 because Dovecot was using the password for a second-time check, I re-implemented the IP check that removes the need for double password checking.

Now, at the cost of a couple dns queries, Dovecot does not check the password anymore, which helps in two ways:
- it divides the checking time by a factor of 2
- it allows PBKDF2, which is significantly faster (factor of 10 at least).

Finally, I added the ability for the user password to be dynamically updated upon login if the database version is in an older format. This way, people can now switch to PBKDF2 and quickly see a decent performance improvement.

PR #668, opened 2018-10-18
Feat setup

ionutfilip:
#### We changed the following:
- setup/Dockerfile
 - instead of pulling from git it will copy the files into the container
- setup/flavors/compose/docker-compose.yml
  - upgrade from version 2 to 3.6. Note that version 3.6 was the most recent version that worked
- setup/server.py
  - hardcoded "master" as the only available version
- setup/templates/steps
  - moved compose flavor related file under the compose dir
- extended documentation in download page

#### Removed the following variables from .env file and used them directly on docker-compose.yml file:
- $ROOT
- $TLS_FLAVOR
- $ADMIN
- $PUBLIC_IPV4/6 (bind4/6)
*Note that there are only displayed on the env file.*

#### Variables which are not asked in the setup page (but exists in the env file)
- message_size_limit (defalut : 50000000)
- fetchmail_delay (def: 600)
- recipient_delimiter (def: +)
- welcome_enabled
- recaptcha_public_key
- recaptcha_private_key
- password_scheme (def: SHA512-CRYPT)
- real_ip_header
- real_ip-from
- compression
- compression_level
- reject_unlisted_recipient

#### Questions:
- Authentication rate limit is mandatory? For now we set default values, but how it's supposed to behave if you don't set them? (Also there is a constraint for entering only numerical values and a check for positive numbers)
- On setup page there some constraints added ( regex pattern ). For now, root path should follow the /dir/subdir type. It won't work for windows path. Any suggestions are welcome


muhlemmer:
Todo from my side:
- [ ] Upgrade the env file after #667 is merged
- [ ] Re-arrange the order of entries in the env file to be consistent with the order in the setup page

setup/server.py
```diff
@@ -68,7 +68,8 @@ def wizard():
         @bp.route("/submit_flavor", methods=["POST"])
         def submit_flavor():
             data = flask.request.form.copy()
-            return flask.render_template('wizard.html', flavor=data["flavor"])
+            flavor_files = sorted(os.listdir(path + "/" + version + "/templates/steps/" + data["flavor"]))
+            return flask.render_template('wizard.html', flavor=data["flavor"], flavor_files=flavor_files)
```
muhlemmer: @kaiyou, we are trying to pass two variables down to the template here.The `flavor` and a list containing the steps for that flavor: `flavor_files`. Is there a more elegant way we could do this?

muhlemmer: @ionutfilip, lets rename `flavor_files` to `steps`.

kaiyou: I do not believe there is a more elegant way, it is already quite acceptable :)

kaiyou:
I just tested the wizard and looks good. Really good. I will comment in more details in the code.

setup/flavors/compose/setup.html
```diff
@@ -30,7 +30,20 @@
...
+Before you can use Mailu, you must create the primary administrator user account. This should be {{ postmaster }}@{{ domain }}. Use the following command, changing PASSWORD to your liking:
+
+<pre><code>docker-compose exec admin python manage.py admin {{ postmaster }} {{ domain }} PASSWORD
+</pre></code>
+
+Login the the admin interface to change the password for a safe one, at
+{% if admin_enabled %}
+one of the hostnames ({{ hostnames }}){{ admin_path }}.
+{% else %}
+http://127.0.0.1:8080 (only directly from the host running docker).
```
kaiyou: Because we know the hostnames, we can actually generate a proper URI here.

setup/templates/steps/compose/01_config.html
```diff
@@ -0,0 +1,67 @@
...
+<p>Before starting some variables must be set</p>
+
+<div class="form-group">
+  <label>Root path: </label>
+<!--   Validates path -->
+  <input class="form-control" type="text" name="root" value="/mailu" required pattern="^/[-_A-Za-z0-9]+(/[-_A-Za-z0-9]*)*">
+</div>
+
+<div class="form-group">
+  <label>Main mail domain</label>
```
kaiyou: We should probably write some details about what the main mail domain is. Maybe even stop calling it that name: it is the server display name, maybe.

setup/templates/steps/compose/01_config.html
```diff
@@ -0,0 +1,67 @@
...
+<!--   Validates domain name -->
+  <input class="form-control" type="text" name="domain" placeholder="e.g. mailu.io" 
+  		required pattern="^([a-z0-9]+(-[a-z0-9]+)*\.)+[a-z]{2,}$">
+</div>
+
+<div class="form-check form-check-inline">
```
kaiyou: I do not think this belongs here. It is a very optional and quite experimental feature, far from the basic settings from this first section.

setup/templates/steps/compose/01_config.html
```diff
@@ -0,0 +1,67 @@
...
+  </label>
+</div>
+
+<div class="form-group">
+  <label>Postmaster local part (will append the main mail domain)</label>
+  <input class="form-control" type="text" name="postmaster" value="admin" required>
+</div>
+
+<div class="form-group">
+  <label>Choose how secure connections will behave </label>
```
kaiyou: Definitely not agreeing with the label. Except for ``notls`` (which we ought to have a discussion about), the security level does not really change when changing this. Maybe "Choose how you wish to handle security (TLS) certificates" and add a link to the documentation for more details.

kaiyou:
More generally, I am really willing to just acknowledge progress on this matter and merge once it is stable, publish it and wait for feedback from the community. Except for the vocal ones, my comments are mostly cosmetic and should not prevent us from merging and show some progress toward 1.6 :)

setup/flavors/compose/docker-compose.yml
```diff
@@ -11,41 +12,39 @@ services:
...
   front:
     image: mailu/nginx:{{ version }}
     restart: always
     env_file: {{ env }}
-    env:
-      - TLS_FLAVOR={{ tls_flavor or 'letsencrypt' }}
-      - ADMIN={{ expose_admin or 'no' }}
+    environment:
+      - TLS_FLAVOR={{ tls_flavor }}
+      - ADMIN={{ admin_enabled or 'false' }}
```
muhlemmer: @kaiyou these variables where mentioned here from the original template. Is there any good reason for doing so? I believe the should go back into the `mailu.env` file. We also will need those for all the other flavors.

muhlemmer:
So, we are done with this for the weekend. We have done the cosmetic updates, some finalization of the docker stack flavor, some testing etc.

There are still things that can be improved, maybe functionality added. Stack probably needs more testing and fine tuning.

However, for now I want to propose to merge this PR. So that others can also test and update the code where needed.

Issue #662, opened 2018-10-16
Admin not accesible after setting to True (301)


muhlemmer:
**Browsers:** Firefox, Chromium
I found on my testing server, when `ADMIN=false` and browsing to domain.com/admin, or any other address, the browser is redirected to /webmail.

When later setting `ADMIN=true`, and `docker-compose down / up`, the browser is still redirecting. However, opening an "unaffected" browser window, admin console shows up.

Using the debug window, it seems the browser has a cached `301: moved permanently`. This cannot be cleared by a <kbd>CTRL + F5</kbd>, which normally flushes web caches.

**Firefox**
![image](https://user-images.githubusercontent.com/5411563/47016884-8e084000-d159-11e8-980a-52343e16d483.png)

**Chromium**
![image](https://user-images.githubusercontent.com/5411563/47017740-b2651c00-d15b-11e8-9274-35696924208d.png)

1. Should be really be redirecting all request paths to webmail? Maybe only request to https://domain.com/. Perhaps is better just to send a  404 in case the address in invalid
2. If we are re-directing admin, perhaps it is better to use a 307 (Temporary redirect?)

**Workaround:**
![image](https://user-images.githubusercontent.com/5411563/47017128-20a8df00-d15a-11e8-891a-d17c0a616a62.png)

muhlemmer:
I want to advocate removing the redirects altogether. A wrong URL should just give a 404. This is just a matter of taste I guess.

HorayNarea:
I agree with @muhlemmer, we should not try to guess what a user maybe could want and force them to go there :grinning:

Instead mabe have a 404 with a list of possible pages? For example:
```
Sorry we didn't find that page, did you mean any of those:
{{ admin_page_link }}
{{ webmail_link }}
```

muhlemmer:
I wanted to create a nice 404 page logo. But inkscape keeps crashing on me.

PR #669, opened 2018-10-18
Fix nginx conf if webmail is on root path

hoellen:
This removes the rewrites in the nginx config, if the webmail is running on root path (e.g. mail.example.com) because it breaks the request.


muhlemmer:
In principle this works (and I'll approve it). But I want to advocate removing the redirects altogether. See issue #662.

PR #670, opened 2018-10-18
Refactor the admin architecture and configuration management

kaiyou:
Until now, the application was entirely generated by the init script, and configuration was only provided as environment variables.

This PR suggests generating the application in a function (namely ``create_app``) and writing everything else as Flask extensions. This clarifies a lot of code inside the admin container, and makes some room for further improvement.

Among those improvements, the configuration is replaced in this version with a simple class that mimics the behavior of previous Flask configuration updated from env variables. It should evolve to the point where configuration can be modified from the Web interface, at least for items that do not impact the other containers.

Currently this builds and run but I still have an issue with the way Flask login interacts with Flask SQLAlchemy, as it does not seem to be using the application context properly, and thus fails when requesting SQLite which is not thread-safe.


muhlemmer:
Just a suggestion. At this moment we are using `manage.py` to create the first user account. The first admin user should be equal to `$POSTMASTER`.

We could use the setup utility to generate a password, store it encrypted in `mailu.env` as `$TEMP_PASS` and inform the user on the final page. Then, when admin is started and doesn't find the first user, insert `$POSTMASTER` with `$TEMP_PASS`. User can now login using the password and change it.

This way, there can be no mismatch in `$POSTMASTER` and the chosen username and the over-all complexity of installing Mailu is less.

kaiyou:
Sorry it took me so long to finish this up, I was struggling with sqlalchemy contexts when mixed with flask login. This is now ready for merge I believe.

muhlemmer:
This seems to be chrom(ium) specific... When sending the login from chromium no redirect:

```
admin_1      | 79.115.207.234 - - [05/Dec/2018:20:56:28 +0000] "POST /ui/login?next=ui.index HTTP/1.0" 200 4395 "https://mail.usrpro.io/admin/ui/login?next=ui.index" "Mozilla/5.0 (X11; Fedora; Fedora; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/70.0.3538.110 Safari/537.36"
```

From firefox:
```
admin_1      | 79.115.207.234 - - [05/Dec/2018:20:56:51 +0000] "POST /ui/login?next=ui.index HTTP/1.0" 302 227 "https://mail.usrpro.io/admin/ui/login?next=ui.index" "Mozilla/5.0 (X11; Fedora; Linux x86_64; rv:63.0) Gecko/20100101 Firefox/63.0"
admin_1      | 79.115.207.234 - - [05/Dec/2018:20:56:51 +0000] "GET /ui/ HTTP/1.0" 302 253 "https://mail.usrpro.io/admin/ui/login?next=ui.index" "Mozilla/5.0 (X11; Fedora; Linux x86_64; rv:63.0) Gecko/20100101 Firefox/63.0"
admin_1      | 79.115.207.234 - - [05/Dec/2018:20:56:51 +0000] "GET /ui/user/settings HTTP/1.0" 200 7533 "https://mail.usrpro.io/admin/ui/login?next=ui.index" "Mozilla/5.0 (X11; Fedora; Linux x86_64; rv:63.0) Gecko/20100101 Firefox/63.0"
```
To be continued...

muhlemmer:
After doing some `docker-compose down`, database deletion and `docker-compose up -d` the problem seems to have disappeared. I think/hope that the issue was caused because of all the migrations between branches and probably a confusion in session management caused this. My test server does get confused from time to time :wink: 

I will drop by blocking review and go back to approving it.

Issue #564, opened 2018-08-09
Creating a FAQ


muhlemmer:
Just a suggestion, there are a number of issues open marked as `question`. It clobbers the issues list with items that are not really important.

We could transfer them to a FAQ page on the website, mailu.io and close them off here once answered and migrated.

kaiyou:
Agreed, we just need a simple documentation page for these, which is as simple as creating an md file. Then we need to list interesting issues that need integration in the faq.

muhlemmer:
So by creating a md file, you mean it will be Github hosted and not included in the documentation site? (Which is .rst file based).

As for nominating: on the issue which would be interesting for inclusion, just mention this issue. This can then work as a meta-issue. People can then collaborate to do the conversion. It can be helpful to reopen nominated issues until inclusion is done.

We can work of this query link: https://github.com/Mailu/Mailu/issues?utf8=%E2%9C%93&q=is%3Aissue+label%3Aquestion. The link itself might also be included in the FAQ for reference.

kaiyou:
Ok for mentioning this issue. And I meant a rst file indeed, for proper inclusion in the docs and website :)

khannurien:
Is there a public repo where we can contribute to this FAQ?

muhlemmer:
It is in my intention to provide a FAQ page as scaffolding in the near future, here in the GitHub repository. It will be a .rst document as part of the main documentation package.


On September 29, 2018 12:18:47 PM GMT+02:00, Vincent <notifications@github.com> wrote:
>Is there a public repo where we can contribute to this FAQ?
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/Mailu/Mailu/issues/564#issuecomment-425633981

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

muhlemmer:
I've started work on the documentation. Just have a small list of docs remarks I want to resolve first, then I will work on the FAQ.

muhlemmer:
Due to the tag migration, some tags were lost. Today I've went through:
- All open and closed issues without tags
- All issues with tag ``topic/docs``
- All issues with tag ``type/question``

All the issues I see fit for inclusion in the FAQ are labeled ``faq/include``. Some of them are related and might be condensed to a single entry. I'll remove the tags from issues I'm done with.

[...]

Issue #332, opened 2017-11-25
Running mailu without docker's iptables creates an open relay


pgeorgi:
When disabling iptables in docker, its forwarding proxy process takes over. This creates the situation that every incoming connection on port 25 seems to come from the local network (docker's 172.17.x.x) and is accepted.

IMHO there should be some mechanism to check for that situation and refuse service in such cases, but right now I'm at loss what that could look like.

kaiyou:
This is actually quite a worrying security issue. Probably we should not even include anything by default in the trusted networks, as authentication already does most of the job.

domnoble:
yea this is a worrying security issue, for the past month or so ive had a bunch of IPs trying to gain access to the mailserver via port 25, there was a whole block of about 20 IPs in ukraine, which lead back to another IP in canada, i run everything behind a separate firewall so i can see what is going on

yesterday morning at about 05:00 i got so many port scans my internet literally started to lag, with them originating from israel, germany and china, i was able to block all the IPs which appeared to be port scanning but then today going over the logs i can now see the docker internal IP now doing some weird port scanning, i assume its down to this, sure enough if i shut down all the containers for mailu, the port scanning from the docker IP stops

im assuming the only sollution to this would be to try and install iptables on the host

kaiyou:
Well, I think you misunderstood the issue. Portscans and authentication bruteforce is common and actually pretty normal on the internet (at least for portscans). It's my opinion only, but I don't think you should spend too much energy trying to track them down (I personally get hundreds bruteforce attempts per day).

That being said, the current issue is in your configuration: you should not have the Docker proxy enabled  unless on a testing network. It is not a supported Docker setup and not one that will play well with many images, including Mailu. The recommended solution is to enable the Docker iptables configuration in your ``daemon.json``.

This issue will be resolved when the default config does not pose a risk for people still running a Docker proxy, or when the documentation states clearly enough that there is a risk of open relay.

DarioFra:
Hi I moved my mailu server from a docker linux host to a docker-desktop v3.2.2 (using WSL2 / ubuntu windows) host  and the server turned in to an open relay. (did not care to check due to lack of imagination I guess)

Seems the windows system does source NAT and the mailu_smtp_1 container gets source addresses from the docker network, as opposed to real external ones when running on a linux docker installation using iptables. (same with the mailu_front nginx container)

A workaround (for now) is to add a line to this file
[...]

Issue #177, opened 2017-03-11
2 Mailu servers for redundancy - Best approach


SaiGonSeamus:
We have two Mailu instances being set up. One in house, one in the cloud to act as a back up.

Is DNS MX the best approach or are there others that are preferable?

I haven't seen anything here on this issue, what are others doing?

Regards

S

zjr8:
Im all cloud based at the moment but am currently implementing a high avaliability setup using HA Proxy. It has the ability todo failover routing and supports smtp / dovecot. Still have a few bugs to workout in my config but HA proxy is working with starttls support (ssl is semi working but need to sortout a dovecot auth issues as the HA Proxy is doing ssl termination and confusing dovecot). Postfix is relaying outgoing email through amazon ses to help ensure client email reaches is destination as ive had issue with ec2  ip reputation before. I should note the entire setup works and is easier with an ELB (ssl termination even working) but it doesnt seem to support a method of restoing original ip to dovecot/postfix (which ha proxy does)

SaiGonSeamus:
Thanks for this, this looks like the way to go. 

Do you have HA Proxy set up on a separate server in a different geographic region?

Is there some documentation you would recommend on setting up HA Proxy to work with Mailu?

SaiGonSeamus:
We never implemented this at the time but want to put a solution in place now. Has anything changed? Is HA Proxy still the best route to take? Has anyone any suggestions or advice?

kaiyou:
It really depends on what your HA requirements are. Do you want to avoid losing any email, do you want to avoid losing the ability to read them? How long can you sustend a service disruption?

kaiyou:
Issue #272 could be a first step towards a highly available frontend for multiple backends.

SaiGonSeamus:
We would like to reconfigure our infrastructure to take advantage of this now. 

Broadly we're try to address 2 issues with this. Deliverability & Resilience. 

Ideally we want our primary backend to reside in our DMZ for security, control, cost etc. We have a KVM cluster with ZFS, RAID, Ceph on the way. 

Even though our public IP space is classed as commercial, sometimes foreign networks have it classed as residential adsl with a poor reputation as their data is out of date, there is also a 3 day turnaround from the ISP on adding, removing or changed PTR records. 

We now have two separate servers with hosting companies here and one as a test server in the DMZ, we would like to migrate them all to one domain. 

We also have had a lot of political interference with hosting companies abroad. they just block SMTP with no warning or notice. 

[...]

PR #677, opened 2018-10-22
Implement a FAQ

muhlemmer:
This is as far as I could get with the FAQ. Content is based on the [faq/include](https://github.com/Mailu/Mailu/issues?utf8=%E2%9C%93&q=is%3Aissue+label%3Afaq%2Finclude) tag plus some own inspiration.

Included items should be taged as faq/done, but Github is having some issues and I can't update the tags for now. TBD. Some issues remain to be included. This is either because I don't have any more inspiration to write a good answer for them, or some of them are pending a final solution.

For now I want to merge the FAQ and I (or anybody else) can continue work on it at a later stage.


docs/faq.rst
```diff
@@ -0,0 +1,276 @@
...
+You should be aware that creating, maintaining and expanding a mail server
+distribution requires a lot of effort. Mail servers are highly exposed to hacking attempts,
+open relay scanners, spam and malware distributors etc. We need to work in a safe way and
+have to prevent pushing out something quickly.
+
+**TODO: Move the next section into the contributors part of docs**
+We currently maintain a strict work flow:
+
+#. Someone writes a solution and sends a pull request;
+#. We use Travis-CI fore some very basic building and testing;
```
HorayNarea:
```suggestion
#. We use Travis-CI for some very basic building and testing;
```

muhlemmer:
This closes #564, #520, #332, #177.

docs/faq.rst
```diff
@@ -0,0 +1,276 @@
...
+
+  docker-compose logs front | less -R
+  docker-compose exec front less /var/log/letsencrypt/letsencrypt.log
+
+Common problems:
+
+- Port 80 not reachable from outside.
+- Faulty DNS records: make sure that all ``HOSTNAMES`` have **A** (IPv4) and **AAAA** (IPv6)
+  records, pointing the the ``BIND_ADDRESS4`` and ``BIND_ADDRESS6``.
+- DNS cache not yet expired. It might be that old / faulty DNS records are stuck in a cache
```
hacor: It might be handy to provide [MXToolbox](https://mxtoolbox.com) here to let users test their settings themselves. I used it a lot to check my complete config.

muhlemmer: Hi, I already provided a link to MXToolbox in the setup page in an earlier commit. However, it is probably not live yet.
https://github.com/Mailu/Mailu/blob/master/docs/setup.rst

I hope that I can assume users will first read the setup guide, before coming to the FAQ. :roll_eyes:

hacor: Indeed 👍

docs/faq.rst
```diff
@@ -0,0 +1,276 @@
...
+````````````````````````
+Fail2Ban is not included in Mailu. Fail2Ban needs to modify the host's IP tables in order to
+ban the addresses. We consider such a program should be run on the host system and not
+inside a container. The ``front`` container does use authentication rate limiting to slow
+down brute force attacks.
+
+We *do* provide a possibility to export the logs from the ``front`` service to the host.
+For this you need to set ``LOG_DRIVER=journald`` or ``syslog``, depending on the log
+manager of the host. You will need to setup the proper Regex in the Fail2Ban configuration.
+Be aware that webmail authentication appears to come form the Docker network,
```
hoellen:
```suggestion
Be aware that webmail authentication appears to come from the Docker network,
```

PR #678, opened 2018-10-23
Added kubernetes CNI support for Mailu

hacor:
This adaptations make it possible to use Mailu stable on Kubernetes using a special CNI (network plugin). 
Sometimes problems occurred when logging in to IMAP, the reasons for this were the following:

- The Nginx Ingress Controller uses `hostNetwork: true` to be able to open up the mail ports on your nodes
- This way the NGinx pods don't receive an IP in the `POD_ADDRESS_RANGE`
- They try to login to the IMAP container using the network bridge (eg: `10.2.0.0` address)
- This is OK, but in one specific situation, when the user accesses the NGinx Container on the same node that is running the IMAP container, it uses it's own address of the network bridge, which is on a different subnet as the `POD_ADDRESS_RANGE` to access the imap pod (in my case it was `10.1.6.0/24). 
- By providing a specific `POD_ADDRESS_RANGE` in the configmap, using commas it gets added in the IMAP pod and the connections are accepted!

Docs are updated
 
Signed-off-by: Hans Cornelis <hacornelis@gmail.com>

### DIFF
--- a/docs/kubernetes/mailu/configmap.yaml
+++ b/docs/kubernetes/mailu/configmap.yaml
@@ -53,6 +53,11 @@
     # This way we can make use of the advantages of the cert-manager deployment
     KUBERNETES_INGRESS: "true"
 
+    # POD_ADDRESS_RANGE is normally provided by default with Kubernetes
+    # Only use this value when you are using Flannel, Calico or a special kind of CNI
+    # Provide the IPs of your network interface or bridge which is used for VXLAN network traffic
+    # POD_ADDRESS_RANGE: 10.2.0.0/16,10.1.6.0/24
+
     ###################################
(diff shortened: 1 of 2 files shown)


muhlemmer:
I think nobody with write access uses kubernetes. Therefore I think it would be wise this is functional reviewed by either @consideRatio or @briantopping. If everything is okay, we (official reviewers) can give the final go on the merge.

briantopping:
Happy to help. I will spin up some new virtual machines to test this.

consideRatio:
I won't manage to squeeze in deployment testing of code yet, but I'm happy to give some feedback to k8s code. Note that I'm still learning a lot about it myself.

PR #679, opened 2018-10-23
Added the option to choose your own RSPAMD password

hacor:
As promised to @muhlemmer I made a separate PR to add the option to choose your custom rspamd password. Although it was not really clear if this was done using the admin interface or not.

Signed-off-by: Hans Cornelis <hacornelis@gmail.com>


muhlemmer:
I don't really understand why we would want an additional password for the Rspamd interface?

Issue #554, opened 2018-08-06
[rspamd] DNS query blocked on multi.uribl.com


sam2kb:
```
(controller) <3nxzfe>; monitored; rspamd_monitored_dns_cb: DNS query blocked on multi.uribl.com (127.0.0.1 returned), possibly due to high volume
(rspamd_proxy) <f69f64>; surbl; surbl_test_url: disable surbl multi.uribl.com as it is reported to be offline
```
I understand that requests are blocked because of the use of public DNS resolvers like 8.8.8.8 or 1.1.1.1 If that's so, is there a way to use my own DNS nameserver across all MailU services or just in the antispam container?

sam2kb:
If you wonder, the number of emails/day is under 100 so there's no way I'm hitting a rate-limit

sam2kb:
The above mentioned unbound image didn't want to resolve any DNS requests, so I just hooked up a different image and it works well now

```
surbl; process_dns_results: <0.0.0.A3.1D42F4BFB3ABFB4.32F383@mail.ketodiet808.us> domain [ketodiet808.us.multi.uribl.com] is in surbl multi.uribl.com(2)
```

sam2kb:
```
version: '3.6'
services:
  unbound:
    container_name: unbound
    #image: ofthesun9/unbound-mailu:1.1
    image: mvance/unbound:latest
    restart: unless-stopped
    ports:
    - 0.0.0.0:53:53/tcp
    - 0.0.0.0:53:53/udp
networks:
  default:
    external: true
    name: mail_default
```

ofthesun9:
Hi,

Sorry to hear that it was'nt working for you...
Could you please provide the logs of you attempt to use it ?

muhlemmer:
@muhlemmer (For own reference)

PR #681, opened 2018-10-23
Unbound DNS as optional service

muhlemmer:
This PR supersedes #385 and #640. It aims to solve the problems discussed in #554.

This will add the unbound image and run it as service "resolver". It is an optional service and can only be included using the new setup application. Using the "resolver" service, will force the use of a fixed subnet. We have provided a default subnet which should stay clear of the Docker ranges.

Some key changes since the last PR #640:

- Unbound is an optional optional service;
- Unbound limits itself to allow only connections from `$SUBNET`, using jinja for config generation;
- Unbound is used only on services that do lookups: `smtp`, `fetchmail`,  `antivirus` and `antispam`.


kaiyou:
This looks good to me and I am so glad we landed the setup wizard.

PR #682, opened 2018-10-23
Further improve setup and neccesary re-write of the test suite

ionutfilip:
### Did following changes:
- created documentation for adding flavors and steps
- added log driver in docker-compose.yml and mailu.env
- added flag for project name in commands from compose setup page
- added replicas step and variables for stack flavor


muhlemmer:
Sorry for the mess with the commits here, we're refactoring the test suite to follow the new format of compose files. But I need the output that is created by this branch in our test suite branch.

muhlemmer:
I've taken out the commits that interfered with the test building. I was creating an unbelievable mess.

muhlemmer:
Updated summary: we wanted to create new test cases based on output of the setup utility. It was impossible to do the necessary changes in two branches and create separate pull requests. So we put everything on one pile. Most of the changes are visible in the commit messages.

One important note: in order to make the tests more reliable, we wanted to implement healthcheck status in the script. We have re-written the test script in python in order to use the Docker API for health status polling. Also, the test structure is changed to allow for the multiple docker-compose files, generated by setup.

Testing now also allows for "hooks". Every test variation is in it's own directory, called by the test name. Any `.sh` or `.py` file in the directory will be automatically run after `docker-compose up -d` and the set timeout.

On the next iteration we will start to implement test hooks and some documentation for other devs on how to do this. For now I consider this PR ready for merge.

muhlemmer:
Since this was not yet reviewed, we have added the next batch of work on the test-suite to this PR.

1. Make TLS work with self-singed certs
2. Creating users
2. Sending / receiving mail for authenticated user
3. Sending and blocking a test virus.

It would be wise to get some reviewing on the way for merge. Otherwise this just keeps growing to a big beast ;). @Mailu/contributors

muhlemmer:
I think the review on this can be simplified:
- The test scripts are actually tested by Travis, no real need to build /run it locally
- Setup can be run locally:

```
git checkout usrpro/feat-setup
cd setup
docker-compose run
```
This builds and runs the setup locally. It can now be accessed through localhost:8000 in the browser.

PR #685, opened 2018-10-24
Feat rainloop 2fa

muhlemmer:
This supersedes #264. TODO: create a FAQ entry (in this branch) denoting the shortcomings of 2FA in webmail.


hoellen:
Maybe we can introduce a switch (as opt-in) in the admin panel, so that the "normal" password is only allowed to use it from the docker network (webmail). Any other client (via imap) has then to use an Authentication token. So the 2fa is a real security feature. And not only "feeling" secure.

muhlemmer:
@hoellen I'm not sure I follow. Webmail is not only accessible from inside the network, it's exposed through front?

hoellen:
No I mean if you enable 2fa on webmail. And someone steal your password it is still possible to use this password via another client (e.g. thunderbird). So the 2fa is useless.
But with an option in the admin interface we could enable the "normal" password **only** for webmail. Then you have to use authentication-tokens for every other imap client. (btw: that's the way nextcloud does it, too)

(With docker network I meant, that for the admin container the webmail container is the internal docker network. And any other imap client is outside the docker network. So that could be a way for the implementation.) 

Is it now more clear? Hopefully this explains it better.

hoellen:
Yeah for sure that's right. Just wanted to give some input for this. So that we have this (for later) in mind.

PR #692, opened 2018-10-30
store rspamd history in redis - fix #561

### LINKED ISSUE #561: [rspamd] Preserve antispam history on container restart
Is there any reason why we don't keep the history? I tried enabling it with the following override, but it didn't work.
Do I need to enable **history_redis** module first or is there anything else I'm missing?

overrides/rspamd/history_redis.inc
```
history_redis {
  servers = "redis";
}
```

HorayNarea:


kaiyou:
Okay with this, maybe we should run some tests and advertise how much it stores on drive in the docs.

HorayNarea:
This only saves the last 200 lines¹ of history as a ZSTD-compressed JSON… so it should be fine as long as you have a few MBs of free space :D

¹ maybe we should make this configurable?

PR #702, opened 2018-11-03
Add docker image prefix

dani909:
Allow prefixing docker images as they may collide with other images.
For example the nginx image may collide with another nginx image on some personal docker hub accounts.

This PR allows setting an image prefix so
`DOCKER_ORG=someuser DOCKER_PREFIX=mailu- docker-compose ...`
would tag the nginx image as `someuser/mailu-nginx` and it will not collide anymore and states that it is intended for use with Mailu.

In case nothing is provided for `DOCKER_PREFIX` it will obviously tag it as before.


muhlemmer:
Note to self: templates for setup utility need adjustment to reflect this change.

hoellen:
Resolved the merge conflict.

PR #709, opened 2018-11-18
1.5 htaccess php7

rbarazzutti:
The `.htaccess` file needs to be modified when php7 is being used (as stated in the .htaccess file comment).

Otherwise, we don't benefit from the Roundcube provided config and thus changing `upload_max_filesize` or `post_max_size` in `.htaccess` won't have any effect since this block is being ignored.


rbarazzutti:
Hi,

Actually, this fix is for the "1.5" branch. On the master, it has been solved already.

Thanks

Issue #343, opened 2017-12-03
Build a setup page


kaiyou:
A setup page could be exposed, that just asks a few questions and provides a custom compose/kubernetes/whatever configuration file. This requires a simple web app and a way to export configuration templates for various Git branches.

kaiyou:
A very first version is available in the repository. The framework is ready, it basically lacks the actual configuration.

muhlemmer:
@kaiyou 
cc: @ionutfilip 

Do you have some kind of todo for this?

We would like to work on the standard compose deployment a bit and later inclusion of swarm. Swarm would be much more complicated, as there are a lot of choises to be made. Hence we want to start simple to get familiar with the program.

kaiyou:
I would say:

[ ] make sure the base python scripts run properly and maybe write some documentation for them
[ ] (probably replace the ``setup.py`` script with simply a build per branch, just like documentation, but we can discuss it later)
[ ] write the actual forms in ``steps/`` to ask the proper questions for a new server configuration to be generated
[ ] write the actual results in ``flavors/`` using jinja

muhlemmer:
Transcript of our discussion on Matrix yesterday, for future reference.

> **muhlemmer:**
> kaiyou, We'll start work on the setup utility. Just a question: Do we want to get rid of variables in docker-compose.yml? Ask all the questions in the utility and provide a hard-coded version.
> This woudl make sence, since docker stack deploy does not support loading variables from .env to use in a docker-compose.yml.
>
> Second question: can we upgrade to Docker compose file version `>3?` Start with a common base for Docker Swarm
> **kaiyou:**
> Hardcoding sounds ok even if providing both options is possible using flavors.
> And of course yes we can (and should) update the compose format
**muhlemmer**:
kaiyou. We are having some hickups. At this moment the Dockerfile seems to pull everything from ``GitHub:master.`` Including the app itself?
> Unless we are missing something, there is no way we can test any local changes untill merged.
> So we want to take this out and just use the files relative to the Dockerfile
> **kaiyou**:
> As I was mentioning on the issue, currently the prepare script does the sane kind of complex job as was the docs versioning stuff before. I believe this can be removed altogether and replaced with multiple tags of the docker image.
> **muhlemmer**:
> Aha, clear :D

muhlemmer:
@kaiyou. So we have a working example running at http://setup.usrpro.io:8000. The compose file and env files generated work well.

[...]

PR #713, opened 2018-11-26
nginx: Allow extending config with overrides

pgeorgi:
To facilitate extending nginx, the default redirect at / can be disabled, even if
the default remains at redirecting to the webmailer.

The extensions are within the host scope and are read from
$ROOT/overrides/nginx/*.conf.


docs/compose/.env
```diff
@@ -97,6 +97,9 @@ COMPRESSION_LEVEL=
 # Web settings
 ###################################
 
+# Module to forward to from /
+WEB_DEFAULT=webmail
```
muhlemmer: There is not any other module to choose from. If any other string is set here, the forward will not happen anymore. For instance, setting this to `admin` will not have the advertised effect and a 404 will be served.

Perhaps it's better to make this a `boolean` type and call it something like `WEBMAIL_REDIRECT`.

core/nginx/conf/nginx.conf
```diff
@@ -84,9 +84,11 @@ http {
       }
       {% else %}
 
+      include /overrides/*.conf;
+
       # Actual logic
       {% if WEBMAIL != 'none' %}
-      {% if WEB_WEBMAIL != '/' %}
+      {% if WEB_WEBMAIL != '/' and WEB_DEFAULT == 'webmail' %}
```
muhlemmer:
```
{% if WEB_WEBMAIL != '/' and WEBMAIL_REDIRECT %}
```
See comment on `docs/compose/.env`.

docs/compose/.env
pgeorgi: Under this scheme it may be feasible to point to admin (which this change doesn't support yet, I admit) or future additions if mailu's scope expands.

Making this a boolean means that we're locking ourselves in this corner: what would happen if, in some future version, both WEBMAIL_REDIRECT and ADMIN_REDIRECT (or CALENDAR_REDIRECT or whatever else may come up) are set to true?

If you prefer, I'll change it to the boolean scheme.

muhlemmer: Personally, I'm not a fan of redirects. See issue #662 ;). I was looking into removing it altogether. However, making it a user's choice is a much better option.

The reason why I suggested a `boolean` type, is that the current proposal seems a bit messy.
- We have `WEB_WEBMAIL=/webmail` with slash
- We have `WEB_DEFAULT=webmail` without slash

Now, a novice user comes along and changed the path to the webmail service. I _might_ think to change `WEB_DEFAULT` to the the same string. He does not know that the string in this variable need to match exactly to a module name.

So there are 3 real options worth considering:

1. Keep your approach and document clearly what are the valid module names. Create `if` statements for the other modules. (Personally, I think this is a dirty approach)
2. Lock it down in a boolean. Simple and straightforward.
3. (New idea) Call it `WEBROOT_REDIRECT` and redirect literally to this string. If its empty, don't redirect.  This way users can customize however they wish. They can even redirect to an external site. It creates flexibility if we will ever implement the support of [multiple webmails](https://github.com/Mailu/Mailu/issues/343#issuecomment-430784490).

pgeorgi: I like the WEBROOT_REDIRECT option and implemented it. In the spirit of "one change per commit" I kept the default behavior the same, ie. it redirects to /webmail by default (since the default behavior also includes providing webmail).

muhlemmer:
Sorry for the mess, all done now.

Issue #83, opened 2016-10-19
Add support to openldap


thiagoalmeidasa:
Would be nice to integrate this stack with an external openldap to auth smtp and imap.
Have you thought about it?

kaiyou:
I did, and I dismissed the idea at first, the spirit was to get a simple mail server up and running. Most database structures are thus based on SQLite and are pretty much *-SQL compatible but OpenLDAP compatibility is another story.

Using Openldap only for authentication would be an option, but then Dovecot could inherit the responsibility of authenticating users or so. For this to work properly, I believe that a central service - let's say a central API - would have to provide the authentication socket to all services and then handle external authentication if required, plus all domain-specific data structures.

It is technically possible, but I would not expect it to be available any time soon, the first strong requirement is a clean central API service. Let's keep the issue open for discussion.

jcgruenhage:
Is there any progress in decoupling authentication? I want to migrate my mail server to mailu, but I really don't want to maintain multiple account stores.

benyanke:
Would also appreciate LDAP integration. Don't have a ton of time to help implement, but would help test and use it when it hits prod.

benyanke:
I'd also like to throw in a plug for the project I maintain, [glauth](https://github.com/glauth/glauth). It has a docker image, and could perhaps be integrated in.

BirgerK:
I'm also very interested into this feature.

I took a look into the code. Do I get it right, there is already the "admin-internal"-service which provides - at least for dovecot - the auth-mechanism.
I think the LDAP-integration could be implemented in that service. If that's all I would do it.

Am I right?

jcgruenhage:
@BirgerK If you just want to use LDAP for authenticating users that are already known to mailu, then that should be nearly all yes. The one missing thing would be config through the UI.

BirgerK:
What did you mean by "config through UI"?

- are the mailu-config-params set through UI? (I thought it's configured through Env-aprams)
or
- do you mean how the UI authenticates itself?

jcgruenhage:
I mean that currently accounts are managed using the UI. The UI should still be able to manage things about user accounts that are not done by LDAP, and managing the LDAP configuration itself could also be done through the web UI

BirgerK:
Hi again,
I'm currently implementing the LDAP-auth-mechanism into the admin-internal-service. By that, every service which authenticates by the admin-internal-service, will be authenticated against LDAP.

[...]

PR #715, opened 2018-11-28
add ldap-authentication

BirgerK:
Like it was discussed in https://github.com/Mailu/Mailu/issues/83:

I implemented the LDAP-authentication as additional Authentication-Provider beneath the existing SQL(ite)-authentication.
When the LDAP-config is set, every HTTP-Authentication-Call on the Core-Admin-Service will first authenticate against LDAP. If there wanted user could not be found there, SQLite will be used as a fallback Authentication-Provider.

TODOs:
- [ ] Document the configuration params


muhlemmer:
Thanks for this PR! I'm afraid this will collide with PR #670, where the config management is changed. So there are two options here:
1. Base the modifications on PR #670 and block this PR untill #670 is reviewed and merged; Or
2. Review and merge this PR as-is. Modify the works in PR #670 afterwards.

The actual outcome will probably depend which PR will pass review first.

I will try to do functional testing of this PR after the weekend. Cheers!

BirgerK:
I added some more TODOs. I didn't know that not only the authentication is done by the "Admin"-service. It also serves some Mailbox-discovery and stuff.

That should all be completed by some LDAP-queries, but it is not implemented, yet.

I didn't recognize the complexity of the "Admin"-service and this feature.

By that, I won't complete this feature.

muhlemmer:
> I didn't know that not only the authentication is done by the "Admin"-service. It also serves some Mailbox-discovery and stuff.

Yes, all queries are done through `admin`. This includes acceptance of incoming mail, aliases and forwarding. And probably a whole bunch more. This was exactly the motivation behind my [reply in 83](https://github.com/Mailu/Mailu/issues/83).

I do think that your approach is not too far off. In theory, you could ldap let to take care of authentication only, and write the results to the session. After that business should continue as usual. You just have to find a way to keep LDAP users in sync with the Mailu ones. That could be as easy as creating a user in the Mailu database upon first successful login.

> By that, I won't complete this feature.

Too bad to hear that.

muhlemmer:
@BirgerK, should be close this PR?

BirgerK:
This code could be kept if anybody wants to complete it.

If nobody cares, this PR can be closed.

Yorny2101:
Hi everyone,

I noticed that about six years ago, a solution for integrating LDAP authentication was proposed in PR #715. @muhlemmer pointed out that it might conflict with PR #670, which aimed to change the configuration management. He suggested two options: either base the modifications on PR #670 or review and merge PR #715 as-is and then modify PR #670 accordingly.

Given the current state of Mailu, I'm wondering how I can adapt the original LDAP integration solution from PR #715 to the current codebase. Any guidance or advice on how to proceed would be greatly appreciated.

Thanks in advance for your help!

PR #717, opened 2018-11-30
Add postgresql database option

muhlemmer:
This PR adds the option of using a postgresql database. Related to #420. This will provide two ways of doing this. Both ways are included in setup.

1. Set you own connection parameters in `mailu.env`. The user has to make sure the database server is correctly configured.
2. Use the image provided by mailu, which will be configured automatically.

The provided image has the following features:

- Automatic creation of users, db etc.
- Connection password is `SECRET_KEY` and is used by admin by default
- TCP connections are only allowed from the mailu `SUBNET`
- Healthcheck
- Automatic minutely _"wal archiving"_ and weekly `pg_basebackup`.
- Automatic cleaning of `wal_archive` and "base backups". Two versions always remain available
- When `/data` is empty and backups are present, the backups are restored automatically. Useful in swarm environments, since the /data directory should not be on any network filesystem (performance)

**Note:** Postgresql does not support `colate="NOCASE"`. We used the `citext` extension to achieve the same effect. However, this means we had to manually hack the migration scripts and models to conditionally use `citext`. 

## Blocked label
The works in this PR are based on:

- [x] PR #670;
- [x] PR #682.

Therefore it cannot be merged until those are. 

## Wip label
Following still needs to be done:
- [x] Documentation
- [ ] Automated test
- [x] Testing against external server


kaiyou:
Damn, I just checked and had a non-commited branch containing some of the model changes required for ``NOCASE``, sorry this did not come through. I'll handle #670 and come back to this see if our changes are identical :)

kaiyou:
Ok, there are some changes that go opposite from the generic spirit of #670. The idea was especially not to embed anything configuration specific in the models or in the migrations. I will try and improve upon that, I hope that generic SQLALchemy types can achieve the same level of functionality.

Otherwise, even though I am not a postgres expert and cannot comment much on the proposed internal image, it looks good.

kaiyou:
If you do not mind, I'll start another branch, cherry pick the postgresql part and update the models and migrations separately.

muhlemmer:
Yeah, I didn't like to hack the stuff in like I did. It's messy. Probably everything should be regular strings and the case handling can be done in the app, instead of the database. Probably just a matter of lower-casing the variable that comes from the signup and login forms.

However, we will remain the issue of the migrations. Even if we choose to change to `db.string` without `colate`, it will just create an additional migration layer, it won't undo the earlier migration scripts. As such, migration scripts will fail on postgresql and it will end up with an incomplete table definition. (psql is rejecting the queries containing `colate=nocase`)

We could also choose to maintain separate migration path for every DB flavor. In that case we can start with a clean slate on postgresql. And probably prevent us from having the same shit all over when somebody implements mysql.

muhlemmer:
By the way, 

> If you do not mind, I'll start another branch, cherry pick the postgresql part and update the models and migrations separately.

I'll do that. I already started to handle some conflicts just now.

PR #721, opened 2018-12-03
Move the Mailu Docker network to a fixed subnet

muhlemmer:
This will make network configuration and host based authentication more robust, across different deployment platforms. The options `RELAYNETS` and`POD_ADDRESS_RANGE` are kept for compatibility. However, their usage have become optional.

Works in this PR are based on PR #670 and therefore this will remain blocked until that one is merged.


muhlemmer:
I will re-submit this after cleaning up the mess :stuck_out_tongue:

Issue #430, opened 2018-03-30
Problems with certificates while running Mailu under nginx-proxy/letsencrypt-companion


flayks:
I found Mailu a good solution to host my emails as I'm trying to rebuild my server with Docker, but I get several issues about getting it running properly, especially with the certificates.
Experimented with `TLS_FLAVOR` in my `.env` file, set to `cert` or `mail` but the admin only works with `notls`.

I run my own nginx container for accessing the admin at `mails.server.site.com`, but I always get a 502 error or a 403 with mail or cert option.

Also, when I run the containers, I get this error from the `mailu-front`:
> **mailu-front  | Missing cert or key file, disabling TLS**
> mailu-front  | 2018/03/30 10:20:37 [notice] 7#7: using the "epoll" event method
> mailu-front  | 2018/03/30 10:20:37 [notice] 7#7: nginx/1.12.2
> mailu-front  | 2018/03/30 10:20:37 [notice] 7#7: OS: Linux 4.9.20-std-1
> mailu-front  | 2018/03/30 10:20:37 [notice] 7#7: getrlimit(RLIMIT_NOFILE): 1048576:1048576
> mailu-front  | 2018/03/30 10:20:37 [notice] 7#7: start worker processes
> mailu-front  | 2018/03/30 10:20:37 [notice] 7#7: start worker process 9
> mailu-front  | 2018/03/30 10:20:37 [notice] 7#7: start worker process 10
> mailu-front  | 2018/03/30 10:20:37 [notice] 7#7: start worker process 11
> mailu-front  | 2018/03/30 10:20:37 [notice] 7#7: start worker process 12

The certificates generated by [docker-letsencrypt-nginx-proxy-companion](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion) are actually present on `/etc/certs/mail.site.com` and working, but when I try to symlink anything it doesn't work and I get the same message for the front container.

Did I missed something? I looked at all the issues here about nginx-proxy, reverse proxies or certificates, read the docs carefully, but nothing sounds to work in my situation.

### My configs
**.env**
```
###################################
# Common configuration variables
###################################

# Set this to the path where Mailu data and configuration is stored
ROOT=/data/tools/mailu

# Mailu version to run (1.0, 1.1, etc. or master)
VERSION=1.5

# Set to a randomly generated 16 bytes string
SECRET_KEY=key

# Address where listening ports should bind
BIND_ADDRESS4=127.0.0.1
[...]

PR #724, opened 2018-12-05
Documentation and examples for traefik

Nebukadneza:
_This is a Work in Progress and Requesting for Comments. Please don’t merge, but critique to all your hearts content ^_^._

Mailu is well suited in containerized setups for selfhosting various sevices. In todays container infrastructures, traefik is the next big reverse proxy / ingress solution next to nginx. As such, it’d be great if Mailu could work well behind traefik — at least for the HTTP part.

I’m aiming to provide:
- Documentation on how to set up Mailu behind traefik, and how to cope with traefiks … finicky parts
- Sample configuration files for relevant parts (⇒ Please tell me if you prefer extra files, commented-out parts in the already existing files, or only text in the documentation for this)
- An additional optional container to extract certificates out of traefik, put them into front and reload fronts nginx

I believe with this, mailu can become even more user-friendly (despite they might be advanced users, with bigger containerized-setups), and play a bigger role in more different setups.


muhlemmer:
Also see issue #430 for an extended Traefik discussion.

Nebukadneza:
Thanks, I didn’t stumble upon this. Quickly added an example configuration that may not be complete yet, but does work. For the weekend i plan to re-implement cert-dumper in python so it’ll fit the mailu style better.

Would you rather have the certdumper-container access `docker.sock` (write, ugh) and `exec nginx -s reload`, or add a small daemon that watches `cert.pem/key` inside of `front`?

Best & Thanks
-Dario

muhlemmer:
> or add a small daemon that watches cert.pem/key inside of front?
I would prefer that. It will be more implementation agnostic. There have been talks between @HorayNarea  and myself in implementing such a thing based on `inotify`. Just didn't got around on doing it yet.

Disclaimer: not sure is `inotify` works when a file is written from outside the container. You might just have to create loop which is checking the modification time, when it does not.

Nebukadneza:
Hi,

I’m now running this setup successfully on master for some time. We have the certificate-reloader in master, I adapted the most popular traefik certificate-dumping dockerimage to be more convenient to use with mailu — and I plan on maintaining this image too.

With these set, I’ve updated examples and documentation text again. From my side this seems somewhat finished.
Please let me know your thoughts on this ….

Best ’n Thanks,
-Dario

docs/compose/treafik/docker-compose.yml
```diff
@@ -0,0 +1,139 @@
...
+      - "$BIND_ADDRESS6:993:993"
+      - "$BIND_ADDRESS6:995:995"
+      - "$BIND_ADDRESS6:25:25"
+      - "$BIND_ADDRESS6:465:465"
+      - "$BIND_ADDRESS6:587:587"
+    volumes:
+      - "$ROOT/certs:/certs" # Mount both certs directory (for dhparams.pem) and your domains key
+      - "$ROOT/overrides/nginx:/overrides"
+      - /data/traefik/ssl/mail.your.doma.in.crt:/certs/cert.pem
+      - /data/traefik/ssl/mail.your.doma.in.key:/certs/key.pem
```
muhlemmer: Instead of using `mail.your.doma.in.key` and `mail.your.doma.in.crt` you could define a variable called `TRAEFIK_HOST` and do something like:
```
      - /data/traefik/ssl/$TRAEFIK_HOST.crt:/certs/cert.pem
      - /data/traefik/ssl/$TRAEFIK_HOST.key:/certs/key.pem
```
It's bit cleaner and people can just add it to their `.env` or just change the variable to a hard coded string. (However they wish).

docs/compose/treafik/docker-compose.yml
```diff
@@ -0,0 +1,139 @@
...
+      - "80" # Let’s not expose 80 or 443 on host, since that’s taken by traefik
+      - "$BIND_ADDRESS4:110:110"
+      - "$BIND_ADDRESS4:143:143"
+      - "$BIND_ADDRESS4:993:993"
+      - "$BIND_ADDRESS4:995:995"
+      - "$BIND_ADDRESS4:25:25"
+      - "$BIND_ADDRESS4:465:465"
+      - "$BIND_ADDRESS4:587:587"
+      - "$BIND_ADDRESS6:80:80"
+      - "$BIND_ADDRESS6:443:443"
```
muhlemmer: You missed a spot. I don't think you want to expose 80 and 443 ports on IPv6 either?

docs/compose/treafik/docker-compose.yml
```diff
@@ -0,0 +1,139 @@
...
+  front:
+    image: mailu/nginx:$VERSION
+    restart: always
+    env_file: .env
+    logging:
+      driver: $LOG_DRIVER
+    labels: # Traefik labels for simple reverse-proxying
+      - "traefik.enable=true"
+      - "traefik.port=80"
+      - "traefik.frontend.rule=Host:mail.your.doma.in"
```
muhlemmer: See comment regarding variable, down this page.

docs/compose/treafik/docker-compose.yml
```diff
@@ -0,0 +1,139 @@
...
+      # This may be needed (plus defining mailu_default external: true) if traefik lives elsewhere
+      # networks:
+      # - mailu_default
+
+  # The certdumper is external, but adapted for using traefik with mailu
+  certdumper:
+    restart: always
+    image: nebukadneza/traefik-certdumper:latest
+    environment: # Make sure this is the same as the main=-domain in traefik.toml
+      - DOMAIN=mail.your.doma.in
```
muhlemmer: We already use `DOMAIN` for the primary mail domain. Although this does overwrite that, it will lead to confusion. I have a comment lower on this page to use something like `TRAEFIK_HOST` in `.env`. That would also take care of passing it down to the container.

docs/compose/treafik/docker-compose.yml
```diff
@@ -0,0 +1,139 @@
...
+      - "/data/traefik/acme.json:/acme.json"
+      - "/data/traefik/traefik.toml:/traefik.toml"
+      # This may be needed (plus defining mailu_default external: true) if traefik lives elsewhere
+      # networks:
+      # - mailu_default
+
+  # The certdumper is external, but adapted for using traefik with mailu
+  certdumper:
+    restart: always
+    image: nebukadneza/traefik-certdumper:latest
```
muhlemmer: I would prefer that you include the Dockerfile into Mailu, so that we can peer-review it and users don't need to pull anything from outside our "trusted" repository.

docs/compose/treafik/docker-compose.yml
```diff
@@ -0,0 +1,139 @@
...
+      - "$ROOT/dav:/data"
+
+  admin:
+    image: mailu/admin:$VERSION
+    restart: always
+    env_file: .env
+    volumes:
+      - "$ROOT/data:/data"
+      - "$ROOT/dkim:/dkim"
+      - /var/run/docker.sock:/var/run/docker.sock:ro
```
muhlemmer: The socket is no longer used and this mount needs to be taken out. It was supposed to be taken out on master (I've seen it in a PR) but somehow it got stuck. I've just taken it out in master, #768. If you can please take care of this one?

docs/compose/treafik/traefik.toml
```diff
@@ -0,0 +1,33 @@
...
+  [entryPoints.https.tls]
+
+[docker]
+endpoint = "unix:///var/run/docker.sock"
+watch = true
+exposedByDefault = false
+
+# Make sure we get acme.json saved, and onHostRule enabled
+[acme]
+email = "root@mail.doma.in"
```
muhlemmer: Don't know if Traefic supports variables for its toml files?

muhlemmer:
O, and by the way it seems that you didn't configure your local git client properly... You should at least make sure that the local account uses the same e-mail address as registered on GitHub. This will not prevent the merge, but the commits will not count in your profile on GH. The greyed-out avatar ussualy gives this hint:
image
More info: https://help.github.com/articles/why-are-my-contributions-not-showing-up-on-my-profile/

docs/compose/treafik/traefik.toml
Nebukadneza: Unfortunately I couldn’t figure out how — to the best of my knowledge, it’s not possible.
While adding a templater would be possible, I think the common deployment scenario is to have traefik already set-up outside of mailu, and this already configured. Maybe I should make this clearer here?

Nebukadneza:
Thank you for your comments! I tried to include everything as clearly as possible.

muhlemmer:
There are still some errors as `dumpcerts.sh` still returns `0` on certain `jq` failures. A fix has been submitted upstream: containous/traefik#4347

Nebukadneza:
Thanks for your review and fixes on this!

outbackdingo:
so does this mean we can build master now with mailu / traefik web front end ???

muhlemmer:
yes

PR #727, opened 2018-12-06
Move the Mailu Docker network to a fixed subnet.

muhlemmer:
This will make network configuration and host based authentication
more robust, across different deployment platforms.
The options `RELAYNETS` and`POD_ADDRESS_RANGE` are kept for compatibility.
However, their usage have become optional.

This is a re-submission of PR #721


muhlemmer:
Sorry, tests were failing. This is fixed now and ready for a new review ;) @HorayNarea

muhlemmer:
This whole singe review thing does not really work :stuck_out_tongue_closed_eyes: Merging manually.

ofthesun9:
We need to point out that SUBNET variables need to be set to get imap authentification working....
For the ones pulling the latest master without noticing the change in .env and docker-compose.yml to introduce SUBNET will get imap authentification failing.

PR #728, opened 2018-12-06
Fix autobuild

muhlemmer:
After the merges from today, the travis tests on master are failing. A change in user creation syntax and a simultaneous change in the test suite caused this.

This PR rectifies the syntax in the test suite.


muhlemmer:
ooh crap... mergify blocks itself when rules are contradicting.

muhlemmer:
Nevermind that... it's github branch protection at play here.

PR #729, opened 2018-12-06
Add postgresql database option

muhlemmer:
This PR adds the option of using a postgresql database. Related to #420. This will provide two ways of doing this. Both ways are included in setup.

- Set you own connection parameters in mailu.env. The user has to make sure the database server is correctly configured.
- Use the image provided by mailu, which will be configured automatically.

The provided image has the following features:

- Automatic creation of users, db etc.
- Connection password is SECRET_KEY and is used by admin by default
- TCP connections are only allowed from the mailu SUBNET
- Healthcheck
- Automatic minutely "wal archiving" and weekly pg_basebackup.
- Automatic cleaning of wal_archive and "base backups". Two versions always remain available
- When /data is empty and backups are present, the backups are restored automatically. Useful in swarm environments, since the /data directory should not be on any network filesystem (performance)

This PR supersedes #717. This one has a cleaner commit history. Also, all the CIText conditional stuff has been taken out as per suggestion of @kaiyou.

## TODO:
- [ ] Find a better way of handling NOCASE
- [ ] make the migrations to be accepted by postgresql
- [x] Inclusion in test suite


muhlemmer:
Copied response from #717:

However, we will remain the issue of the migrations. Even if we choose to change to db.string without colate, it will just create an additional migration layer, it won't undo the earlier migration scripts. As such, migration scripts will fail on postgresql and it will end up with an incomplete table definition. (psql is rejecting the queries containing colate=nocase)

We could also choose to maintain separate migration path for every DB flavor. In that case we can start with a clean slate on postgresql. And probably prevent us from having the same shit all over when somebody implements mysql.

kaiyou:
I tried to go the pragmatic way and started editing the migration scripts.

The overall idea is: 
 - keep the collate on the column definition, so that runtime queries can use the collate on supported backends, this will ensure forward support of existing SQLite databases
 - remote the collate from column creation in migrations, it will not alter the SQLite behavior even for new databases (runtime collate is used instead), and existing collate on existing databases is harmless, other new databases will simply have a default collate at table creation
 - lowercase every domain name and email when sending it to the database, so that new databases, no matter the collate, will only store and compare lowercase strings for these fields, thus removing any need for specific collate on postgres or MySQL.

I am testing this on my own branch and can open a PR afterward. Then it is a matter of merging our two pieces of work, your postgresql container and setup scripts, plus the admin part with now collate.

muhlemmer:
Sounds good!...

Something we also came across. Some e-mail columns where colate=NOCASE and had a foreign constraint to columns without such option. When we implemented CIText PostgreSQL was kindly rejecting the constraints on type mis-match. This will probably not expose in your approach, but might be buggy if not handled correctly. Probably we'll get noisy database constraint errors when a user will try to create a cased e-mail address. One table will get a case-less version and the other one cased.

https://github.com/Mailu/Mailu/blob/c6dbcc26da6baacb36133d5df740bd29a82fb0d7/core/admin/mailu/models.py#L462

kaiyou:
I just finished testing https://github.com/kaiyou/Mailu/tree/feat-multiple-db, it runs fine on SQLite, PostgreSQL and MySQL (only tested the admin UI). I still have to run functional tests but I don't think it broke anything.

I will then rebase it on top of your work.

PR #731, opened 2018-12-07
Dynamic attachment size limit

### LINKED ISSUE #542: Increase attachment size limit
Hello, 

How can web change the max file limit of attachment in Roundcube ?

Version 1.5

ionutfilip:
Max attachment size will be 33% smaller that MESSAGE_SIZE_LIMIT defined in env file.
This closes #542.

### DIFF
--- a/docs/compose/.env
+++ b/docs/compose/.env
@@ -61,6 +61,7 @@ ANTIVIRUS=none
 
 # Message size limit in bytes
 # Default: accept messages up to 50MB
+# Max attachment size will be 33% smaller
 MESSAGE_SIZE_LIMIT=50000000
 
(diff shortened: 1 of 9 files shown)


muhlemmer:
[Background](https://docs.microsoft.com/en-us/exchange/mail-flow/message-size-limits?view=exchserver-2019) on the 33% limit reduction:

> For any message size limit, you need to set a value that's larger than the actual size you want enforced. This accounts for the Base64 encoding of attachments and other binary data. Base64 encoding increases the size of the message by approximately 33%, so the value you specify should be approximately 33% larger than the actual message size you want enforced. For example, if you specify a maximum message size value of 64 MB, you can expect a realistic maximum message size of approximately 48 MB.

PR #732, opened 2018-12-07
Add certificate watcher for external certs to reload nginx

Nebukadneza:
In case of TLS_FLAVOR=[mail,cert], the user supplies their own certificates.
However, since nginx is not aware of changes to these files, it cannot
reload itself e.g. when the certs get renewed.

To solve this, let’s add a small daemon in the place of
`letsencrypt.py`, which uses a flexible file-watching framework and
reloads nginx in the case the certificates change ….

### DIFF
--- a/core/nginx/certwatcher.py
+++ b/core/nginx/certwatcher.py
@@ -1,0 +1,63 @@
+#!/usr/bin/python3
+"""
+Certificate watcher which reloads nginx or reconfigures it, depending on what
+happens to externally supplied certificates. Only executed by start.py in case
+of TLS_FLAVOR=[mail, cert]
+"""
+
+from os.path import exists, split as path_split
+from os import system
+import time
+from watchdog.observers import Observer
+from watchdog.events import FileSystemEventHandler, FileDeletedEvent, \
+    FileCreatedEvent, FileModifiedEvent, FileMovedEvent
+
+class ChangeHandler(FileSystemEventHandler):
+    "watchdog-handler listening on any event, executing the correct configuration/reload steps"
+    @staticmethod
+    def reload_nginx():
+        "merely reload nginx without re-configuring everything"
+        if exists("/var/run/nginx.pid"):
+            print("Reloading a running nginx")
+            system("nginx -s reload")
+
+    @staticmethod
+    def reexec_config():
+        "execute a reconfiguration of the system, which also reloads"
+        print("Reconfiguring system")
+        system("/config.py")
+
+    def on_any_event(self, event):
+        "event-listener checking if the affected files are the cert-files we're interested in"
+        if event.is_directory:
+            return
+
+        filename = path_split(event.src_path)[-1]
+        if isinstance(event, FileMovedEvent):
+            filename = path_split(event.dest_path)[-1]
+
+        if filename in ['cert.pem', 'key.pem']:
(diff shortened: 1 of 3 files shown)


core/nginx/certwatcher.py
```diff
@@ -0,0 +1,59 @@
...
+
+if __name__ == '__main__':
+    OBSERVER = Observer()
+    HANDLER = ChangeHandler()
+    OBSERVER.schedule(HANDLER, "/certs", recursive=False)
+    OBSERVER.start()
+    while True:
+        time.sleep(1)
+    OBSERVER.stop()
+    OBSERVER.join()
```
muhlemmer: 1. The `while True` loop will prevent `OBSERVER.stop()` and `OBSERVER.join()` from ever executing. Even is if the script is killed.
2. I don't realy like the CAPS code. I don't really know what is best practice in python globally, but in this project everything is lower-case.

Nebukadneza: 2. ⇒ Okay, it was a `pylint` recommendation, i believe based on PEP-8(?), but not sure about that.
Thanks for 1. ⇒ I overlooked that one.

core/nginx/certwatcher.py
```diff
@@ -0,0 +1,63 @@
...
+    def on_any_event(self, event):
+        "event-listener checking if the affected files are the cert-files we're interested in"
+        if event.is_directory:
+            return
+
+        filename = path_split(event.src_path)[-1]
+        if isinstance(event, FileMovedEvent):
+            filename = path_split(event.dest_path)[-1]
+
+        if filename in ['cert.pem', 'key.pem']:
```
HorayNarea: If I understand this right, nginx will be reloaded/restarted twice if you change the key because a key-rotation implies a cert-rotation -> 2 events get handled

Maybe just watch `cert.pem`? Or am I missing something?

Nebukadneza: The events are queued, and the queue is handled in series, as far as i understand watchdog. 

I chose to watch both, in case some user-sided script is either slow handling them, or is broken and only handles one of them correctly. In this case, the desired behavior is to act upon changes for both of them. Given how quickly config.py executes, and how swiftly nginx reloads internally, I didn’t see much harm in two consecutive quick reloads. However, I do agree with you that it’s not paticularly elegant ….

The better solution would be to wait for a small time (1-3sec or so?) after each event received, queue up the events we receive ourselves, and then act upon the »harshest« of them, disregarding the others. However, this seemed a bit over the top for me, given the already explained very quick reloading times ….

I hope that’s okay with you …

Nebukadneza: Yep, I’m pretty sure that can’t happen ^_^. By the way: big :+1: for such detailed reviews, even of small bits!

muhlemmer:
On a side note: please don't force push a branch that is already submitted in a PR. If someone would have cloned your branch for testing, that person will no longer be able to pull the latest version, due to conflicts. It happened to me the other day on another PR and it was deeply frustrating.

I was guilty to this crime too in the past, but now I know better :wink:

Nebukadneza:
Thank you!

Zetten:
Is there any chance this commit could be cherry-picked to the 1.6 branch? Or is there a new Mailu release planned in the next weeks/months?

Thanks!

Issue #734, opened 2018-12-08
nginx: [emerg] BIO_new_file("/certs/dhparam.pem") failed (SSL: error:02FFF002:system library:func(4095):No such file or directory:fopen('/certs/dhparam.pem', 'r') error:20FFF080:BIO routines:CRYPTO_internal:no such file)


mnooblet:
after installing and switching from notls to letsencrypt my frontend container crashes und keeps restarting with this error..

I can't even attach a console to look into the container

I noticed that the dhparam.pem file is missing in /mailu/certs/letsencrypt/live/mailu

Any ideas?

muhlemmer:
#735 is merged. Image on Docker Hub. Please do:
```
docker-compose pull
docker-compose up -d
```
This should pull the new Nginx image and recreate front.

mnooblet:
yey just pulled the image, seems to work now, thx

PR #735, opened 2018-12-08
fix dhparam.pem shadowing in 1.5

HorayNarea:
In 1.5 if Mailu is setup with `TLS_FLAVOR=notls` and later switched to `cert` or `letsencrypt` `/certs/dhparam.pem` is missing because the directory/volume is created (empty) on first start.

This will lead to the error reported in #734


muhlemmer:
My bet... tested this fix successfully. I will merge it now.

PR #749, opened 2018-12-15
1.5 autobuild

muhlemmer:
In preparation of a multi-version documentation setup, I considered it would be helpful to move 1.5 to travis building also. This is a simplified back port, without any testing.

@kaiyou, If you agree with this, I'll disable the autobuild on Docker hub, before merging this PR. This is to prevent collisions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,2 +1,29 @@
+sudo: required
+services: docker
+addons:
+  apt:
+    packages:
+      - docker-ce
+
+env:
+  - MAILU_VERSION=$TRAVIS_BRANCH
+language: python
+python:
+  - "3.6"
+install:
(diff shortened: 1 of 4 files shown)


muhlemmer:
Good idea ;). Radicale does not build anymore on alpine 3.7, so I've moved it to edge.

kaiyou:
It seems to run just fine on my side with the images built from ``tests/build.yml``, so let's go through with it. This way, we can also benefit from cron builds on 1.5.

muhlemmer:
Ok, I'll disable the build on docker hub and I'll do the merge when done.

kaiyou:
Once done, I believe the compose file from the master branch should run just fine. You actually did 90% of the job setting this up :)

muhlemmer:
I'll probably need to do some small mods to the compose file (as I found with setup). I'll keep the working version on master.

Done with docker hub, will merge this now.

PR #767, opened 2018-12-19
allow all characters for username in dovecot

### LINKED ISSUE #443: ^_^@domain.com is not recognized as a valid email
I cannot login via webmail nor receive email. However ^_^@domain.com should be a valid email address.

hoellen:
fix #443

### DIFF
--- a/core/dovecot/conf/dovecot.conf
+++ b/core/dovecot/conf/dovecot.conf
@@ -64,6 +64,7 @@ plugin {
 ###############
 # Authentication
 ###############
+auth_username_chars =
 auth_mechanisms = plain login
 disable_plaintext_auth = no
 


core/dovecot/conf/dovecot.conf
```diff
@@ -64,6 +64,7 @@ plugin {
 ###############
 # Authentication
 ###############
+auth_username_chars =
```
muhlemmer: Sorry, I'm confused. In the issue you proposed a whole range of characters, and this line is empty? Empty means all? (Normal convention would be none)

hoellen: In the issue I proposed a workaround just for the special case. But we want to allow any character in dovecot and according to the doc the variable should set to empty:
```
# List of allowed characters in username. If the user-given username contains
# a character not listed in here, the login automatically fails. This is just
# an extra check to make sure user can't exploit any potential quote escaping
# vulnerabilities with SQL/LDAP databases. If you want to allow all characters,
# set this value to empty.
#auth_username_chars = abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ01234567890.-_@
```
FTR: Since we use sqlalchemy the chars will be escaped and sql injection should not be a problem.

PR #769, opened 2018-12-20
Fix setup filegen

### LINKED ISSUE #762: Setup: generated files missing
Traefik is ending up in a looped redirect when trying to download the generated files. Flask blueprint context to be modified to simplify the forwarding and eliminate the need for redirects.

muhlemmer:
Fixes #762. WIP cause I still have some doubts about static file routing in traefik.


muhlemmer:
This is ready for review. For testing purposes, this branch is currently hosted on https://setup.mailu.io/fix-setup-filegen/ and is the default for https://setup.mailu.io.

Once this is reviewed, I will put both version containers to master until we have the 1.6 release.

setup/flavors/compose/setup.html
```diff
@@ -11,8 +11,8 @@
 to read and check the configuration variables generated by the wizard.</p>
 
 <pre><code>cd {{ root }}
-curl {{ url_for('.file', uid=uid, filepath='docker-compose.yml', _external=True) }} > docker-compose.yml
-curl {{ url_for('.file', uid=uid, filepath='mailu.env', _external=True) }} > mailu.env
+wget {{ url_for('.file', uid=uid, filepath='docker-compose.yml', _external=True) }}
+wget {{ url_for('.file', uid=uid, filepath='mailu.env', _external=True) }}
```
Nebukadneza: This isn’t in this changeset, but … do I see this right that we let the user download this as `mailu.env`, but in `docker-compose.yml` refer to the env-file as `.env`, without mentioning the fact that the downloaded file should be renamed? Maybe we could guide the user some more at this point…?

setup/server.py
```diff
@@ -10,7 +10,9 @@
 import ipaddress
 
 
-app = flask.Flask(__name__)
+version = os.getenv("this_version")
+static_url_path = "/" + version + "/static"
+app = flask.Flask(__name__,static_url_path=static_url_path)
```
Nebukadneza: nitpick: missing space after `,`, but really, only a nitpick

setup/flavors/compose/setup.html
muhlemmer: In the generated `docker-compose.yml` we now use `mailu.env`:
https://github.com/usrpro/Mailu/blob/c1bfde4e150922745373f848cba7ae58694136f9/setup/flavors/compose/docker-compose.yml#L1

Historically, `.env` is parsed by the `docker-compose` command and any variables in `docker-compose.yml` would get evaluated accordingly. Simultaneously we were passing the same `.env` to the containers.

However, `docker stack deploy` does not support parsing `.env` for variables in `docker-compose.yml`. And since we now have a setup, the user can configure anything from there and the results get hard-coded to the rendered `docker-compose.yml` file. This way we keep implementation more or less the same between the 2 docker flavors.

Because of this design decision, we are no longer bound to the `.env` file name. We can call it whatever we want, as long as we pass the correct name to the containers. We choose to rename it to `mailu.env` because:

1. It's no longer a hidden file, so easier to find and modify for users
2. Makes people aware that we do things differently with setup. (The generated `docker-compose.yml` is not compatible with and old-school `.env`)

muhlemmer: https://github.com/Mailu/Mailu/blob/d483ef3c2a1df250eec6b9457f2270735f0bc909/setup/flavors/compose/docker-compose.yml#L1

Nebukadneza: Whoops, i seem to have overlooked that line/fact. Sorry!

PR #772, opened 2018-12-21
Add docker image prefix in tests and setup utility

dani909:
As stated in https://github.com/Mailu/Mailu/pull/702#issuecomment-437295564 the docker prefix variable also has to be in the docker-compose file of the setup utility and the tests.
Sorry that i forgot them in #702 and was unresponsive for a while.
Also the default value of an empty string is provided because docker-compose is complaining that it isn't set.


muhlemmer:
Thanks for picking this up!

dani909:
Just updated my branch from master

Issue #758, opened 2018-12-18
[Help wanted]: write CHANGELOG.md


muhlemmer:
There are a lot of new features introduced since 1.5.1 and most of them in the last period. Nothing of this was documented in the change log. We could use help if someone could go through the merge pull request and write anything notable in [CHANGELOG.md](https://github.com/Mailu/Mailu/blob/master/CHANGELOG.md).

Investigating the network graph, I believe we need to go back as far as #329.

PR #782, opened 2018-12-30
Changelog for release 1.6

hoellen:
I went through the issues and PRs since 2017-11-21 and wrote a first changelog for the `1.6` release.
From now on we should extend the changelog when we have new features/changes.

*No guarantee of correctness and completeness :P*


hoellen:
#758

muhlemmer:
O my, nice job! I wouldn't blame you if there was a missing entry ;)

tobias992:
Hello,
there are any plans for final 1.6?

tobias992:
Cool,
thank you for the quick answer.
I an looking for an alternative for Mailcow.
Then i found your Mailserver but with last update in 2017. This was the reason for my question.
I tested a little bit with the demo server. Do you have a function for set a rate limit per mail address?


~ Tobias Fronk

⁣BlueMail for Android herunterladen ​

hoellen:
There is a rate limit per source IP address:
https://github.com/Mailu/Mailu/blob/356012a672335a291366a79c363cc47c24ef0930/docs/compose/.env#L42-L43

tobias992:
Okay,
does i understand right that this set the  limit per domain?

⁣BlueMail for Android herunterladen ​

hoellen:
It sets the limit per Client IP address.

tobias992:
I dont understand, sorry for my bad english.
I have an example. I have a shared hosting Webserver with domain1.com (ip: 1.2.3.4) and domain2.com (ip: 1.2.3.4). The Mailserver is setup with Mailu on ip: 5.6.7.8. Limit is set 1000/h.
How the limit is working? All Domains from 1.2.3.4 has the limit 1000/h or both...

In Mailcow i can set a limit per Domain/E-Mail Address... thats why i am ask.

hoellen:
`User A` (client IP) can connect 1000/h to your Mailu setup (doesn't matter which domain: domain1.com or domain2.com)
A second `user B` can also connect 1000/h to your mailu instance for domain1.com and domain2.com in the same time.
And so on... Is this more clear?

PR #785, opened 2018-12-30
Feature: send auto reply with displayed name 

TheLegend875:
This pull request makes it possible to set a displayed name in the UI which is then used by the auto-reply messages.


Nebukadneza:
Tested this, and it works just fine :+1: . Thanks a lot.
As far as I’m familiar with Mailus code-practices, the code looks good too.

PR #799, opened 2019-01-05
don't allow negative values on domain creation/edit

### LINKED ISSUE #798: It's possible to set negative values as mailbox/alias limit for a domain
When adding a new domain it's possible to set negative values for the mailbox/alias limits. While this is rather a **low priority/aesthetic** issue, imho it would be great to disable negative values, since they add no benefit for a user.

grafik

### Steps to reproduce
1) Go to the domain site
2) Add a new domain and edit it
-> It's possible to add a negative amount of users/aliases

hoellen:
fixes #798 
This disable the negative values on domain creation/edit, but also changes previous "unlimited" behavior. That is the reason for the "review/need2"label. 

The "new" unlimited for `max_users` and `max_aliases` is -1 instead of 0. So now we can disable aliases and/or users for a specific domain.


hoellen:
done

core/admin/migrations/versions/fc099bd15cbe_.py
```diff
@@ -0,0 +1,65 @@
...
+                domain_table.c.max_users == 0
+                ).values(
+                    max_users=-1
+                )
+            )
+    connection.execute(
+            domain_table.update().where(
+                domain_table.c.max_aliases == 0
+                ).values(
+                    max_aliases=-1
```
Nebukadneza: Can we maybe help users who have accidentally mis-configured their domains by forcing everything `<=0` to `-1`? Although then the downgrade path wouldn’t be clean anymore …. And maybe the user would be surprised since while he accidentally entered a negative number, he didn’t necessarily *mean* infinity?
Oh well, is hard — I guess staying as deterministic like this is the best after all …

hoellen: Yes, I had this thought too.  But then I thought that maybe previously someone wanted to disable aliases for a domain (`0` was infinity, and e.g. -5 disabled aliases). 
And also the use case you mentioned is against the `<=0` migration. So I think you're right, we should keep the way it is ...
Thanks for the additional input!

Issue #808, opened 2019-01-07
Fetchmail sslproto configuration (typo?)


rueckix:
The fetchmail container repeatedly fires the following warning

`fetchmail: Invalid SSL protocol 'AUTO' specified, using default (SSLv23).`

From the fetchmail manpage, this appears to relate to the `--sslproto` switch. 

IMHO the default should be an all lowercase `auto`. Otherwise, an improvement suggestion would be to make this configurable through the admin interface.

PR #809, opened 2019-01-07
Update fetchmail.py

rueckix:
## What type of PR?
Typo in fetchmail configuration flag sslproto

## What does this PR do?
Change sslproto=AUTO to sslproto=auto 

### Related issue(s)
#808 

## Prerequistes
None

### DIFF
--- a/services/fetchmail/fetchmail.py
+++ b/services/fetchmail/fetchmail.py
@@ -22,7 +22,7 @@ poll "{host}" proto {protocol}  port {port}
     is "{user_email}"
     smtphost "{smtphost}"
     {options}
-    sslproto 'AUTO'
+    sslproto 'auto'
 """
 
 


Nebukadneza:
Sorry for the late reaction! Does this indeed fix the problem for you? In my test, and apparently also others in the matrix chat, `fetchmail` whines about `auto` as well as `AUTO`, which is kind of a bummer. The underlying problem is also that fetchmail hasn’t gotten a release in ~5 years(?) …

rueckix:
Apparently, it does not solve the issue.  Never trust the man page.. it seems. I looked into the fetchmail code and indeed, this does not appear to make a difference on the latest official release. Glad to see that newer releases are in the works. 

Checking this switch again, I am wondering if "AUTO" is a safe default choice. It makes fetchmail fall back to SSL23, which uses deprecated SSLv23 methods in OpenSSL.
(see https://gitlab.com/fetchmail/fetchmail/blob/RELEASE_6-3-26/socket.c#L928)

Same issue for 'ssl2' and `ssl3`.

The only safe option, from a security point of view, is `sslproto 'tls1'`

Opinions?

muhlemmer:
TLS 1.0 is deprecated from June 30, 2018: https://en.wikipedia.org/wiki/Transport_Layer_Security#TLS_1.0.

Although, some "mayor" providers will still support it until march 2020. Mailu doesn't:
https://github.com/Mailu/Mailu/blob/effaaa90621b79ee3fd2c4c47ec4393ea381e66b/core/nginx/conf/tls.conf#L1

So setting `sslproto 'tls1'` might work in the cases of some of the "mayor" service providers like Google and some other non-security aware mail servers. But it won't work against Mailu or any other up-to-date server.

The opposite argument would be that currently Fetchmail doesn't work at all, and this setting would at least make it work in some cases. So to conclude, I'm indifferent in this matter. Maybe some other opinions from @Nebukadneza, @kaiyou or @HorayNarea?

muhlemmer:
Fetchmail has our full attention and we'll try to fix it as soon as fetchmail allows us. For now this PR will be closed.

PR #814, opened 2019-01-09
Deliver mails to alias-stripped-of-delimeter, even if catchall exists

### LINKED ISSUE #813: Wildcard aliases override concrete aliases + delimiter
After #778 , the following scenario behaved in a non-intuitive way:
Aliases:
* `foo@xxx.tld` ⇒ user**A**
* `%@xxx.tld` ⇒ user**B**
Sending to both `foo@xxx.tld` and `random@xxx.tld` work, however `foo+spam@xxx.tld` gets delivered to user**B**. This is not intuitive to the user.

Nebukadneza:
This fixes delivery to an alias minus recipient delimiter in cases where a
wildcard alias would also match. For example,
* foo@xxx.tld
* %@xxx.tld
Sending to foo+spam@xxx.tld would get eaten by the catchall before this fix.
Now, the order of alias resolution is made clearer.

closes #813

## What type of PR?
bug-fix

## What does this PR do?
This re-orders the resolution of aliases, such that more specific aliases are preferred over wildcards — even if it means stripping the recipient-delimited part.

### Related issue(s)
closes #813

## Prerequistes
  - Not required: Bugfix restores intuitive behavior
- [X] Unless it's docs or a minor change: place entry in the [changelog](CHANGELOG.md), under the latest un-released version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -111,6 +111,7 @@ v1.6.0 - unreleased
 - Bug: Fix rainloop permissions ([#637](https://github.com/Mailu/Mailu/issues/637))
 - Bug: Fix broken webmail and logo url in admin ([#792](https://github.com/Mailu/Mailu/issues/792))
 - Bug: Don't recursivly chown on mailboxes ([#776](https://github.com/Mailu/Mailu/issues/776))
+- Bug: Don't deliver alias and recipient-delimiter to overwhelming wildcard-alias ([#813](https://github.com/Mailu/Mailu/issues/813))
 
 v1.5.1 - 2017-11-21
 -------------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -111,7 +111,6 @@ v1.6.0 - unreleased
 - Bug: Fix rainloop permissions ([#637](https://github.com/Mailu/Mailu/issues/637))
 - Bug: Fix broken webmail and logo url in admin ([#792](https://github.com/Mailu/Mailu/issues/792))
 - Bug: Don't recursivly chown on mailboxes ([#776](https://github.com/Mailu/Mailu/issues/776))
-- Bug: Don't deliver alias and recipient-delimiter to overwhelming wildcard-alias ([#813](https://github.com/Mailu/Mailu/issues/813))
 
 v1.5.1 - 2017-11-21
 -------------------
--- a/core/admin/mailu/models.py
+++ b/core/admin/mailu/models.py
@@ -260,24 +260,14 @@ class Email(object):
 
     @classmethod
     def resolve_destination(cls, localpart, domain_name, ignore_forward_keep=False):
-        localpart_stripped = None
-        stripped_alias = None
-
+        localpart_maybe_split = localpart
         if os.environ.get('RECIPIENT_DELIMITER') in localpart:
-            localpart_stripped = localpart.rsplit(os.environ.get('RECIPIENT_DELIMITER'), 1)[0]
+            localpart_maybe_split = localpart.rsplit(os.environ.get('RECIPIENT_DELIMITER'), 1)[0]
 
-        pure_alias = Alias.resolve(localpart, domain_name)
-        has_wildcard = pure_alias and '%' in pure_alias.email and pure_alias.wildcard
-        stripped_alias = Alias.resolve(localpart_stripped, domain_name)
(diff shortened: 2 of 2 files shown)


core/admin/mailu/models.py
```diff
@@ -261,14 +261,23 @@ def resolve_domain(cls, email):
...
+        stripped_alias = Alias.resolve(localpart_stripped, domain_name)
+
+        if pure_alias and not has_wildcard:
+            return pure_alias.destination
+        elif not pure_alias and stripped_alias:
+            return stripped_alias.destination
+        elif has_wildcard and localpart_stripped and stripped_alias:
+            return stripped_alias.destination
+        elif pure_alias:
+            return pure_alias.destination
```
hoellen: This seems way to complicated to me. Why we don't just always split the localpart at the recipient delimiter and make one alias lookup? 

With your example from #813: 
- `foo@xxx.tld` ⇒ User A
- `%@xxx.tld` ⇒ User B

If we now send an email to ` foo+spam@xxx.tld` after the split we look for the alias `foo@xxx.tld`. And if someone sends directly to `foo@xxx.tld` the split doesn't change anything and we resolve `foo@xxx.tld`. 
So we could reduce this to a few lines of code and do just one database lookup.

Or am I missing here something? :man_shrugging:

Nebukadneza: My thought was that maybe a user could want to re-route `foo+spam@xxx.tld` a a later time due to his/her changed personal preferences … then an alias like that should take precedence over `alias foo with delimiter stripped`. If my thought on this is a bit far-fetched, I’ll gladly build it to look easier again.

CHANGELOG.md
```diff
@@ -111,6 +111,7 @@ v1.6.0 - unreleased
 - Bug: Fix rainloop permissions ([#637](https://github.com/Mailu/Mailu/issues/637))
 - Bug: Fix broken webmail and logo url in admin ([#792](https://github.com/Mailu/Mailu/issues/792))
 - Bug: Don't recursivly chown on mailboxes ([#776](https://github.com/Mailu/Mailu/issues/776))
+- Bug: Don't deliver alias and recipient-delimiter to overwhelming wildcard-alias ([#813](https://github.com/Mailu/Mailu/issues/813))
```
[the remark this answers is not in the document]
Nebukadneza: Ack, sorry

core/admin/mailu/models.py
hoellen: I think we shouldn't allow this kind of alias. Postfix doesn't allow the delimiter in aliases too. So imo we should do it the "postfix-way".

Nebukadneza:
This yields almost the desired behavior, as described above, sans the »i want to override an alias that has been used with delimiter so far to be something else« case — which you are right, might not be as important …

core/admin/mailu/models.py
```diff
@@ -260,13 +260,12 @@ def resolve_domain(cls, email):
 
     @classmethod
     def resolve_destination(cls, localpart, domain_name, ignore_forward_keep=False):
-        localpart_stripped = None
+        localpart_maybe_split = localpart
```
hoellen: This breaks the user query in line 274. I would suggest to leave the name of the variable as `localpart_stripped`.

core/admin/mailu/models.py
```diff
@@ -261,14 +261,21 @@ def resolve_domain(cls, email):
...
         if os.environ.get('RECIPIENT_DELIMITER') in localpart:
             localpart_stripped = localpart.rsplit(os.environ.get('RECIPIENT_DELIMITER'), 1)[0]
 
-        alias = Alias.resolve(localpart, domain_name)
-        if not alias and localpart_stripped:
-            alias = Alias.resolve(localpart_stripped, domain_name)
-        if alias:
-            return alias.destination
+        pure_alias = Alias.resolve(localpart, domain_name)
+        pure_alias_has_wildcard = pure_alias and '%' in pure_alias.email and pure_alias.wildcard
```
hoellen: Sorry, one last thing for further clean up: 
I think we can get rid of line 270 and use in L273 just `pure_alias.wildcard`

Nebukadneza: Yep, you’re right. If the user clicks 'sql like syntax' without actually using it, it’s probably better the behavior is clear instead of smart — where smart has been = check if `%` is actually used. thx!

hoellen:
Nice job!

Issue #1, opened 2016-02-20
Manage user authentication and permissions


kaiyou:
Currently no authentication is implemented. Multiple issues will have to be tackled:

 - complete permission scheme or simple admin role plus admins per domain?
 - how to store user passwords (shared format between Flask-admin and dovecot)?
 - how should the initial use be created?

kaiyou:
Fix by 4576e1f5c2bd8fbdd8c706f435c9e4718dbc384d.

PR #826, opened 2019-01-11
Reverse the resolution order of user and alias

### LINKED ISSUE #815: Configuration to rank users higher than aliases
Currently, when resolving the localpart of a domain, `admin` first tries aliases, and if no alias is found, users are searched:
```
        alias = Alias.resolve(localpart, domain_name)
        if not alias and localpart_stripped:
            alias = Alias.resolve(localpart_stripped, domain_name)
        if alias:
            return alias.destination

        user = User.query.get('{}@{}'.format(localpart, domain_name))
        if not user and localpart_stripped:
            user = User.query.get('{}@{}'.format(localpart_stripped, domain_name))
        if user:
            if user.forward_enabled:
                destination = user.forward_destination
                if user.forward_keep or ignore_forward_keep:
                    destination.append(user.email)
            else:
                destination = [user.email]
        return destination
```
This may be counter-intuitive to some users: After all, wildcards often behave the way that they are considered only after all more-specific options are considered and dismissed. Here, it’s vice-versa: If there’s a catchall alias (`%@xxx.tld`), then all users under `xxx.tld` are not reachable anymore.
One use-case that is not possible here would be: A domain-owner would like catchall for himself, but give specific names to services, relatives, friends, etc.: The only workaround for this is currently to create these accounts on another domain, and then create an alias on `xxx.tld` to them.

Since changing the behavior of that function might break existing users setups, I’d follow the suggestion of `braintopping` from matrix: Using a kind of feature-flag or configuration variable to switch the behavior of preferring users vs. preferring aliases.

What does everyone think about that? I’d be grateful for some feedback~.

Nebukadneza:
Since it’s common for wildcard~ish systems to prefer concrete objects over
wildcards, and aliases can be broad-wildcards (think catchall, %@xxx.tld), it
may be more intuitive for users that user-names rank higher than aliases. This
makes it impossible for user-names to be unreachable, since they can be
completely overridden by a catchall otherwise.

This changes default behavior, and is not configurable.

closes #815

## What type of PR?

### Related issue(s)
- Mention an issue like: #001

## Prerequistes

### DIFF
--- a/core/admin/mailu/models.py
+++ b/core/admin/mailu/models.py
@@ -266,16 +266,6 @@ class Email(object):
         if os.environ.get('RECIPIENT_DELIMITER') in localpart:
             localpart_stripped = localpart.rsplit(os.environ.get('RECIPIENT_DELIMITER'), 1)[0]
 
-        pure_alias = Alias.resolve(localpart, domain_name)
-        stripped_alias = Alias.resolve(localpart_stripped, domain_name)
-
-        if pure_alias and not pure_alias.wildcard:
-            return pure_alias.destination
-        elif stripped_alias:
-            return stripped_alias.destination
-        elif pure_alias:
-            return pure_alias.destination
-
         user = User.query.get('{}@{}'.format(localpart, domain_name))
         if not user and localpart_stripped:
             user = User.query.get('{}@{}'.format(localpart_stripped, domain_name))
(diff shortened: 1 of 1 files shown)


muhlemmer:
> Behavior is, i think, in line with 1.5, where postfix still directly queried the database, so — my gut feeling would be not to update the changelog(?)

Fine by me

PR #831, opened 2019-01-13
Unify and coerce booleans from env used in admin

### LINKED ISSUE #830: With DOMAIN_REGISTRATION set to `false`, the registration ui is enabled
Looks like it doesn’t matter what it’s set to — as long as its string pythonically evaluates to `True`, it’ll be shown. That’s AFAIK every string not `""`, right?

That said, it’s probably true for many other Boolean config flags, as no data-type aware casting is done in `core/admin/mailu/configuration.py`. Maybe a more general solution would be nice?

Nebukadneza:
At some places, the string that DOMAIN_REGISTRATION is got used like a boolean
(an easy misassumption to make while in python and dealing with the config
dict), making `DOMAIN_REGISTRATION=False` act as a truthy value. To stop such
future problems from happening, coerce environment config strings to real
bools.

closes #830

## What type of PR?
Bug-fix
### Related issue(s)
closes #830

## Prerequistes
  - not needed
- [X] Unless it's docs or a minor change: place entry in the [changelog](CHANGELOG.md), under the latest un-released version.


ionutfilip:
Looks good

PR #788, opened 2018-12-31
Multiple databases support and postgresql image

muhlemmer:
This PR adds the option of using Mysql and Postgresql databases. Migrations and types had to be changed in various models. Credit to @kaiyou for doing this.

Users can now choose to configure an external database, use the default SQLite or the included postgresql image.

The provided Postgresql image has the following features:

- Automatic creation of users, db etc.
- Connection password is SECRET_KEY and is used by admin by default
- TCP connections are only allowed from the mailu SUBNET
- Healthcheck
- Automatic minutely "wal archiving" and weekly pg_basebackup.
- Automatic cleaning of wal_archive and "base backups". Two versions always remain available
- When /data is empty and backups are present, the backups are restored automatically. Useful in swarm environments, since the /data directory should not be on any network filesystem (performance)

This PR supersedes #729 and it's a merge from @kaiyou's work. Local building is tested, bu we will commence functional tests after the new year. If everything goes well, this can still be included in 1.6.

## Todo:
- [x] Entry in changelog

## Testing checklist

- [x] Make it work with default `.env` and `docker-compose.yml` from master
- [x] Check correctness of docs

- [x] SQLite migrations: Fresh install
- [x] SQLite migrations: Upgrade from `1.5`
- [x] SQLite migrations: Upgrade from `master`
- [x] SQLite run: Admin functionality
- [x] SQLite run: Send / receive mail
- [x] SQLite run: Send to alias
- [x] SQLite run: Auto forward
- [x] SQLite run: Auto reply
- [x] SQLite run: Fetchmail

- [x] PostgreSQL migrations: Fresh install
- [x] PostgreSQL run: Admin functionality
- [x] PostgreSQL run: Send / receive mail
- [x] PostgreSQL run: Send to alias
- [x] PostgreSQL run: Auto forward
- [x] PostgreSQL run: Auto reply
- [x] PostgreSQL run: Fetchmail

- [x] MySQL migrations: Fresh install
- [x] MySQL run: Admin functionality
- [x] MySQL run: Send / receive mail
- [x] MySQL run: Send to alias
- [x] MySQL run: Auto forward
- [x] MySQL run: Auto reply
- [x] MySQL run: Fetchmail

- [x] PostgrSQL image: WAL archiving
- [x] PostgrSQL image: Auto backups
- [x] PostgrSQL image: Auto cleanup
- [x] PostgrSQL image: Restore operation

PR #835, opened 2019-01-13
Change quota columns type to BigInteger 

hoellen:
## What type of PR?
bug-fix

## What does this PR do?
As described in this [comment](https://github.com/Mailu/Mailu/pull/788#issuecomment-453690849) we need a bigger type for columns which store the size of quota. 
With this PR I can successfully migrate an existing sqlite3 installation to MySQL/MariaDB. :tada: 

@muhlemmer  #788 is accidentally merged already and this PR does not do a migration. But since external databases are considered as experimental and it is only merged a few days it is OK for me. So we don't have to mess up the migration. 

## Prerequistes

### DIFF
--- a/core/admin/mailu/models.py
+++ b/core/admin/mailu/models.py
@@ -131,7 +131,7 @@ class Domain(Base):
         backref=db.backref('manager_of'), lazy='dynamic')
     max_users = db.Column(db.Integer, nullable=False, default=-1)
     max_aliases = db.Column(db.Integer, nullable=False, default=-1)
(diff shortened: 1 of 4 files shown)


Nebukadneza:
Sorry for my bad understanding of the sqlalchemy/alembic migrations … Since you change existing migrations, that does mean that for example my dogfooding instance on current `master` would need these manually updated?

On a fresh installation they work fine ….

hoellen:
@Nebukadneza I did not do a migration, because it is only for new external databases (e.g. MySQL) and for this I oriented myself at #788. Since this feature is very new and experimental I think we can just change it for new created (external) databases.

Issue #841, opened 2019-01-17
Admin container SQLite database URI is incorrect


nathan-sain:
PR #788 changed [core/admin/mailu/configuration.py](https://github.com/Mailu/Mailu/blob/master/core/admin/mailu/configuration.py#L69) to set the SQLite database URI as 'sqlite:////{DB_HOST}' which results in the database being stored in the file /database of the admin container. This breaks persistence across container rebuilds when following the documented practice of mounting a persistent volume on /data.

PR #842, opened 2019-01-17
Update SQLite DB file in DEFAULT_CONFIG and DB_TEMPLATES

nathan-sain:
## What type of PR?

bug-fix
## What does this PR do?

Fix for issue: #841

### DIFF
--- a/core/admin/mailu/configuration.py
+++ b/core/admin/mailu/configuration.py
@@ -19,7 +19,7 @@ DEFAULT_CONFIG = {
     'DB_HOST': 'database',
     'DB_NAME': 'mailu',
     'SQLITE_DATABASE_FILE':'data/main.db'
-    'SQLALCHEMY_DATABASE_URI': '',
+    'SQLALCHEMY_DATABASE_URI': 'sqlite:////data/main.db',
     'SQLALCHEMY_TRACK_MODIFICATIONS': False,
     # Statistics management
     'INSTANCE_ID_PATH': '/data/instance',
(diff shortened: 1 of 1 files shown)


core/admin/mailu/configuration.py
```diff
@@ -18,7 +18,8 @@
     'DB_PW': None,
     'DB_HOST': 'database',
     'DB_NAME': 'mailu',
-    'SQLALCHEMY_DATABASE_URI': 'sqlite:////data/main.db',
+    'SQLITE_DATABASE_FILE':'data/main.db'
+    'SQLALCHEMY_DATABASE_URI': '',
```
muhlemmer: Tests fail on this line:
https://travis-ci.org/Mailu/Mailu/builds/480729143#L3158

muhlemmer:
The following code only gets executed when `DB_FLAVOR` is set in the environment:

https://github.com/Mailu/Mailu/blob/6dea8b422a525f1cf03d1afc181128cae2050635/core/admin/mailu/configuration.py#L93-L95

From that, may I assume you generated your config with setup utility? Our "Classic" `.env` does not have the variable set. So that snippet never gets executed. This also makes the auto build fail.

In order to keep backwards compatibility, `SQLALCHEMY_DATABASE_URI` cannot be left empty by default.

core/admin/mailu/configuration.py
nathan-sain: OK. I updated the patch to leave the SQLite URI as the value for SQLALCHEMY_DATABASE_URI

Issue #858, opened 2019-01-21
Using setup to generate docker-compose has no antispam option


bladeswords:
There is no antispam option in the docker compose setup wizard.  The value in the resulting docker environment file is always:

`ANTISPAM=none`

Not sure if this would actually change if the antispam container/service is deployed as I found this value and changed it.

bladeswords:
To add the same comment which was added to the pull request.

I think my description was incomplete.

My issue was with the setup wizard, for example - http://setup.mailu.io/1.6/file/867459ae-365a-4830-9c0d-2e007db01997/mailu.env

Then under 'Optional features' anti-spam is listed with the value of 'none'.  This gives the impression that it is in fact optional, and that the feature has been disabled.  But you rightfully state @muhlemmer , anti-spam can't be optional.

![image](https://i.imgur.com/ODrYpIA.png)

So this could be better addressed by changing the mailu.env template to reflect that this isn't optional and that setting it to 'none' will still result in a anti-spam feature being installed?

PR #859, opened 2019-01-21
Add in antispam option in online setup wizard

bladeswords:
## What type of PR?

bug-fix

## What does this PR do?
My attempt at fixing a bug - may not be complete
### Related issue(s)
- Issue raised is #858 

## Prerequistes

- [ X ] In case of feature or enhancement: documentation updated accordingly
- [ X ] Unless it's docs or a minor change: place entry in the [changelog](CHANGELOG.md), under the latest un-released version.

### DIFF
--- a/setup/flavors/compose/mailu.env
+++ b/setup/flavors/compose/mailu.env
@@ -64,8 +64,9 @@ WEBDAV={{ webdav_enabled or 'none' }}
 # Antivirus solution (value: clamav, none)
 #ANTIVIRUS={{ antivirus_enabled or 'none' }}
 
-#Antispam solution (value: rspamd, none)
-ANTISPAM={{ antispam_enabled or 'none'}}
+# This feature is required for postfix to function
+# Anti-spam solution 
+ANTISPAM=true
 
 ###################################
 # Mail settings
--- a/setup/flavors/compose/mailu.env
+++ b/setup/flavors/compose/mailu.env
@@ -64,7 +64,7 @@ WEBDAV={{ webdav_enabled or 'none' }}
 # Antivirus solution (value: clamav, none)
 #ANTIVIRUS={{ antivirus_enabled or 'none' }}
 
-#Antispam solution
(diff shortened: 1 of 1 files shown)


muhlemmer:
Anti-spam can't be optional. Postfix will fail when it can't connect to rspamd... (Or have you made it run successfully somehow?)

bladeswords:
Hmmm, I think my description was incomplete.  Also, perhaps it would be better to change the generated mailu.env file then what is proposed in this patch.

My issue was with the setup wizard, for example - http://setup.mailu.io/1.6/file/867459ae-365a-4830-9c0d-2e007db01997/mailu.env

Then under 'Optional features' anti-spam is listed with the value of 'none'.  This gives the impression that it is in fact optional, and that the feature has been disabled.  But you rightfully state @muhlemmer , it can't be optional.

image

So this could be better addressed by changing the mailu.env template to reflect that this isn't optional and that setting it to 'none' will still result in a anti-spam feature being installed?

Nebukadneza:
Sorry to come with a discussion so late to the game, but … do I see it correctly that the `ANTISPAM` environment-variable is neither checked nor used anywhere anymore? Rather than setting it to a lighthouse-value just to be seen, should we maybe remove it?

bladeswords:
I was actually thinking that myself - maybe there are plans in the future to allow it to be customised, but if it isn't optional...just remove it

muhlemmer:
We chose to let commented-out values in the `mailu.env` file, so that people with past experience of Mailu are informed what the value is set to in `docker-compose.yml`. And yet, they would see that changing it in `mailu.env` is pointless, as it is commented out.

The fact that `antispam` is set to `none` is a remainder from a very early  and experimental version of setup (we forgot to take it out).

Perhaps its indeed better to take out all the commented fields, as it already starts to lead to confusion.

muhlemmer:
I've created a general issue to cleanup `mailu.env`, as a reminder. This PR can now be closed.

PR #877, opened 2019-01-24
fix some cli commands in the docs

### LINKED ISSUE #849: Cannot create admin password on kubernetes
I cannot do python manage.py admin root domain.ltd passw0rd.
Once into the pod I had to switch to the /app/mailu path.
python command is not available but I tried with python3, pythin3.6 and python3.6m and each time I got the message; 

```
  File "manage.py", line 1, in <module>
    from mailu import models
ModuleNotFoundError: No module named 'mailu'
```

### LINKED ISSUE #925: command documentation uses `_` where `-` should be used
https://mailu.io/1.6/cli.html shows commands

```
alias
alias_delete
user
user_import
user_delete
config_update
```

where `flask mailu` shows the following subcommands:

```
Commands:
  admin          Create an admin user
  advertise      Advertise this server against statistic services.
  alias          Create an alias
  alias-delete   delete alias
  config-update  sync configuration with data from YAML-formatted stdin
  domain
  setlimits      Set domain limits
  setmanager     Make a user manager of a domain
  user           Create a user
  user-delete    delete user
  user-import    Import a user along with password hash.
```

I think the documentation must be updated to use `-` rather than `_`.

hoellen:
## What type of PR?
documentation

## What does this PR do?
Since we updated the python package click to 7.0, commands with underscores in name are replaced by a dash. This PR fixes this in the documentation. 
It also fixes two `python manage.py [...]` leftovers. 

In addition I think we should rename the arguments now as well and replace underscores with dashes, e.g. `--max_users` to `--max-users`. So we keep it consistent. What do you think? I'd update this PR if you agree on this. 

### Related issue(s)
closes #849
closes #925

## Prerequistes

## Todo
- [x] #911 
- [x] #926 
- [x] rename options (`_` to `-`)
- [x] domain command

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -4,11 +4,11 @@ Mailu command line
 Managing users and aliases can be done from CLI using commands:
 
 * alias
-* alias_delete
+* alias-delete
 * user
-* user_import
-* user_delete
-* config_update
+* user-import
+* user-delete
+* config-update
 
 alias
 -----
@@ -18,12 +18,12 @@ alias
   docker-compose exec admin flask mailu alias foo example.net "mail1@example.com,mail2@example.com"
 
 
-alias_delete
+alias-delete
 ------------
 
 .. code-block:: bash
 
-  docker-compose exec admin flask mailu alias_delete foo@example.net
+  docker-compose exec admin flask mailu alias-delete foo@example.net
 
 user
 ----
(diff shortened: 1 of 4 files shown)


muhlemmer:
> In addition I think we should rename the arguments now as well and replace underscores with dashes, e.g. --max_users to --max-users. So we keep it consistent. What do you think? I'd update this PR if you agree on this.

I agree.

upMKuhn:
`flask mailu admin root domain.com password`

This command doesn't work inside a kubernetes pod. It comes back with ` (sqlite3.OperationalError) database is locked ` :(

docs/cli.rst
```diff
@@ -32,30 +32,30 @@ user
...
-user_import
+user-import
 -----------
 
 primary difference with simple `user` command is that password is being imported as a hash - very useful when migrating users from other systems where only hash is known.
 
 .. code-block:: bash
 
-  docker-compose run --rm admin python manage.py user --hash_scheme='SHA512-CRYPT' myuser example.net '$6$51ebe0cb9f1dab48effa2a0ad8660cb489b445936b9ffd812a0b8f46bca66dd549fea530ce'
+  docker-compose run --rm admin flask mailu user-import --hash_scheme='SHA512-CRYPT' myuser example.net '$6$51ebe0cb9f1dab48effa2a0ad8660cb489b445936b9ffd812a0b8f46bca66dd549fea530ce'
```
zoldar: Syntax for this command seems to be invalid. AFAIK that should be:

```bash
docker-compose run --rm admin flask mailu user-import myuser example.net '$6$51ebe0cb9f1dab48effa2a0ad8660cb489b445936b9ffd812a0b8f46bca66dd549fea530ce' 'SHA512-CRYPT' 
```

I have made a PR for that particular change (https://github.com/Mailu/Mailu/pull/923) before stumbling upon this one. Please feel free to close mine and incorporate the change to reduce the noise.
```

File: docs/cli.rst [hunk 2]
```diff
@@ -32,30 +32,30 @@ user
...
-config_update
+config-update
 -------------
 
 The sole purpose of this command is for importing users/aliases in bulk and synchronizing DB entries with external YAML template:
 
 .. code-block:: bash
 
-  cat mail-config.yml | docker-compose exec admin flask mailu config_update --delete_objects
+  cat mail-config.yml | docker-compose exec admin flask mailu config-update --delete_objects
```
tgpfeiffer: In my environment, this requires `-T` option for `docker-compose exec`, see #926.

Also, it seems that `--delete_objects` is not a boolean flag, but requires some arguments:
> Error: --delete_objects option requires an argument

PR #844, opened 2019-01-17
IPv6 stuff

muhlemmer:
## What type of PR?

(Feature, enhancement, bug-fix, documentation) -> A bit of everything

## What does this PR do?

Document how to use ipv6nat. This, however triggers some kind of flaky behavior with the Docker DNS resolver, resulting in lookup failures inside Podop. So this is fixed in theis PR as well... We now do `admin` resolving once, during startup. (As we did with other services).

Note: `docker-compose.yml` downgrade is necessary, as IPv6 settings are not supported by the Docker Compose file format 3 :disappointed:  

Credit to @cooperaj for finding this ipv6nat tool. And @robbertkl for writing it in the first place!

### Related issue(s)
- Fixes #827 
- Hopefully helps with #829 and #834

## Prerequistes
Before we can consider review and merge, please make sure the following list is done and checked.
If an entry in not applicable, you can check it or remove it from the list.

- [x] In case of feature or enhancement: documentation updated accordingly
- [x] Unless it's docs or a minor change: place entry in the [changelog](CHANGELOG.md), under the latest un-released version.

PR #834, opened 2019-01-13
allow ipv6 addresses for tokens

hoellen:
## What type of PR?
Allows you to create an Authentication Token with IPv6 address restriction.

enhancement
## What does this PR do?
closes #829 

## Prerequistes
Before we can consider review and merge, please make sure the following list is done and checked.
If an entry in not applicable, you can check it or remove it from the list.

- [x] In case of feature or enhancement: documentation updated accordingly
- [x] Unless it's docs or a minor change: place entry in the [changelog](CHANGELOG.md), under the latest un-released version.

Issue #829, opened 2019-01-12
Cannot create an Authorisation Token for an IPv6 address


HorayNarea:
On `/admin/ui/token/create` when one enters an IPv6 address it responds with "invalid ip address" because wtforms.validators.IPAddress understands IPv4 only.

I'm also not sure if the other parts that handle auth tokens support IPv6.

hoellen:
Can you please test #834, since I don't have an IPv6 address or proxy to test it.

HorayNarea:
Well… Now I can setup an auth token for a IPv6 address but authentication via the token doesn't work:

`front_1      | 2019/01/13 18:46:04 [info] 11#11: *108 client login failed: "Authentication credentials invalid" while in http auth state, client: 192.168.203.1 using starttls, server: 0.0.0.0:587, login: "root@example.com"`

I'm not sure why the IPv4 client address is shown even if I explictly connect to the IPv6 address :/

hoellen:
`192.168.203.1` is the docker network gateway address. I think that's another issue, so restricting tokens with an ipv4 addresses shouldn't work either, because mailu don't get the real ip of the client. :see_no_evil:

HorayNarea:
Nope, IPv4 works fine.
If I connect to the servers IPv4 address then the correct IPv4 address of the client appears in the logs.

hoellen:
Ok it seems that docker routes IPv6 traffic through a userland proxy. I don't have IPv6 on dockers internal network enabled, maybe this could help?
I find many posts out there, which have the same problem. For the most people this seems to work: 
- enable ipv6 on docker
- set container privilege to `network_mode: host`

But I think that binding the host network to our front container is too overkill and we're loosing the network isolation.
Maybe here is someone with more experience in it and can share useful information.

muhlemmer:
#827 is probably related. I'm seeing also refused connections on IPv6. I'm investigating this. Prbably #827 needs to be resolved before this one.

muhlemmer:
This is a good summary of everything that's wrong with Docker's IPv6 stack:
https://github.com/robbertkl/docker-ipv6nat/blob/master/README.md

A possible workaround is to use that mentioned tool. I've verified Mailu works on IPv6, however you'll be exposed to #827.

PR #886, opened 2019-01-25
Ipv6 support

### LINKED ISSUE #827: Authentication issues when using IPv6 via robbertkl/ipv6nat
I'm using robbertkl/ipv6nat to simplify my IPv6 management and it's causing issues with auth due to the new $SUBNET requirement for things. 

TLDR. I believe that IPv6 causes issues for containers that communicate over IPv6 since the authentication is based on the IPv4 $SUBNET property.

Using robbertkl/ipv6nat lets me treat IPv6 on my server in an identical fashion to my normal IPv4 services. It does standard NAT using the iptables firewall so that opening ports on a container works for both standards without changes.

The issue is that, on an apparent random basis (I think) the internal traffic uses the ULA local addresses to communicate amongst containers and as this does not match the $SUBNET things sometimes don't work. The most noticeable service with this issue is the Rspamd UI which spews near constant error messages, which when viewed in the inspector, are 403 errors. 

My (abbreviated) compose file is:

```
version: '2.1'

services:

  front:
    image: mailu/nginx:$VERSION
    restart: always
    env_file: .env
    logging:
      driver: $LOG_DRIVER
    ports:
    - "80:80"
    - "443:443"
    - "110:110"
    - "143:143"
    - "993:993"
    - "995:995"
    - "25:25"
    - "465:465"
    - "587:587"
    volumes:
      - "$ROOT/certs:/certs"
      - "$ROOT/overrides/nginx:/overrides"

... other services ...

networks:
  default:
    enable_ipv6: true
    ipam:
      driver: default
      config:
        - subnet: $SUBNET
        - subnet: $SUBNET6
```

As you can see I've added some IPv6 config to the network which is needed for the ipv6nat container to be able to route to the services.

The reason I setup the system this way (and not even because I tend to always do it) is because without doing this the standard docker-compose file you recommend would have my IPv6 services be inaccessible. I was not able to have any service connect to them. This is on a standard Ubuntu 18.04 VM with the latest vendor Docker installed and the UFW firewall running. Using the ipv6nat tool allows me to keep the firewall without changes as docker then manages the IPv6 the same way as IPv4.

muhlemmer:
## What type of PR?

(Feature, enhancement, bug-fix, documentation) -> A bit of everything

## What does this PR do?

Document how to use ipv6nat. This, however triggers some kind of flaky behavior with the Docker DNS resolver, resulting in lookup failures between containers.  So all resolving needs to be done during container startup/configuration.

In order not to pollute every single start.py file, we've created a small library called [Mailu/MailuStart](https://github.com/Mailu/MailuStart). As an addition, this library also defines the template generation function, including its logging facility.

Note: `docker-compose.yml` downgrade is necessary, as IPv6 settings are not supported by the Docker Compose file format 3 :disappointed:  

### Related issue(s)
Supersedes  PR #844
- Fixes #827 
- Hopefully helps with #829 and #834

## No backport yet

This PR directly imports MailuStart from git. This makes it a bit more simple to implement on the short term an do some testing and probably some future improvements. When everything is proved stable, we will create a proper PyPi package with versioning and consider back porting.

## Prerequistes


muhlemmer:
Please note that the build failure is a result of an old `.travis.yml` in this branch and can be ignored. (Bors build succeeded).

PR #890, opened 2019-01-26
Make WEBROOT_REDIRECT configurable in setup

### LINKED ISSUE #860: Using setup to generate docker-compose WEBROOT_REDIRECT improvement
When following the setup wizard and enabling a webmail client, the optional path can be set
image1

When the generated compose file is created, it would be useful to have the `WEBROOT_REDIRECT` set to the webmail client path that was configured.
image2

Additionally, it might be desired that the redirect goes to a different path, so maybe that could be exposed in the setup wizard?

Nebukadneza:
## What type of PR?
enhancement

## What does this PR do?
Add a configuration-field for WEBROOT_REDIRECT to the setup utility.

### Related issue(s)
closes #860

## Prerequistes
  - Minor enhancement to setup … i’d say not required?

### DIFF
--- a/setup/flavors/compose/mailu.env
+++ b/setup/flavors/compose/mailu.env
@@ -110,7 +110,13 @@ COMPRESSION_LEVEL={{ compression_level }}
 ###################################
 
 # Path to redirect / to
+{% if webroot_redirect == '' %}
+# You did not set any webroot redirect URL.
+# We default to /webmail to start error-free, regardless if you have webmail configured
 WEBROOT_REDIRECT=/webmail
+{% else %}
+WEBROOT_REDIRECT={{ webroot_redirect }}
+{% endif %}
 
 # Path to the admin interface if enabled
 WEB_ADMIN={{ admin_path }}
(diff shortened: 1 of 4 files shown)


muhlemmer:
I'm gonna have to think about this one a bit.. I still believe redirects are evil, as that the browser chaches a `302` status forever. (Can't be cleared by a simple `ctrl+F5`). A novice user that makes an error setting a redirect will get stuck with it and probably ends up opening an issue.

I'm not saying this PR is bad or anything, I just see a range of issues appearing since we made the redirect path settable... I'm just wondering, how many variations are there and what do the users really want? Where will the confusion end?

Before we made it settable, I was advocating of taking out the redirects all together, but then this setting came along and seemed a nice hybrid. It created only havoc ever since...

kaiyou:
Honestly I do not see the harm in making this configurable. Maybe we could simply provide examples in the field description so people know what they should set to redirect to the admin for instance.

kaiyou:
Given #1151 and if we agree on the principles, I vote we close this one. We need to configure this in the admin container, and have a static redirect in the nginx for simplicity sake.

PR #891, opened 2019-01-27
Update fetchmail to selfbuilt 7.0.0-alpha

### LINKED ISSUE #808: Fetchmail sslproto configuration (typo?)

Nebukadneza:
## What type of PR?
enhancement

## What does this PR do?
Fetchmail in alpine is ~5 years old — and doesn’t support current SSL/TLS
variants anymore. This especially leads to our own fetchmail not being able to
pull mail from mailu itself. Since no new fetchmail release is on the horizon,
let’s build the lastest distribution artifact — which strangely is not
6.4.0-snapshot, but 7.0.0-alpha — ourselves.

### Related issue(s)
closes #808 

## Prerequistes

### DIFF
--- a/services/fetchmail/Dockerfile
+++ b/services/fetchmail/Dockerfile
@@ -1,13 +1,30 @@
+# First stage: Build
+FROM alpine:3.8 as builder
+
+# build dependencies
+RUN apk add --no-cache curl tar xz autoconf git gettext build-base openssl openssl-dev
+
+RUN curl 'https://netcologne.dl.sourceforge.net/project/fetchmail/branch_6.3/fetchmail-7.0.0-alpha6.tar.xz' | tar xJ
+RUN cd fetchmail-7.0.0-alpha6 && \
+    ./configure --with-ssl --prefix /usr/local --disable-nls && \
+    make
+
 FROM alpine:3.8
+
+
 # python3 shared with most images
 RUN apk add --no-cache \
     python3 py3-pip bash \
   && pip3 install --upgrade pip
+
 # Image specific layers under this line
-RUN apk add --no-cache fetchmail ca-certificates \
+RUN apk add --no-cache ca-certificates openssl \
  && pip3 install requests
 
+COPY --from=builder /fetchmail-7.0.0-alpha6/fetchmail /usr/local/bin
 COPY fetchmail.py /fetchmail.py
+
+RUN adduser -D fetchmail
 USER fetchmail
 
 CMD ["/fetchmail.py"]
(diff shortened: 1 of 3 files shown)


muhlemmer:
bors try

muhlemmer:
I tried this and it seems to work okay :+1: 
I bombed the account to be fetched with 1000 mails, which runs a few minutes and fetchmail timeout set to 60 seconds. And no errors occurred.

Fetched account was on test.mailu.io and the fetchmail instance on my test server.

I do have to note that using the `STARTTLS` port (143) didn't work. Don't know if this a fetchmail thing or the way the python script is launching fetchmail. Or maybe we never intended `STARTTLS`?

I would not have anything against merging this. Fetchmail-6.3 is broken. This seems to work. The known bugs in https://gitlab.com/fetchmail/fetchmail/blob/master/NEWS don't seem to affect us. It's just that it's alpha so not very tested by a bigger audience.

I'm open for a second opinion from anybody here. @kaiyou, @Mailu/contributors?

Nebukadneza:
Again, sorry for insane delay. After looking at the fetchmail source again, i think you’re right — `wrapped` isn’t right. I guess either not setting an `sslmode` or setting it to `starttls=` either `may` or `must` would be right. I didn’t fully follow the codepath for `may` yet, and am unsure what happens on a ssl-only connection/port in that case. I’ll report back … hopefully a bit quicker.

Nebukadneza:
Hi,

after a longer-than-expected while of dogfooding on this, it seems stable enough to me. Sorry for the force-push, but to resolve the merge conflict in regard to towncrier, actually rebasing seemed best to me.

kaiyou:
@hoellen do you have time to review this?

services/fetchmail/Dockerfile
```diff
@@ -1,13 +1,30 @@
+# First stage: Build
+FROM alpine:3.8 as builder
+
+# build dependencies
+RUN apk add --no-cache curl tar xz autoconf git gettext build-base openssl openssl-dev
+
+RUN curl 'https://netcologne.dl.sourceforge.net/project/fetchmail/branch_6.3/fetchmail-7.0.0-alpha6.tar.xz' | tar xJ
```
hoellen:
```suggestion
RUN curl -L 'https://netcologne.dl.sourceforge.net/project/fetchmail/branch_6.3/fetchmail-7.0.0-alpha6.tar.xz' | tar xJ
```
Can only build it with the `-L` flag, because sourceforge returns a 302 (moved).

kaiyou:
@Nebukadneza, any feedback on the suggestion?

Nebukadneza:
Sorry, had lost track of it :( … makes sense, accepted it ^_^

PR #907, opened 2019-02-04
Don't generate the clamav configuration if ANTIVIRUS is none.

mvaled:
## What type of PR?

bug-fix

## What does this PR do?

Avoid rspamd to try to connect to clamav if you choose ANTIVIRUS="none".  Otherwise the rspamd will try to connect to none, timing out several times and leading to poor performance.

I have a server without CLAMAV; and I'm getting times up to 45s:

```
$ grep 'CLAMAV' ~/downloads/logs-from-antispam-in-mailu-security-5d75fb987-kv958.txt | grep -o 'time: [^m]*'
time: 45011.089
time: 45126.002
time: 45002.024
time: 45037.436
time: 45006.775
...
```

Mails for which clamav is not used range from a tens of miliseconds to a few hundred.

## Prerequistes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ v1.6.1 - unreleased
 - Enhancement: Create an Authentication Token with IPv6 address restriction ([#829](https://github.com/Mailu/Mailu/issues/829))
 - Bug: Fix creating new fetched accounts
 - Enhancement: Missing wildcard option in alias flask command ([#869](https://github.com/Mailu/Mailu/issues/869))
+- Bug: Fix poor performance if ANTIVIRUS is configured to none.
 
 v1.6.0 - 2019-01-18
 -------------------
--- a/services/rspamd/conf/antivirus.conf
+++ b/services/rspamd/conf/antivirus.conf
@@ -1,6 +1,8 @@
+{% if ANTIVIRUS == 'clamav' %}
 clamav {
   attachments_only = true;
   symbol = "CLAM_VIRUS";
   type = "clamav";
   servers = "antivirus:3310";
 }
+{% endif %}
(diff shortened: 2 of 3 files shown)


mvaled:
I fail to see why the tests are failing.  @Nebukadneza would look into this?

Nebukadneza:
Hi, sorry for the late reply. I hope to have some time by weekend, then i’ll gladly look. Sorry for delaying this :/

mvaled:
Also, we've noticed that the provided antivirus.conf would not work on Kubernetes.  The name "antivirus" would not get properly resolved.  Now, since both rspamd and clamav are deployed in the same pod, we configured to be "127.0.0.1:3310" so that it works.  But I think we must modify the template to take that into account.

Nebukadneza:
Hi,

please don’t ask me why this worked before. For some reason, the `filters`-test, which pipes the `eicar` testfile through mailu and checks that `antivirus` catches it … didn’t … have … antivirus … enabled?!

What I did was:
```
diff --git a/tests/compose/filters/mailu.env b/tests/compose/filters/mailu.env
index 25c7c13..e165fee 100644
--- a/tests/compose/filters/mailu.env
+++ b/tests/compose/filters/mailu.env
@@ -60,7 +60,7 @@ WEBMAIL=none
 WEBDAV=none
 
 # Antivirus solution (value: clamav, none)
-#ANTIVIRUS=clamav
+ANTIVIRUS=clamav
 
 #Antispam solution
 ANTISPAM=none
```

Instead of opening my own PR for this, would you add it to yours? Thanks~

mvaled:
Nebukadneza <notifications@github.com> writes:
> Instead of opening my own PR for this, would you add it to yours? Thanks~

Done.

mvaled:
:fireworks:

PR #917, opened 2019-02-06
Add kubernetes support to setup utility

daniel0611:
## What type of PR?

Feature

## What does this PR do?

This pr allows the setup utility to generate kubernetes manifests based on #576.
There are some adjustments to the manifests e.g.
- Upgrading all deployments from `extensions/v1beta1` to `apps/v1`.
- Update all nginx-ingress annotations to match the newer versions.
  This enables version 0.22.0 and up to be used.
- Adding 2 default cert-manager issuers. One for staging and one for production
- Updating the ConfigMap from old variables like `POD_ADDRESS_RANGE` to newer ones like `SUBNET`.
- Setting the storage-class on the PersistentVolumeClaims to standard to be more generic.

To allow dynamic generation of these manifests a new flavor was added to the setup utility. It covers almost as many settings as the compose flavor, but some features like databases other than sqlite are not implemented.

## Prerequisites


setup/static/render.js
```diff
@@ -2,17 +2,21 @@ $(document).ready(function() {
 	if ($("#webmail").val() == 'none') {
 		$("#webmail_path").hide();
 		$("#webmail_path").attr("value", "");
+        $("#webmail_host").parent().hide();
```
ionutfilip: For styling purposes i would suggest to keep the same indentation here. Beside that looks good

dani909: Sorry was using spaces instead of tabs. Fixed :smiley:

kaiyou:
Anyone around here can provide feedback on k8s?

muhlemmer:
CC: @hacor, @micw, @blinkeye; there are the names I could quickly scramble.

Friendly request to check / test this PR. Setup can easily be run local:
````bash
cd Mailu/setup
export DOCKER_ORG="mailutest"
export MAILU_VERSION="pr-917"
docker-compose up
````

Browse to: http://localhost:8001 and select Kubernetes. Choose your options and see if you get something you'd expect / valid.

micw:
I am actually running a mailu test setup on kubernetes and have a few issues which I think they should be fixed before we "officially" support kubernetes setups:

- there are currently some hardcoded service names that don't work in all setups #1014
- hostnames are resolved to ips on service startup which breaks service discovery when such an ip changes #1114
- url rewriting is not very portable and will work only on nginx ingress #1117 / #1021 / #1045
- the security of rspamd ingress works only on nginx, on other ingress it will be exposed unprotected #1121

The kubernets yaml files are a starting point but not very optimal. E.g. env vars are shared via config map. That means that services are not re-deployed or restarted on env changes. Certs that are shared between services should be defined as own resource and the secret should be used by the different ingresses.

Generally I think that such complex deployments whould be done using helm. In this case the setup tool for kubernetes would just create a values.yaml file to parameterize the helm chart. It's very uncommon to use a tool to create yamls for kubernetes deployments.

I'm working on a clean helm chart which I will release public when the issues above are closed.

Regards,
Michael.

hacor:
Hello folks,

sorry to dig in so late, I had to move my on-premise cluster to Digital Ocean and it seemed that all their IP addresses are on public spam lists, so running a mail server on their infrastructure was impossible. That's why I lost track of Mailu for some time.
From what I see that's happening, I think it should be interesting to discuss a more global path first before trying to implement new features (K8s based)

From my point of view:
- `HELM`: really interesting! I use it but I never dived into writing helm charts myself.
- `NGINX/TRAEFFIK` (and many others): This is not a discussion which concerns Mailu. That's why I made following separation in my kubernetes manifests:

- All HTTP traffic (ports 443 and 80) was done by 'standard' ingresses which should work for Traeffik or Nginx (when the UI can be moved to a different subdomain, rather than a subpath. Minor changes to mailu? Not sure). This choice should be up to the user.
- All other mail traffic (25, 993, etc etc) should be done by a separate NGINX controller which only listens to these ports, so it should never interfere with whatever ingress controller your kubernetes cluster is running. This part I would keep, because it's Mailu specific and it doesn't bother about the  chosen HTTP(S) ingress controller

So my question would be: is it possible to provide the admin container with an env variable `mail-admin.example.com` which routes directly to `mail.example.com/admin/ui`. If we can do this, it should work for every ingress controller out there.

All the best!

hacor

micw:
Hello @hacor,

thank you for the long reply. I absolutely agree, we should have an internal nginx for HTTP that handles url rewriting, rspamd authentication and all other stuff that is beyond default ingress stuff so that the primary ingress is free to choose. On other platforms, the main nginx does this (along with mail inbound) and I think we should enable this for on kubernetes too.

One exception is SSL termination. The internal Nginx on k8s should only serve on HTTP since SSL terminationis usually done by the main ingress.

When my other PRs are merged (almost done) and my mail system works, I can look into this and create a PR. This way we can solve #1117 / #1021 / #1045 and #1121 at once.

Best regards,
Michael.

micw:
I have created PR #1158 that should solve all the mentioned issues with ingress.

micw:
@hacor #1158 got finally merged. This opens a bunch of oportunities. I have pushed my helm chart at https://github.com/micw/charts/tree/mailu/incubator/mailu - give it a try if you like.
We can also focus on generating k8s yamls with the setup tool now if that makes sense. Would it be an option to run "helm template" from setup to generate the yaml? That way we'd maintain only one source.

micw:
@kaiyou I'd like to close this one in favor of the helm chart.

kaiyou:
Agreed

PR #931, opened 2019-02-13
Resolve webmail in admin

### LINKED ISSUE #716: Admin whitelist_webmail hard coded
Hi,
I'm trying to send transactional email with Mailu SMTP from a backend server and it takes more than 15s.
I tracked down one problem in core/admin/mailu/internal/__init__.py:
def whitelist_webmail() uses socket.gethostbyname("webmail")

In my docker configuration there is no "webmail" host so the function socket.gethostbyname return nothing after 5s which slows down a lot the request /internal/auth/email
When I set "webmail" to a fake ip on the admin server the /internal/auth/email returns immediately.

Maybe it would be better to define a list of hostnames in the configuration file instead of using a hard coded "webmail" value. What do you think?

Thanks Mailu for the great work!
JB

ionutfilip:
## What type of PR?
bug-fix

## What does this PR do?
Implement mailustart to resolve webmail in admin

### Related issue(s)
Fix #716 

## Prerequistes


core/admin/mailu/configuration.py
```diff
@@ -79,6 +80,8 @@ def resolve_host(self):
         self.config['HOST_POP3'] = resolve(self.config['HOST_POP3'])
         self.config['HOST_AUTHSMTP'] = resolve(self.config['HOST_AUTHSMTP'])
         self.config['HOST_SMTP'] = resolve(self.config['HOST_SMTP'])
+        if self.config["WEBMAIL"] != "none":
+            self.config["HOST_WEBMAIL"] = resolve(self.config['HOST_WEBMAIL'])
```
hoellen: Maybe it's a bit too strict: But above this lines we use single quotes. Sorry :see_no_evil:

muhlemmer:
bors +retry

muhlemmer:
bors retry

muhlemmer:
whatever man.
bors try

ionutfilip:
bors retry

Issue #941, opened 2019-02-18
postfix.cf is not read


flopes89:
I want to replace my current mail server (hosted on shared webhosting) with a mailu instance on my own server. So far everything looks dandy (awesome work!), however for the time being I want to copy all mails from "mailu-server" to my "old mail-server" for some time until I'm confident to switch.

For this, I figured I'd add an `always_bcc` address to postfix that points to an email on my old server so I can be sure to catch all emails if something goes wrong (along with an `MX` record with lower priority).

I added the `always_bcc` to `~/mailu/mounted/overrides/postfix.cf` and recreated the container from scratch (deleted image and ran `docker-compose up -d smtp`).

I can see that the file is mounted correctly:

```
$ docker-compose exec smtp cat /overrides/postfix.cf
always_bcc=mail@old-server.example.com
```

However, checking `postconf` reveils the configuration is not read:

```
$ docker-compose exec smtp postconf | grep bcc
always_bcc =
recipient_bcc_maps =
sender_bcc_maps =
```

What am I missing / doing wrong  here?

flopes89:
Just checked, executing

```
$ docker-compose exec smtp postconf -e 'always_bcc=mail@old-server.example.com'
```

in the running container works as expected:

```
$ docker-compose exec smtp postconf | grep bcc
always_bcc = mail@old-server.example.com
recipient_bcc_maps =
sender_bcc_maps =
```

I suspect it has something to do with the setup. However there are no errors in the startup logs. Not sure how to proceed here.

flopes89:
Further investigation revealed the error to be in `/etc/postfix/main.cf`. This file ends like so:

```
$ docker-compose exec smtp cat /etc/postfix/main.cf | tail -n 10
###############

smtpd_milters = inet:antispam:11332
milter_protocol = 6
milter_mail_macros = i {mail_addr} {client_addr} {client_name} {auth_authen}
milter_default_action = tempfail

###############
# Extra Settings
###############
```

Notice that there is **no final blank line**. However as I see that `Mailu/core/postfix/conf/main.cf` does in fact contain empty lines in the end.

I suspect that the jinja conversion somehow removes these empty lines, but I'm not sure yet why...

flopes89:
As expected, it's related to jinja, which strips newlines at the end of templates unless configured with 

> `keep_trailing_newline`
> Preserve the trailing newline when rendering templates. The default is False, which causes a single newline, if present, to be stripped from the end of the template.

[...]

PR #945, opened 2019-02-19
Add extra newline to main.cf

flopes89:
This should prevent jinja from stripping the newline, which causes overrides to be appended after the comment section

## What type of PR?

Bugfix

## What does this PR do?

Adds a new empty newline a the end of `conf/main.cf` so prevent jinja from stripping it, by which overrides done with `postconf -e` are not appended correctly.

### Related issue(s)

see #941

## Prerequistes

### DIFF
--- a/core/postfix/conf/main.cf
+++ b/core/postfix/conf/main.cf
@@ -112,3 +112,4 @@ milter_default_action = tempfail
 ###############
 # Extra Settings
 ###############
+


kaiyou:
This is a duplicate of #991. Sorry for the delays, only getting back at reviewing PRs.

kaiyou:
Huh, #991 is based on 1.6 and this one on Master, let's switch :smile:

PR #956, opened 2019-02-28
remove "ANTISPAM" from "mailu.env"

ahwebd:
"ANTISPAM" env variable is not used anymore, and should be removed from "mailu.env" to avoid confusion.
see commit: https://github.com/Mailu/Mailu/commit/1e3392e4175ff41d0dfc61c8fba40914e7d179ea


kaiyou:
PR #939 already provides this and actually removes all unused options

PR #967, opened 2019-03-11
Ability to use both webmails simultaneously (Rainloop and RoundCube)

mariooos:
## What type of PR?

Enhancement (maybe feature)

## Prerequistes
- [ ] Unless it's docs or a minor change: add [changelog]

### DIFF
--- a/core/admin/mailu/configuration.py
+++ b/core/admin/mailu/configuration.py
@@ -43,9 +43,10 @@ DEFAULT_CONFIG = {
     'DEFAULT_QUOTA': 1000000000,
     # Web settings
     'SITENAME': 'Mailu',
-    'WEBSITE': 'https://mailu.io',
+    'WEBSITE': None,
     'WEB_ADMIN': '/admin',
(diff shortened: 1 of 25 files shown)


micw:
I really like the idea. Unfortunately, the PR is quite outdated. Especially the resolution of HOST_* and *_ADDRESS has changed. @mariooos would you mind to rebase this? I also think that this PR should be aligned with #890 
Alternatively, "/" will be replaced by a static website, showing the options.

Nebukadneza:
This functionality is a enhancement for the user, although it adds additional complexity of which I don’t know if we can currently cope with it.

Sorry for being such a silent project and not appreciating your contribution. Would you still be willing to continue this enhancement?

Diman0:
We discussed this PR during the last Mailu meeting. This adds too much complexity to Mailu for only an edge case. Mailu is designed to only support one Webmail. We are not planning to change this.

you can simply run an extra non-Mailu webmail image and connect it via IMAP to Mailu, if you have the need for more different webmail clients.

PR #985, opened 2019-04-01
Permit raspberry pi (and other architectures) builds

abondis:
## What type of PR?

Enhancement

## What does this PR do?

Add an option to select base images and permit building for different CPU architectures.

### Related issue(s)
N/A

## Prerequistes

- [X] documentation updated accordingly

### DIFF
--- a/CHANGELOG.md.20190330
+++ b/CHANGELOG.md.20190330
@@ -1,1 +1,0 @@
-find . -iname 'dockerfile' -exec sed -i "s/FROM.*/ARG DISTRO=alpine:3.8\nFROM \$DISTRO/" '{}' \;
--- a/CHANGELOG.md.20190330
+++ b/CHANGELOG.md.20190330
@@ -1,0 +1,1 @@
+find . -iname 'dockerfile' -exec sed -i "s/FROM.*/ARG DISTRO=alpine:3.8\nFROM \$DISTRO/" '{}' \;
--- a/core/nginx/Dockerfile
+++ b/core/nginx/Dockerfile
@@ -1,4 +1,5 @@
-FROM alpine:3.10
+ARG DISTRO=alpine:3.10
+FROM $DISTRO
 # python3 shared with most images
 RUN apk add --no-cache \
     python3 py3-pip git bash \
--- a/docs/rpi_build.rst
+++ b/docs/rpi_build.rst
@@ -1,0 +1,19 @@
+.. _rpi_build:
+
+Building for a Raspberry Pi
+===========================
+
+The build does not need to be done on the Pi.
+
+To do so, go to ``tests/`` and call ``./build_arm.sh``, it will build all necessary images for arm.
+
+To push the locally built images to a remote server, run ``./deploy_to_pi.sh``. Docker 18.09+ is needed to use ``-H ssh://<user>@<host>``.
+
+Adjustments
+-----------
+
+``build_arm.sh`` uses some variables passed as ``build-arg`` to docker-compose:
+
+- ``DISTRO``: is the main distro used (ie: alpine:3.8)
+- ``PHP_DISTRO``: is used for the ``rainloop`` and ``roundcube`` images
+- ``EDGE_DISTRO``: is used for ``radicale`` as edge has dulwich and radicale as packages
--- a/docs/rpi_build.rst
+++ b/docs/rpi_build.rst
@@ -5,15 +5,21 @@ Building for a Raspberry Pi
 
 The build does not need to be done on the Pi.
 
-To do so, go to ``tests/`` and call ``./build_arm.sh``, it will build all necessary images for arm.
+To do so, go to ``tests/`` and call ``./build_arm.sh``, it will build all
+necessary images for arm.
 
-To push the locally built images to a remote server, run ``./deploy_to_pi.sh``. Docker 18.09+ is needed to use ``-H ssh://<user>@<host>``.
+To push the locally built images to a remote server, run ``./deploy_to_pi.sh``.
+Docker 18.09+ is needed to use ``-H ssh://<user>@<host>``.
 
 Adjustments
 -----------
 
 ``build_arm.sh`` uses some variables passed as ``build-arg`` to docker-compose:
 
-- ``DISTRO``: is the main distro used (ie: alpine:3.8)
-- ``PHP_DISTRO``: is used for the ``rainloop`` and ``roundcube`` images
-- ``EDGE_DISTRO``: is used for ``radicale`` as edge has dulwich and radicale as packages
+- ``ALPINE_VER``: version of ALPINE to use
+- ``DISTRO``: is the main distro used. Dockerfiles are set on Alpine 3.10, and
+  build script overrides for ``balenalib/rpi-alpine:3.10``
+- ``QEMU``: Used by webmails dockerfiles. It will add ``qemu-arm-static`` only
+  if ``QEMU`` is set to ``arm``
+- ``ARCH``: Architecture to use for ``admin``, and ``webmails`` as their images
+  are available for those architectures.
--- a/optional/radicale/Dockerfile
+++ b/optional/radicale/Dockerfile
@@ -1,4 +1,5 @@
-FROM alpine:3.10
+ARG DISTRO=alpine:3.10
+FROM $DISTRO
 
 RUN echo "@testing http://nl.alpinelinux.org/alpine/edge/testing" >> /etc/apk/repositories \
  && apk add --no-cache radicale@testing curl bash
--- a/setup/Dockerfile
+++ b/setup/Dockerfile
@@ -1,12 +1,12 @@
-ARG DISTRO=alpine:3.8
+ARG DISTRO=alpine:3.10
 FROM $DISTRO
 
 RUN mkdir -p /app
 WORKDIR /app
 
 COPY requirements.txt requirements.txt
-RUN apk add --no-cache curl python3 py3-pip
-RUN pip3 install -r requirements.txt
+RUN apk add --no-cache curl python3 py3-pip \
+    && pip3 install -r requirements.txt
 
 COPY server.py ./server.py
 COPY main.py ./main.py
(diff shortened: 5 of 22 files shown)


abondis:
it's strange, bors says 'build succeeded' but it seems like it doesn't report it's status ...

kaiyou:
It does if you click the link in the comment. Bors merely creates branches and merges them, Travis is the one to run the checks.

kaiyou:
This is pretty cool, I have a few comments and questions at this point.

1. Really sorry about it, but we are migrating to alpine 3.10 and if merged first it might have some (minor) impact on your PR, starting with not using edge anymore.

2. Why are you copying qemu in the php images? I am probably missing something there.

3. You are changing quite a lot of things in the admin Dockerfile, some of them which do not seem to be related to arm. Could you detail why?

abondis:
Hi, thank you for the review, here are the answers : 

1. Changing to 3.10 should be ok, I'll take a look when I have time

2. the qemu file is needed when you build for arm from a non arm plateform. Most of my changes are just permitting a variable to override the image to use so the build script can use an arm image. The one I used in the build script already has the qemu except for the php image and I could not find any other way to do that.

3. In admin I think I only changed the image name and added the variable. Maybe you meant the nginx one ? I don't remember exactly, I recall having issues but can't recall why.
Another thing I don't know, is how to make the CI actually run the arm build as for now it still only calls the normal x86 build ..

muhlemmer:
> In admin I think I only changed the image name and added the variable. Maybe you meant the nginx one ? I don't remember exactly, I recall having issues but can't recall why.
Another thing I don't know, is how to make the CI actually run the arm build as for now it still only calls the normal x86 build ..

I don't believe travis-CI has anything else then X86_64. They've closed an issue without fix regarding this travis-ci/travis-ci#3376

core/nginx/Dockerfile
```diff
@@ -1,13 +1,15 @@
...
+    python3 py3-pip git bash
+# \ && pip3 install --upgrade pip
 # Shared layer between rspamd, postfix, dovecot, unbound and nginx
 RUN pip3 install git+https://github.com/usrpro/MailuStart.git#egg=mailustart
 # Image specific layers under this line
-RUN apk add --no-cache certbot nginx nginx-mod-mail openssl curl \
- && pip3 install idna requests watchdog
+RUN apk add --no-cache certbot nginx nginx-mod-mail openssl curl
+RUN apk del --no-cache py3-openssl
+RUN pip3 install idna requests watchdog
```
muhlemmer: Most changes in this file have nothing to do ARM support. Please revert non-relevant changes. Also, using multiple RUN` statements will create new layers and additional image size. Where possible `RUN` operations need to remain grouped using `&& \` syntax.

docs/rpi_build.rst
```diff
@@ -0,0 +1,19 @@
...
+To push the locally built images to a remote server, run ``./deploy_to_pi.sh``. Docker 18.09+ is needed to use ``-H ssh://<user>@<host>``.
+
+Adjustments
+-----------
+
+``build_arm.sh`` uses some variables passed as ``build-arg`` to docker-compose:
+
+- ``DISTRO``: is the main distro used (ie: alpine:3.8)
+- ``PHP_DISTRO``: is used for the ``rainloop`` and ``roundcube`` images
+- ``EDGE_DISTRO``: is used for ``radicale`` as edge has dulwich and radicale as packages
```
muhlemmer: Edge is no longer used:
https://github.com/Mailu/Mailu/blob/13d6cf2fdbcf9b4e3bd31d5f8db467ee340cbc74/optional/radicale/Dockerfile#L1

optional/radicale/Dockerfile
```diff
@@ -1,7 +1,9 @@
-FROM alpine:edge
+ARG EDGE_DISTRO=alpine:edge
+FROM $EDGE_DISTRO
```
muhlemmer: Edge is no longer used:
https://github.com/Mailu/Mailu/blob/13d6cf2fdbcf9b4e3bd31d5f8db467ee340cbc74/optional/radicale/Dockerfile#L1

optional/radicale/Dockerfile
```diff
@@ -1,7 +1,9 @@
...
 
 RUN echo "@testing http://nl.alpinelinux.org/alpine/edge/testing" >> /etc/apk/repositories \
- && apk add --no-cache radicale@testing py-dulwich@testing curl bash
+  && apk add --no-cache py3-dulwich@testing curl bash
+RUN pip3 install radicale
```
muhlemmer: Why drop official distro package for pip one? This change should not be related to ARM. If it is an improvement by itself, please open a specific pull request.

setup/Dockerfile
```diff
@@ -1,11 +1,12 @@
...
+FROM $DISTRO
 
 RUN mkdir -p /app
 WORKDIR /app
 
 COPY requirements.txt requirements.txt
-RUN apk add --no-cache curl \
- && pip install -r requirements.txt
+RUN apk add --no-cache curl python3 py3-pip
+RUN pip3 install -r requirements.txt
```
muhlemmer: Keep these actions in a single `RUN` statement, in order to avoid additional layers.

tests/build_arm.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash -x
+
+ALPINE_VER="3.8"
+DISTRO="balenalib/rpi-alpine:$ALPINE_VER"
+# Used for Radicale
+EDGE_DISTRO="balenalib/rpi-alpine:edge"
+# Used for webmails
```
muhlemmer: Edge is no longer used:
https://github.com/Mailu/Mailu/blob/13d6cf2fdbcf9b4e3bd31d5f8db467ee340cbc74/optional/radicale/Dockerfile#L1

tests/build_arm.sh
```diff
@@ -0,0 +1,16 @@
...
+PHP="arm32v7/php:7.2-apache"
+
+# use qemu-*-static from docker container
+docker run --rm --privileged multiarch/qemu-user-static:register
```
muhlemmer: As I understand correctly from your comment, the `qemu-user-static` binary is already present in the OS images you are pulling? Why does this need to be run?

webmails/rainloop/Dockerfile
```diff
@@ -1,4 +1,10 @@
-FROM php:7.2-apache
+ARG PHP_DISTRO=php:7.2-apache
+
+FROM $PHP_DISTRO
+# FIXME: hardcoded since : --from won't work with variables
+# and with multistage build --from does not seem to copy the binary
+COPY --from=balenalib/rpi-alpine:3.8 /usr/bin/qemu-arm-static /usr/bin/qemu-arm-static
```
muhlemmer: I'm aware of the limitations you expressed:
>the qemu file is needed when you build for arm from a non arm plateform. Most of my changes are just permitting a variable to override the image to use so the build script can use an arm image. The one I used in the build script already has the qemu except for the php image and I could not find any other way to do that.

However, this will affect non-arm build as well and therefore cannot be merged as-is. You'll have to find another solution. Perhaps a custom php-arm image containing the binary?

webmails/roundcube/Dockerfile
```diff
@@ -1,4 +1,10 @@
-FROM php:7.2-apache
+ARG PHP_DISTRO=php:7.2-apache
+
+FROM $PHP_DISTRO
+# FIXME: hardcoded since : --from won't work with variables
+# and with multistage build --from does not seem to copy the binary
+COPY --from=balenalib/rpi-alpine:3.8 /usr/bin/qemu-arm-static /usr/bin/qemu-arm-static
```
muhlemmer: Same issue as my rainloop comment

abondis:
@muhlemmer actually the problem is not TravisCI: the build for Raspberry Pi can be done from an x86_64 machine, that is why there is some qemu related files. The problem is just that I don't know what to configure where to make it run the command :)

webmails/rainloop/Dockerfile
abondis: @muhlemmer is it an issue to have the binay in non-arm machines as it won't interact with the machine when it's not running arm binaries ? It's not that big if we consider the size of the apache-php stack, and I don't think having an extra binary in that kind of setup changes much.

abondis: @muhlemmer what if I add it for the build and remove it at the end, would that be ok ?

abondis:
Ah yes I'm going to take a look at this this week. Thanks.

fergalmoran:
Does... does... does this mean what I think it means?

abondis:
I don't know, what do you think it means ? It build the docker images to
run mailu on a raspberry pi ( and probably other ARM boards).
I still have to make the build pass.

fergalmoran:
@abondis Sorry - I've been following this PR with interest for a while...
I was just getting excited that the build succeeded....

kaiyou:
Oh dear. It passed. :clap:

CHANGELOG.md.20190330
```diff
@@ -0,0 +1 @@
+find . -iname 'dockerfile' -exec sed -i "s/FROM.*/ARG DISTRO=alpine:3.8\nFROM \$DISTRO/" '{}' \;
```
muhlemmer: Sorry, I might have missed this one before, but I believe this file does not belong here?

muhlemmer: Functional testing went fine :+1:  So ready to approve if this little piggie is clarified.

abondis: Yep, my bad, I should not have commited this. Thanks !

setup/Dockerfile
```diff
@@ -1,11 +1,12 @@
...
```
kaiyou: Small notice, this should not be split in two layera.

abondis:
Good catch, I thought I didn't modify it, but I guess it's some extra from
the rebase ... I'll try to take fix it soon. Thanks.

PR #991, opened 2019-04-09
Added extra empty line at the end of file

zxpower:
## What type of PR?

Bug-fix

## What does this PR do?

In case if you have postfix.cf in overrides folder then without extra empty line first value from postfix.cf gets added in the comment line.


kaiyou:
This looks okay, could you please rebase it against master, and we'll backport it to 1.6?

kaiyou:
In fact this is a duplicate of #945 which is based on master, so closing here.

PR #1006, opened 2019-05-06
Refactor admin using webpack

kaiyou:
## What type of PR?

Refactoring

## What does this PR do?

This PR brings some refactoring to the admin container :
- remove the dependency to mailustart and replace it with socrate
- remove static assets from the repo and use Webpack for building them


kaiyou:
I guess I need to update the changelog as well.

PR #1028, opened 2019-05-21
Fixed Nginx configuration documentation regarding reverse proxy

oli-ver:
The current version of the example configuration seems not to work. In my opinion the letsencrypt certificates have to be configured and http has to be forwarded as well to allow obtaining letsencrypt certificates.

This configuration works for Nginx as reverse proxy before Mailu (I am serving mailu on a subdomain mail.myhost.com).

Did I not understand something, or is it possible the documentation was missing this?

## What type of PR?

documentation

## What does this PR do?

Fixes the documentation to allow serving Mailu behind a reverse proxy.

### Related issue(s)
No issue created, just tried to share the experience I had when installing mailu.

## Prerequistes

### DIFF
--- a/docs/reverse.rst
+++ b/docs/reverse.rst
@@ -33,7 +33,7 @@ The simplest and safest option is to modify the port forwards for Mailu Web fron
     volumes:
       - "$ROOT/certs:/certs"
 
-Then on your own frontend, point to these local ports. In practice, you only need to point to the HTTPS port (as the HTTP port simply redirects there). Here is an example Nginx configuration:
+Then on your own frontend, point to these local ports. You need to point to the HTTP and HTTPS port to ensure that letsencrypt is able to obtain certificates from the url http://your-domain/.well-known. All other requests will be automatically forwarded to https. Here is an example Nginx configuration (assuming you mapped your certs volume to /srv/docker/mailu/certs/):
 
 .. code-block:: nginx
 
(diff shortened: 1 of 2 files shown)


kaiyou:
The configuration was indeed not intended for a letsencrypt setup. It looks okay, except the TLS configuration is different from the internal Mailu settings. For consistency, I would use the same.

Nebukadneza:
For the convenience, to be found here:
https://github.com/Mailu/Mailu/blob/8ec31df1396297e2d7588fc3c6d49d22dc1d7d03/core/nginx/conf/tls.conf

muhlemmer:
The source repository went away and the PR has been open for long without reply. I'm voting to close,

@oli-ver, do you still intend to implement any changes to can merge this?

oli-ver:
@muhlemmer No you can close it. I just wanted to make you aware that setting up the server according to the docs did only work after some changes. Even now not everything works well, the letsencrypt certificates seem not to be updated automatically. Did I perhaps miss something in the docs especially explaining how to setup letsencrypt in mailu?

oli-ver:
Should I create a separate issue for the configuration issue? Mailu shows the following error message:

```
front_1      | Failed authorization procedure. webmail.myhost.com (http-01): urn:acme:error:connection :: The server could not connect to the client to verify the domain :: Fetching https://127.0.0.1/.well-known/acme-challenge/<the-challenge>: Invalid host in redirect target "127.0.0.1". Only domain names are supported, not IP addresses
```

I don't know where this comes from as in the .env file host names and urls seem to be set correctly.

kaiyou:
That last bit is worrying, please create an issue indeed.

delijati:
`  proxy_set_header Host $http_host;` is missing in this PR also

PR #1030, opened 2019-05-23
Update user_external example for nextcloud

kesselb:
## What type of PR?

documentation

## What does this PR do?

Update the user_external example for Nextcloud due an upstream change. PHP will remove the imap extension. Newer user_external releases requires a different configuration for imap.

### Related issue(s)
- https://github.com/nextcloud/user_external/issues/52

## Prerequistes

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -265,7 +265,7 @@ I want to integrate Nextcloud 15 (and newer) with Mailu
   
 
 If a domain name (e.g. example.com) is specified, then this makes sure that only users from this domain will be allowed to login.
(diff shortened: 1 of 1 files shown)


kesselb:
Please review @violoncelloCH :ping_pong:

violoncelloCH:
so nextcloud/user_external#65 is merged and released in v0.6.3

kesselb:
Are these required? The default should be fine for most user?

violoncelloCH:
no as they are at the end and as they have a default value they are not required... I just thought you might want to add them to the documentation here as well, because they might be useful...

kesselb:
Have a look @violoncelloCH @Nebukadneza :construction_worker_man:

violoncelloCH:
@kesselb Looks good! (only Nextcloud writes itself with a lowercase "c", doesn't it?)

Issue #641, opened 2018-10-09
mailu_front_1: Error starting userland proxy: listen tcp [::]:995: bind: cannot assign requested address


Jan-Bart:
Hi,

I'm getting this error when listening on both, ipv6 and ipv4 addresses:
```
mailu_front_1: Error starting userland proxy: listen tcp [::]:995: bind: cannot assign requested address
```

I know per #321 that you can fix this by removing either BIND_ADDRESS6 or BIND_ADDRESS4.
And indeed, that fixes the issues but is rather a workaround than a fix.

What's the proper way to deal with this issue?
I'm running docker-compose version 1.17 and do not use 0.0.0.0 as ipv4 address ( #537 )
Using `notls` or `letsencrypt` doesn't make any difference ( #378 )

Related issues but with workarounds:
#464
#422

kaiyou:
The proper way is to set an actual IPv6 address instead of ``::``. Mailu is meant to be listening on a specific IP address, which is the address of your MX record.

Jan-Bart:
But you can't use both IPv4 and IPv6?

I'm using an actual IP address for both IPv4 and IPv6.

Jan-Bart:
In that case: I think I found a bug :)

kaiyou:
What is weird is that docker tries to listen to ``::`` which should not happen if you properly set ``BIND_ADDRESS6``. Could you share at least this part of your ``.env``? (slightly change your IP if you do not wish to share it)

Jan-Bart:
(IP is slightly changed)

```
# Address where listening ports should bind
BIND_ADDRESS4=00.00.137.241
BIND_ADDRESS6=3b00:4f9:c010:1f0::


# Main mail domain
```

muhlemmer:
I can reproduce this when using the <link> scoped ip6 address. 

@Jan-Bart : Run the following, replacing `eth0` to the interface you want to use:
```
[tim@mars ~]$ ifconfig eth0
eth0: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
        inet 5.189.138.127  netmask 255.255.255.0  broadcast 5.189.138.255
        inet6 2a02:c207:2019:6443::1  prefixlen 64  scopeid 0x0<global>
        inet6 fe80::250:56ff:fe3c:b223  prefixlen 64  scopeid 0x20<link>
        ether 00:50:56:3c:b2:23  txqueuelen 1000  (Ethernet)
[...]

PR #1033, opened 2019-05-26
use .env file to provide external IP address

belegnar:
## What type of PR?

enhancement

## What does this PR do?

prettyfies docker-compose.yml been generated

## Prerequistes


Nebukadneza:
I know that `.env` is somewhat the standard, and evaluated by default, but since a while the the `docker-compose.yml` offered by the setup utility sets the env-file to `mailu.env`. I agree that pulling out the IP to a single place is quite nice, but why should we add a third file, `.env` instead of using the configuration-bearing `mailu.env` which already exists? Would you also be happy with putting it in `mailu.env`?

Thanks for your addition and cleanups though ^_^.

belegnar:
Because `mailu.env` is not available _before_ build, so you can't reference vars from it in `docker-compose.yaml`
https://vsupalov.com/docker-arg-env-variable-guide/

muhlemmer:
> Because `mailu.env` is not available _before_ build, so you can't reference vars from it in `docker-compose.yaml`
> https://vsupalov.com/docker-arg-env-variable-guide/

Variables in setup are appointed in the python app. Why the additional `.env`? How is the additional env parsed from inside the app?

core/postfix/start.py
```diff
@@ -36,7 +36,9 @@ def start_podop():
 
 if os.path.exists("/overrides/postfix.cf"):
     for line in open("/overrides/postfix.cf").read().strip().split("\n"):
-        os.system('postconf -e "{}"'.format(line))
+        line = line.strip()
+        if line and line[0] != "#":
+            os.system('postconf -e "{}"'.format(line))
```
muhlemmer: What is the puprose of this change? it is not described in the PR title or body? This should go in a separate PR.

docs/faq.rst
```diff
@@ -242,6 +242,9 @@ correct syntax. The following file names will be taken as override configuration
 
 *Issue reference:* `206`_.
 
+ keep in mind, that commands from ``postfix.cf`` are injected via system call to ``postconf -e``
+ so symbols been used should be escaped
+
```
muhlemmer: I'm very OK with including this into the FAQ, but has nothing to do with the topic of this PR? Perhaps belongs in a PR with the changes in `core/postfix/start.py`?

setup/flavors/compose/.env
```diff
@@ -0,0 +1,5 @@
+{% if bind4 %}
+EXTERNAL_IP={{ bind4 }}
+{% else %}
+EXTERNAL_IP=0.0.0.0
+{% endif %}
```
muhlemmer: We don't [officially support](https://mailu.io/1.7/compose/setup.html#bind-address) running on `0.0.0.0`. As such, I don't think we should provide this as a default:

> If the address is not configured directly (NAT) on any of the network interfaces or if you would simply like the server to listen on all interfaces, use 0.0.0.0 and ::. Note that running is this mode is not supported and can lead to [issues](https://github.com/Mailu/Mailu/issues/641).

kaiyou:
Sorry about my current opinion, but really, if the only gain is to be able to parametrize the IP address, it is probably not worth it multiplying the number of config files the admin has to manage.

We could discuss going back to `.env` in a separate issue/discussion.

micw:
I reviewed it and came to the same conclusion: We should close this in favor of separate PRs with a clean scope.

muhlemmer:
Closing as discussed [here](https://matrix.to/#/!RJFCFtixHgPhzacdhW:tedomum.net/$1570831252137459GIvHn:matrix.org?via=tedomum.net&via=matrix.org&via=ghostdub.de)

PR #1039, opened 2019-06-02
upgrade alpine base-image

HorayNarea:
Upgrades the alpine image from 3.8 to 3.9 on all containers __except__ postgres

Postgres container upgrade would pull in a new major version of postgres which requires some (manual) work the make to old data readable by the new version

Also I changed `libressl` to `openssl` because alpine 3.9 has switched back to the OG openssl :D

### DIFF
--- a/core/admin/Dockerfile
+++ b/core/admin/Dockerfile
@@ -1,17 +1,18 @@
-FROM alpine:3.8
+FROM alpine:3.9
 # python3 shared with most images
 RUN apk add --no-cache \
     python3 py3-pip git bash \
   && pip3 install --upgrade pip
+# Shared layer between admin, rspamd, postfix, dovecot, unbound and nginx
 RUN pip3 install git+https://github.com/usrpro/MailuStart.git#egg=mailustart
 # Image specific layers under this line
 RUN mkdir -p /app
 WORKDIR /app
 
 COPY requirements-prod.txt requirements.txt
-RUN apk add --no-cache libressl curl postgresql-libs mariadb-connector-c \
+RUN apk add --no-cache openssl curl postgresql-libs mariadb-connector-c \
  && apk add --no-cache --virtual build-dep \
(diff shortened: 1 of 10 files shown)


blinkeye:
FYI: Alpine 3.10 was released a few days ago: https://alpinelinux.org/posts/Alpine-3.10.0-released.html

blinkeye:
@HorayNarea : do you intentionally reference a forked MailuStart repo (https://github.com/usrpro/MailuStart.git#egg=mailustart) in the Dockerfiles?

blinkeye:
@HorayNarea : the update to Alpine:3.9 breaks postfix. 
```
sh: /usr/lib/postfix/post-install: not found
sh: /usr/lib/postfix/master: not found
rsyslogd 8.40.0: running as pid 1, enabling container-specific defaults, press ctl-c to terminate rsyslog
: environment variable TZ is not set, auto correcting this to TZ=UTC [v8.40.0 try https://www.rsyslog.com/e/2442 ]
:  [origin software="rsyslogd" swVersion="8.40.0" x-pid="1" x-info="https://www.rsyslog.com"] start
```
We also need to change 2 path of files run by the `start.py`.

I've created a new PR https://github.com/Mailu/Mailu/pull/1050 based on your changes here and rebased against the current master (and solved a conflict).

HorayNarea:
> @HorayNarea : do you intentionally reference a forked MailuStart repo (https://github.com/usrpro/MailuStart.git#egg=mailustart) in the Dockerfiles?

Well yes but actually no. :D
I didn't change which MailuStart is used…

> @HorayNarea : the update to Alpine:3.9 breaks postfix.

I… uhh… well… I thought I tested it, but obviously I did it wrong. Maybe shouldn't do important things when I haven't slept enough. :upside_down_face:


Since your PR fixes stuff I will close mine. Thanks for cleaning up after me :kissing_heart: :dagger:

PR #1042, opened 2019-06-07
Postsrsd to avoid SPF rejects when Mailu used as MTA

### LINKED ISSUE #328: Support for SRS
See https://en.wikipedia.org/wiki/Sender_Rewriting_Scheme and https://github.com/roehling/postsrsd

jbmolle:
This PR should resolve #328
It implements Postsrsd with Mailu Postfix so that emails from alias are not rejected under SPF protocol.

### DIFF
--- a/core/postfix/conf/main.cf
+++ b/core/postfix/conf/main.cf
@@ -112,4 +112,7 @@ milter_default_action = tempfail
 ###############
 # Extra Settings
 ###############
-
+sender_canonical_maps = tcp:localhost:10001
+sender_canonical_classes = envelope_sender
+recipient_canonical_maps = tcp:localhost:10002
+recipient_canonical_classes = envelope_recipient, header_recipient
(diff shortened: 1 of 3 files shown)


core/postfix/conf/main.cf
```diff
@@ -112,4 +112,7 @@ milter_default_action = tempfail
 ###############
 # Extra Settings
 ###############
-
+sender_canonical_maps = tcp:localhost:10001
```
kaiyou: I would put these under a previous section specific to SRS instead of just extra settings.

core/postfix/start.py
```diff
@@ -50,7 +51,9 @@ def start_podop():
 
 convert("/conf/rsyslog.conf", "/etc/rsyslog.conf")
 
-# Run Podop and Postfix
+# Generate SRS Secret file, start SRS, run Podop and Postfix
```
kaiyou: I feel bad about this. Maybe we could use native python (os.urandom and files)?

Nebukadneza:
The more current #1349 looks less invasive, and generally like a leaner solution to me. If we go ahead merging #1349 , would it be okay if we closed this one after that? Sorry for sounding so harsh, and nevertheless thanks a lot for the work put into this @jbmolle !

muhlemmer:
@jbmolle thank you for your submission. We have implemented SRS by a different method in #1349. This PR will now be closed, hopefully without hard feelings. We would like to welcome you for contributions in the future!

Issue #1021, opened 2019-05-15
Kuebernetes Admin UI URL in redirect loop


trilitheus:
I'm attempting to get this set up on Kubernetes to evaluate.

Unfortunately, when trying to hit the admin, I get in a redirect loop to /admin/ui/

```
$ curl -I -L https://mail.example.com/admin/ui/                                                            

HTTP/1.1 301 MOVED PERMANENTLY
Date: Wed, 15 May 2019 19:15:56 GMT 
Content-Type: text/html; charset=utf-8
Content-Length: 291 
Connection: keep-alive
Location: https://mail.example.com/admin/ui/
Strict-Transport-Security: max-age=15724800; includeSubDomains
...
...
...
HTTP/1.1 301 MOVED PERMANENTLY
Date: Wed, 15 May 2019 19:15:58 GMT
Content-Type: text/html; charset=utf-8
Content-Length: 291
Connection: keep-alive
Location: https://mail.example.com/admin/ui/
Strict-Transport-Security: max-age=15724800; includeSubDomains

curl: (47) Maximum (50) redirects followed
```

I think this is a combination of the admin flask app and the nginx-ingress redirects (although that may be wrong as I've no experience with flask)

I'm using a previously installed nginx-ingress (from https://github.com/kubernetes/kops/blob/master/addons/ingress-nginx/v1.6.0.yaml) and am running kubernetes as this is Kubernetes installed via KOPS on AWS.

The ingresses seem to be configured correctly, SSL certs in palce etc and annotiation prefix made sure to match.

This does not appear to be a http -> https loop, all redirects are to and from https

Any assistance would be greatly appreciated.

hacor:
Late reply, but anyway.
I figured out the error you have and a PR is created. 

You can see it's only a minor change in the `mailu-admin-ui-ingress` and the `mailu-admin-static-ingress`

Hopefully this solves your problem!
All the best,

Hacor

trilitheus:
Hi I can confirm that ths changes have sorted this particular issue - may thanks for the update

micw:
@trilitheus Could you check if https://github.com/Mailu/Mailu/pull/1117 solves this issue too on nginx? I have changed the admin ui to listen on /admin/ui so that no path rewriting is required. This makes it more portable between different ingress controllers.

PR #1045, opened 2019-06-18
Kubernetes manifest update

hacor:
## What type of PR?

Enhancement

## What does this PR do?
It fixes the infinite redirect loops in Kubernetes Admin interface. 
The change is to add the complete uri string to the rewrite rule.
Updated antispam URL

### Related issue(s)
#1021

### DIFF
--- a/docs/compose/setup.rst
+++ b/docs/compose/setup.rst
@@ -31,7 +31,7 @@ Sets the ``TLS_FLAVOR`` to one of the following
 values:
 
 - ``cert`` is the default and requires certificates to be setup manually;
-- ``letsencrypt`` will use the *Letsencrypt!* CA to generate automatic ceriticates;
+- ``letsencrypt`` will use the *Letsencrypt!* CA to generate automatic certificates;
 - ``mail`` is similar to ``cert`` except that TLS will only be served for
   emails (IMAP and SMTP), not HTTP (use it behind reverse proxies);
 - ``mail-letsencrypt`` is similar to ``letsencrypt`` except that TLS will only be served for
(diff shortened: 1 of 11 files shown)


blinkeye:
May I ask what you plan to do with this specific PR? You updated the K8s manifests to reference the 1.6 images, so, in the current state this would be a bugfix/hotfix of the 1.6 release branch. But then you starting point is wrong (because on the 1.6 release the manifests already reference the 1.6 images).

If, on the other hand (and what I assume) you intend to merge this PR to the master I would assume that you should not reference the 1.6 images  but the current development master images?

kaiyou:
@micw, could you review this work and either provide some feedback or make sure that your helm chart builds on it.

I am not sure if we should keep both a chart and set of k8s configs, any advice?

micw:
The changes here are very specific for nginx ingress controller and would eventually break others like traefik. Unfortunately I don't have nginx ingress (only traefik), so I cannot verify this.
What I can do is to test if this would still work on traefik (which btw. does not have the issue described here).

kaiyou:
Any hint on the matter would be great. I'll try and test this tomorrow but I cannot promise anything.

micw:
I tried to get this work on traefik too but I could not find a portable way. So it will either work with nginx ingress or with traefik ingress, so the solution is far from being optimal.

The best solution would be to have the admin service listening on /admin (and not on /). In this case, we would not require path rewriting and we can use standard ingress definitions.

micw:
@hacor I have created PR https://github.com/Mailu/Mailu/pull/1117 which addresses this issue in a different way.  I have tested it using traefik ingress. Can you test it with nginx? A ready-to-use docker image is built at 

micwy/mailu-test-builds:admin--admin-ui-path

PR #1048, opened 2019-06-23
Refactor admin using webpack

kaiyou:
## What type of PR?

Refactoring

## What does this PR do?

This PR brings some refactoring to the admin container :
- remove the dependency to mailustart and replace it with socrate
- remove static assets from the repo and use Webpack for building them


kaiyou:
Still an issue when checking the webmail host from the admin container. My bad for accidentally removing a condition.

hoellen:
bors try

kaiyou:
Oh, thank you so much. And I approve your changes. Now missing a review :)

Nebukadneza:
Thank alot, this makes the whole admin UI so much cleaner!

One thing I noticed while trying out is that I cannot enable automatic reply anymore. When activating the checkbox, I get a `ReferenceError: $ is not defined at reply:1:18`, and the disabled text/input-fields just stay disabled. On 1.6, they will turn active.
Another small nitpick is that here, the flip-sidebar-away-button and sidebar caption are kinda misaligned and badly readable. But that’s a really minor nitpick.

Thanks a bunch!
-Dario

hoellen:
I added the missing `app.css` and fixed the `ReferenceError: $ is not defined at reply` error with help from [StackOverflow](https://stackoverflow.com/a/46937798). It has something to do that jQuery is not exposed properly.

Maybe someone with more webpack knowledge can verify it a little bit more. But it is working for me now.

kaiyou:
I would prefer not to have to expose a global `$`, but I guess it is the only simple way without doing more refactoring for now. I hope this can help merge the PR and move forward, thank you so much :)

hoellen:
bors try

PR #1050, opened 2019-06-23
Upgrade alpine to 3.9

blinkeye:
@HorayNarea wrote:

> Upgrades the alpine image from 3.8 to 3.9 on all containers except postgres
>
> Postgres container upgrade would pull in a new major version of postgres which requires some (manual) work the make to old data readable by the new version
> 
> Also I changed libressl to openssl because alpine 3.9 has switched back to the OG openssl :D


HorayNarea:
Closing in favour of #1052 but thanks anyway! :)

PR #1052, opened 2019-06-23
Upgrade alpine to 3.10

### LINKED ISSUE #1049: Feature Request: drop rsyslogd dependency for postfix
Postfix v3.4 has a new feature to directly log to stdout without the need of a rsyslogd dependency.

See http://www.postfix.org/MAILLOG_README.html#log-to-stdout)  for reference.

This feature was introduced with 3.4 and would be available in the Alpine v3.10 version (they have Postfix v3.4.5)

### LINKED ISSUE #1051: Regular alpine base-image update
To receive current updates for upstream projects, security fixes and new features such as mentioned in #1049 , let’s upgrade the alpine baseimage to latest 3.10

Nebukadneza:
## What type of PR?
enhancement / bug-fix

## What does this PR do?
Upgrade the alpine base image to 3.10 and clean up ensuing problems. Also directly uses postfix foreground-running with stdout logging.

### Related issue(s)
closes #1049
closes #1051

Note: This is a duplicate effort of #1050 #1039 …

## Prerequistes

With these images, I have tested manually:
- Email receive to user
  - on main domain
  - on additional domain
  - via an alias on main-domain
  - via an alias on additional domain
  - via catchall
- Email reject
  - of eicar testfiles
- Email sending
- Fetchmail from legacy POP
- Front LE certificates
- Simple overriding for postfix (only postfix.cf), nginx, dovecot, rspamd
- Creating a CalDAV calendar and CardDAV entry using davx5

I have not (yet) tested:
- certdumper
- databases other than sqlite

Todo:
- [x] Get rid of podop at `git+https://github.com/Nebukadneza/Podop.git@fix_py37` once https://github.com/Mailu/Podop/pull/6 is merged
- [x] Bend Mailustart back to original repository
- [x] Test more (? also help wanted ?)


blinkeye:
I've done a local test with this feature branch and tested the setup on a private K8s cluster.

I have a small addition (namely to upgrade to PHP 7.3 for the webmail frontends), see [blinkeye/HorayNarea-feat-upgrade-alpine](/blinkeye/Mailu/tree/HorayNarea-feat-upgrade-alpine).

Roundcube supports PHP 7.3 since v1.3.8  (https://roundcube.net/news/2018/10/26/update-1.3.8-released). What do you say?

blinkeye:
I think we need more work for the upgrade to Alpine:3.10. The /overrides approach seems not to work for dovecot v2.3.6 anymore. What you see in the logs is:
```
doveconf: Warning: /overrides/dovecot.conf line 7: Global setting mail_max_userip_connections won't change the setting inside an earlier filter at /etc/dovecot/dovecot.conf line 98 (if this is intentional, avoid this warning by moving the global setting before /etc/dovecot/dovecot.conf line 98)
``` 
Can be double checked with: `doveconf -n | grep -i Warning`.

This comes from an upstream change: https://github.com/dovecot/core/commit/87404eae4581d7ef834f490507503e59a500066e . Looking at that change closely it seems it has been merged since a long time, but I don't see the warning or error with Alpine:3.8 (with dovecot v2.3.5.1).

blinkeye:
Ok, my mistake was to just put a global with this setting inside `/overrides/dovecot.conf`, when you put it in the proper namespace:
```
protocol imap {
  mail_max_userip_connections = 110
}
```
the warning is gone.

Nebukadneza:
Hi,

thanks for catching PHP, I’ll try cherrypicking your commit to upgrade. Also good point with the overrides … I’ll override something in every service and see what happens.

core/postfix/start.py
```diff
@@ -48,12 +48,9 @@ def start_podop():
     os.system("postmap {}".format(destination))
     os.remove(destination)
 
-convert("/conf/rsyslog.conf", "/etc/rsyslog.conf")
-
 # Run Podop and Postfix
 multiprocessing.Process(target=start_podop).start()
 if os.path.exists("/var/run/rsyslogd.pid"):
```
kaiyou: These should not be required anymore

core/admin/Dockerfile
```diff
@@ -1,17 +1,18 @@
-FROM alpine:3.8
+FROM alpine:3.10
```
kaiyou: Maybe we could use an ARG for these so we can mimic the behavior from the arm PR.

core/dovecot/Dockerfile
```diff
@@ -1,14 +1,15 @@
...
 # python3 shared with most images
 RUN apk add --no-cache \
     python3 py3-pip git bash \
   && pip3 install --upgrade pip
-# Shared layer between rspamd, postfix, dovecot, unbound and nginx
-RUN pip3 install git+https://github.com/usrpro/MailuStart.git#egg=mailustart
+# Shared layer between admin, rspamd, postfix, dovecot, unbound and nginx
+RUN pip3 install git+https://github.com/Mailu/MailuStart.git#egg=mailustart
+# Shared layer between dovecot and postfix
+RUN pip3 install git+https://github.com/Nebukadneza/Podop.git@fix_py37
```
kaiyou: I will merge your PR so we don't have to pip install from Github, which is terrible.

kaiyou:
Hm. Could you PR on Podop ? :smiley:

Nebukadneza:
Hi,

the Podop-PR is already open:
https://github.com/Mailu/Podop/pull/6

Nebukadneza:
Hi,

I’ve rebased on the newest fetchmail change in master, verified it can now connect to another mailu and a external service, updated radicale too, and verified webdav works using davx5.
Podop is now still github, but the official mailu-repo at least — waiting for @kaiyou for pypi upload ^_^.

This is getting better slowly ^_^

Nebukadneza:
… pulling out seems so fiddly, i think it’s better to get in the upgrade first / quicker, and then later on pull out a set of build-variables into some kind of build configuration. That’s worthy of its own effort though, and i’d feel bad to mix simple updating, and refactoring the build scripts too much …
Hope this is ok for you

kaiyou:
Definitely great going forward and postponing this refactoring.

Nebukadneza:
:+1: — i also just tested this some more, put it on my dogfooding box, and it’s been going great for the day. Open for review ^_^

kaiyou:
The code looks right. I will publish a new version of Podop later, we should go forward with this in the meantime.

I will run it on my testing box before approving.

core/admin/Dockerfile
```diff
@@ -1,17 +1,18 @@
...
 # python3 shared with most images
 RUN apk add --no-cache \
     python3 py3-pip git bash \
   && pip3 install --upgrade pip
-RUN pip3 install git+https://github.com/usrpro/MailuStart.git#egg=mailustart
+# Shared layer between admin, rspamd, postfix, dovecot, unbound and nginx
+RUN pip3 install git+https://github.com/Mailu/MailuStart.git#egg=mailustart
```
blinkeye: Is this really needed? https://github.com/Mailu/Mailu/blob/master/core/admin/start.py does not use any (shared) functionality?

blinkeye:
I have a question about fetchmail: we are using an alpha version which we also have to mini-patch. Have you ever thought about using [getmail](http://pyropus.ca/software/getmail/CHANGELOG) which looks actively maintained and is Python based? From the looks of it there's no public github repository (only source tar balls available), but it's also used in official distros like Ubuntu (https://packages.ubuntu.com/xenial/mail/getmail4), **and**, the most recent version would be available in Arch Linux Edge (https://pkgs.alpinelinux.org/package/edge/testing/aarch64/getmail), it also seems to be no effort to install: [Arch Linux Build](https://git.alpinelinux.org/aports/tree/testing/getmail/APKBUILD)

core/admin/Dockerfile
HorayNarea: You are right.
And if this layer gets removed we should also change the comment in the other dockerfiles :)

(Maybe we should create mailu base-images so we could do
`FROM mailubase+python+mailustart`
or something, instead of the same text again and again in so many files…)

kaiyou: Creating a base image is certainly an approach. I would like to have time and provide Socrate properly before then, so we can properly pull it from pypi.

blinkeye:
For several days now I had a mixed prod setup of mailu:1.6 and some of the changes from this PR (I initially did not upgrade the start.py with the [Mailu/MailuStart](https://github.com/Mailu/MailuStart) refactoring):

- upgraded rspamd from alpine:3.8 to alpine:3.10 (rspamd v1.9.4)
- upgraded postfix from mailu:1.6 to alpine:3.10
- upgraded dovecot from mailu:1.6 to alpine:3.10
- upgraded roundcube from mailu:1.6 to alpine:3.10 (with all the dovecot changes from this PR)
- upgraded admin from mailu:1.6 to alpine:3.10 (with all the admin changes from this PR)
- upgraded fetchmail from mailu:1.6 to alpine:3.10 (with all the fetchmail changes from this PR)

I didn't see any problems with a small (distributed 3 node K8s cluster) setup of 5 users and 2 domains (about 2k messages so far on this setup).

What I irregularly but repeatedly noticed on the postfix container was a couple of minutes with:
```
ERROR:root:Error when processing request
Jul 08 02:50:40 mail postfix/trivial-rewrite[730]: warning: socketmap:unix:/tmp/podop.socket:domain socketmap server temporary error: unknown error
```
What I've noticed is that rspamd loses it's history if the container is restarted. I'm not sure if that's supposed to happen? But I think that's not related to the upgrade.

And: I think we have to provide a few changes/adaption to keep the K8s working. For example we introduced a  new environment variables `ADMIN_ADDRESS` which is now necessary in some `start.py`.

**Now I've also upgraded to the latest start.py from this PR. Will let you know of any issues.**

kaiyou:
Getmail ils certainly interesting, but I believe it is a job for another PR :)

HorayNarea:
> What I've noticed is that rspamd loses it's history if the container is restarted. I'm not sure if that's supposed to happen? But I think that's not related to the upgrade.

That's not supposed to happen and was already broken before the upgrade ;D

As far as I understand the rspamd docs we already do exactly what should be done to save the history across restarts, I think the problem has something to do with redis not saving its data correctly but I didn't have enough time (and motivation, to be honest.) to look into it.

blinkeye:
It's not directly related to this PR, but I noticed that we're missing some of the changes introduced in the startup behaviour for all other containers in fetchmail. We're still using the hardcoded url: `http://admin/internal/fetch`:

- [fetchmail.py](https://github.com/Mailu/Mailu/blob/master/services/fetchmail/fetchmail.py#L47)
- [fetchmail.py](https://github.com/Mailu/Mailu/blob/master/services/fetchmail/fetchmail.py#L85)

I can't commit on this PR. Do you prefer the necessary changes in another PR or do you want to make them yourself?

kaiyou:
I believe borg merges with master when testing, so they should get in when we finally merge.

core/admin/Dockerfile
Nebukadneza: IMO what @kaiyou says makes sense, if going the baseimage approach this would either be a separate effort/PR, or a good chance to combine with his socrate efforts, which seem great to me …

Nebukadneza: … and all of us didn’t notice admin uses `MailuStart` in 
https://github.com/Mailu/Mailu/blob/master/core/admin/mailu/configuration.py#L2
:joy: :joy: :joy:

hoellen: But with #1048 @kaiyou removed `MailuStart`. So I would prefer to merge it first and then rebase this PR here. 
@Nebukadneza maybe you can look into #1048, wheter the issues you had are now resolved and approve the PR :+1:

Nebukadneza: ah, and then rebase there … that makes sense. Will do!

Nebukadneza:
updated, despite a bit chaotically, and working again. yay ^_^

Issue #987, opened 2019-04-01
Demonstration server - Admin credentials do not seem to work.


d19dotca:
Hello,

I've been trying to use the Mailu demo server for checking out features and stuff and familiarizing myself with it before deploying it in my Docker environment. Unfortunately for several days now I have been unable to login with the credentials provided in the documentation. Can anyone please double-check this and reset if necessary?

From the docs:

Server name : test.mailu.io
IP address : 173.249.45.89
Webmail : https://test.mailu.io/webmail/
Admin UI : https://test.mailu.io/admin/
Admin login : admin@test.mailu.io
Admin password : letmein

glaszig:
problem persists. can somebody check or update the docs? thanks.

kaiyou:
We are currently trying to get in touch with people managing that test server. Otherwise we will setup a new one in the coming weeks.

fightchwang:
anybody contribute a practical username/password to login webmail? cause the credentials of admin seems not working. thx so much

Nebukadneza:
Hi,
we’ll try to set up a new server under the control of the currently active people soon. Please have some patience, and sorry for the delay!

muhlemmer:
Sorry about that guys, the demo server has been reset.

It seems the daily Cron job is not firing somehow.

muhlemmer:
Ok, the cron seemed to be defined in the wrong place. That is fixed now also. It should reset every night at 03:00 utc

glaszig:
well, still cannot log into the [admin panel](https://test.mailu.io/admin/ui/login) with the [credentials provided](https://mailu.io/1.6/demo.html).

<img width="352" alt="Bildschirmfoto 2019-07-07 um 12 33 06" src="https://user-images.githubusercontent.com/201135/60771272-6aa06f00-a0b3-11e9-98ad-6e7acf776050.png">

muhlemmer:
Ok, double checked and the cron job for resetting worked fine last night at 3 AM. I will close this issue now.

muhlemmer:
@glaszig, sorry completely missed your post. The thing is, after a reset the credentials work fine. I've checked in numerous times. I think someone is dicking around and changing the password all the time.

hoellen:
@muhlemmer Since #1067 is merged, can you please set a reset of the credentials every few minutes?
We receive continuous feedback about wrong access data. Thanks.

muhlemmer:
Yeah, that's done since so time now. Closing...

[...]

PR #1067, opened 2019-07-08
Update password in commandline

### LINKED ISSUE #1066: Update user password in commandline
A feature request, the ability to set/reset a user's password from the Mailu command line. Similar to the [user](https://mailu.io/master/cli.html#user) command:

```
docker-compose exec admin flask mailu password myuser example.net 'password123'
```

My particular interest is for the Demo server, some people find it necessary to change the admin password from time to time. Executing a full reset takes over a minute, in which period the Demo is not available. I don't want to execute it more frequent than daily. It would be nice to can just call above command every few minutes and always override the admin's password with the correct one. See also #987.

hoellen:
## What type of PR?

Feature

## What does this PR do?
Adds the ability to set/reset a user's password from the Mailu command line. Similar to the user command:
```
docker-compose exec admin flask mailu password myuser example.net 'password123'
```
With this we can reset the demo server password every few minutes (#987)

### Related issue(s)
closes #1066

## Prerequistes

### DIFF
--- a/core/admin/mailu/manage.py
+++ b/core/admin/mailu/manage.py
@@ -86,6 +86,26 @@ def user(localpart, domain_name, password, hash_scheme=None):
 
 
 @mailu.command()
+@click.argument('localpart')
+@click.argument('domain_name')
+@click.argument('password')
+@click.argument('hash_scheme', required=False)
+@flask_cli.with_appcontext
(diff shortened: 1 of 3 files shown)


hoellen:
bors try

PR #1078, opened 2019-07-17
expose pop3 port, remove imap-ssl from dovecot in kubernetes manifest

sbias:
## What type of PR?
bug-fix

## What does this PR do?
Exposes provided pop3 service in service definition to get useable by front

### DIFF
--- a/docs/kubernetes/mailu/imap.yaml
+++ b/docs/kubernetes/mailu/imap.yaml
@@ -32,6 +32,7 @@ spec:
         ports:
           - containerPort: 2102
           - containerPort: 2525
+          - containerPort: 110
           - containerPort: 143
           - containerPort: 993
           - containerPort: 4190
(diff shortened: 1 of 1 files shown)


kaiyou:
This is related to #1111. @micw, could review? You are probably more of a k8s expert than any of us.

kaiyou:
Adding a priority label if we can merge it in 1.7.

micw:
@kaiyou I found that issue too when creating my helm chart. The fix is valid and should be merged. I don't have approval permissions.

The ports should also be named in the deployment (not only in the service). That's not required but should be done for readability and consistency.

Maybe we should also remove the imap-ssl port exposure because it's completely unused (ssl termination is done by nginx).

sbias:
You are absolutely correct, the port should be added in the deployment too. At the same time we can add names to the ports there to make them consistent. Should i add it in this PR?

micw:
> Should i add it in this PR

I think so, then we have this done too and can close #1111 when this is merged.

sbias:
Ok, i will check imap-ssl removal too. If ssl is terminated in nginx there is no reason to map it in this service.

micw:
Thumbs up from my side. Thank you!

PR #1094, opened 2019-08-02
replace hardcoded attachment size in roundcube's .htaccess

blinkeye:
with a dynamically calculated value based on the environment value MESSAGE_SIZE_LIMIT

Current defaults from roundcube-1.3.9:

- upload_max_filesize=5M
- post_max_size=6M

## What type of PR?
Bug Fix

## What does this PR do?
Fixes the attachment size limit of Roundcube to be (dynamically) based on the environment variable `MESSAGE_SIZE_LIMIT` also in `.htaccess` file. Current behaviour: ignores the environment variable and limits the attachment/message size to 5MB/6MB respectively.

### DIFF
--- a/webmails/roundcube/start.py
+++ b/webmails/roundcube/start.py
@@ -7,9 +7,15 @@ from socrate import conf
 
 log.basicConfig(stream=sys.stderr, level=os.environ.get("LOG_LEVEL", "WARNING"))
 
-os.environ["MAX_FILESIZE"] = str(int(int(os.environ.get("MESSAGE_SIZE_LIMIT"))*0.66/1048576))
+max_filesize = str(int(int(os.environ.get("MESSAGE_SIZE_LIMIT"))*0.66/1048576))
+os.environ["MAX_FILESIZE"] = max_filesize
 
 conf.jinja("/php.ini", os.environ, "/usr/local/etc/php/conf.d/roundcube.ini")
+
+# Replace default values in /var/www/html/.htaccess (upload_max_filesize=5M, post_max_size=6M)
+os.system("sed -E -i 's/^php_value[[:space:]]+upload_max_filesize .*$/php_value   upload_max_filesize   %sM/g' /var/www/html/.htaccess" % (max_filesize))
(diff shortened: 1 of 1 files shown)


kaiyou:
I agree with the idea but not with the realization. We tend to use jinja for config templating and never sed magics.

The php.ini right above your code has all the bits required already.

micw:
Verified, could be closed because it's already implemented.

muhlemmer:
> Verified, could be closed because it's already implemented.

AlexFBP:
"Well... Yes, but actually, no"...? The values are being ignored because of the values inside the /var/www/html/.htacess, so it is needed to update not just the php.ini but also to modify the .htacess (So, i personally agree with the current title of this PR)

Or that, at least for a "fresh" (~2 weeks ago) mailu v1.7 setup with almost no changes. This appeared to me when writing a new email:

image

By attaching a bash session to the roundcube container, and creating something like ...

```bash
echo "<?php phpinfo();" > information.php
```

... to see what happened, appeared to me that, yes, the global value were 31MB, but, the local value is 5MB (needless to say this php file was `chmod -x`ed after final testings)

```bash
root@e4f798636d46:/var/www/html# cat /usr/local/etc/php/php.ini-production | grep upload_max_filesize
upload_max_filesize = 2M
root@e4f798636d46:/var/www/html# cat /usr/local/etc/php/php.ini-development | grep upload_max_filesize
upload_max_filesize = 2M
root@e4f798636d46:/var/www/html# cd /usr/local/etc/php/conf.d/; cat *.ini | grep upload_max_filesize
upload_max_filesize = 31M
root@e4f798636d46:/usr/local/etc/php/conf.d# cd -
/var/www/html
root@e4f798636d46:/var/www/html# cat .htaccess | grep size
php_value   upload_max_filesize   5M
php_value   post_max_size         6M
```

As an (ugly maybe?) workaround for v1.7, by just commenting this at the beginning of the .htaccess (note I commented also `memory_limit`, ignored too, in global it was 128M) like ...:

```apache
...
# php_value   upload_max_filesize   5M
# php_value   post_max_size         6M
# php_value   memory_limit          64M
...
```

... and then reloading with `/etc/init.d/apache2 reload` it given me:

image

so this was the need to modify the .htaccess, that afterall it was fixed in #1249 so, I guess it will be in the upcoming stable release...

PR #1104, opened 2019-08-11
Improve traefik-certdumper - run.sh and documentation

boldt:
## What type of PR?

Bug fix and documentation

## What does this PR do?

On the first run, the `run.sh` script shows the error `diff: can't stat '/output/cert.pem': No such file or directory`, because the file does not exist in the folder `/output` yet. This bugfix ensures, that the diff is only called, when all required files are available.

Additionally, the corresponding documentation was updated, since the `run.sh` writes the two files `cert.pem` and `key.pem`, and not `your.doma.in.crt` and `your.doma.in.key`. Also some volume fixes.

## Prerequistes

### DIFF
--- a/docs/reverse.rst
+++ b/docs/reverse.rst
@@ -178,7 +178,7 @@ One such example is ``mailu/traefik-certdumper``, which has been adapted for use
     volumes:
       # Folder, which contains the acme.json
       - "/data/traefik:/traefik"
-      # Folder, where your.doma.in.crt and your.doma.in.key will be written
+      # Folder, where cert.pem and key.pem will be written
       - "/data/mailu/certs:/output"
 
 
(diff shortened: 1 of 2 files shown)


docs/reverse.rst
```diff
@@ -176,20 +176,22 @@ One such example is ``mailu/traefik-certdumper``, which has been adapted for use
...
+Yay! Now let’s mount this to our ``front`` container like:
 
 .. code-block:: yaml
 
     volumes:
-      - "$ROOT/overrides/nginx:/overrides"
-      - /data/traefik/ssl/$TRAEFIK_DOMAIN.crt:/certs/cert.pem
-      - /data/traefik/ssl/$TRAEFIK_DOMAIN.key:/certs/key.pem
+      - /data/mailu/certs:/certs
+      - /data/mailu/certs:/certs
```
Nebukadneza: same line twice?

boldt: Good match. This is wrong. Probably a C&P issue. You can remove that doubled line, or I can update my PR.

boldt: Fixed.

PR #1106, opened 2019-08-14
nginx config.py: Don't resolve hosts

micw:
## What type of PR?

bug-fix

## What does this PR do?

It disables static name resolution in config.py because both, docker and kubernetes dynamically updates host ips. Nginx uses the resolver to properly resolve hostnames at runtime.

### Related issue(s)

- none

## Prerequistes

None, it's a minor change.


micw:
I close this and will provide a new PR that makes this behavior configurable

PR #940, opened 2019-02-15
Fix host variables

ionutfilip:
## What type of PR?
bug-fix

### Related issue(s)
- Fix #884

## Prerequistes
Before we can consider review and merge, please make sure the following list is done and checked.
If an entry in not applicable, you can check it or remove it from the list.

- [x] In case of feature or enhancement: documentation updated accordingly
- [x] Unless it's docs or a minor change: place entry in the [changelog](CHANGELOG.md), under the latest un-released version.

PR #1114, opened 2019-08-16
Resolve HOST to ADDRESS only if ADDRESS is not already set

### LINKED ISSUE #1113: Make dns name to ip resolving at startup configurable
Most services at mailu resolve hostnames to IP addresses at startup. On dynamic environments like Kubernetes, IP addresses for hostnames might change which will break a running deployment.
The behaviour should be configurable via "RESOLVE_SERVICE_HOSTNAMES" variable which defaults to true for backward compatibility.

micw:
## What type of PR?

bug-fix

## What does this PR do?

~Makes the rsolving from hosts to ips at startup configurable~

I rewrote the pull request after #940 was merged. Now it resolves HOSTs to ADDRESSes only of ADDRESSes are not already set. So on kubernetes we can jsut set the address and have working service discovery.

### Related issue(s)
- closes #1113

## Prerequistes

~Minor change, backward compatible~
Changelog will be added

### DIFF
--- a/core/admin/mailu/configuration.py
+++ b/core/admin/mailu/configuration.py
@@ -60,7 +60,9 @@ DEFAULT_CONFIG = {
     'HOST_SMTP': 'smtp',
     'HOST_AUTHSMTP': 'smtp',
     'HOST_ADMIN': 'admin',
+    'ANTISPAM': 'none',
     'HOST_ANTISPAM': 'antispam:11334',
+    'WEBMAIL': 'none',
     'HOST_WEBMAIL': 'webmail',
     'HOST_WEBDAV': 'webdav:5232',
     'HOST_REDIS': 'redis',
@@ -79,18 +81,26 @@ class ConfigManager(dict):
         'mysql': 'mysql://{DB_USER}:{DB_PW}@{DB_HOST}/{DB_NAME}'
     }
 
-    HOSTS = ('IMAP', 'POP3', 'AUTHSMTP', 'SMTP', 'REDIS')
-    OPTIONAL_HOSTS = ('WEBMAIL', 'ANTISPAM')
-
     def __init__(self):
         self.config = dict()
 
-    def resolve_host(self):
-        optional = [item for item in self.OPTIONAL_HOSTS if item in self.config and self.config[item] != "none"]
-        for item in list(self.HOSTS) + optional:
-            host = 'HOST_' + item
-            address = item + '_ADDRESS'
-            self.config[address] = system.resolve_address(self.config[host])
+    def get_host_address(self, name):
+        # if MYSERVICE_ADDRESS is defined, use this
+        if '{}_ADDRESS'.format(name) in os.environ:
+            return os.environ.get('{}_ADDRESS'.format(name))
+        # otherwise use the host name and resolve it
+        return system.resolve_address(self.config['HOST_{}'.format(name)])
+
+    def resolve_hosts(self):
+        self.config["IMAP_ADDRESS"] = self.get_host_address("IMAP")
+        self.config["POP3_ADDRESS"] = self.get_host_address("POP3")
+        self.config["AUTHSMTP_ADDRESS"] = self.get_host_address("AUTHSMTP")
+        self.config["SMTP_ADDRESS"] = self.get_host_address("SMTP")
+        self.config["REDIS_ADDRESS"] = self.get_host_address("REDIS")
+        if self.config["WEBMAIL"] != "none":
+            self.config["WEBMAIL_ADDRESS"] = self.get_host_address("WEBMAIL")
+        if self.config["ANTISPAM"] != "none":
+            self.config["ANTISPAM_ADDRESS"] = self.get_host_address("ANTISPAM")
 
     def __coerce_value(self, value):
         if isinstance(value, str) and value.lower() in ('true','yes'):
@@ -106,7 +116,7 @@ class ConfigManager(dict):
             key: self.__coerce_value(os.environ.get(key, value))
             for key, value in DEFAULT_CONFIG.items()
         })
-        self.resolve_host()
+        self.resolve_hosts()
 
         # automatically set the sqlalchemy string
         if self.config['DB_FLAVOR']:
--- a/core/admin/mailu/internal/nginx.py
+++ b/core/admin/mailu/internal/nginx.py
@@ -3,6 +3,9 @@ from flask import current_app as app
 
 import re
 import urllib
+import ipaddress
+import socket
+import tenacity
 
 
 SUPPORTED_AUTH_METHODS = ["none", "plain"]
@@ -88,4 +91,18 @@ def get_server(protocol, authenticated=False):
             hostname, port = extract_host_port(app.config['AUTHSMTP_ADDRESS'], 10025)
         else:
             hostname, port = extract_host_port(app.config['SMTP_ADDRESS'], 25)
+    try:
+        # test if hostname is already resolved to an ip adddress
+        ipaddress.ip_address(hostname)
+    except:
+        # hostname is not an ip address - so we need to resolve it
+        hostname = resolve_hostname(hostname)
     return hostname, port
+
+@tenacity.retry(stop=tenacity.stop_after_attempt(100),
+                wait=tenacity.wait_random(min=2, max=5))
+def resolve_hostname(hostname):
+    """ This function uses system DNS to resolve a hostname.
+    It is capable of retrying in case the host is not immediately available
+    """
+    return socket.gethostbyname(hostname)
--- a/core/dovecot/start.py
+++ b/core/dovecot/start.py
@@ -21,12 +21,13 @@ def start_podop():
     ])
 
 # Actual startup script
-os.environ["FRONT_ADDRESS"] = system.resolve_address(os.environ.get("HOST_FRONT", "front"))
-os.environ["REDIS_ADDRESS"] = system.resolve_address(os.environ.get("HOST_REDIS", "redis"))
-os.environ["ADMIN_ADDRESS"] = system.resolve_address(os.environ.get("HOST_ADMIN", "admin"))
-os.environ["ANTISPAM_ADDRESS"] = system.resolve_address(os.environ.get("HOST_ANTISPAM", "antispam:11334"))
+
+os.environ["FRONT_ADDRESS"] = system.get_host_address_from_environment("FRONT", "front")
+os.environ["REDIS_ADDRESS"] = system.get_host_address_from_environment("REDIS", "redis")
+os.environ["ADMIN_ADDRESS"] = system.get_host_address_from_environment("ADMIN", "admin")
+os.environ["ANTISPAM_ADDRESS"] = system.get_host_address_from_environment("ANTISPAM", "antispam:11334")
 if os.environ["WEBMAIL"] != "none":
-    os.environ["WEBMAIL_ADDRESS"] = system.resolve_address(os.environ.get("HOST_WEBMAIL", "webmail"))
+    os.environ["WEBMAIL_ADDRESS"] = system.get_host_address_from_environment("WEBMAIL", "webmail")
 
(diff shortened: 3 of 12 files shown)


kaiyou:
This sounds good to me, but I am not in favor of backporting immediately to 1.7. I would first give it a chance in master, then if we validate that the behavior is unchanged in a default setup maybe backport it for larger scale testing.

kaiyou:
Another review is more than welcome :)

muhlemmer:
Conflicts with most of the things done in #940.

micw:
> Conflicts with most of the things done in #940.

No worries, I'll rebase on master when #940 is merged.

core/dovecot/start.py
```diff
@@ -21,11 +21,20 @@ def start_podop():
...
+os.environ["ADMIN_ADDRESS"] = os.environ.get("HOST_ADMIN", "admin")
+os.environ["WEBMAIL_ADDRESS"] = os.environ.get("HOST_WEBMAIL", "webmail")
+
+# resolve hosts to ip addresses
+if os.environ.get('RESOLVE_SERVICE_HOSTNAMES','true').lower() in ('true','yes'):
+    os.environ["FRONT_ADDRESS"] = system.resolve_address(os.environ.get("FRONT_ADDRESS"))
+    os.environ["REDIS_ADDRESS"] = system.resolve_address(os.environ.get("REDIS_ADDRESS"))
+    os.environ["ADMIN_ADDRESS"] = system.resolve_address(os.environ.get("ADMIN_ADDRESS"))
+    if os.environ["WEBMAIL"] != "none":
+        os.environ["WEBMAIL_ADDRESS"] = system.resolve_address(os.environ.get("WEBMAIL_ADDRESS"))
```
muhlemmer: I'm not sure what the logic is here? In #940 we tried to achieve some consistency:
- `HOST_*` caries unresolved host names with optional port suffix
- `*_ADDRESS` caries resolved IP addresses with optional port suffix

In the above code I'm reading: 
1. Assign the unresolved hostname into (global) address.
2. if `RESOLVE_SERVICE_HOSTNAMES` is `true` (default) or `yes`:
2.1. Overwrite address with a resolved address

Perhaps something like:
````
os.environ.["FRONT_ADDRESS"] = os.environ.get("FRONT_ADDRESS",  system.resolve_address(os.environ.get("HOST_FRONT", "front"))
````
Using this logic, 
1. If the user sets `FRONT_ADDRESS` it is used without resolving
2. If `FRONT_ADDRESS` is emtpy, the resolver gets called:
2.1. If `HOST_FRONT` is set, that host name will be resolved
2.2. If `HOST_FRONT` is empty, the default host name `front` will be resolved.

This way we don't have to mangle between variables, or declare a new variable like `RESOLVE_SERVICE_HOSTNAMES`

File: core/nginx/config.py [hunk 2]
```diff
@@ -14,14 +14,20 @@
...
+args["WEBDAV_ADDRESS"] = args.get("HOST_WEBDAV", "webdav:5232")
+
+# resolve hosts to ip addresses
+if os.environ.get('RESOLVE_SERVICE_HOSTNAMES','true').lower() in ('true','yes'):
+    args["ADMIN_ADDRESS"] = system.resolve_address(args.get("ADMIN_ADDRESS"))
+    args["ANTISPAM_ADDRESS"] = system.resolve_address(args.get("ANTISPAM_ADDRESS"))
+    if args["WEBMAIL"] != "none":
+        args["WEBMAIL_ADDRESS"] = system.resolve_address(args.get("WEBMAIL_ADDRESS"))
+    if args["WEBDAV"] != "none":
+        args["WEBDAV_ADDRESS"] = system.resolve_address(args.get("WEBDAV_ADDRESS"))
```
muhlemmer: See  `core/dovecot/start.py`

core/postfix/start.py
```diff
@@ -26,10 +26,19 @@ def start_podop():
...
+os.environ["ADMIN_ADDRESS"] = os.environ.get("HOST_ADMIN", "admin")
+os.environ["ANTISPAM_ADDRESS"] = os.environ.get("HOST_ANTISPAM", "antispam:11332")
+os.environ["LMTP_ADDRESS"] = os.environ.get("HOST_LMTP", "imap:2525")
+
+# resolve hosts to ip addresses
+if os.environ.get('RESOLVE_SERVICE_HOSTNAMES','true').lower() in ('true','yes'):
+    os.environ["FRONT_ADDRESS"] = system.resolve_address(os.environ.get("FRONT_ADDRESS"))
+    os.environ["ADMIN_ADDRESS"] = system.resolve_address(os.environ.get("ADMIN_ADDRESS"))
+    os.environ["ANTISPAM_ADDRESS"] = system.resolve_address(os.environ.get("ANTISPAM_ADDRESS"))
+    os.environ["LMTP_ADDRESS"] = system.resolve_address(os.environ.get("LMTP_ADDRESS"))
```
muhlemmer: See  `core/dovecot/start.py`

services/rspamd/start.py
```diff
@@ -9,10 +9,15 @@
...
-if "HOST_REDIS" not in os.environ:
-    os.environ["REDIS_ADDRESS"] = system.resolve_address(os.environ.get("HOST_REDIS", "redis"))
+# set default values
+os.environ["FRONT_ADDRESS"] = os.environ.get("HOST_FRONT", "front")
+os.environ["REDIS_ADDRESS"] = os.environ.get("HOST_REDIS", "redis")
+
+# resolve hosts to ip addresses
+if os.environ.get('RESOLVE_SERVICE_HOSTNAMES','true').lower() in ('true','yes'):
+    os.environ["FRONT_ADDRESS"] = system.resolve_address(os.environ.get("FRONT_ADDRESS"))
+    os.environ["REDIS_ADDRESS"] = system.resolve_address(os.environ.get("REDIS_ADDRESS"))
```
muhlemmer: See  `core/dovecot/start.py`

muhlemmer:
Thanks for this PR, I imagine functionality like this is most welcome. I've left some comments that reflect my opinions. But you know what they say about opinions ;). Anyway, it's open to discussion.

core/admin/mailu/configuration.py
```diff
@@ -66,7 +66,8 @@
     'HOST_REDIS': 'redis',
     'HOST_FRONT': 'front',
     'SUBNET': '192.168.203.0/24',
-    'POD_ADDRESS_RANGE': None
+    'POD_ADDRESS_RANGE': None,
+    'RESOLVE_SERVICE_HOSTNAMES': os.environ.get('HOST_SMTP', 'true')
```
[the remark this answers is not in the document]
micw: Copy+Paste error, forgot to replace the variable. Shall be RESOLVE_SERVICE_HOSTNAMES

micw:
@muhlemmer I wrote that PR before #940 was merged. I will rethink and rewrite it when I understand fully why the hosts are resolved at all. IMO hosts should not be resolved at startup at all because it breaks service discovery on both, docker and kubernetes.

core/dovecot/start.py
micw: I have changed it accordingly

core/nginx/config.py
micw: Solved too

core/postfix/start.py
micw: solved too

services/rspamd/start.py
micw: also solved.

micw:
@muhlemmer I added the resolver code to socrate which got released yesterday. Then I re-worked to PR to use the new socrate. Could you please review again?
Thank you!

Michael.

muhlemmer:
I'm okay with how this looks. Thanks.

> Changelog will be added

Please do so. And perhaps the docs need an update on how we are parsing /resolving the `HOST` and `ADDRESS` variables.

We will retry the build and start testing. If all goes well and the Changelog and Docs are done, I will approve it.

----
bors try

core/admin/mailu/configuration.py
```diff
@@ -79,18 +81,26 @@ class ConfigManager(dict):
...
-    def resolve_host(self):
-        optional = [item for item in self.OPTIONAL_HOSTS if item in self.config and self.config[item] != "none"]
-        for item in list(self.HOSTS) + optional:
-            host = 'HOST_' + item
-            address = item + '_ADDRESS'
-            self.config[address] = system.resolve_address(self.config[host])
+    def get_host_address(self, name):
+        # if MYSERVICE_ADDRESS is defined, use this
+        if '{}_ADDRESS'.format(name) in os.environ:
+            return os.environ.get('%_ADDRESS' % name)
```
ionutfilip: Tested and looks like python doesn't like this syntax.
```
File "/app/mailu/configuration.py", line 119, in init_app
self.resolve_hosts()
File "/app/mailu/configuration.py", line 95, in resolve_hosts
self.config["IMAP_ADDRESS"] = self.get_host_address("IMAP")
File "/app/mailu/configuration.py", line 90, in get_host_address
return os.environ.get('%_ADDRESS' % name)
ValueError: unsupported format character '_' (0x5f) at index 1
```

Also tested only this function in a python shell and got same error

micw: @ionutfilip Thank you for pointing this out. Looks like I forgot this code when I reworked the PR

kaiyou:
This looks great now, and the feeling of finally unifying this is somehow satisfying. I have to run it a little bit before approving, probably in a couple hours.

muhlemmer:
New issues. In `mailu.env` we have set:
````
IMAP_ADDRESS=imap
LMTP_ADDRESS=imap:2525
POP3_ADDRESS=imap
SMTP_ADDRESS=smtp
AUTHSMTP_ADDRESS=smtp
ADMIN_ADDRESS=admin
#HOST_ANTISPAM=antispam:11334
WEBMAIL_ADDRESS=webmail
WEBDAV_ADDRESS=webdav:5232
REDIS_ADDRESS=redis
FRONT_ADDRESS=front
````
So the intention is that all hostnames are not resolved in startup, but at each connection. Now I can't login through imap and webmail (errors are the same in both cases)

````
front_1      | 2019/08/31 09:57:20 [error] 9#9: *595 auth http server 127.0.0.1:8000 sent invalid server address:"imap" while in http auth state, client: 192.168.203.11, server: 0.0.0.0:10143, login: "___@___"
````

Is there something wrong with our test variables? In which configuration did this work on your side?

I also found a bug related to antispam, but that already existed before this PR and I will announce it in a separate issue.

**Comment by micw** [comment]
@muhlemmer Looks like I broke my test system with a combination of too many different services from different pull requests. I try to fix it to give you feedback. I'm quite busy this weekend, so it might take a few days.

**Comment by micw** [comment]
@muhlemmer I got the system running again, now replacing the services one by one with my builds for #1114.

1. Replaced Postfix. Env for postfix is:
```
bash-5.0# env|grep -e _ADDRESS -e HOST_
POD_ADDRESS_RANGE=10.42.0.0/16
LMTP_ADDRESS=mailu-dovecot:2525
ANTISPAM_ADDRESS=mailu-rspamd:11332
FRONT_ADDRESS=mailu-front
ADMIN_ADDRESS=mailu-admin
```
works

**Comment by muhlemmer** [comment]
bors try

File: core/admin/mailu/internal/nginx.py [hunk 7]
```diff
@@ -88,4 +91,18 @@ def get_server(protocol, authenticated=False):
...
             hostname, port = extract_host_port(app.config['SMTP_ADDRESS'], 25)
+    try:
+        # test if hostname is already resolved to an ip adddress
+        ipaddress.ip_address(hostname)
+    except:
+        # hostname is not an ip address - so we need to resolve it
+        hostname = resolve_hostname(hostname)
     return hostname, port
+
+@tenacity.retry(stop=tenacity.stop_after_attempt(100),
```
kaiyou: I do not like this here, I would have prevered something more generic on the path to getting back to resolving things dynamically. But I guess it is a necessary evil toward it :)

PR #1117, opened 2019-08-20
Admin ui path

micw:
## What type of PR?

enhancement

## What does this PR do?

Makes the mapping of the admin ui configurable. For kubernetes it should run at /admin/ui, not at /ui so that default ingress settings without rewriting can be used.

### Related issue(s)

- should replace PR #1045

## Prerequistes

### DIFF
--- a/core/admin/mailu/__init__.py
+++ b/core/admin/mailu/__init__.py
@@ -42,7 +42,7 @@ def create_app_from_config(config):
 
     # Import views
     from mailu import ui, internal
-    app.register_blueprint(ui.ui, url_prefix='/ui')
+    app.register_blueprint(ui.ui, url_prefix=app.config.get('ADMIN_UI_PATH'))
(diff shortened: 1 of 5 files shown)


micw:
Docker image for testing is available at micwy/mailu-test-builds:admin--admin-ui-path

kaiyou:
As mentioned in the issue, I am really skeptical about the amount of complexity we add to Mailu related to k8s integration. In the present case, wouldn't pointing the ingress to the nginx front container solve the issue about path mapping? This, and we might one day rely on that nginx for specific purposes like IP filtering or what else.

micw:
@kaiyou I do agree. We currently rely too much on specific ingress features which might and will break things (I saw also a kubernetes setup pull request which updated nginx annotations because they have changed).

So the best way to deal with it in a portable manner would probably be to relay only on standard ingress features  and to everything special within the "application layer" - which would be the front controller in this case.

I will look onto it and provide a PR soon.

PR #1128, opened 2019-08-22
[SUGGESTION] fix for nginx-0.22.0 controller

sbias:
## What type of PR?
bug-fix

## What does this PR do?
As of a breaking change in nginx ingress controller 0.22 (see https://github.com/kubernetes/ingress-nginx/releases/tag/nginx-0.22.0 ) the rewrite annotation syntax has changed and results in an endless 308 redirect for admin ans static resources. Sadly this change breaks nginx-controllers < 0.22 .

### DIFF
--- a/docs/kubernetes/mailu/admin-ingress.yaml
+++ b/docs/kubernetes/mailu/admin-ingress.yaml
@@ -35,7 +35,7 @@ metadata:
   annotations:
     kubernetes.io/tls-acme: "true"
     certmanager.k8s.io/cluster-issuer: letsencrypt-stage
-    ingress.kubernetes.io/rewrite-target: "/ui"
+    ingress.kubernetes.io/rewrite-target: "/ui/$1"
     ingress.kubernetes.io/configuration-snippet: |
       proxy_set_header X-Forwarded-Prefix /admin;
   labels:
@@ -51,7 +51,7 @@ spec:
   - host: "mail.example.com"
     http:
       paths:
-      - path: "/admin/ui"
+      - path: "/admin/ui/?(.*)"
         backend:
           serviceName: admin
           servicePort: 80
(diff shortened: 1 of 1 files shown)


blinkeye:
Thanks @sbias for this PR. If I'm not mistaken this relates or is actually an improved fix similiar to #1045. I myself updated to nginx-0.24.1 and needed this change (but waited for the other PR to get merged).

sbias:
@blinkeye yes, you are right, sorry, didn't found that pr. So, this is a duplicate, feel free to close it.

micw:
@sbias @kaiyou Can we close this now since #1158 is merged?

sbias:
didn't seen #1158 yet, let me check that the weekend please. currently i don't see any authentication for rspamd anymore (was done in ingress) - maybe its done in the front-controller now (?)

sbias:
... nevermind, found it here: https://github.com/Mailu/Mailu/blob/master/core/nginx/conf/nginx.conf#L130
#1158 solves this, so i'm fine to close this PR

micw:
I have moved all special configs from ingress to nginx with this PR. This allows to use any ingress (usually nginx or trafik) and should fix almost all related issues.
If you are on kubernetes, consider to give my helm chart (https://github.com/micw/charts/tree/mailu/incubator/mailu) a try - feedback is highly welcome.

I have no perms to close this issue, so please close it ;)

sbias:
Okey, i close it, if @kaiyou has a veto we can reopen it anytime, but i think everything is done in your PR @micw

btw, i think that is a good choice to we do not replicate the logic for kubernetes anymore but use the same way like its running in plain docker. just unsure if i like the hostPorts :)
... Sorry, not using helm (yet) :)

micw:
What do you prefer over host ports? Service with nodePort doesn't give you the right ports and net=host is a bad idea.

PR #1148, opened 2019-08-29
update roundcube 1.3.10

HorayNarea:


kaiyou:
Is there any security issue fixed by this release? I would restrain from backporting otherwiae

HorayNarea:
the release notes state
`It contains fixes to several bugs backported from the master branch including minor security fixes`

Why would you restrain from backporting non-securitiy related upstream dotdot-releases? Aren't they generally considered safe to upgrade to because they should only include bug-fixes?

kaiyou:
> the release notes state
> 
> ```
> It contains fixes to several bugs backported from the master branch including minor security fixes
> 
> […]
> 
> This version is considered stable and we recommend to update all productive installations of Roundcube
> ```
> 
> Why would you restrain from backporting non-securitiy related upstream dotdot-releases? Aren't they generally considered safe to upgrade to because they should only include bug-fixes?

Sorry, I had not realized it was a minor release. Backporting sounds ok in that case indeed.

Issue #1133, opened 2019-08-22
Architecture : find some simplicity


kaiyou:
Spoiler : this issue and the opinion I express might very unpopular.

Something struck me while I was crawling back from the dead and working on Mailu again a couple months ago: it has become complex. In fact I believe it has become so complex that we urgently need to stop piling up on layers of features until we take a step back or two; debugging some of the issues made me realize that the current complexity is putting the project at risk under maintenance cost every time we try and add anything to the mix.

Below I suggest a few measures that we need to discuss in order to decrease complexity while not losing too many features. The idea is not to act immediately, but to first define a couple guidelines of what we should now avoid and/or fix, then open related issues.

1. Have one documented architecture, and only one.

    Currently Mailu has multiple flavors, including k8s and compose. For some flavors we ship a front container that proxifies everything, for some other we only route mail traffic through that front and use native features (ingress) to handle the rest.

    This will backfire, and it will backfire hard. Next time we rely on some critical feature of the front container, let's say for authenticating a sensitive component like an AV dashboard, people running under k8s will end up exposing unauthenticated bits on the Internet.

    We need to decide if we want to have a single front container and delegate some critical features to it like authentication and filtering. If we do, all flavors should use it and none should work around it, at least not in a supported way.

2. Use a decent naming convention and build on it

    We name containers differently than images, mostly for historical reasons and some kind of genericity for webmails. This is obsolete, especially if we want to support multiple webmails. Plus, it adds complexity when naming services for internal communications.

    We have many paths for web access and so many names that can be overridden. This is too much configuration. We should have simple rules, like "a container web app should be exposed through `/<container>`, like `/admin` or `/roundcube` or `/rspamd`", and stick to it, stop allowing cosmetic overrides that do not actually improve anything.

3. Stop resolving all the things

    We have start scripts that do dns resolution with a retry logic, and will fail at startup when something does not resolve yet, or block for minutes because of timeouts. Why do we need to resolve at startup? because dns is sometimes unreliable for a couple minutes.

    Our current setup works really bad on a swarm, and terribly bad of k8s because of this.

    I believe the solution is to go back to resolving live, and even resolve more (there are still some hardcoded ips in compiles configs that could go away). Maybe add some useful caching in podop or in admin so we do not stress the dns too much, then spend time understanding why dns fails sometimes. But definitely stop working around dns issues by disabling dns at runtime.

4. Push complexity to the admin API

[...]

PR #1151, opened 2019-08-31
Start writing the contribution guidelines

kaiyou:
## What type of PR?

Documentation

## What does this PR do?

It provides a first suggested version of contribution guidelines for the project. These are very open do discussion. This was suggested in #1133.

## Prerequistes

### DIFF
--- a/docs/contributors/guidelines.rst
+++ b/docs/contributors/guidelines.rst
@@ -10,7 +10,7 @@ constraints, some are architectural opinions.
 
 If you contribute to this project, please read through them all and read the
 architectural documentation. Reviewers will check that your contributions
-match these guidelines amont other things.
+match these guidelines among other things.
 
 If you feel one of the guidelines is wrong or incomplete, please come and
 discuss on our messaging channel, then open a pull request including the
@@ -74,7 +74,11 @@ antispam should be in a separate container.
 
 A container is developped as a singled directory under the proper category in
 the main repository, the only exception being service containers that should
-only use official Docker images.
+only use official Docker images. Categories are:
+
+- core, for mandatory components
+- optional, for optional components
+- webmail, for webmails
 
 A container image name must explicitely state the technology being used.
 Container versions are synchronized and all containers are always built at
@@ -82,7 +86,7 @@ once. The service name associated in the Compose file or Kubernetes configuratio
 should match the container image name.
 
 How configuration should be managed
-``````````````````````````````````
+```````````````````````````````````
 
 Anything that is static should be managed using environment variables.
 Configuration files should be compiled at runtime by the container `start.py`
@@ -97,7 +101,7 @@ admin API. The `podop` package binds mail specific sofware (Postfix and Dovecot
 at the moment) to the admin API, other containers should use specific API calls.
 
 What traffic should go through the nginx container
-````````````````````````````````````````````````to
+``````````````````````````````````````````````````
 
 All of it. All traffic, including HTTP(S), IMAP, SMTP, POP3, should go through
 the front container.
@@ -119,7 +123,10 @@ should be accessed at ``/<container_name>/<traffic_type_a>`` and
 ``/admin/api``.
 
 Traffic directed to ``/`` should be routed to an endpoint on the admin container
-responsible for dynamically redirecting it.
+responsible for dynamically redirecting it. That last redirect must use a
+302.
+
+Any traffic to a non existing endpoint must return a 404.
 
 
 How authentication should work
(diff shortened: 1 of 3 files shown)


docs/contributors/guidelines.rst
```diff
@@ -0,0 +1,205 @@
...
+the one that supports overriding the more configuration.
+
+In case the tool does not support conditional inclusion, we can add the
+override logic in the `start.py` script.
+
+How much should configuration be documented
+``````````````````````````````````````````
+
+We should not keep default documentation included by the distribution when
+providing configuration files.
```
muhlemmer: I'm curious about the rationale here. I might have quite some understanding how most parts in Mailu work. But I find it deeply frustrating to go through these bare-bones config files, for instance during bug-hunting. Its a real time waster to use Google in trying to understand what the config is doing.

It might be even worse for new users and potential contributors. Our online documentation is more or less how-to based, with a FAQ based on past incidents and there is small configuration reference. There is no resource telling anybody what most of the internal config means.

This might make the learning curve to steep for users that are experiencing issues or are looking to contribute.

docs/contributors/guidelines.rst
```diff
@@ -0,0 +1,205 @@
...
+------
+
+Coding standards
+````````````````
+
+All Python code should comply with PEP-8. We should review our code using
+pylint.
+
+We should comply with architectural recommendation from the Flask
+documentation.
```
muhlemmer: recommendation**s**

docs/contributors/workflow.rst
```diff
@@ -1,25 +1,8 @@
...
-The mailserver is designed as a whole, some images are therefore not best
-suited for reuse outside this project. All images should however follow
-Docker best practices and be as generic as possible :
-
- - even if not suited for reuse, they should be simple enough to
-   fit as base images for other projects,
- - interesting settings should be available as environment variables
- - base images should be well-trusted (officiel Alpine or Debian for instance).
-
-.. _git_workflow:
```
muhlemmer: https://github.com/Mailu/Mailu/blob/master/docs/contributors/environment.rst#git links to this deleted reference.

docs/contributors/workflow.rst
```diff
@@ -53,7 +36,7 @@ master directly if you find this appropriate. Still, keep in mind that:
   either by the name of the Github issue or a short and meaningful name.
 
 PR Workflow
-````````````
+----------
```
muhlemmer: Underline too short

docs/contributors/guidelines.rst
```diff
@@ -0,0 +1,205 @@
...
+What is a Mailu container
+`````````````````````````
+
+A Mailu container should provide one service and run one type of process only.
+A new Webmail should be in a separate container, a new antivirus or a new
+antispam should be in a separate container.
+
+A container is developped as a singled directory under the proper category in
+the main repository, the only exception being service containers that should
+only use official Docker images.
```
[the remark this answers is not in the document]
kaiyou: Yes I mean redis, maybe later some others (no specific plans, but heh, could happen). Maybe utility containers would avoid confusion?

docs/contributors/guidelines.rst
```diff
@@ -0,0 +1,205 @@
...
+go directly from the nginx container to the target container.
+
+All traffic to a container should be accessed at ``/<container_name>`` for a given
+container. If multiple types of traffic are routed for a specific container, they
+should be accessed at ``/<container_name>/<traffic_type_a>`` and
+``/<container_name>/<traffic_type_b>``, for instance ``/admin/ui`` or
+``/admin/api``.
+
+Traffic directed to ``/`` should be routed to an endpoint on the admin container
+responsible for dynamically redirecting it.
```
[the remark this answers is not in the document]
kaiyou: Nginx is able to match exact path, and my intention is to only redirect `/`, not subpaths. Maybe I should mention this as well.

docs/contributors/guidelines.rst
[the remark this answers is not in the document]
muhlemmer: I'm thinking of the current state of the `services` directory in the repo:
- Fetchmail -> Is actually `optional`
- Rspamd -> Postfix breaks without it, perhaps should be `core`?
- Unbound -> Is actually `optional`

So perhaps we can define 3 groups, instead of 4:
1. Core
2. Optional
3. Webmails

What's your view on this?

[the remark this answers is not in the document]
kaiyou: Agreed with the 3-directory approach :)

docs/contributors/guidelines.rst
kaiyou: I agree we should add more documentation to our config files. But I believe our documentation should state what our specific config does and why we wrote it that way, not include the list of available options and their meaning for every available setting.

kaiyou:
Sorry about the typos, I will fix them asap. I am also thinking of adding to the workflow how we should use Github issues, what do you think?

One of the ideas is to more clearly use projects for long term changes that are actually implemented by many issues.

docs/contributors/guidelines.rst
muhlemmer: Yes, you are right.

docs/contributors/guidelines.rst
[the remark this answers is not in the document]
muhlemmer: Yes, that would be better

muhlemmer:
> One of the ideas is to more clearly use projects for long term changes that are actually implemented by many issues.

Projects would be great! I'm currently working on an improvement plan around Mailu storage vs scaleability. And I was already breaking my head over managing the improvement process.

Projects also allow for adding cards without issues. Easy if you just have some quick reminders or tasks and not spamming the issue list with it.

In my opinion issues are the interface with the user community. A place for Bugs and Suggestions. Internal processes and reminders would only clutter the overview.

How I usually check FOSS projects for fitness:
1. What was the last commit on master? (is the project mainained)
2. Check the issue list: are there many critical bugs? Are the maintainers responding?
3. Recently merge pull request: is there a healthy relationship with the community?

So IMHO keeping the issue list a clean place is important for attracting more users (== more potential contributors).

kaiyou:
Let's try projects then. I'll try and write a separate PR that amends the workflow and details how we manage issues, task lists, and reviews. Feel free to beat me at it if you can write anything before sunday :)

docs/contributors/guidelines.rst
```diff
@@ -0,0 +1,205 @@
...
+Why these guidelines
+--------------------
+
+This page is written as an evolving memo of the main development guidelines
+that we follow or should follow as contributors. Some are basic development
+constraints, some are architectural opinions.
+
+If you contribute to this project, please read through them all and read the
+architectural documentation. Reviewers will check that your contributions
+match these guidelines amont other things.
```
pierreozoux: Among

docs/contributors/guidelines.rst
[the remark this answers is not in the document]
pierreozoux: In 12factor.net , they use the term backing service.

Maybe you can rephrase it and call them stock backing service as opposed to customized.

Stock would be redis because you consume the image as is.
And customized, would be postfix as you highly customize it (in the event there would be an official postfix docker image)

kaiyou:
The remaining unresolved comment was actually resolved I believe. Ready to move forward :)

PR #1183, opened 2019-09-27
Fix rspamd-learn when moving mail from/to junk folder

### LINKED ISSUE #1177: Mailu 1.7 : Issue when moving mail to the Junk folder
When I move a mail to the Junk Folder, i see the following in the logs:
imap_1       | Sep 22 16:51:26 imap: Error: cannot stat file mailu
imap_1       | Sep 22 16:51:26 imap: Error: cannot stat file mailu
imap_1       | Sep 22 16:51:26 imap: Error: tee: /dev/fd/63: I/O error
imap_1       | Sep 22 16:51:26 imap(xxx@yyy.net)<27629><hn93GCeTsresEgAC>: Info: program exec:/conf/bin/spam (27655): Terminated with non-zero exit code 1
imap_1       | Sep 22 16:51:26 imap(xxx@yyy.net)<27629><hn93GCeTsresEgAC>: Info: sieve: left message in mailbox 'Junk'

Nebukadneza:
Before, the ham/spam scripts got the rspamd-ip/port from the environment.
However, when checking the environment of these processes now, it seems
cleared. Maybe the new dovecot version now clears environment? — I couldn’t
find a hint.

In any case, using the common mechanism of injecting the ip/port from where
it’s definately known by the already-used jinja2-mechanism seems reasonably
safe.

## What type of PR?
bug-fix

## What does this PR do?
Instead of relying on dovecot passing our environment cleanly to sieve-called scripts, this explicitly injects the antispam ip/port into the spam/ham scripts used when moving files from/to the spam-folder. This required some management of the files, such as setting proper permissions after the jinja-run.

### Related issue(s)
fixes #1177 

## Prerequistes

### DIFF
--- a/core/dovecot/start.py
+++ b/core/dovecot/start.py
@@ -37,7 +37,7 @@ os.makedirs("/conf/bin", exist_ok=True)
 for script_file in glob.glob("/conf/*.script"):
     out_file = os.path.join("/conf/bin/", os.path.basename(script_file).replace('.script',''))
     conf.jinja(script_file, os.environ, out_file)
-    os.chmod(out_file, 0555)
+    os.chmod(out_file, 0o555)
 
 # Run Podop, then postfix
 multiprocessing.Process(target=start_podop).start()
--- a/core/dovecot/start.py
+++ b/core/dovecot/start.py
@@ -33,14 +33,11 @@ if os.environ["WEBMAIL"] != "none":
 for dovecot_file in glob.glob("/conf/*.conf"):
     conf.jinja(dovecot_file, os.environ, os.path.join("/etc/dovecot", os.path.basename(dovecot_file)))
 
-try:
-    os.mkdir("/conf/bin")
-except FileExistsError:
-    pass
+os.makedirs("/conf/bin", exist_ok=True)
 for script_file in glob.glob("/conf/*.script"):
     out_file = os.path.join("/conf/bin/", os.path.basename(script_file).replace('.script',''))
     conf.jinja(script_file, os.environ, out_file)
-    os.chmod(out_file, stat.S_IRUSR | stat.S_IXUSR | stat.S_IRGRP | stat.S_IXGRP | stat.S_IROTH | stat.S_IXOTH)
+    os.chmod(out_file, 0555)
 
 # Run Podop, then postfix
 multiprocessing.Process(target=start_podop).start()
--- a/core/dovecot/start.py
+++ b/core/dovecot/start.py
@@ -1,7 +1,6 @@
 #!/usr/bin/python3
 
 import os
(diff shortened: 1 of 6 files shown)


core/dovecot/start.py
```diff
@@ -32,6 +33,15 @@ def start_podop():
 for dovecot_file in glob.glob("/conf/*.conf"):
     conf.jinja(dovecot_file, os.environ, os.path.join("/etc/dovecot", os.path.basename(dovecot_file)))
 
+try:
```
kaiyou: Makedirs sounds more fitted for this and will remove the need for exception handling.

core/dovecot/start.py
```diff
@@ -32,6 +33,15 @@ def start_podop():
...
+    os.mkdir("/conf/bin")
+except FileExistsError:
+    pass
+for script_file in glob.glob("/conf/*.script"):
+    out_file = os.path.join("/conf/bin/", os.path.basename(script_file).replace('.script',''))
+    conf.jinja(script_file, os.environ, out_file)
```
kaiyou: I personally find this harder to read than plain old 0555.

core/dovecot/start.py
```diff
@@ -32,6 +33,15 @@ def start_podop():
 for dovecot_file in glob.glob("/conf/*.conf"):
     conf.jinja(dovecot_file, os.environ, os.path.join("/etc/dovecot", os.path.basename(dovecot_file)))
 
+try:
+    os.mkdir("/conf/bin")
+except FileExistsError:
+    pass
+for script_file in glob.glob("/conf/*.script"):
```
kaiyou: We should not write these files directly to the source directory: everything in that directory is seen as executable in sieve scripts. I guess there is very little chance the source jinja contains an exploitable vulnerability without actually passing to a jinja parser, but you never really are too safe with whitelists.

kaiyou: Never mind, it seems I can't read, or distinguish /conf from /conf/bin :unamused:

core/dovecot/start.py
Nebukadneza: you are, of course, right. Thanks, I didn’t know makedirs had `exist_ok` …

core/dovecot/start.py
Nebukadneza: mhh, ack too. Note that python2 != python3 in leading `0` being enough for octal … \*grml\*

Nebukadneza:
(Note there’s also #1184 against the 1.7 branch, where this is broken for actual users)

kaiyou:
Really last comment since we are being picky, you do not seem to need stat anymore. Once solved I'll approve them both for merge.

PR #1196, opened 2019-10-07
Change default password scheme to PBKDF2 (#1194)

### LINKED ISSUE #1194: Make PBKDF2 the default password scheme

hoellen:
## What type of PR?
enhancement

## What does this PR do?
This PR change the default password scheme to PBKDF2. It is already changed in some places (e.g. [docs/compose/.env](https://github.com/Mailu/Mailu/blob/master/docs/compose/.env#L142)).

### Related issue(s)
closes #1194 

## Prerequistes


kaiyou:
This looks good to me :)

nextgens:
https://github.com/Mailu/Mailu/issues/1194#issuecomment-761564557 adding this here to ensure it's seen @kaiyou  @hoellen

PR #1208, opened 2019-10-11
Persist mailqueue

### LINKED ISSUE #1161: SMTP mail queue is not persistent
While writing some proposals for Mailu storage, I actually found that the `/queue` directory for postfix is not defined as a volume. This might result in a loss of mail in case of container shutdown/crash/reboot.

https://github.com/Mailu/Mailu/blob/1f0b45e47b44d892305c59521a6b30386048370c/core/postfix/conf/main.cf#L14

In the blame I found a commit that added this in the configuration, but the Dockerfile or `docker-compose.yml` never got updated. I believe this still needs to be done.

Related: #211

micw:
## What type of PR?

bug-fix

## What does this PR do?

Makes postfix mailqueue presistent (for docker, swarm and kubernetes)

### Related issue(s)
- closes #1161

## Prerequistes

### DIFF
--- a/docs/kubernetes/mailu/smtp.yaml
+++ b/docs/kubernetes/mailu/smtp.yaml
@@ -27,9 +27,6 @@ spec:
             memory: 2Gi
             cpu: 500m
         volumeMounts:
-          - mountPath: /data
-            name: maildata
-            subPath: maildata
           - mountPath: /queue
             name: maildata
             subPath: mailqueue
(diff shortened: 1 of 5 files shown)


docs/kubernetes/mailu/smtp.yaml
```diff
@@ -30,6 +30,9 @@ spec:
           - mountPath: /data
             name: maildata
             subPath: maildata
```
muhlemmer: `/data` is correctly removed in Dockerfile, since it is obsolete. I think it should be removed from here as well?

micw: Right, I've change it.

PR #1209, opened 2019-10-11
Use POD_ADDRESS_RANGE for postfix mynetworks if set

### LINKED ISSUE #1210: Vacation mails not working on kubernetes
On kubernetes, vacation mails sent by dovecot get rejected by postfix. The reason is that the mails are sent unauthenticated but postfix does not thread them as local mails. This is due to a missing entry in mynetworks which should contain POD_ADDRESS_RANGE if set.

micw:
## What type of PR?

bug-fix

## What does this PR do?

On Postfix, we already use POD_ADDRESS_RANGE for smtpd_authorized_xclient_hosts. This must also be used for mynetworks, otherwise it's not possible to send mails from other mailu pods (e.g. vacation mails from dovecot).

### Related issue(s)

- closes #1210

## Prerequistes

### DIFF
--- a/core/postfix/conf/main.cf
+++ b/core/postfix/conf/main.cf
@@ -17,7 +17,7 @@ queue_directory = /queue
 message_size_limit = {{ MESSAGE_SIZE_LIMIT }}
 
 # Relayed networks
-mynetworks = 127.0.0.1/32 [::1]/128 {{ SUBNET }} {{ RELAYNETS }}
+mynetworks = 127.0.0.1/32 [::1]/128 {{ POD_ADDRESS_RANGE or SUBNET }} {{ RELAYNETS }}
 
 # Empty alias list to override the configuration variable and disable NIS
 alias_maps =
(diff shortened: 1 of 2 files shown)


muhlemmer:
If memory serves me well, I proposed to use `SUBNET` to replace `POD_ADDRESS_RANGE` Since `POD_` is a k8s specific thingy and non-intuitive for regular Docker deployments. At the time we've agreed to keep `POD_ADDRESS_RANGE` around in order not to instantly break k8s.

So I'm not sure if we should continue to further implement `POD_ADDRESS_RANGE`? Wouldn't it be possible to use a `SUBNET` variable in k8s instead?

micw:
I think the reason that it's used comes from that:

```
./docs/kubernetes/mailu/configmap.yaml:    # POD_ADDRESS_RANGE is normally provided by default with Kubernetes
```

I have checked this, It's not set on any of my kubernetes setups. I also googled for that variable name, the only hits I get are from mailu.

So I think this decision was made by a wrong assumption. I'm for removing it entirely and write a changelog with a "breaking change" warning (can we do that?).

muhlemmer:
> (can we do that?).

In my opinion yes, but maybe its better to vote on it.

kaiyou:
My thumb up is vote in favor, anything that can reduce the number of variables and keep things consistent :)

PR #1214, opened 2019-10-13
Move services into core and optional

muhlemmer:
## What type of PR?

Enhancement

## What does this PR do?

The current state of the services directory in the repo:
- Fetchmail -> Is actually **optional**
- Rspamd -> Postfix breaks without it, should be **core**
- Unbound -> Is actually **optional**

This PR moves those services into the proper core or optional directory. From now on we will have just 3 categories of images:
1. Core
2. Optional
3. Webmails

### Related issue(s)
- [Discussed](https://github.com/Mailu/Mailu/pull/1151#discussion_r320936937) in PR 1151

## Prerequistes

- [x] In case of feature or enhancement: documentation updated accordingly **(I could not find any docs describing the repo layout. Please correct me if I was wrong)**


towncrier/newsfragments/1214.enhancement
```diff
@@ -0,0 +1 @@
+Move rspamd, fetchmail and unbound services into core and optional
```
hoellen: Same [comment](https://github.com/Mailu/Mailu/pull/1221/files#r334758529) on file extension as for PR #1221.

muhlemmer: Renamed to `.misc`

kaiyou:
Blocked by the related issue if we are to be consistent, ready to merge otherwise :)

muhlemmer:
> Blocked by the related issue if we are to be consistent, ready to merge otherwise :)

I'm guessing you mean PR #1151? In that case: all good to go!

PR #1221, opened 2019-10-14
Dovecot: Delete obsolete data volume

muhlemmer:
## What type of PR?

Enhancement

## What does this PR do?

The /data volume was used in the past for Dovecot to access the main.db SQLite file. This is no longer the case and the volume has become obsolete.

## Prerequistes


towncrier/newsfragments/1221.enhancement
```diff
@@ -0,0 +1 @@
+Dovecot: Delete obsolete data volume
```
hoellen: We usually don't use the `.enhancement` extension for [towncrier](https://mailu.io/1.7/contributors/guide.html#changelog). I'm not sure if towncrier supports custom extensions? Maybe `.removal` or `.misc` fits better.

muhlemmer: Changed to `.removal`

PR #1224, opened 2019-10-14
RFC: Mailu directory structure

muhlemmer:
## What type of PR?
RFC / design documentation

## What does this PR do?
Describes a proposal to restructure the `/mailu` directories to allow for easier and more clear configuration in replicated environments.

It proposes the following layout:

````
/mailu
├── config
│   ├── dovecot
│   ├── postfix
│   ├── rainloop
│   ├── redis
│   ├── roundcube
│   │   └── gpg
│   ├── rspamd
│   └── share
│       ├── certs
│       └── dkim
├── data
│   ├── admin
│   ├── rainloop
│   ├── roundcube
│   └── rspamd
├── local
│   ├── clamav
│   └── mailqueue
└── mail
````

Where in replicated environments:

- `/mailu/config/`: should be a small, low performant and shared filesystem.
- `/mailu/data`: should be avoided. More work will need to be done to configure external DB servers for relevant services. Ideally, this directory should only exist on docker-compose deployments.
- `/mailu/local/`: Should exist only on local file systems of worker nodes.
- `/mailu/mail`: A distributed filesystem with sufficient performance and storage requirements to hold and process all user mailboxes. Ideally only Maildir without indexes.

### DIFF
--- a/design/mailu-directory-structure.md
+++ b/design/mailu-directory-structure.md
@@ -1,0 +1,199 @@
+# RFC: Mailu directory structure
+The current layout of the Mailu directory structure can be improved to allow for easier replicated deployments, like Docker Swarm and Kubernetes. Please read https://usrpro.com/publications/mailu-persistent-storage/ for the background motivation of this RFC.
+
+## Scope
+This document only describes the re-arrangement of the `$ROOT` Mailu filesystem as-is. This means moving around of current files and directories. The linked article also proposes more advanced improvements. Those are not included in this RFC and need to be evaluated and implemented independently. However, the changes proposed in this document should make such improvements easier.
+
+Currently some services are wrongfully sharing mountpoints or have unused volumes declared. Those are also taken care of in this RFC.
+
+## Compatibility
+If these changes were to be accepted, it will break compatibility with previous Mailu version `<=1.7`. As such, we will propably increment to version `2.0`.
+
+## Root
+The root of the Mailu filesystem is located at `/mailu` by default. For simplicity we will assume this location throughout the document. Within `/mailu` we will aim to define 3 main sub-directories:
+
+### Config
+
+- Path: `/mailu/config/`
+
+Small config bearing files, sometimes shared between multiple services. The performance and storage needs for this filesystem are low. Availability is important for correct functioning of the mail server. No file locking issues are expected from concurrent access. A basic (and redundant) filesystem should suffice.
+
+#### Dovecot
+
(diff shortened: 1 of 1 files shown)


kaiyou:
All of this sounds great, and pushes into the right direction.

I see one requirement that needs to be specified: we will first need to remove the webmail genericity and have more specific mount points for the webmail.

micw:
@muhlemmer I'd say this structure is already possible since everything is in docker volumes. So it's just a matter of setting the right path mappings on your docker-compose.yaml.
For a replicated setup, having mail on a shared path is probably not enough:
- mail spool should be on high-available storage (makes less sense to protect/replicate mails in maildirs but not in in/out spool) so it should not be in local
- Database storage needs to be high-available (e.g. using replicated mysql or postgress, not sqlite3) - can't see the store for it in your proposal
- redis should be ha (using sentinel), clients needs to speak the redis-sentinel protocol and be configured accordingliy. This means that redis-store should be in local.

micw:
@muhlemmer Thank you for pointing things out. I don't know much of Mailu's history and decisions, so this helps me a lot to understand why things are as they are.
I strongly agree that providing extended DB/Redis HA is out of Mailu's scope.

I have read many helm charts, the common way is (e.g. for mysql): if an external mysql address is set, use it. otherwise deploy an internal single-node mysql.

lub:
> I'm not against your proposal per se, it's just that I don't think Postfix likes its queue being shared by multiple instances, but I will need to investigate this.

fyi it seems like it is not supported.


> The reason is that you cannot share Postfix queues among multiple running Postfix instances. You can use NFS to switch a Postfix mail queue from one NFS client to another one, but only one NFS client can access a Postfix mail queue at any particular point in time. 

http://www.postfix.org/NFS_README.html

Diman0:
I like this. It provides a better folder structure then we currently have. The current structure is not self-explanatory like this new structure is. But it does require a lot of documentation changes that instructs how to migrate your data folder to the new structure. 
Existing documentation, like for example the overrides faq, must also be modified.
This would really be a 2.0 release (as you already mentioned).

Diman0 (approved):
We discussed this during the last Mailu meeting. The proposed structure looks good. The proposed structure is excellent.
As you already stated, implementing this needs coordination with the helm-charts repo. This is something for a 2.0 release.

We will accept this and PR and we will come back to this once we are ready for a 2.0 release.

nextgens (approved):

PR #1241, opened 2019-11-06
Change extensions/v1beta to apps/v1 to be compliant with Kubernetes 1.16

### LINKED ISSUE #1237: Kubernetes: apiVersion extensions/v1beta1 is now apps/v1
As of Kuberentes v1.16 the api-version extensions/v1beta1 does not work anymore. 

It sould be updated to apps/v1.

WebSpider:
## What type of PR?

Enhancement

## What does this PR do?

Changes to Kubernetes YAML files to be able to use this in Kubernetes 1.16

### Related issue(s)
- Fixes #1237

## Prerequistes

- N/A In case of feature or enhancement: documentation updated accordingly

### DIFF
--- a/docs/kubernetes/mailu/ingress.yaml
+++ b/docs/kubernetes/mailu/ingress.yaml
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: Ingress
 metadata:
   name: mailu-ingress
--- a/docs/kubernetes/mailu/admin.yaml
+++ b/docs/kubernetes/mailu/admin.yaml
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: mailu-admin
--- a/docs/kubernetes/mailu/fetchmail.yaml
+++ b/docs/kubernetes/mailu/fetchmail.yaml
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: mailu-fetchmail
--- a/docs/kubernetes/mailu/imap.yaml
+++ b/docs/kubernetes/mailu/imap.yaml
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: mailu-imap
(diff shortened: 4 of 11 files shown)


kaiyou:
This looks fine to me. @micw would have time to review as well?

micw:
It will break compatibility with older kubernetes versions and not everyone has already updated. On the helm chart we can make it contitional to the k8s version or available APIs. For the static yamls the user must edit it accordingly.
Since the next release it in a few months (giving users time to upgrade their k8s), I'm fine with that breaking change.

kaiyou:
@micw could you update your review for the merge to go?

micw:
Already did, just forgot to approve

garykrige:
Does this actually work for anyone? I get `error validating data: ValidationError(Deployment.spec): missing required field "selector" in io.k8s.api.apps.v1.DeploymentSpec;`

docs/kubernetes/mailu/ingress.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
```
garykrige: This does not work. It should be `networking.k8s.io/v1beta1`

micw:
Oh, right. k8s is now also a bit picky about correct selectors ;)
Looks like we need one more PR.

docs/kubernetes/mailu/ingress.yaml
micw: Can you confirm that this works on 1.16? extensions/v1beta1 still works for me and it's going to finally change in 1.20 (https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/).

But apps/v1 is definitely wrong. I must have overlooked this when reviewing the change.

micw:
You can find verified-working yamls at https://github.com/Mailu/helm-charts/tree/master/mailu/templates. Would you mind to create a new PR against master containing your findings?

docs/kubernetes/mailu/ingress.yaml
garykrige: yeah I've just upgraded to 1.16 and its working

garykrige:
I'll make a PR at some point, but its frustrating for broken config to pass so many checks

kaiyou:
k8s config still does not undego any automated test (and is not displayed as properly supported yet). If you find it frustrating, by all mean help us through the review process :)

PR #1215, opened 2019-10-13
Allow specifying the traefik version for cert dumping

timoschwarzer:
## What type of PR?

Enhancement

### Related issue(s)
- #1011 

## Prerequistes
Before we can consider review and merge, please make sure the following list is done and checked.
If an entry in not applicable, you can check it or remove it from the list.

- [X] In case of feature or enhancement: documentation updated accordingly
- [X] Unless it's docs or a minor change: add [changelog](https://mailu.io/master/contributors/guide.html#changelog) entry file.

PR #1247, opened 2019-11-13
Fix traefik-certdumper for Traefik v2

oal:
Running Traefik v2, the current container doesn't export certificates from acme.json anymore. Can be fixed by adding `--version v2`, but I'm not sure if that should be the default now, or an env variable? Feel free to merge if v2 format should be the default moving forward.

Without `--version v2`, this is what I get:
```
certdumper_1  | Wed Nov 13 20:59:22 UTC 2019 Dumping certificates
certdumper_1  | Wed Nov 13 20:59:22 UTC 2019 Certificate or key differ, updating
certdumper_1  | mv: can't rename '/tmp/work/mydomain.com/*.pem': No such file or directory
```

## What type of PR?
bug-fix

## What does this PR do?
Makes the container work with Traefik v2

### DIFF
--- a/optional/traefik-certdumper/run.sh
+++ b/optional/traefik-certdumper/run.sh
@@ -3,7 +3,7 @@
 function dump() {
     echo "$(date) Dumping certificates"
 
-    traefik-certs-dumper file --crt-name "cert" --crt-ext ".pem" --key-name "key" --key-ext ".pem" --domain-subdir --dest /tmp/work --source /traefik/acme.json > /dev/null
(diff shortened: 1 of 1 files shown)


kaiyou:
bors try

kaiyou:
Now that it builds, my only request is that we specify in the docs that traefik 2 is expected when reverse proxying Mailu.

Nebukadneza:
There is also https://github.com/Mailu/Mailu/pull/1215 which is a duplicate of this — 1215 allows specifying the traefik version via env in the dumper parameters, which does not break the setup for traefik 1 users …

kaiyou:
Agreed. I'll review #1225 right now.

kaiyou:
Closing this as duplicate

PR #1257, opened 2019-11-18
Allow local IPv6 connections to Postgres

tulir:
## What type of PR?

bug-fix

## What does this PR do?

Fix postgres connection not working when IPv6 is enabled


kaiyou:
Simple enough and useful until we retire that postgres image. Thank you!

kaiyou:
Anyone else?

PR #1266, opened 2019-11-21
Install py3-multidict from repository before installing socrate

micw:
to avoid the need of gcc during build

## What type of PR?

bug-fix

## What does this PR do?

Install py3-multidict on Docker images where socrate is used. Otherwise the build fails on socrate >=2.5 due to missing gcc

### DIFF
--- a/core/dovecot/Dockerfile
+++ b/core/dovecot/Dockerfile
@@ -2,7 +2,7 @@ ARG DISTRO=alpine:3.10
 FROM $DISTRO
 # python3 shared with most images
 RUN apk add --no-cache \
-    python3 py3-pip git bash py3-multidict \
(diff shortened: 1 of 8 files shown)


kaiyou:
Build seems broken on my side. I have to investigate a bit.

micw:
> Build seems broken on my side. I have to investigate a bit.

For bors an me it works :-D

kaiyou:
I does work after resetting my Docker setup. Weird but at least it builds and the fault was on my side. Okay with merging :)

micw:
Seems that there's another changed dep meanwhile which requires gcc

hoellen:
bors try

hoellen:
lazy bot 🙄 

bors r+

PR #1267, opened 2019-11-21
Update roundcube to 1.4.0 and enable new elastic skin

hoellen:
## What type of PR?
enhancement

## What does this PR do?
This PR brings the new roundcube 1.4.0 release with the new elastic skin.
grafik
grafik

As an idea we could integrate the Mailu logo in a next step to customize the webmail client a bit.

## Prerequistes

### DIFF
--- a/webmails/roundcube/config.inc.php
+++ b/webmails/roundcube/config.inc.php
@@ -34,6 +34,10 @@ $config['smtp_pass'] = '%p';
 $config['managesieve_host'] = $imap;
 $config['managesieve_usetls'] = false;
 
+// Customization settings
+$config['support_url'] = getenv('WEB_ADMIN') ? '../..' . getenv('WEB_ADMIN') : '';
+$config['product_name'] = 'Mailu Webmail';
+
 // We access the IMAP and SMTP servers locally with internal names, SSL
 // will obviously fail but this sounds better than allowing insecure login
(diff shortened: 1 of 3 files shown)


HorayNarea:
1.4.1 is released already, could you update the PR?

webmails/roundcube/config.inc.php
```diff
@@ -34,6 +34,10 @@
 $config['managesieve_host'] = $imap;
 $config['managesieve_usetls'] = false;
 
+// Customization settings
+$config['support_url'] = getenv('WEB_ADMIN') ? '../..' . getenv('WEB_ADMIN') : '';
```
HorayNarea: Why do you set the URL where the webmail user is supposed to find support for their webmail experience to the admin interface? I'm not sure this would help anyone :/

hoellen: It's not the support for the "webmail experience", it's more that you find out about Mailu Admin Page (settings, information). And there is also a another link that redirects to the Mailu documentation for help. I found this very use- and helpful.

HorayNarea: Oh… sorry, then I missunderstood the comment in roundcubes default config.

hoellen:
#1266 is now merged, so retrigger merge:
bors r+

Issue #1226, opened 2019-10-14
Roundcube: support multiple DB backends


muhlemmer:
Roundcube currently has a SQLite file under: `/mailu/webmail/roundcube.db`. In order to allow for distributed deployment targets better, it should be possible to configure it to use an external database instead.

Config should be rendered in the setup as follows:

- User selects default `sqlite` DB backend --> use `roundcube.db`
- User selects stock PostgreSQL or MariaDB --> configure default connection parameters (pending #1216 and #1217)
- User selects external DB --> custom connection parameters

micw:
Could roundcube use the same mysql database that mailu admin uses or would they interfere with each other?

micw:
@muhlemmer What do you think about this solution?

https://github.com/micw/Mailu/commit/253e261a362f139240d6e67c50f4dcdfa7b71339

mamiapatrick:
Hello
i'm facing the same issues, cannot configure roundcube to use external database  i got a message for a database error connection failed with $config['db_dsnw'] = 'sqlite:////data/roundcube.db'; somes updates on that issue?

muhlemmer:
@micw I think the approach is good. And it could go into a PR. I think at first only document the capabilities in the reference docs and merge it like that.
And when we get around to #1216, #1217, integrate it into setup.

muhlemmer:
When I replied I wasn't aware that #1268 was already submitted and merged. Looked over it. This issue can be closed.

PR #1268, opened 2019-11-21
Roundcube db

### LINKED ISSUE #1157: Make roundcube log to the console
Roundcube is currently hard to debug because it logs into a special folder (/var/www/html/logs). It should log to stdout/stderr instead.

micw:
## What type of PR?

feature

## What does this PR do?

- makes roundcube work with mysql
- runs db init/upgrade scripts on startup
- redirects roundcube logs to stdout

### Related issue(s)
- preparations to solve #1226
- closes #1157 (side effect ;-) )

## Prerequistes

### DIFF
--- a/webmails/roundcube/Dockerfile
+++ b/webmails/roundcube/Dockerfile
@@ -21,7 +21,7 @@ ENV ROUNDCUBE_URL https://github.com/roundcube/roundcubemail/releases/download/1
 RUN apt-get update && apt-get install -y \
       zlib1g-dev libzip4 libzip-dev \
       python3-jinja2 \
- && docker-php-ext-install zip \
+ && docker-php-ext-install zip pdo_mysql \
  && echo date.timezone=UTC > /usr/local/etc/php/conf.d/timezone.ini \
  && rm -rf /var/www/html/ \
  && cd /var/www \
--- a/webmails/roundcube/Dockerfile
+++ b/webmails/roundcube/Dockerfile
@@ -19,9 +19,9 @@ RUN pip3 install socrate
 ENV ROUNDCUBE_URL https://github.com/roundcube/roundcubemail/releases/download/1.3.10/roundcubemail-1.3.10-complete.tar.gz
 
 RUN apt-get update && apt-get install -y \
-      zlib1g-dev libzip4 libzip-dev \
+      zlib1g-dev libzip4 libzip-dev postgresql-dev \
       python3-jinja2 \
- && docker-php-ext-install zip pdo_mysql \
+ && docker-php-ext-install zip pdo_mysql pdo_pgsql \
  && echo date.timezone=UTC > /usr/local/etc/php/conf.d/timezone.ini \
  && rm -rf /var/www/html/ \
  && cd /var/www \
--- a/webmails/roundcube/Dockerfile
+++ b/webmails/roundcube/Dockerfile
@@ -19,7 +19,7 @@ RUN pip3 install socrate
 ENV ROUNDCUBE_URL https://github.com/roundcube/roundcubemail/releases/download/1.3.10/roundcubemail-1.3.10-complete.tar.gz
 
 RUN apt-get update && apt-get install -y \
-      zlib1g-dev libzip4 libzip-dev postgresql-dev \
+      zlib1g-dev libzip4 libzip-dev libpq-dev \
       python3-jinja2 \
  && docker-php-ext-install zip pdo_mysql pdo_pgsql \
  && echo date.timezone=UTC > /usr/local/etc/php/conf.d/timezone.ini \
--- a/webmails/roundcube/start.py
+++ b/webmails/roundcube/start.py
@@ -9,6 +9,29 @@ log.basicConfig(stream=sys.stderr, level=os.environ.get("LOG_LEVEL", "WARNING"))
 
 os.environ["MAX_FILESIZE"] = str(int(int(os.environ.get("MESSAGE_SIZE_LIMIT"))*0.66/1048576))
 
+db_flavor=os.environ.get("ROUNDCUBE_DB_FLAVOR","sqlite")
+if db_flavor=="sqlite":
+    os.environ["DB_DSNW"]="sqlite:////data/roundcube.db"
+elif db_flavor=="mysql":
+    os.environ["DB_DSNW"]="mysql://%s:%s@%s/%s" % (
+        os.environ.get("ROUNDCUBE_DB_USER","roundcube"),
+        os.environ.get("ROUNDCUBE_DB_PW"),
+        os.environ.get("ROUNDCUBE_DB_HOST","database"),
+        os.environ.get("ROUNDCUBE_DB_NAME","roundcube")
+        )
+elif db_flavor=="postgresql":
+    os.environ["DB_DSNW"]="pgsql://%s:%s@%s/%s" % (
+        os.environ.get("ROUNDCUBE_DB_USER","roundcube"),
+        os.environ.get("ROUNDCUBE_DB_PW"),
+        os.environ.get("ROUNDCUBE_DB_HOST","database"),
+        os.environ.get("ROUNDCUBE_DB_NAME","roundcube")
+        )
+else:
+    print("Unknown ROUNDCUBE_DB_FLAVOR: %s",db_flavor)
+    exit(1)
+
+
+
 conf.jinja("/php.ini", os.environ, "/usr/local/etc/php/conf.d/roundcube.ini")
 
 # Fix some permissions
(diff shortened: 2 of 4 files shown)


hoellen:
Is it really necessary to have an extra Roundcube database setting? I think it would be easier (and less confusing) to use the database which is already defined in `DB_HOST` and maybe create there an extra database for Roundcube.

micw:
> Is it really necessary to have an extra Roundcube database setting? I think it would be easier (and less confusing) to use the database which is already defined in `DB_HOST` and maybe create there an extra database for Roundcube.

For the internal database, it's perfectly fine to have one database host with mutiple databases (even one database would do, using table prefixes). For external database I can see no reason to enforce the same host. But we could simply point default the ROUNDCUBE_DB_HOST to DB_HOST.

micw:
Rebased to master

micw:
@hoellen This PR is open for a while. I'd be very happy if you'd find some time to review it ;)
I run this in production since 2 months but that requires me to run a custom-build image which I'd like to cleanup.

webmails/roundcube/start.py
```diff
@@ -4,16 +4,61 @@
...
+elif db_flavor=="postgresql":
+    os.environ["DB_DSNW"]="pgsql://%s:%s@%s/%s" % (
+        os.environ.get("ROUNDCUBE_DB_USER","roundcube"),
+        os.environ.get("ROUNDCUBE_DB_PW"),
+        os.environ.get("ROUNDCUBE_DB_HOST",os.environ.get("DB_HOST","database")),
+        os.environ.get("ROUNDCUBE_DB_NAME","roundcube")
+        )
+else:
+    print("Unknown ROUNDCUBE_DB_FLAVOR: %s",db_flavor)
+    exit(1)
```
kaiyou: sys.exit would make sense here

webmails/roundcube/start.py
```diff
@@ -4,16 +4,61 @@
 import logging as log
 import sys
 from socrate import conf
+import subprocess
 
 log.basicConfig(stream=sys.stderr, level=os.environ.get("LOG_LEVEL", "WARNING"))
 
 os.environ["MAX_FILESIZE"] = str(int(int(os.environ.get("MESSAGE_SIZE_LIMIT"))*0.66/1048576))
 
+db_flavor=os.environ.get("ROUNDCUBE_DB_FLAVOR",os.environ.get("DB_FLAVOR","sqlite"))
```
kaiyou: General advice, please follow PEP008, which means spaces around assignment, comparison, etc.

kaiyou: Variable names should probably be unified with the plugin stuff, so we maintain some consistency

webmails/roundcube/start.py
```diff
@@ -4,16 +4,61 @@
...
+if db_flavor=="sqlite":
+    os.environ["DB_DSNW"]="sqlite:////data/roundcube.db"
+elif db_flavor=="mysql":
+    os.environ["DB_DSNW"]="mysql://%s:%s@%s/%s" % (
```
kaiyou: Please use string.format, which is a more standard way of formatting strings nowadays.

webmails/roundcube/start.py
```diff
@@ -4,16 +4,61 @@
...
+    result=subprocess.check_output(["/var/www/html/bin/initdb.sh","--dir","/var/www/html/SQL"],stderr=subprocess.STDOUT)
+    print(result.decode())
+except subprocess.CalledProcessError as e:
+    if "already exists" in e.stdout.decode():
+        print("Already initialzed")
+    else:
+        print(e.stdout.decode())
+        quit(1)
+
+try:
```
kaiyou: Maybe (unsure) we could run both commands inside the same try block since exception handling is identical.

webmails/roundcube/start.py
```diff
@@ -4,16 +4,61 @@
...
+    print("Initializing database")
+    result=subprocess.check_output(["/var/www/html/bin/initdb.sh","--dir","/var/www/html/SQL"],stderr=subprocess.STDOUT)
+    print(result.decode())
+except subprocess.CalledProcessError as e:
+    if "already exists" in e.stdout.decode():
+        print("Already initialzed")
+    else:
+        print(e.stdout.decode())
+        quit(1)
```
kaiyou: Same as above, sys.exit would fit

webmails/roundcube/start.py
```diff
@@ -4,16 +4,61 @@
...
+
+try:
+    print("Upgrading database")
+    subprocess.check_call(["/var/www/html/bin/update.sh","--version=?","-y"],stderr=subprocess.STDOUT)
+except subprocess.CalledProcessError as e:
+    quit(1)
```
kaiyou: Same as above, sys.exit would fit

webmails/roundcube/start.py
```diff
@@ -4,16 +4,61 @@
...
+
+# Tail roundcube logs
+subprocess.Popen(["tail","-f","-n","0","/var/www/html/logs/errors"])
```
kaiyou: I dislike this tbh, since we have done our best to limit the number of forked utility processes until now, even reducing it sometimes.

If these logs really cannot be sent through Apache php module, then okay, because we do need to output them.

kaiyou:
I wrote a couple comments, sorry that my client spammed with as many reviews.

Additional thought: I am more and more disturbed by the fact we construct a dsn string from a complex set of variables, while we could expose the dsn variable, lower the number and complexity of settings ("if flavor is this, then you need to set that, etc...")

What would you think about a single variable with a default value : `WEBMAIL_ROUNDCUBE_DSNW`?

This would also apply to the admin container ofc.

micw:
@kaiyou I'd also be fine with the DB_DSNW as an env variable, it would just move the construction from docker container into setup or the helm chart and would give much more flexibility.
Shall we do it together with this PR or seperate into a different PR?

webmails/roundcube/start.py
micw: The issue is that logging from php as apache module or fcgi is really ugly. I tried recently to achieve the same for a different tool and eventually went back to tailing the logfile.
Unfortunately I committed this in this PR, if you want I can also pull it out into a new one.

webmails/roundcube/start.py
micw: exit(), quit() and sys.exit does the same (https://stackoverflow.com/questions/19747371/python-exit-commands-why-so-many-and-when-should-each-be-used). I'll choose one for consistency, if you prefer sys.exit, i use this.

webmails/roundcube/start.py
micw: It's slightly different. I can do it in one but with the current implementation, it's more clear which call causes which behaviour

webmails/roundcube/start.py
micw: I'll change it accordingly (including the things you mentioned above)

webmails/roundcube/start.py
kaiyou: The issue you referenced makes a good case about sys.exit and I agree it is the only one we should use.

webmails/roundcube/start.py
kaiyou: Okay :)

kaiyou:
Depends if you think it is a complex change. If it only takes you so much time, I would go for it now, otherwise in a separate PR so we can merge this and see things clearer.

micw:
@kaiyou I unfortunately have little time these days, so I'd go with this PR and change to DB_DSNW later

micw:
@kaiyou Can you approve please?

webmails/roundcube/start.py
```diff
@@ -4,16 +4,61 @@
...
+if db_flavor=="sqlite":
+    os.environ["DB_DSNW"]="sqlite:////data/roundcube.db"
+elif db_flavor=="mysql":
+    os.environ["DB_DSNW"]="mysql://%s:%s@%s/%s" % (
+        os.environ.get("ROUNDCUBE_DB_USER","roundcube"),
+        os.environ.get("ROUNDCUBE_DB_PW"),
+        os.environ.get("ROUNDCUBE_DB_HOST",os.environ.get("DB_HOST","database")),
+        os.environ.get("ROUNDCUBE_DB_NAME","roundcube")
+        )
+elif db_flavor=="postgresql":
```
Nebukadneza: Do we need to install the php-ext for postgres too? If so, i think its missing from the dockerfile above …

micw: I think so, yes

micw:
bors try

webmails/roundcube/Dockerfile
```diff
@@ -19,9 +19,9 @@ RUN pip3 install socrate
 ENV ROUNDCUBE_URL https://github.com/roundcube/roundcubemail/releases/download/1.3.10/roundcubemail-1.3.10-complete.tar.gz
 
 RUN apt-get update && apt-get install -y \
-      zlib1g-dev libzip4 libzip-dev \
+      zlib1g-dev libzip4 libzip-dev postgresql-dev \
```
Nebukadneza: i think its `libpq-dev` — not 100% sure though

micw:
bory try

micw:
bors try

micw:
@Nebukadneza @kaiyou Should be fine now.

PR #1269, opened 2019-11-22
Add Fail2Ban section

### LINKED ISSUE #1263: Properly configure Fail2Ban
Hi,

Maybe this is not the correct place and should I instead ask this at Fail2Ban, but I would like to try and explain anyway.

After a fresh installation the last couple of days, I was getting the occasional authentication problem, where I had to put in my password for all accounts. Finally I was able to pinpoint it, at least for the webmail, to an

`"Authentication rate limit from one source exceeded" while in http auth state`

error. In the mailu issues I found several topics with the same problems and in one of them the tip to up the AUTH_RATELIMIT in the mailu.env. This seems to solve the problem, but opens me up to another issue: increased brute force attempts.

I already had investigated this as well, and found information on Fail2Ban in your FAQ and several Github issues. Eventually I installed and configured it on the host as described here: https://github.com/Mailu/Mailu/issues/584#issuecomment-433702594

This blacklists the IPs on the host, as seen with

`fail2ban-client status bad-auth`

but the attempts keep happening in the containers. This is on a single host, single (virtual) NIC and IP. What am I doing wrong?

iohenkies:
## What type of PR?
documentation

## What does this PR do?
This PR adds documentation to setup Fail2Ban to secure Mailu.

### Related issue(s)
- closes #1263

## Prerequistes

edit [@hoellen]: Give information about the PR and link issue.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -443,8 +443,68 @@ down brute force attacks.
 We *do* provide a possibility to export the logs from the ``front`` service to the host.
 For this you need to set ``LOG_DRIVER=journald`` or ``syslog``, depending on the log
 manager of the host. You will need to setup the proper Regex in the Fail2Ban configuration.
-Be aware that webmail authentication appears to come from the Docker network,
-so don't ban those addresses!
+Below an example how to do so. Be aware that webmail authentication appears to come from the
+Docker network, so don't ban those addresses!
+
+Assuming you have a working Fail2Ban installation on the host running your Docker containers,
+follow these steps:
+
+1. In the mailu docker-compose set the logging driver of the front container to journald
+
+.. code-block:: bash
+
+  logging:
+    driver: journald
+
+2. Add the /etc/fail2ban/jail.d/bad-auth.conf
+
+.. code-block:: bash
+
+  # Fail2Ban configuration file
+  [Definition]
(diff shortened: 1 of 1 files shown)


kaiyou:
Our mistake, this is based on 1.7. Could you rebase against master ?

hoellen:
I merge this PR, since two other already approved it and its only documentation.

bors r+

docs/faq.rst
```diff
@@ -443,8 +443,68 @@ down brute force attacks.
...
+follow these steps:
+
+1. In the mailu docker-compose set the logging driver of the front container to journald
+
+.. code-block:: bash
+
+  logging:
+    driver: journald
+
+2. Add the /etc/fail2ban/jail.d/bad-auth.conf
```
NeroPcStation: wrong file path |   should be /etc/fail2ban/filter.d/bad-auth.conf

yeawellwhatever, line 459: you probably meant `/etc/fail2ban/filter.d/bad-auth.conf` ?

NeroPcStation: > you probably meant `/etc/fail2ban/filter.d/bad-auth.conf` ?

Its fixed see https://github.com/Mailu/Mailu/pull/1333

PR #1278, opened 2019-12-06
Limiter implementation

### LINKED ISSUE #634: Ratelimit only failed authentications
currently the ratelimiting happens for all authentications, it would make much more sense to limit only if it's a brute-force == many failed logins

with the current implementation i get complains from some of my users who send (legitimate) newsletters and it feels wrong to set the limit higher because then brute-forcing is much easier

### LINKED ISSUE #1195: Better handling of authentication limit
As discussed in #301 and  #686 the auth limit causes trouble in some scenarios. A probably simple solution would be to only count failed authentications in this limit. To make it still work against brute-force attacks, the behavior should be like:

- If an auth fails, increment the counter
- If the counter is above a certain threshold, reject any auth (valid as well as invalid)
- Don't increment the counter for valid authentication requests

Edit:
- Webmail auth rate limit should be done at HTTP level in nginx so that the limit applies on a per-ip base.

micw:
## What type of PR?

## What does this PR do?

Adds a custom limter based on the "limits" lirary that counts up on failed auths only

### Related issue(s)
- closes #1195
- closes #634

## Prerequistes

### DIFF
--- a/core/admin/mailu/configuration.py
+++ b/core/admin/mailu/configuration.py
@@ -32,6 +32,7 @@ DEFAULT_CONFIG = {
     'POSTMASTER': 'postmaster',
     'TLS_FLAVOR': 'cert',
     'AUTH_RATELIMIT': '10/minute;1000/hour',
+    'AUTH_RATELIMIT_SUBNET': True,
     'DISABLE_STATISTICS': False,
     # Mail settings
     'DMARC_RUA': None,
--- a/core/admin/mailu/internal/__init__.py
+++ b/core/admin/mailu/internal/__init__.py
@@ -1,4 +1,4 @@
-from flask_limiter import RateLimitExceeded
+from mailu.limiter import RateLimitExceeded
 
 from mailu import utils
 from flask import current_app as app
@@ -18,15 +18,6 @@ def rate_limit_handler(e):
     if int(flask.request.headers['Auth-Login-Attempt']) < 10:
         response.headers['Auth-Wait'] = '3'
     return response
-
-
-@utils.limiter.request_filter
-def whitelist_webmail():
-    try:
-        return flask.request.headers["Client-Ip"] ==\
-            app.config["HOST_WEBMAIL"]
-    except:
-        return False
 
 
 from mailu.internal.views import *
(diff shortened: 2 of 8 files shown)


kaiyou:
From my understanding of the limits module, because we do not want to reject legitimate imap clients that just open many connections (webmails), we should not raise when hit returns False, only when the check fails.

micw:
Let's assume a limit of 3. With the current code users get:

invalid login 1 -> no limit exception
invalid login 2 -> no limit exception
invalid login 3 -> raises the exception on hit
valid login -> still a limit exeception

Or:
invalid login 1 -> no limit exception
invalid login 2 -> no limit exception
valid login -> login ok

If I don't raise on hit:

invalid login 1 -> no limit exception
invalid login 2 -> no limit exception
invalid login 3 -> no limit exception
valid login -> limit exeception

Or:
invalid login 1 -> no limit exception
invalid login 2 -> no limit exception
valid login -> login ok

Does not make much difference.

kaiyou:
After rethinking it myself, because you only hit on error, the difference is negligeable indeed.

kaiyou:
bors try

micw:
@kaiyou Please have a look to https://github.com/alisaifee/flask-limiter/issues/234 - the author of the lib is about to add this function to the annotations. Does this help or do we still require a programmatic way (e.g. for rate-limiting mails and other things)?

kaiyou:
I like your implementation for its simplicity and the fact that we can check the counter value in a fairly straightforward way.

 If they add the `penalty_condition` I think it would cover most cases. Using the library will be at the cost of a slightly more complex code though.

I would personally go for it and have them maintain the limiter logic then have us focus on delivering email :p

kaiyou:
Anyway depending on the timeframe, we might still want to merge this PR in the meantime.

micw:
From my side it can be merged. I put the roundcube limiter in a different PR.

micw:
@Nebukadneza @kaiyou Can we go with that?

kaiyou:
Yep, lets see if the test succeds then merge.

PR #1285, opened 2019-12-15
Unset HOST_ANTISPAM in k8s yamls

micw:
## What type of PR?

documentation

## What does this PR do?

Remove HOST_ANSTISPAM from k8s docs, add a warning that it will break the deployment.

## Prerequistes

### DIFF
--- a/docs/kubernetes/mailu/configmap.yaml
+++ b/docs/kubernetes/mailu/configmap.yaml
@@ -164,6 +164,7 @@
     HOST_WEBMAIL: "webmail.mailu-mailserver.svc.cluster.local"
     HOST_ADMIN: "admin.mailu-mailserver.svc.cluster.local"
     HOST_WEBDAV: "webdav.mailu-mailserver.svc.cluster.local:5232"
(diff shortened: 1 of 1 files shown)


kaiyou:
Mmmh. Wouldn't backporting the fix almost as straightforward?

Diman0 (approved):

micw:
This issue is related to https://github.com/Mailu/Mailu/pull/1716 which fixed this entirey.

IMO rather than merge this, we should deprecate the outdated yamls and add the helm chart to the docs (which also allows to create up-to-date yamls)

Diman0:
@micw  You state that PR #1716 fixes this completely. Does this not mean you can close this PR and we should get PR #1716 implemented?

>IMO rather than merge this, we should deprecate the outdated yamls and add the helm chart to the docs (which also allows to create up-to-date yamls)
kubernetes is not my thing (no knowledge). If this is a good idea, I think you should create an issue for this (if it doesnt exist already) where this can be discussed.

micw:
Yes, I still think this should be solved in general, so #1716 is the way to go.
For replacing k8s yamls with the helm chart, there is #1451

Issue #1292, opened 2019-12-24
Recipient address rejected: unverified address: Address lookup failure;


SunMar:
Hi,

As mentioned in #38, after upgrading to the latest version recently my catch-all address didn't work anymore, whereas previously it was working fine. The postfix logs showed:

```
Recipient address rejected: unverified address: Address lookup failure;
```

I've been able to resolve it by adding a `postfix.cf` override where I removed `reject_unverified_recipient` from `smtpd_client_restrictions`.

SunMar:
I see the directive was added in this commit: https://github.com/Mailu/Mailu/commit/175349a224c210bceefe918dbe16d66738de8104

mindrunner:
Does this apply to 'latest' only, or also to '1.7' tag?

kaiyou:
It is really interesting that the behavior is only recent. No recent version of postfix seem to change this specifically.

Address verification uses persistent storage however, to avoid spamming remote hosts, and we made some changes there, maybe persisting the queue directory has an impact even if I doubt it.

Anyway, address verification was naively enabled at the time because most docs said it was okay. I have read much documentation against today, especially that it does not help much with anything in general.

I would be quite inclined to disable it altogether, sender and recipient, and see if spam rate actually increases or not.

vogelfreiheit:
Confirming this. Best way to fix this when using a stable tag? How do I store a persistent main.cf override?

kaiyou:
Given the impact, it should be backported to 1.7 once we merge

Nebukadneza:
I don’t quite understand how this breaks catchalls … as they’re also working fine for me currently on `1.7`. On the other hand, I have no strong opinion for/against that check, since i also think most spam is not caught by this, but in later stages (at some computational cost though).
@kaiyou can you oversee this better? If so, from my side, I’d be fine with approve+merge …

mathew3:
I have encountered the same issue (1.7). Interestingly it affected addresses that worked just fine before. Resolved by adjusting `postfix.cf` according to @SunMar 
 
> I've been able to resolve it by adding a `postfix.cf` override where I removed `reject_unverified_recipient` from `smtpd_client_restrictions`.

bd8392:
same issue here

Aaron-Ritter:
Hi, this has as well impact in to relay domains by the looks of it, a server with a configured relay domain has no clue about the final receiving addresses.

Nebukadneza:
Hi There,

[...]

PR #1293, opened 2019-12-24
Remove `reject_unverified_recipient` from `smtpd_client_restrictions`

SunMar:
## What type of PR?

Bug-fix

## What does this PR do?

It removes recipient verification, as it broke my catch-all address.

Fix for #1292, though I'm not sure if this is the right way to fix the issue. It was added in 175349a224c210bceefe918dbe16d66738de8104.

### Related issue(s)
Fix for #1292 and a revert of 175349a224c210bceefe918dbe16d66738de8104.

### DIFF
--- a/core/postfix/conf/main.cf
+++ b/core/postfix/conf/main.cf
@@ -104,7 +104,6 @@ smtpd_client_restrictions =
   reject_non_fqdn_sender,
   reject_unknown_sender_domain,
   reject_unknown_recipient_domain,
-  reject_unverified_recipient,
   permit
 
 smtpd_relay_restrictions =


yohannprigent:
This would be awesome because relayed domains feature is broken without this.

kaiyou:
As stated in the issue, I am really inclined to removing it. How much feedback so you have on the impact?

I will apply on my own server as well.

kaiyou:
@SunMar I have tested on my own server and this looks okay. Do you have prod feedback as well?

Nebukadneza:
I don’t quite understand how this breaks catchalls … as they’re also working fine for me currently on 1.7. On the other hand, I have no strong opinion for/against that check, since i also think most spam is not caught by this, but in later stages (at some computational cost though).
@kaiyou can you oversee this better? If so, from my side, I’d be fine with approve+merge …

foosinn:
I ran into that error for mails that are forwarded from my old gmail account. I would love to see a 1.7 backport.

Thanks!

SunMar:
@Nebukadneza I am also a bit flustered on why this is breaking catchalls, especially since it's not a recent configuration change, but it is the only solution I could find to fix my catch all, which did break after upgrading

SunMar:
@kaiyou sorry for the late reply, so far I haven't experienced any issues anymore in my production environment, everything has been running fine. also I'm not sure how to disable sender verification, since I can't find any reference to `reject_unverified_sender` or [`smtpd_sender_restrictions`](http://www.postfix.org/postconf.5.html#smtpd_sender_restrictions) if that's what you mean.

PR #1296, opened 2019-12-26
fetchmail: print unhandled exceptions, but don't crash

### LINKED ISSUE #1295: fetchmail crashes due to exception
```
  File "/fetchmail.py", line 93, in <module>
    run(os.environ.get("DEBUG", None) == "True")
  File "/fetchmail.py", line 47, in run
    fetches = requests.get("http://admin/internal/fetch").json()
  File "/usr/lib/python3.7/site-packages/requests/models.py", line 897, in json
    return complexjson.loads(self.text, **kwargs)
  File "/usr/lib/python3.7/json/__init__.py", line 348, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3.7/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.7/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

Al2Klimov:
fixes #1295

### DIFF
--- a/optional/fetchmail/fetchmail.py
+++ b/optional/fetchmail/fetchmail.py
@@ -8,6 +8,7 @@ import subprocess
 import re
 import requests
 import sys
+import traceback
 
 
 FETCHMAIL = """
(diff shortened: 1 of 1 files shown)


Nebukadneza:
Hi, and thanks for the fix!

I must say that i find the solution a little bit roundabout, although it works. Most developers probably go and look for a custom excepthook assignment somewhere upon seeing it called manually, which might lead to confusion. IMO printing the exception itself or using `traceback.print_exc()` would be clearer. Finally, but that is nitpicking, couldn’t we try/except the request in `run` itself, and bail out on error?
Otoh, I agree that pokemon exception-handling (»gotta catch them all …«) in the `while True` DOES make sense.

Would you be willing to modify this once more? ^_^

Al2Klimov:
Hello @Nebukadneza!

Have I fixed all of your not-just-nice2haves?

Best,
AK

PR #1299, opened 2019-12-27
Don't remove the address extension in postfix

### LINKED ISSUE #982: sieve filtering based on recipient_delimiter no longer works in 1.6
I have a sieve rule setup to automatically sort mails into folders based on the part after the recipient delimiter, so mail to `foo+bar@example.com` would get sorted into folder `bar` provided it exists.
The rule for this is as follows:
```
require ["envelope","fileinto","subaddress","variables"];
if envelope :detail :matches "to" "*"
{
	fileinto "${1}";
}
```

This worked fine in Mailu 1.5, but ever since upgrading to 1.6 all those mails get filed just into `INBOX`. Sieve evaluation itself is working, as other rules get applied properly.

RobertMe:
## What type of PR?
Bugfix

## What does this PR do?
Currently when the mail address is looked up by Postfix (using the admin
part) the address extension is removed. This is due to the address
extension being removed to look up the user, and afterwards returning
the users mail address. But by not returning the mail address including
the address extension it also isn't part anymore in the LMTP
communication to Dovecot. So Dovecot doesn't know about the extension,
and in turn the address extension can't be used in Sieve mail filtering.

This change fixes that by returning the original address by just
concatinating the "localpart" and domain again when the user is found.

### Related issue(s)
Fixes #982

## Prerequistes

### DIFF
--- a/core/admin/mailu/models.py
+++ b/core/admin/mailu/models.py
@@ -277,12 +277,14 @@ class Email(object):
         if not user and localpart_stripped:
             user = User.query.get('{}@{}'.format(localpart_stripped, domain_name))
         if user:
+            email = '{}@{}'.format(localpart, domain_name)
+
             if user.forward_enabled:
                 destination = user.forward_destination
                 if user.forward_keep or ignore_forward_keep:
-                    destination.append(user.email)
+                    destination.append(email)
             else:
-                destination = [user.email]
+                destination = [email]
             return destination
 
         pure_alias = Alias.resolve(localpart, domain_name)


kaiyou:
That's a go. I will probably rewrite the entire function sometimes since it has become cluttered but that is no fault of yours.

RobertMe:
The function indeed might need some love.

When I first started fixing this I also had a go at saving the address extension and adding it back later. That might be a nice feature for aliases where you can have an option to keep the address extension after applying the alias. So say that alias@example.com delivers the mail to hello@example.com. That when you have this option enabled a mail send to alias+foo@example.com is delivered to hello+foo@example.com and the address extension is thus carried over.

But then you can of course also get more complicated requests (or issues). If the alias delivers to an address which already has an address extension that gives a conflict (should it use the extension as defined, or based on the incoming mail, or even add an extra option to merge them both).

Issue #1301, opened 2019-12-29
Increasing DKIM key size to 2048


abelhooge:
As it stands right now the automatically generated DKIM key is of a length of 1024. 

According to multiple articles, including one from [techrepublic](https://www.techrepublic.com/article/how-to-increase-your-dkim-key-to-2048-bits-for-google-apps/), a key size of less than 2048 is considered by some providers as inadequate. 

Changing the default generation key size should not cause any problems with backward compatibility, since it would only affect newly generated keys and not already existing ones. 

I'm interested to see if there is any support for this and whether there are any interesting perspectives on this matter.

kaiyou:
Agreed. We would need to check that the provided DNS entries are still valid though.

I remember that larger key size means multiple DNS records sometimes.

kaiyou:
Cc @t0mra when you find the time to make a PR :)

bidetaggle:
Hi, sorry if my question is stupid. When will it be merged in the branch 1.7? Or can I make the same PR for this branch? 😊

kaiyou:
It will probably never be merged in 1.7, unless we find it is blocking in many cases. We like to keep our stable branches stable. But 1. there is a PR in progress to start releasing often and smaller 2. we should be releasing 1.8 soonish? it would sound good anyway.

PR #1319, opened 2020-01-16
Upgrading to a 2048-bit DKIM key

radtkedev:
## What type of PR?

Enhancement

## What does this PR do?

Upgrades to a 2048-bit DKIM key from the currently used and outdated 1024-bit DKIM key.

### Related issue(s)
- #1301

### DIFF
--- a/core/admin/mailu/dkim.py
+++ b/core/admin/mailu/dkim.py
@@ -5,7 +5,7 @@ They are thus represented as ASCII armored PEM.
 from OpenSSL import crypto
 
 
-def gen_key(key_type=crypto.TYPE_RSA, bits=1024):
+def gen_key(key_type=crypto.TYPE_RSA, bits=2048):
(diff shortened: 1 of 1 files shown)


kaiyou:
As stated in the issue, we should check that this does not affect DNS records. Are they still short enough to fit in a single record?

kaiyou:
Here is an example article mentioning the issue: http://hack.limbicmedia.ca/how-to-split-dns-dkim-records-properly/

HorayNarea:
> Are they still short enough to fit in a single record?

I'm using 2048-bit DKIM keys on a non-Mailu server in a single record just fine, so yeah!

4096-bit keys would be the harder case because then it doesn't fit in a single record _and_ is bigger then DNS over UDP without EDNS0 allows :sweat_smile:

kaiyou:
Okay, then we are a go.

PR #1297, opened 2019-12-26
Added SOLR support

ajgon:
## What type of PR?

Feature

## What does this PR do?
Adds support for full text search via SOLR to dovecot (imap).

## Prerequistes
Before we can consider review and merge, please make sure the following list is done and checked.
If an entry in not applicable, you can check it or remove it from the list.

- [x] In case of feature or enhancement: documentation updated accordingly
- [ ] Unless it's docs or a minor change: add [changelog](https://mailu.io/master/contributors/guide.html#changelog) entry file.

PS. changelog guide is 404 😏

PR #1320, opened 2020-01-19
Add xapian full-text-search plugin to dovecot

Nebukadneza:
## What type of PR?
Enhancement

## What does this PR do?
Currently we are not able to offer our users a FTS experience after the
demise of lucene due to unfixed coredumps with musl/alpine.
We now add lucene, the only remaining maintained small/lean FTS plugin
for dovecot. It is quite simple to add to our stack: A two-stage docker
build is used to compile the fts plugin in the first stage, and copy
over only the resulting plugin-artifact to the second stage, which is
our usual dovecot container. Configuration is also minimal.

There was a upstream issue where bodies were not able to be searched for subwords, but fortunately it was fixed quite quickly. We currently need to wait for a new release to use a stable tag in our `Dockerfile`.

### Related issue(s)
- https://github.com/Mailu/Mailu/pull/1176
- https://github.com/Mailu/Mailu/pull/1297
- https://github.com/Mailu/Mailu/issues/751
- **Upstream-issues which is the cause for the `TODO` in the `Dockerfile`**: https://github.com/grosjo/fts-xapian/issues/32

## Prerequistes
- [ ] Wait for upstream to prepare new release after https://github.com/grosjo/fts-xapian/issues/32 — so that we can use a stable tag in our `Dockerfile`

### DIFF
--- a/core/dovecot/Dockerfile
+++ b/core/dovecot/Dockerfile
@@ -1,5 +1,5 @@
-ARG DISTRO=alpine:3.10 as builder
-FROM $DISTRO
+ARG DISTRO=alpine:3.10
+FROM $DISTRO as builder
 WORKDIR /tmp
 RUN apk add git build-base automake autoconf libtool dovecot-dev xapian-core-dev icu-dev
 RUN git clone https://github.com/grosjo/fts-xapian.git \
--- a/core/dovecot/Dockerfile
+++ b/core/dovecot/Dockerfile
@@ -4,7 +4,7 @@ WORKDIR /tmp
 RUN apk add git build-base automake autoconf libtool dovecot-dev xapian-core-dev icu-dev
 RUN git clone https://github.com/grosjo/fts-xapian.git \
   && cd fts-xapian \
-  # && git checkout 1.2.7 \ # TODO: when upstream releases 1.2.7, use that
+  && git checkout 1.2.7 \
   && autoreconf -vi \
   && PANDOC=false ./configure --with-dovecot=/usr/lib/dovecot \
   && make \
--- a/core/dovecot/Dockerfile
+++ b/core/dovecot/Dockerfile
@@ -1,4 +1,4 @@
-ARG DISTRO=alpine:3.10
+ARG DISTRO=alpine:3.10 as builder
 FROM $DISTRO
 WORKDIR /tmp
 RUN apk add git build-base automake autoconf libtool dovecot-dev xapian-core-dev icu-dev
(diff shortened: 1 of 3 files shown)


Nebukadneza:
To continue the discussion in the previous PRs: I think that mailu is best served with a small and lean solution. Solr and Elastic might both be stable, perform well, but they’re also fat, harder to set up, and might need dedicated maintenance from users in future upgrades. Kind of like the default sqlite vs. external DBs like postgres … where the user needs manual steps at a major-version upgrade of postgres.

So, similar to sqlite, I’d like a FTS solution that’s hands-off, integrated with dovecot (i.e., not a separate daemon/container/…), and will get the basic job done. Whether it’d perform on a scale of thousands of users is another question, and in my opinion not one that mailu should answer.

So … it’d be great if you could share your thoughts on this as well, whether both my view and contribution would be acceptable ^_^

core/dovecot/Dockerfile
```diff
@@ -13,9 +24,11 @@ RUN pip3 install "podop>0.2.5"
 
 # Image specific layers under this line
 RUN apk add --no-cache \
-  dovecot dovecot-lmtpd dovecot-pop3d dovecot-submissiond dovecot-pigeonhole-plugin rspamd-client \
+  dovecot dovecot-lmtpd dovecot-pop3d dovecot-submissiond dovecot-pigeonhole-plugin rspamd-client xapian-core \
   && mkdir /var/lib/dovecot
 
+COPY --from=0 /usr/lib/dovecot/lib21_fts_xapian_plugin.* /usr/lib/dovecot/
```
HorayNarea: I'm not sure which way is best practice but at least I think it's clearer to name the build-stage container and do `--from=buildcontainername`, If you agree could you change that? If you disagree it's fine too :D

Nebukadneza: Thanks for the hint and sorry for the late reply. That is indeed much clearer, and I wasn’t aware yet.

Nebukadneza:
bors try

Nebukadneza:
bors try

parisni:
> To continue the discussion in the previous PRs: I think that mailu is best served with a small and lean solution. Solr and Elastic might both be stable, perform well, but they’re also fat, harder to set up, and might need dedicated maintenance from users in future upgrades. Kind of like the default sqlite vs. external DBs like postgres … where the user needs manual steps at a major-version upgrade of postgres.
> 
> So, similar to sqlite, I’d like a FTS solution that’s hands-off, integrated with dovecot (i.e., not a separate daemon/container/…), and will get the basic job done. Whether it’d perform on a scale of thousands of users is another question, and in my opinion not one that mailu should answer.
> 

IMO it would be great to allow to choose among xapian and solr. On the exact same design than sqllite / postgres. In this case the solr external instance has to be setup and maintained by the user. Still I wonder if this would allow to deal with thousand of users ; there might exiat other part of mailu that are limiting such amount of users ?

Issue #545, opened 2018-08-01
Roundcube plugin management


SunMar:
Hi,

It would be very nice if it's possible to add our own plugins to Roundcube from https://plugins.roundcube.net/. There are some great plugins on there :).

micw:
I'd like to pick up this issue because there are some pending feature requests for roundcube plug-ins:

- carddav (https://github.com/Mailu/Mailu/issues/1230, https://github.com/Mailu/Mailu/issues/1116)
- pgp (https://github.com/Mailu/Mailu/issues/506)

Personally I'd like to see https://plugins.roundcube.net/packages/melanie2/infinitescroll in the plug-in list because it improves UX a lot (any my users are used to have it and now complain about the old-school-style pagination) ;)

I think having an env var and installing the plug-ins in the running docker container is an anti-pattern (containers should be immutable, preparation errors should only happen on build time). So my idea is to install all popular plug-ins and allow to enable them via env vars.

Ninjiner:
@micw Great to hear this. I would also like to see the PGP feature.
What's the status with this issue?

micw:
No prgress yet, just collecting some feedback ;)

smoebody:
actually roundcube returns an error message when trying to open a mail with a pgp-signature.

Nebukadneza:
Hi There,

The `Mailu`-Project is currently in a bit of a bind! We are short on man-power, and we need to judge if it is possible for us to put in some work on this issue.

To help with that, we are currently trying to find out which issues are actively keeping users from using `Mailu`, which issues have someone who want to work on them — and which issues may be less important. These a less important ones could be discarded for the time being, until the project is in a more stable and regular state once again.

In order for us to better assess this, it would be helpful if you could put a **reaction on this post** (use the :smiley: icon to the top-right).
- 👍️ if you **need this** to be able to use Mailu. Ideally, you’d also be able to test this on your installation, and provide feedback …
- 🎉 if you find it a **nice bonus**, but no deal-breaker
- 🚀 if you want to **work on it yourself**!
We want to keep this **voting open for 2 weeks** from now, so please help out!

harveyconnor:
Any progress on getting functionality to install roundcube plugins?
My idea was to mount the volume however that does not work because it is done at build time.

mbergann:
I got it working:

at first, I switched into my webmail-container and copied the /var/www/html/plugins directory to /data and also the config-file /var/www/html/config/config.inc.php

[...]

PR #1298, opened 2019-12-26
Added carddav-plugin for roundcube webmail

sholl:
## Feature

This PR enables the carddav contacts plugin for integration remote contact-repositories based on CardDAV.

## What does this PR do?

This PR enables the carddav contacts plugin for integration remote contact-repositories based on CardDAV.

### Related issue(s)
- Related #1230, at least for CardDAV.

## Prerequistes
Before we can consider review and merge, please make sure the following list is done and checked.
If an entry in not applicable, you can check it or remove it from the list.

- [x] In case of feature or enhancement: documentation updated accordingly (not needed since the plugins of webmail is not mentioned in the docs.)
- [x] Changelog-entry added

PR #1326, opened 2020-01-26
Allow user to install, remove and activate roundcube plugins

jawabuu:
## What type of PR?
Feature

## What does this PR do?
This PR allows users to install RoundCube Plugins and activate them using environment variables.
It is designed to be entirely optional and to not interfere with the core roundcube installation.
At this time, it only installs plugins that are available on https://plugins.roundcube.net/ using composer method.
The user is required to provide the following environment variables;
**INSTALL_RC_PLUGINS**
**REMOVE_RC_PLUGINS**
**ACTIVATE_RC_PLUGINS**
All the variables above are space delimited strings, e.g.
```
INSTALL_RC_PLUGINS="kolab/calendar roundcube/carddav"
REMOVE_RC_PLUGINS="johndoh/contextmenu johndoh/swipe"
ACTIVATE_RC_PLUGINS="calendar carddav"
```
You can provide any, all or none of the environment variables.
Providing none of the variables starts up core Mailu as usual.
You can always reset you instance to run core Mailu by removing these environment variables.
Any changes to any of the variables will require the container be restarted.

**Points to consider**

- If acceptable, Mailu core could include git and composer in the roundcube image
- A volume mount pointing to /var/www/html/plugins should be created to persist plugins installed. This can be done from the docker-compose.yml file.
- Complex requirements for plugins e.g. Plugins requiring authentication, Complex versioning e.t.c will be left to the users implementation by creating a custom entrypoint script. We should have a WIKI with user contributed scripts to address these cases.
- The entrypoint script included in Mailu should take utmost care to always guarantee that Mailu will start despite any errors that may occur in plugin installation.
- Finally, the plugin functionality should stay true to Mailu guidelines in the sense that mail is not dependent on any of these plugins but the user nonetheless finds them useful. Therefore the burden for ensuring the plugins work as desired will belong to the user. Mailu should only offer basic infrastructure on how to implement this.

**Update 1**
I have confirmed that mailu/roundcube image already has git installed. In addition to composer I would like to make a suggestion for a minimum set of packages and php extensions that should be baked into the image. This list is derived from packages that roundcube itself suggests.
**Packages**
```
libfreetype6-dev
libicu-dev
libjpeg62-turbo-dev
libldap2-dev
libpng-dev
libpq-dev
libsqlite3-dev
libzip-dev
libmcrypt-dev
python3-jinja2
zlib1g-dev
unzip
libzip4
```
**PHP Extensions**
```
exif
gd 
intl
ldap
pdo_mysql
pdo_pgsql
pdo_sqlite
zip
```
### Related issue(s)
- Related to https://github.com/Mailu/Mailu/pull/1298/
- Closes Issue https://github.com/Mailu/Mailu/issues/545

## Prerequistes
@kaiyou , @sholl

### DIFF
--- a/tests/compose/roundcube/Dockerfile
+++ b/tests/compose/roundcube/Dockerfile
@@ -1,0 +1,39 @@
+FROM mailu/roundcube:master
+
+###Begin Package additions###
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    libfreetype6-dev \
+    libicu-dev \
+    libjpeg62-turbo-dev \
+    libldap2-dev \
+    libpng-dev \
+    libpq-dev \
+    libsqlite3-dev \
+    libzip-dev \
+    libmcrypt-dev \
+    python3-jinja2 \
+    gpg \
+    zlib1g-dev \
+    unzip \
+    libzip4 \
+    ; \
+    debMultiarch="$(dpkg-architecture --query DEB_BUILD_MULTIARCH)"; \
+    docker-php-ext-configure gd --with-freetype-dir=/usr/include/ --with-png-dir=/usr --with-jpeg-dir=/usr; \
+    docker-php-ext-configure ldap --with-libdir="lib/$debMultiarch"; \
+    docker-php-ext-install \
+    exif \
+    gd \
+    intl \
+    ldap \
+    pdo_mysql \
+    pdo_pgsql \
+    pdo_sqlite \
+    zip \
+    ; \
+    rm -rf /var/lib/apt/lists/*
+
+RUN curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/bin --filename=composer
+###End Package additions###
+
+CMD /start.py
+HEALTHCHECK CMD curl -f -L http://localhost/ || exit 1
--- a/tests/compose/roundcube/docker-compose.yml
+++ b/tests/compose/roundcube/docker-compose.yml
@@ -79,11 +79,23 @@ services:
 
   # Webmail
   webmail:
-    image: ${DOCKER_ORG:-mailu}/${DOCKER_PREFIX:-}roundcube:${MAILU_VERSION:-master}
-    restart: always
+    image: ${DOCKER_ORG:-mailu}/${DOCKER_PREFIX:-}roundcube:rc
+    build: .
+    # restart: always
     env_file: mailu.env
+    environment:
+      - INSTALL_RC_PLUGINS=sblaisot/automatic_addressbook roundcube/carddav johndoh/contextmenu
+      # Plugins that generate errors: endroid/qr-code
+      # Plugins that require mysql: texxasrulez/calendar
+      # Optional env_variable to remove undesired plugins
+      # - REMOVE_RC_PLUGINS=texxasrulez/calendar
+      - IMAP_ADDRESS=imap
+      - SMTP_ADDRESS=smtp
+    ports:
+      - "811:80"
     volumes:
       - "/mailu/webmail:/data"
+      - "./plugins.py:/var/www/html/plugins.py"
     depends_on:
       - imap
 
--- a/tests/compose/roundcube/mailu.env
+++ b/tests/compose/roundcube/mailu.env
@@ -16,7 +16,7 @@
 
 # Mailu version to run (1.0, 1.1, etc. or master)
 #VERSION=master
-
+VERSION=1.7
 # Set to a randomly generated 16 bytes string
 SECRET_KEY=PGGO2JRQ59QV3DW7
 
@@ -140,3 +140,8 @@ REAL_IP_FROM=
 
 # choose wether mailu bounces (no) or rejects (yes) mail when recipient is unknown (value: yes, no)
 REJECT_UNLISTED_RECIPIENT=
+
+INITIAL_ADMIN_ACCOUNT=admin
+INITIAL_ADMIN_DOMAIN=mailu.io
+INITIAL_ADMIN_PW=pass
+INITIAL_ADMIN_MODE=update
--- a/tests/compose/roundcube/plugins.py
+++ b/tests/compose/roundcube/plugins.py
@@ -1,0 +1,96 @@
+#!/usr/bin/python3
+
+import os
+import re
+import sys
+import subprocess
+import shutil
+from socrate import conf
+import logging as log
+
+
+# Assign plugin env variables to local variables
+REMOVE_RC_PLUGINS   = os.environ.get("REMOVE_RC_PLUGINS")
+INSTALL_RC_PLUGINS  = os.environ.get("INSTALL_RC_PLUGINS")
+ACTIVATE_RC_PLUGINS = os.environ.get("ACTIVATE_RC_PLUGINS", \
+  re.sub('[^\s]+?/', '', INSTALL_RC_PLUGINS) \
+  if INSTALL_RC_PLUGINS else None)
+plugin_error        = False
+# Check Composer
+if shutil.which('composer') is not None:
+  print("Composer Installed")
+else:
+  print("Composer Not Found. Installing..")
+  os.system("curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/bin --filename=composer")
+
+# Create composer.json if it does not exist
+if os.path.isfile('composer.json'):
+    print ("composer.json exists")
+else:
+    print ("Creating composer.json")
+    os.system("cp composer.json-dist composer.json")
+
+# Remove plugins from composer.json if REMOVE_RC_PLUGINS environment variable is set
+if REMOVE_RC_PLUGINS:
+  print(f'Removing plugins {REMOVE_RC_PLUGINS}')
+  subprocess.check_call(['php', '-d', 'disable_functions=', '/usr/bin/composer', '--no-update', '--no-interaction', 'remove'] \
+  + f'{REMOVE_RC_PLUGINS}'.split(),stderr=subprocess.STDOUT)
+
+
+# Add plugins to composer.json if INSTALL_RC_PLUGINS environment variable is set
+if INSTALL_RC_PLUGINS:
+  print(f'Adding plugins {INSTALL_RC_PLUGINS}')
+  subprocess.check_call(['php', '-d', 'disable_functions=', '/usr/bin/composer', '--prefer-dist', '--prefer-stable', '--no-update', \
+  '--update-no-dev', '--no-interaction', '--optimize-autoloader', '--apcu-autoloader', 'require'] \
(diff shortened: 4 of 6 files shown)


micw:
Hello,
thank you very much for this PR. Personally I think, adding stuff to a docker container should be a build-time action and not a run-time action. So I'd really prefer a solution where most common plug-ins are already installed and just needs to be enabled and have the option to install a plug-in at run-time only as the last option if a rarely used plug-in is missing.

jawabuu:
> Hello,
> thank you very much for this PR. Personally I think, adding stuff to a docker container should be a build-time action and not a run-time action. So I'd really prefer a solution where most common plug-ins are already installed and just needs to be enabled and have the option to install a plug-in at run-time only as the last option if a rarely used plug-in is missing.

Hey @micw 
You are correct in terms of adding files to a container.
Here's how I think we can address this;
1. Mailu includes plugins it deems necessary for properly functioning mail.
2. Mailu can add to this list upon request. Think of the cardav plugin https://github.com/Mailu/Mailu/pull/1298/
So this PR does not interfere with any of this workflow.
However;
It is impossible or very difficult to come to a consensus of what plugins to include or leave out.
Additionally, there are too many plugins available to include all of them in the image.

My compromise is that for users that need additional plugins not provided by core mailu, this can be used.
All other users are unaffected.
Secondly, I have proposed a volume mount of the plugins directory which will avoid modifying the container at runtime.

jawabuu:
bors try

kaiyou:
I do not disagree with the idea, but I strongly dislike that you did not include your work in the current start script, but override the entrypoint instead.

Maybe this could be reworked, and we need to discuss the complexity of adding 3 environment variables. One of the motos of Mailu is to keep it simple. Why cant we just have a single variable, listing plugins that we want installed and enabled?

jawabuu:
Thank you for the feedback @kaiyou
The decision to use the entrypoint was informed by my desire to leave core mailu entirely untouched at least until I received feedback.
We can definitely rework with your suggestions if you are comfortable with changes to the existing code.
In terms of using the 3 environment variables, this was supposed to give flexibility to the user drawing from how plugin management in roundcube works.
In your opinion;
1. Should a user be able to uninstall plugins?
2. Should a user be able to disable installed plugins?
I would be interested to hear what you think of this.
It is however quite simple on my end to install and activate a list of plugins in one step if we use one variable.

kaiyou:
Here is my current intuition: let's start simple but standard.

If a variable named `WEBMAIL_ROUNDCUBE_PLUGINS` could list wanted plugins, enable them if they are already installed and accumulated installing them when required, this would probably cover most use cases.

jawabuu:
Hey @kaiyou
I'll make the changes for your review.
I'm still inclined not to touch any of the existing code yet and use the entrypoint script to ensure compatibility and easy testing of edge cases.
Is this OK?

kaiyou:
At first and for testing purposes it would be okay, but I insist we stick to the model before we merge.

After all, adding an entrypoint or changing the start script will both alter the start up behavior of the container.

jawabuu:
Alright.
Just to confirm what is the current entrypoint?

micw:
@jawabuu I had some more thoughts about this.

1st IMO all prerequisites should be in the docker image (e.g. composer, compser.json, whatever is technically needed to install plugins). That reduces the amount of changes applied to the image at runtime.

Then, I think plugins should be installed the same way at docker build time as at runtime, e.g. by calling your script during build time too. There's one little difference: the plugins that are enabled might differ from the plugins that are installed here.

I could imagine the following:

* Dockerfile runs "WEBMAIL_ROUNDCUBE_PLUGINS=list-of-plugins /your-plugin-installer-script.sh" - your script would run and install all the plugins (if missing)
* after that in Dockerfile we set the same variable to the list of plug-ins we want to enable by default
* When the container is started, your script runs and would do nothing since all plugins are already installed
* Roundcube is started with all plugins enabled with WEBMAIL_ROUNDCUBE_PLUGINS.

So if the container is started with a modified WEBMAIL_ROUNDCUBE_PLUGINS, missing plugins would be installed.

With that we'd have a clearly defined way on how to add/enable plugins during build and during runtime.

(hope that was not too confusing ;-) )

jawabuu:
@micw 
I think I follow :-)
I totally agree with having the prerequisites in the image.
There are already a few plugins enabled in the image at this point namely;
`archive zipdownload markasjunk managesieve enigma`
This can always be updated to whatever people feel is important.
On your second point, the script does not touch any of the plugins that come with the image.
It also skips any that are already installed.

In terms of simplicity, I think that users ought to interact with the built mailu/roundcube image and not have to build it again unless in the rare cases where they want to extend it to create a custom image.
We can update the docs to reflect this.
For majority of the cases however we should go with `WEBMAIL_ROUNDCUBE_PLUGINS` environment variable.
Depending on what plugins you finally agree to add, the change should have no effect on any users.

kaiyou:
On the enreypoint matter, the current entrypoint is the image default, i.e. a shell, that then runs the start script.

If we change the entrypoint, we take the risk that a simple docker run with a specific command will install plugins while the operator only meant to inspect some files or whatever he intended to do, not anticipating that the state will change before he accesses the container.

Then, I disagree about the graceful failure: there is always a way to fail gracefully, which involves proper exception handling, and it is the proper way to go imo.

Finally, we once decided that start scripts were all going to be Python scripts for the sake of simplicity: Python is at the core of Mailu and we do not want to depend on too many script languages for maintaining the beast 😅

jawabuu:
@kaiyou 
Noted your point on having too many scripts. 😄 
Allow me however to make one final case for the custom entrypoint as I believe this is one of the special cases where its benefits outweigh the drawbacks.
This entrypoint will only be activated a if a single  environment variable `WEBMAIL_ROUNDCUBE_PLUGINS` is explicitly set.
https://github.com/jawabuu/Mailu/blob/e93c79f06c6ae7b1e7f7676d939f350eec976787/webmails/roundcube/rc-entrypoint.sh#L7
If the variable is missing, the script will proceed straight away to the command provided in the Dockerfile, docker-run e.t.c
https://github.com/jawabuu/Mailu/blob/e93c79f06c6ae7b1e7f7676d939f350eec976787/webmails/roundcube/rc-entrypoint.sh#L87

So the risk of unintended consequences is removed.

Secondly, due to the process of how plugins are installed. We will still have to invoke the shell in start.py.
i.e.
`composer --`
`curl -o --`
`cp --`
`mv --`
e.t.c.
This is already done here
https://github.com/Mailu/Mailu/blob/0bc10b7bc5b473b8b5f5c3286ba0b73f51fbc3ab/webmails/roundcube/start.py#L14-L19
We will have to use
`os.execv()`
As a very simple example, We can remove the entrypoint and in start.py call
```
import subprocess
subprocess.call("/entrypoint.sh")
```
I'm curious to find out if this adds or removes complexity in your opinion.
Once we settle on the approach, I will make the necessary changes.

jawabuu:
Hey @kaiyou
Could you provide direction so I can make the appropriate changes.

kaiyou:
Sorry about the delay, I was away for a couple days. I really do not see how placing your code in an entrypoint will remove the risk for unintended consequences: any code anywhere, if written to just continue upon error, will just continue upon error. It is not the fact that you use an entrypoint instead of editing the Python that provides this, but the fact that you handle errors properly, which you must do wherever you put the code.

As to calling an external script, it does not really help with the fact that we will have to maintain code using two different technos. Take the front containe for instance, we have external scripts called from the main `start.py`, but those are using Python as well.

It is just a choice we made to use Python for system scripting instead of bash wherever we could.

jawabuu:
Thanks @kaiyou.
For clarity's sake
1. I should use a single env variable: **ROUNDCUBE_PLUGINS**
2. I should script it using Python in start.py

Is this correct?

myteksp:
Guys, any updates on merging this?

Nebukadneza:
Sorry for the period of inactivity!

I agree with @kaiyou, the existing start-script is the place for such functionality, and I believe that scripting such things is more readable and maintainable in python, which is why Mailu uses it for these kind of tasks.

I don’t object the functionality itself, same as @kaiyou .

Sorry for being bold and a tad rude, but: Would you still be interested in continuing this PR?

jawabuu:
Hey @Nebukadneza It's no problem at all :-)
I just wanted to get clarification on the preferred method for implementation before I continued with it.
1. Uuse a single env variable: ROUNDCUBE_PLUGINS
2. Script it using Python in start.py
I would appreciate a final decision before I proceed.

Nebukadneza:
Ah, sorry for not responding that part.

I think that using a single `ROUNDCUBE_PLUGINS` and doing the work in `start.py` is the right approach. Also, as @micw noted, dependencies that are needed in all cases (such as composer) should be installed in the `Dockerfile`, and not on the fly.

It would be a real bonus if we had a mechanism to install plugins at build-time too, like mentioned above. However, as we intend to work iteratively, start-time installation is a viable first step. While it goes against a few best-practices to change the container too much at runtime, this gives our users some rather easy flexibility without having to care for selfbuilt-images.

jawabuu:
> Ah, sorry for not responding that part.
> 
> I think that using a single `ROUNDCUBE_PLUGINS` and doing the work in `start.py` is the right approach. Also, as @micw noted, dependencies that are needed in all cases (such as composer) should be installed in the `Dockerfile`, and not on the fly.
> 
> It would be a real bonus if we had a mechanism to install plugins at build-time too, like mentioned above. However, as we intend to work iteratively, start-time installation is a viable first step. While it goes against a few best-practices to change the container too much at runtime, this gives our users some rather easy flexibility without having to care for selfbuilt-images.

@Nebukadneza Thank you. I shall work on this.

jawabuu:
Hey @Nebukadneza @kaiyou @myteksp 
I have pushed a  rough POC of a python implementation. I decided to make all the changes in the roundcube/tests directory to be as un-intrusive as possible.
I would like the script tested to identify edge-cases (there are a number when it comes to plugins management).
Minimal instructions to test locally;
`cd tests/compose/roundcube`
`docker-compose up -d`
`docker-compose exec webmail ./plugins.py`

Diman0:
This PR is superseded by PR#2035 where we are looking at using the official roundcube docker image. This image supports configuring requested plugins via environment variables.

We will close this PR after PR#2035 has been merged.

Diman0:
With PR#2035 merged it is possible to activate and configure plugins shipped with roundcube.
See https://mailu.io/master/configuration.html#webmail-settings

We are not planning to allow users to install any random roundcube plugin from the roundcube repository using composer. If you have this need, we recommend to use a different roundcube image that supports this. For example the offical rouncube image (https://github.com/roundcube/roundcubemail-docker#installing-roundcube-plugins).

I'm closing this pull request.

PR #1324, opened 2020-01-26
Fix hardcoded reference to admin container.

frgomes:
## What type of PR?

big-fix

## What does this PR do?

Fixes a hardcoded reference to ``admin`` container which may or may not be called ``admin`` in reality, since the user may eventually to name the container otherwise.

### Related issue(s)
- closes #1323 

## Prerequistes
Before we can consider review and merge, please make sure the following list is done and checked.
If an entry in not applicable, you can check it or remove it from the list.

- This PR is intended to be applied over branch 1.7
- Ideally, a new branch (or tag) should be made available
- Minor bug-fix, no additional documentation needed.

PR #1330, opened 2020-01-27
Fix hardcoded reference to admin container.

### LINKED ISSUE #1323: Hardcoded http://admin/ in fetchmail.py
I've tweaked ``docker-compose.yml`` so that all my containers related to ``mailu`` are prefixed, in order to avoid conflict with other containers I start alongside containers relted to ``mailu``.

However, the hardcode ``http://admin/...`` below causes failure in ``fetchmail``, since my container container is now named ``maily-admin`` in my docker-compose.yml, not ``admin`` as the code supposes it should be.

```
./services/fetchmail/fetchmail.py:47:    fetches = requests.get("http://admin/internal/fetch").json()
./services/fetchmail/fetchmail.py:85:            requests.post("http://admin/internal/fetch/{}".format(fetch["id"]),
```

frgomes:
## What type of PR?

big-fix

## What does this PR do?

Fixes a hardcoded reference to ``admin`` container which may or may not be called ``admin`` in reality, since the user may eventually to name the container otherwise.

### Related issue(s)
- closes #1323
- subsumes #1324 

## Prerequisites

- Minor bug-fix, no additional documentation needed.

### DIFF
--- a/optional/fetchmail/fetchmail.py
+++ b/optional/fetchmail/fetchmail.py
@@ -44,7 +44,7 @@ def fetchmail(fetchmailrc):
 
 
 def run(debug):
-    fetches = requests.get("http://admin/internal/fetch").json()
+    fetches = requests.get("http://" + os.environ.get("HOST_ADMIN", "admin") + "/internal/fetch").json()
     smtphost, smtpport = extract_host_port(os.environ.get("HOST_SMTP", "smtp"), None)
     if smtpport is None:
         smtphostport = smtphost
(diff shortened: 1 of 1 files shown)


micw:
Looks good to me. However, behaviour of fetchmail would still differ a lot from the other services which uses the "socreate" library and the behaviour in https://github.com/Mailu/socrate/blob/master/socrate/system.py -> get_host_address_from_environment.

I would really love it if you could add socreate to the fetchmail image and use the function there for host lookups.
Relevant examples:
- https://github.com/Mailu/Mailu/blob/master/core/postfix/Dockerfile#L9
- https://github.com/Mailu/Mailu/blob/master/core/postfix/start.py#L30

Kind regards,
Michael.

kaiyou:
Same comment as @micw, impoting and using socrate would make much sense here.

Nebukadneza:
Sorry for the long silence~

While i agree with @kaiyou and @micw , this is still a improvement compared to the current situation. So, while not perfect, I’d take it in as-is.

@frgomes would you be willing to rebase this on current master?

Issue #1339, opened 2020-02-05
Support for usernames without domain part (default domain)


warmanton:
Looks like that current version of Mailu-1.7 does not support usernames without domain part i.e. with default domain. It is very hard to migrate from some old mail system where 1000 users have configured their MUA without domain name.

Perhaps this is a feature request to have some default domain which is used if user provide username without "@somedomain.com" but "username@somedomain.com" should work too.

Nebukadneza:
Hi There,

The `Mailu`-Project is currently in a bit of a bind! We are short on man-power, and we need to judge if it is possible for us to put in some work on this issue.

To help with that, we are currently trying to find out which issues are actively keeping users from using `Mailu`, which issues have someone who want to work on them — and which issues may be less important. These a less important ones could be discarded for the time being, until the project is in a more stable and regular state once again.

In order for us to better assess this, it would be helpful if you could put a **reaction on this post** (use the :smiley: icon to the top-right).
- 👍️ if you **need this** to be able to use Mailu. Ideally, you’d also be able to test this on your installation, and provide feedback …
- 🎉 if you find it a **nice bonus**, but no deal-breaker
- 🚀 if you want to **work on it yourself**!
We want to keep this **voting open for 2 weeks** from now, so please help out!

Diman0:
After discussing this during the last meeting, we decided we do not want to support this feature.  We do not support  email addresses without the domain part. Using a full email address is a requirement for using Mailu.

nathanielhourt:
I wanted this. So I wrote it myself. Diff from branch `2.0`:

```diff
diff --git a/core/admin/mailu/configuration.py b/core/admin/mailu/configuration.py
index bc91c332..c420ecd2 100644
--- a/core/admin/mailu/configuration.py
+++ b/core/admin/mailu/configuration.py
@@ -36,6 +36,7 @@ DEFAULT_CONFIG = {
     'SECRET_KEY': 'changeMe',
     'DOMAIN': 'mailu.io',
     'HOSTNAMES': 'mail.mailu.io,alternative.mailu.io,yetanother.mailu.io',
+    'DEFAULT_LOGIN_DOMAIN': '',
     'POSTMASTER': 'postmaster',
     'WILDCARD_SENDERS': '',
     'TLS_FLAVOR': 'cert',
diff --git a/core/admin/mailu/sso/forms.py b/core/admin/mailu/sso/forms.py
index c01ef572..99bc704e 100644
[...]

PR #1344, opened 2020-02-06
add DEFAULT_DOMAIN feature

warmanton:
## What type of PR?
feature

## What does this PR do?
PR adds env variable DEFAULT_DOMAIN for domainless login names

### Related issue(s)
- Mention an issue like:  #1339

## Prerequistes
- [*] In case of feature or enhancement: documentation updated accordingly

### DIFF
--- a/core/admin/mailu/configuration.py
+++ b/core/admin/mailu/configuration.py
@@ -43,6 +43,7 @@ DEFAULT_CONFIG = {
     'DKIM_SELECTOR': 'dkim',
     'DKIM_PATH': '/dkim/{domain}.{selector}.key',
     'DEFAULT_QUOTA': 1000000000,
+    'DEFAULT_DOMAIN': None,
     # Web settings
     'SITENAME': 'Mailu',
     'WEBSITE': 'https://mailu.io',
--- a/core/admin/mailu/internal/nginx.py
+++ b/core/admin/mailu/internal/nginx.py
@@ -46,6 +46,10 @@ def handle_authentication(headers):
         raw_password = urllib.parse.unquote(headers["Auth-Pass"])
         password = raw_password.encode("iso8859-1").decode("utf8")
         ip = urllib.parse.unquote(headers["Client-Ip"])
+
+        if not ('@' in user_email ) and app.config["DEFAULT_DOMAIN"]:
+            user_email = user_email + "@" + app.config["DEFAULT_DOMAIN"]
+
         user = models.User.query.get(user_email)
         status = False
         if user:
--- a/core/admin/mailu/internal/views/postfix.py
+++ b/core/admin/mailu/internal/views/postfix.py
@@ -45,6 +45,11 @@ def postfix_sender_login(sender):
     if localpart is None:
         return flask.abort(404)
     destination = models.Email.resolve_destination(localpart, domain_name, True)
+    # postfix has "smtpd_sasl_local_domain = my.defaultdomain.com" it allows to auth
+    # domainless users but it does not work in 'check_sender_access ${podop}senderaccess'.
+    # Have no idea how to do workaround. Only this dirty hack (simply add local part to
+    # the allowed senders list):
+    destination.append(localpart)
     return flask.jsonify(",".join(destination)) if destination else flask.abort(404)
 
 
--- a/core/dovecot/conf/dovecot.conf
+++ b/core/dovecot/conf/dovecot.conf
@@ -6,7 +6,9 @@ protocols = imap pop3 lmtp sieve
 postmaster_address = {{ POSTMASTER }}@{{ DOMAIN }}
 hostname = {{ HOSTNAMES.split(",")[0] }}
 submission_host = {{ FRONT_ADDRESS }}
-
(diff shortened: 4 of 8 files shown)


warmanton:
bors try

mildred:
bors try

kaiyou:
> Did you check this actually work ?
> 
> Also, could it make sense not to introduce a DEFAULT_DOMAIN but use HOSTNAME instead? This would avoid yet another environment variable and avoid complex state where DEFAULT_DOMAIN != HOSTNAME. This would enable the feature by default, would that be a problem ?

In fact, HOSTNAME has a very different semantic, it is the SMTP and HTTP hostname of the server, which does not even needs to be one of the server domains. I think introducting an additional variable is required here until we have a decent in-database config for it.

kaiyou:
About reviewing this myself, I have a couple important things to check before anything else. My basic opinion is that using Postfix and Dovecot  settings for this opens as usual a possible issue with varying semantics, which is possibly why you faced an issue with sender access.

warmanton:
> Did you check this actually work ?
> 
> Also, could it make sense not to introduce a DEFAULT_DOMAIN but use HOSTNAME instead? This would avoid yet another environment variable and avoid complex state where DEFAULT_DOMAIN != HOSTNAME. This would enable the feature by default, would that be a problem ?

Yes, I have checked with changed files. I mount patched files via docker volume and tested imap/smpt with full username and without domain part. Status works for me as expected.

warmanton:
> About reviewing this myself, I have a couple important things to check before anything else. My basic opinion is that using Postfix and Dovecot settings for this opens as usual a possible issue with varying semantics, which is possibly why you faced an issue with sender access.

I think this is postfix issue. Postfix has 'smtpd_sasl_local_domain` option but expects domainless username when does check against `check_sender_access` list but not 'username@domain' which is returned. I.e. if postfix is able to add some default domain if domain is not specified during auth then it should do the same while checking sender access. As I can see it does not add default domain during this check. I am not very familiar with postfix and probably there is an option to do correct sender checking with default domain.

kaiyou:
Okay, thank you for clarifying. What is your opinion on using `DOMAIN` as the default domain by the way?

warmanton:
> Okay, thank you for clarifying. What is your opinion on using `DOMAIN` as the default domain by the way?

`DOMAIN` variable has slightly different meaning. Of cource if system is planned from scratch then `DOMAIN` can be used (b.t.w. any other domain name related variable can be used) but it does not cover this case: I have up and running Mailu system with multiple domains. All works as expected. Then I need to serve yet another some old domain with 100-1000 users whith domainless usernames. `DOMAIN` variable is already has some value so I cannot use it. Users have to reconfigure their MUA or I can configure `DEFAULT_DOMAIN` for them without any any additional reconfiguration on Mailu side. Of cource `DEFAULT_DOMAIN` is "one shot" variable for specific case for some single domain.

So my opinion is to use `DEFAULT_DOMAIN` variable or variable with other not reserved yet name for this feature.

Nebukadneza:
Hi @warmanton,

while i do understand your use-case, and feel it is a legitimate one … I’m a bit torn about this PR. Currently the project struggles with man-power, and this is a big change with a lot of complexity to test and bring into the next release.

You invested so much work in this, so I would feel bad for closing it.

Given my concerns … would you still be interested in rebasing this PR, and continueing it? Or is your use-case already satisfied otherwise? Or … do you have a way to use mailu in its current state, maybe with some means of migration?

Sorry & Best,
-Dario

warmanton:
Hi @Nebukadneza. I do not use Mailu with my DEFAULT_DOMAIN feature (we postponed our mail system migration and use old one). If my PR breaks some others important PRs or conflicts do not merge my PR. I think it can be closed without merge, looks like nobody needs this feature.

Nebukadneza:
Thank you so much for your understanding. FYI, i put a voting up to the issue, so maybe we get feedback from other users on whether this could be relevant to them. Let’s decide after we see what’ll come out.

lub (changes requested):
Hi,

thanks for the PR! I've left a view comments.

Additionally I think we need additional this to enable login with the legacy name in rainloop (or do we not want to also allow it in webmails?)

```
--- a/webmails/rainloop/application.ini
+++ b/webmails/rainloop/application.ini
@@ -13,5 +13,8 @@ allow_gravatar = Off
 enable = On
 allow_sync = On

+[login]
+default_domain = {{ DEFAULT_DOMAIN }}
+
 [plugins]
 contacts_autosave = On
```

I also have trouble with sending as aliases while logged in via DEFAULT_DOMAIN, but couldn't figure out why. I guess the 'hack' is not enough.

It's also not possible to login to the admin gui with only the short user name. Is this fine? If yes I think we should add this to the documentation.  
A newsfragment file for this would also be great, to include it in the changelog.

This also needs a rebase (or manual merge) as far as I can see.

Best,
lub

Diman0:
After discussing this during the last meeting, we decided to close this PR. We are not planning to support email addresses without the domain part. This adds yet more complexity to the project with almost no real gain.

PR #1346, opened 2020-02-06
Parameterize dkim selector

### LINKED ISSUE #1345: Use DKIM_SELECTOR header when signing emails
@micw 
When starting Mailu, It is possible to define an environment variable `DKIM_SELECTOR=my_dkim` to replace the default `DKIM_SELECTOR=dkim`
Keys can be generated successfully on Admin  UI and added to DNS records.
However Mailu still signs email using the default `dkim` key resulting in failed validations.

jawabuu:
## What type of PR?
Enhancement

## What does this PR do?
Configures rspamd to sign emails using user provided `DKIM_SELECTOR` instead of using the default `dkim` selector even when `DKIM_SELECTOR` is defined..
### Related issue(s)
- Auto close an issue like: closes #1345 
This should prove useful to people that have existing mail servers apart from mailu for the same domain.

### DIFF
--- a/core/rspamd/conf/dkim_selectors.map
+++ b/core/rspamd/conf/dkim_selectors.map
@@ -1,0 +1,1 @@
+{{ DOMAIN }} {% if DKIM_SELECTOR %}{{ DKIM_SELECTOR }}{% else %}dkim{% endif %}
--- a/core/rspamd/conf/dkim_signing.conf
+++ b/core/rspamd/conf/dkim_signing.conf
@@ -2,3 +2,7 @@ try_fallback = true;
 path = "/dkim/$domain.$selector.key";
 use_esld = false;
 allow_username_mismatch = true;
+selector = "dkim";
+{% if DKIM_SELECTOR %}
+selector_map = "/etc/rspamd/local.d/dkim_selectors.map";
+{% endif %}"


kaiyou:
I fear that using the map, the selector will only be set for the main domain, won't it? Why not simply set the selector to the given value?

jawabuu:
You are correct.
The map sets the selector to the given value for the **DOMAIN**  variable. Any other domain not in the map will use "dkim" as the selector.
Setting it as the selector the will set it as the selector for all domains the user may add. The choice in my case was informed by a desire to make the assignment explicit for the user.
If it is not an issue assigning it for all the domains, I will make the change.

jawabuu:
Additionally the map also allows flexibility of different selectors for different domains if that is implemented in future.

Nebukadneza:
I’m going to revive this one, as it looks useful. I’m not deep into DKIM myself I have to admit.

Question first: Even though we have been silent for a while, would you be willing to continue this once more?
Also, what’s the current state? The selector is only overridden in the map for `DOMAIN`, i.e., the single main-domain of the installation? This will not change it for the other domains, but leave those in working state?
If so, this is a clear step forward, and I’d be for integrating.

Also:

bors try

jawabuu:
Hey @Nebukadneza I am interested in this pull request
The selector only applies for the MAIN DOMAIN and only if specified to allow mailu to co-exist with other mail-servers.

nextgens:
This has been made obsolete by #2017

PR #1359, opened 2020-02-09
Refactor the rate limiting code

kaiyou:
## What type of PR?

Enhancement

## What does this PR do?

Rate limiting was already redesigned to use Python limits. This
introduced some unexpected behavior, including the fact that only
one criteria is supported per limiter. Docs and setup utility are
updated with this in mind.

Also, the code was made more generic, so limiters can be delivered
for something else than authentication. Authentication-specific
code was moved directly to the authentication routine.

### Related issue(s)

No specific issue.

## Prerequistes

### DIFF
--- a/core/admin/mailu/internal/views/auth.py
+++ b/core/admin/mailu/internal/views/auth.py
@@ -5,21 +5,31 @@ from flask import current_app as app
 import flask
 import flask_login
 import base64
-
+import ipaddress
 
 
 @internal.route("/auth/email")
 def nginx_authentication():
     """ Main authentication endpoint for Nginx email server
     """
-    utils.limiter.check(flask.request.headers["Client-Ip"])
+    limiter = utils.limiter.get_limiter(app.config["AUTH_RATELIMIT"], "auth-ip")
+    client_ip = flask.request.headers["Client-Ip"]
+    if not limiter.test(client_ip):
+        response = flask.Response()
+        response.headers['Auth-Status'] = 'Authentication rate limit from one source exceeded'
+        response.headers['Auth-Error-Code'] = '451 4.3.2'
+        if int(flask.request.headers['Auth-Login-Attempt']) < 10:
+            response.headers['Auth-Wait'] = '3'
+        return response
     headers = nginx.handle_authentication(flask.request.headers)
     response = flask.Response()
     for key, value in headers.items():
         response.headers[key] = str(value)
-    if ("Auth-Status" not in headers) or (headers["Auth-Status"]!="OK"):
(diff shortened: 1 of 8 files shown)


docs/compose/.env
```diff
@@ -68,6 +68,10 @@ ANTIVIRUS=none
 # Max attachment size will be 33% smaller
 MESSAGE_SIZE_LIMIT=50000000
 
+# Message rate limit for outgoing messages
+# This limit is per user
+MESSAGE_RATELIMIT=100/day
+
```
Nebukadneza: I think this slipped in from a rebase orgy together with #1360

docs/configuration.rst
```diff
@@ -57,6 +56,10 @@ The ``MESSAGE_SIZE_LIMIT`` is the maximum size of a single email. It should not
 be too low to avoid dropping legitimate emails and should not be too high to
 avoid filling the disks with large junk emails.
 
+The ``MESSAGE_RATELIMIT`` is the limit of messages a single user can send. This is
+meant to fight outbound spam in case of compromised or malicious account on the
+server.
+
```
Nebukadneza: dito, probably belongs in #1360

core/admin/mailu/internal/views/auth.py
```diff
@@ -5,21 +5,31 @@
...
     for key, value in headers.items():
         response.headers[key] = str(value)
-    if ("Auth-Status" not in headers) or (headers["Auth-Status"]!="OK"):
-        utils.limiter.hit(flask.request.headers["Client-Ip"])
-
+    if ("Auth-Status" not in headers) or (headers["Auth-Status"] != "OK"):
+        limit_subnet = str(app.config["AUTH_RATELIMIT_SUBNET"]) != 'False'
+        subnet = ipaddress.ip_network(app.config["SUBNET"])
+        if limit_subnet or ipaddress.ip_address(client_ip) not in subnet:
+            limiter.hit(flask.request.headers["Client-Ip"])
```
Nebukadneza: architectural nitpick: i wonder if this could (later on) live in the wrapper-code, since it — if i understand the `AUTH_RATELIMIT_SUBNET` setting, that should apply to all places where we might want to rate-limit in the future. The wrapper would then be named something akin to a mailu-specific-wrapper which includes own logic in addition to proxying the call to the lib.

Otoh, since currently this is the only place we really use the limiter, this is quite fine ^_^

Nebukadneza:
One thing I’ve overlooked — I think the `AUTH_RATELIMIT_SUBNET` has never been in any of the env-files, only docs. Even if it has a default … maybe we should add it to the env files, so that it has greater user-visibility?
Your call …

PR #1380, opened 2020-03-01
[Roundcube] DKIM sign message delivery reports

TheLegend875:
This PR enables the From header for message delivery reports in Roundcube.
This ensures that the message delivery report is DKIM signed and therefore not not blocked or considered spam by receiving mailservers.


Nebukadneza:
It looks like clamav update service was unavailable here, causing this to fail (sorry, didn’t look for too long). Let’s retry

bors try

Nebukadneza:
Could you please reference your matching ticket, #1381, like
```
closes #1381
```
Thanks ^_^

ofthesun9:
Added a fragment 1831.fix to update the changelog

PR #1389, opened 2020-03-06
Prefer specific alias over wildcard, regardless of case

### LINKED ISSUE #1387: Case insensitiveness broken for aliases
If there’s a catchall (`foo%`) as well as a specific alias (`foobar`) on a domain, and a mail arrives for that alias, but in a different case (`FOOBAR`), it is delivered to the catchall instead of the specific alias. If the mail arrives in the correct case, the specific alias is chosen. This is inconsistent behavior (… and endlessly unnerving with airlines and the likes), so brace for a hopefully-soon-to-come PR ^_^.

Nebukadneza:
## What type of PR?
bug-fix

## What does this PR do?
Since direct addresses (not aliases) are case-insensitive since a while,
it makes sense for aliases to behave the same. Up until now, a wildcard
alias could trump a alias not-matching-the-case of the incoming address.
This clarifies this behavior.

## Notes
I realize that the if-hell down there isn’t nice. What it is, however, is quite clear and easy to read. I’m hoping that if anyone ever gets confused in the future, this will make the current behavior transparent. For me, that was more important than a minimal amount of statements/branches …

### Related issue(s)
closes #1387

## Prerequistes

### DIFF
--- a/core/admin/mailu/models.py
+++ b/core/admin/mailu/models.py
@@ -451,24 +451,33 @@ class Alias(Base, Email):
                     )
                 )
             ).order_by(cls.wildcard, sqlalchemy.func.char_length(cls.localpart).desc()).first()
-        if alias_preserve_case:
-            return alias_preserve_case
 
-        if localpart:
-            localpart = localpart.lower()
-        return cls.query.filter(
+        localpart_lower = localpart.lower() if localpart else None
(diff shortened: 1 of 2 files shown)


ofthesun9:
I agree that direct addresses and aliases should behave the same.

ofthesun9:
bors retry

PR #1392, opened 2020-03-07
Use environment variables for cert paths/names in nginx certwatcher

### LINKED ISSUE #903: certwatcher.py is not using TLS_CERT_FILENAME nor TLS_KEYPAIR_FILENAME
certwatcher.py is assuming that the certificate and the key are in /certs and named respectively cert.pem and key.pem

However, Mailu offer two environment variables to allow specific path & filename for the certificate and the key which are used in config.py:
TLS_CERT_FILENAME
TLS_KEYPAIR_FILENAME

I would add that in my use case (https-portal as a reverse proxy), those files are not directly in /certs. My environment variables look like:
TLS_CERT_FILENAME=mail.example.net/production/signed.crt
TLS_KEYPAIR_FILENAME=mail.example.net/production/domain.key
Currently, certwatcher.py is monitoring `/certs`, whereas I would need to monitor 
`/certs/mail.example.net/production`

Nebukadneza:
## What type of PR?
bug-fix

## What does this PR do?
Previously, nginx certwatcher would only react to the hardcoded paths. It should have
honored the enviroment variables that are used by config.py too for this.
 
### Related issue(s)
closes #903

## Prerequistes
- [x] no feature or enhancement
- [x] minor/internal change

### DIFF
--- a/core/nginx/certwatcher.py
+++ b/core/nginx/certwatcher.py
@@ -5,8 +5,8 @@ happens to externally supplied certificates. Only executed by start.py in case
 of TLS_FLAVOR=[mail, cert]
 """
 
-from os.path import exists, split as path_split
-from os import system
+from os.path import exists, split as path_split, join as path_join
+from os import system, getenv
 import time
 from watchdog.observers.polling import PollingObserver
 from watchdog.events import FileSystemEventHandler, FileDeletedEvent, \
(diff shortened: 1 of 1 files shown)


Nebukadneza:
clamav update again :( 

bors try

Nebukadneza:
still looks like a temporary failure

bors try

ofthesun9:
Tested on my setup, this works great !
Thanks !

PR #1398, opened 2020-03-09
Update crypto to be modern and inline with tls.conf

bladeswords:
Updated to match tls.conf and be aligned to more modern cryptographic standards and only use currently secure protocols and ciphers.

## What type of PR?

bugfix

## What does this PR do?
Update to use more modern cryptographic techniques
### Related issue(s)
- Addresses comment raised in 4f973f6

## Prerequistes
None


Nebukadneza:
bors try

Nebukadneza:
bors try

Nebukadneza:
sorry for joining the show so late … but actually i’d advise `ssl_prefer_server_ciphers off`. The reasons for the configurations we chose is very well explained by the mozilla(?) folks who maintain the SSL Configuration Generator:
https://github.com/mozilla/server-side-tls/issues/260

I think that we should follow their recommendation, with the reasons gives sound valid to me …

bladeswords:
@Nebukadneza  - you and me both.  That is why this change modified the setting to off.  :)

PR #1399, opened 2020-03-09
Remove SPF type SPF record #1394

### LINKED ISSUE #1394: In accordance with RFC 7208, offer only TXT RRs for SPF
Currently, the `Domain Details` view tells the user to create a `[…] IN SPF […]` record in their DNS zones. According to [RFC 7208](https://tools.ietf.org/html/rfc7208):
>    SPF records MUST be published as a DNS TXT (type 16) Resource Record
>   (RR) [RFC1035] only.  […] Use of alternative DNS RR types was supported in
>   SPF's experimental phase but has been discontinued.

AFAIK the code-change can be small since this is only displayed, but we might want to let users know that they have to (manually) change their setups to stay compliant. Any thoughts on the latter?

bladeswords:
As mentioned in #1394 - In accordance with RFC 7208, offer only TXT RRs for SPF.
Agree with @Nebukadneza - but not sure how to go about telling people to remove the old record...

## What type of PR?

Documentation

## What does this PR do?
Removes the recommendation to add a SPF RR for SPF records, as this is no longer RFC complaint and often causes issues to maintain two records.

### Related issue(s)
- closes #1394

## Prerequistes
None

### DIFF
--- a/core/admin/mailu/ui/templates/domain/details.html
+++ b/core/admin/mailu/ui/templates/domain/details.html
@@ -34,8 +34,7 @@
 <tr>
   <th>{% trans %}DNS SPF entries{% endtrans %}</th>
   <td><pre>
-{{ domain.name }}. 600 IN TXT "v=spf1 mx a:{{ hostname }} -all"
-{{ domain.name }}. 600 IN SPF "v=spf1 mx a:{{ hostname }} -all"</pre></td>
+{{ domain.name }}. 600 IN TXT "v=spf1 mx a:{{ hostname }} -all"</pre></td>
 </tr>
 {% if domain.dkim_publickey %}
 <tr>
--- a/towncrier/newsfragments/1394.bugfix
+++ b/towncrier/newsfragments/1394.bugfix
@@ -1,0 +1,1 @@
+Show SPF records in accordance with RFC 7208: Previously we instructed admins to create SPF and TXT records, where only TXT records are correct now. !! Attention !! You need to manually remove the SPF-typed records and keep only TXT in your DNS.


kaiyou:
Does keeping the old record have a negative impact on existing setup?

Nebukadneza:
i have to admit i don’t know if anyone misinterprets something with the old record in place. essentially, if you’re a spec-nitpicker, it’s not valid to have both as far as i can see.

also, clamav hell again:
bors try

bladeswords:
@kaiyou - not that I know of (but there are some very unusual mail implementations out there), but based on some research a while ago, usually if someone published both SPF and TXT SPF records, they didn't match - which indicated to me that at least one was the wrong policy.  This is more of a human issue not a technical one (and if I trusted it wouldn't be messed up I would advise do both and make sure both are identical)

Nebukadneza:
mh, that sounds reasonable. I’ve also tried digging around a bit to find out if people having only TXT records face problems … couldn’t find any reports, but that’s only one clue, not any real evidence.

I’ve also removed my non-TXT records on my live box, and didn’t face any problems so far.

Still, my gut feeling would be we should only offer TXT, and maybe add something to the changelogs/upgrading that notifies admins they should clean up their DNS.

@bladeswords do you want to add a newsfragment to tell admins they have some manual work, or should I add one to this PR?

bladeswords:
@Nebukadneza - I have no idea what the newsfragment is or looks like - so feed free to add :)

Nebukadneza:
It seems you have disallowed maintainer push on your branch. Do you want to enable it? Alternatively, you can steal the commit from here:
https://github.com/Nebukadneza/Mailu/commit/007f4e3403906147bf10798abbe7e54b585ee280

ofthesun9:
@Nebukadneza : I pushed your commit on @bladeswords push request
@bladeswords the [fragments](https://mailu.io/master/contributors/workflow.html#changelog) are used to build the changelog, and a fragment should be part of every PR.

ofthesun9:
bors try

muhlemmer:
> mh, that sounds reasonable. I’ve also tried digging around a bit to find out if people having only TXT records face problems … couldn’t find any reports, but that’s only one clue, not any real evidence.

I only ever had TXT records. Never any problems from that perspective.

PR #1433, opened 2020-04-05
Adjust kubernetes manifests to K8S 1.18

jjlakis:


micw:
Hello,
thank you for the PR. The old k8s yamls are very likely to be removed in 1.8. There is a helm chart at https://github.com/Mailu/helm-charts and the static yamls will be rendered from the chart (see https://github.com/Mailu/helm-charts/tree/gh-pages/yaml/ for a preview version).

The chart should already work with k8s 1.18.

Nebukadneza:
@micw you can better judge this … is this PR deprecated in favor of the helm-charts repo, and thus we can close it, or is there a reason to keep it open?

jlcox1970:
a bit late in but as the wiki direct users to install using YAML and not helm then all the yaml files need to updated to align with documentation.

PR #1441, opened 2020-04-09
Rsyslog logging for postfix

micw:
## What type of PR?

enhancement

## What does this PR do?
Changes postfix logging from stdout to rsyslog:
* stdout logging still enabled
* internal test request log messages are filtered out by rsyslog
* optional logging to file via POSTFIX_LOG_FILE env variable

## Prerequistes


micw:
bors try

kaiyou:
The thing about this PR is: we historically had rsyslog installed and running in the Postfix image, and we explicitely removed it because we wanted things to be simple: Postfix takes care or delivering mail and writes its log to stdout, like Docker best practice dictate, then Docker takes care of logging, either to file or to anything else.

What is the use case for this, can't it be accomplished with more standard means in the Docker ecosystem?

micw:
I added rsyslog for two reasons. One is that the logs are flooded by log lines caused by test request requests (one every few seconds). It can easily be filtered out by rsyslog.
The other is that especially for mail, logging is often needed for legal or customer service reasons ("was that mail really sent (and when)" or "why did I miss that particular mail"). So the idea is to allow extra logging of postfix maillog into a file (in addition to docker logs).

micw:
I also thought about adding an extra rsyslog container. Drawback is that postfix can either log to stdout or to syslog, so with that we'd loose the stdout logging of the postfix container

kaiyou:
I get this, and I sure agree that rsyslog can be really useful. But what is wrong about setting the log driver to syslog for that container, and let Docker, k8s or the execution stack take care of the log pipeline?

The syslog driver is well documented at: https://docs.docker.com/config/containers/logging/syslog/

It can even be very useful if you want to ship logs outside, or collect the logs from multiple Mailu servers or multiple instances of the SMTP container, then have a single place for filtering/indexing/etc.

micw:
There's just no easy way to achieve this on k8s for a single deployment (https://kubernetes.io/docs/concepts/cluster-administration/logging/). The docker log driver is handled by k8s, kubernetes docs recommends using some extra service for special log handling.
Using rsyslog here would be by far the easiest solution.

micw:
Related: https://github.com/Mailu/Mailu/issues/1741

micw:
@kaiyou Would the feature be acceptable if syslog is optional and default to the old behaviour?

core/postfix/start.py
```diff
@@ -37,6 +37,10 @@ def is_valid_postconf_line(line):
 os.environ["ANTISPAM_MILTER_ADDRESS"] = system.get_host_address_from_environment("ANTISPAM_MILTER", "antispam:11332")
 os.environ["LMTP_ADDRESS"] = system.get_host_address_from_environment("LMTP", "imap:2525")
 
+os.environ["LMTP_ADDRESS"] = system.get_host_address_from_environment("LMTP", "imap:2525")
```
Diman0, line 40: Duplicate line (see line 38 above it). You can remove it.

docs/configuration.rst
```diff
@@ -187,3 +190,15 @@ When using ``*_ADDRESS``, the hostnames must be full-qualified hostnames. Otherw
 resolve the hostnames.
 
 
+Maillog setitngs
```
Diman0, line 193: Spelling error.

Diman0 (dismissed):
See the in-line comments. 

It might be a handy feature as long it is only optionally enabled via mailu.env. It should not be the default.

Diman0:
We discussed this during the last mailu meeting. The functionality must be optional and not the default. This is already implemented in this PR. This PR is much needed for kubernetes users. 

@micw are you still planning on finishing it? Otherwise I will make the requested changes.

Diman0:
I have made the requested changes. I think it can be reviewed now.

Diman0:
I forgot to also add the newsfragment *which micw should already have added!). I guess it is time for some testing.

bors try

micw:
Great that you picked this up. I just saw that rsyslogd supports basic log rotation:
https://www.rsyslog.com/doc/master/tutorials/log_rotation_fix_size.html

It's size based only but maybe we should add this as an option anyways so that we have a fully-working basic solution?

Diman0:
I did not think about log rotation. Normally on a linux box, logrotate is already configured on the /var/log folder.
I agree it is important the logs must be rotated. We don't want a log file that keeps growing exponentially!

I do not think that your approach is viable. Your link uses deprecated functionality:
`As of this writing, it is most likely that they will be replaced by something different in the future.`
See https://www.rsyslog.com/doc/master/configuration/output_channels.html

I think it is wiser to  install the logrotate package and configure this to properly rotate the logs. Alpine also has logrotate available (https://pkgs.alpinelinux.org/packages?name=logrotate&branch=edge). 

It's been a long time ago since I configured log rotate. I will need to have a look at the documentation before I can make the changes. 

@micw Do you agree it is better to use logrotate? Or do you rather use the script you linked earlier?

micw:
Clear for logrotate. Maybe with
POSTFIX_LOG_ROTATE that takes a config for logrotate like "size 100k" or "daily"
POSTFIX_LOG_ROTATE_KEEP that takes an argument for "rotate $POSTFIX_LOG_ROTATE_KEEP"

Diman0:
> Clear for logrotate. Maybe with POSTFIX_LOG_ROTATE that takes a config for logrotate like "size 100k" or "daily" POSTFIX_LOG_ROTATE_KEEP that takes an argument for "rotate $POSTFIX_LOG_ROTATE_KEEP"
> 
> Edit: Syntax is quite simple: https://linux.die.net/man/8/logrotate
> 
>     * need to confgure when to rotate (user configurable)
> 
>     * need to confgure how many to keep (user configurable)
> 
>     * compression yes/no (user configurable or fixed)
> 
>     * some other settings specific to postgres (fixed)
> 
> 
> Edit 2: Postfix specific setting: https://www.cloudinsidr.com/content/set-up-logrotate-for-postfix/

That looks doable. We should always look out for adding to many configuration options. The configuration file (Mailu.env) is already too complicated. Now we had 4 more settings?

I have done some more thinking. This PR will remove all those useless log messages for the healthcheck and logs to standard out. If tests reveal it works fine, then we should maybe always use it. So no extra mailu.env setting for enabling  it, but have it always enabled. That means we will have in the advanced section of the configuration guide:

- POSTFIX_LOG_FILE ( by default empty / disabled )
- POSTFIX_LOG_ROTATE ( with default daily )
- POSTFIX_LOG_ROTATE_KEEP ( with default 365 )

For the last 2 settings we must make clear to what setting it corresponds in the logrotate config file, so the user can read the actual logrotate manual for how to configure the setting.

I think I will have a further look this weekend.

Diman0:
@micw The log to file never worked. You had an error in rsyslog, you logged to {{ LOG_FILE }} instead of {{POSTFIX_LOG_FILE}}. You did not test the PR before pushing it?

I have chosen to enable rsyslog by default. I think it is handy for all users that we do not log the healthcheck messages anymore. E.g. taken from my server (running mailu 1.8):
```
smtp_1       | Dec 01 12:48:30 mail postfix/smtpd[8620]: connect from localhost[127.0.0.1]
smtp_1       | Dec 01 12:48:30 mail postfix/smtpd[8620]: disconnect from localhost[127.0.0.1] quit=1 commands=1
smtp_1       | Dec 01 12:49:00 mail postfix/smtpd[8620]: connect from localhost[127.0.0.1]
smtp_1       | Dec 01 12:49:00 mail postfix/smtpd[8620]: disconnect from localhost[127.0.0.1] quit=1 commands=1
smtp_1       | Dec 01 12:49:30 mail postfix/smtpd[8620]: connect from localhost[127.0.0.1]
smtp_1       | Dec 01 12:49:30 mail postfix/smtpd[8620]: disconnect from localhost[127.0.0.1] quit=1 commands=1
smtp_1       | Dec 01 12:50:00 mail postfix/smtpd[8620]: connect from localhost[127.0.0.1]
smtp_1       | Dec 01 12:50:00 mail postfix/smtpd[8620]: disconnect from localhost[127.0.0.1] quit=1 commands=1
```
It is nice to be rid of this. It is still possible to disable rsyslog. I have chosen to handle the configuration of logrotate via an override. We already have **too many** configuration options in mailu.env. This is also an option that almost nobody will use. I think I picked sensible default values, but we could tweak these.

I have tested myself all the possible configurations:
- Use default settings (nothing configured)
- Log to file
- Use override.

The pull request can be reviewed now. 

bors try

nextgens:
I don't get this.

1) If the default is "ship with a syslogd" IMHO we shouldn't provide options for removing it.
2) Have you considered lighter alternatives to rsyslog? busybox has a syslogd that can remove duplicates if that's all we're after https://github.com/brgl/busybox/blob/master/sysklogd/syslogd.c
3) Logrotate is usually triggered by cron; Are you sure that it works and if so, how?

Diman0:
bors try

Diman0:
crond must still be started:
bash-5.1# /usr/sbin/crond --help
BusyBox v1.33.1 () multi-call binary.

Usage: crond [-fbS] [-l N] [-d N] [-L LOGFILE] [-c DIR]

	-f	Foreground
	-b	Background (default)
	-S	Log to syslog (default)
	-l N	Set log level. Most verbose 0, default 8
	-d N	Set log level, log to stderr
	-L FILE	Log to FILE
	-c DIR	Cron dir. Default:/var/spool/cron/crontabs
bash-5.1#

Diman0:
I double checked that crond is started (only) when POSTFIX_LOG_FILE is set.

bors try

nextgens (approved):
LGTM

PR #1444, opened 2020-04-11
Harden default configuration

Jarel1337:


setup/flavors/compose/docker-compose.yml
```diff
@@ -31,8 +31,8 @@ services:
     {% endif %}
     {% endfor %}
     volumes:
-      - "{{ root }}/certs:/certs"
-      - "{{ root }}/overrides/nginx:/overrides"
+      - "{{ root }}/certs:/certs:ro"
```
HorayNarea, line 34: `/certs` needs to be rw because front also creates/updates the Let's Encrypt certificates if you use them.

Jarel1337: Yes, my mad assuming everyone used their own certificates

muhlemmer:
I took the liberty of adding some documentation regarding this change. @Jarel1337, is what I wrote correct?

muhlemmer:
bors try

Issue #191, opened 2017-04-06
Strange header rewrite


senilio:
Running v1.4, I've noticed that outgoing email gets its Received header rewritten with PRIMARY_HOSTNAME [PUBLIC_IP], literally.

Sending an email to gmail, here's the headers on the receiver side.

Received: from authenticated-user (PRIMARY_HOSTNAME [PUBLIC_IP]) by domain.com (Postfix) with ESMTPSA id 87AD9648 for <recipient@gmail.com>; Thu,
  6 Apr 2017 12:56:52 +0000 (UTC)

Here's the replace taking place in the postfix logs:

smtp_1       | postfix/cleanup[61]: 0EADA648: replace: header Received: from [172.17.7.226] (unknown [81.146.13.44])??by domain.com (Postfix) with ESMTPSA id 0EADA648??for <recipient@gmail.com>; Thu,  6 Apr 2017 12:34:23 +0000 (UTC) from unknown[81.146.13.44]; from=<sender@domain.com> to=<recipient@gmail.com> proto=ESMTP helo=<[172.17.7.226]>: Received: from authenticated-user (PRIMARY_HOSTNAME [PUBLIC_IP])??by domain.com (Postfix) with ESMTPSA id 0EADA648??for <recipient@gmail.com>; Thu,  6 Apr 2017 12:34:23 +0000 (UTC)

I'd expect PRIMARY_HOSTNAME [PUBLIC_IP] to be a real hostname and IP.

Thanks!

kaiyou:
That is actually the intended behavior, in order to keep the full ``Received`` chain while helping authenticated users stay anonymous when sending emails through SMTP or Submission.

You can change this behaviour and build your own version of the ``smtp`` container by editing ``outclean_header_filter``.

senilio:
Thanks for clarifying this!

PR #1448, opened 2020-04-14
Override to use server hostname + IP in outbound headers

dmwilcox:
## What type of PR?
Enhancement

## What does this PR do?
Allows users to set the server IP and hostname in the outgoing "Received" header rather than "PUBLIC_HOSTNAME [PUBLIC_IP]".

### Related issue(s)
https://github.com/Mailu/Mailu/issues/191

## Prerequistes

### DIFF
--- a/core/postfix/conf/outclean_header_filter.cf
+++ b/core/postfix/conf/outclean_header_filter.cf
@@ -4,7 +4,11 @@
 # Remove the first line of the Received: header. Note that we cannot fully remove the Received: header
 # because OpenDKIM requires that a header be present when signing outbound mail. The first line is
 # where the user's home IP address would be.
+{%- if USE_SERVER_OUTCLEAN and HOST_OUTCLEAN and OUTCLEAN_ADDRESS %}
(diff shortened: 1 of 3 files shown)


lub:
Does it make sense to generally use something like `os.environ["HOSTNAMES"].split(",")[0]` + 127.0.0.1/0.0.0.0/..?

For example Exchange Online also seems to just write the server handling the connection instead of the client into the header:

```
Received: from AM0PR05MB4452.eurprd05.prod.outlook.com
 ([fe80::51ec:24fe:245c:527e]) by AM0PR05MB4452.eurprd05.prod.outlook.com
 ([fe80::51ec:24fe:245c:527e%7]) with mapi id 15.20.3391.014; Wed, 16 Sep 2020
 08:08:02 +0000
```

The server information is public anyway and it's probably better than a generic placeholder.

nextgens:
I agree; I think it would make sense to avoid a new configuration flag... and to turn it on by default with a fallback to 127.0.0.1 if the address lookup fails.

nextgens:
@dmwilcox are you still okay to work on this?

devfaz:
Looking forward to see this feature ;)

Issue #707, opened 2018-11-10
Verify DANE when sending emails and use STARTTLS when possible.


jcgruenhage:
According to https://ssl-tools.net/mailservers/jcg.re, mailu seems to not use StartTLS when sending emails. Additionally, https://havedane.net/ indicates that when a DANE record is available but an invalid certificate is used, mailu will still deliver the mail. It makes sense, since it doesn't use StartTLS there is no certificate to validate.

muhlemmer:
Up for testing and submitting tomorrow.

jcgruenhage:
I'm currently running 1.5, what would be the best way to test that?

muhlemmer:
I will test it myself (now actually). I'll use the links you've provided.

jcgruenhage:
So, how did your tests go?

jcgruenhage:
Related to this, there is also MTA-STS that should be checked.

muhlemmer:
> So, how did your tests go?

Not successful. I've set DANE checking on Postfix, but it's still sending mail to a server with broken certs. I've halted work on this for the moment, because some more urgent issues popped up.

You can try it for yourself if you use the following commit: https://github.com/usrpro/Mailu/commit/5b275f1745a41b9e2aa8315cf568507f34dfafb4.

jcgruenhage:
I'm afraid I have more urgent things to do right now too, but thanks for starting to work on this!

ofthesun9:
I started also some tests (reading https://tech.feedyourhead.at/content/postfix-verified-tls-with-dane)
and also found that it is still sending to server with broken certs.

Anyway it seems that some mail domain have a broken implementation of DANE, which prevents mail delivery (if DANE is activated on mail side). There is a "dane fail list" at https://github.com/danefail/list
Before activating DANE on mailu, we shall consider whether we want to incorporate the "fail list" or not...

jcgruenhage:
If this won't be fixed any time soon, then the docs should state it. Currently mailu.io says that TLS is enforced, which while technically true, doesn't give any benefit if people can MitM those connections trivially anyway.

vchoi:
> Before activating DANE on mailu, we shall consider whether we want to incorporate the "fail list" or not...

From the commit history, this list is frequently updated.(https://github.com/danefail/list/commits/master)

If this ever gets implemented, I suggest there should be an automated process to fetch and update the list.

jcgruenhage:
[...]

PR #1478, opened 2020-05-02
Allow to enforce TLS for outbound

micw:
using OUTBOUND_TLS_LEVEL=encrypt (default is 'may')

## What type of PR?

enhancement

## What does this PR do?

Add an option to postfix to enforce outbound traffic to be TLS encrypted.

## Prerequistes

### DIFF
--- a/core/postfix/conf/main.cf
+++ b/core/postfix/conf/main.cf
@@ -52,9 +52,10 @@ tls_high_cipherlist = EDH+CAMELLIA:EDH+aRSA:EECDH+aRSA+AESGCM:EECDH+aRSA+SHA256:
 tls_preempt_cipherlist = yes
 tls_ssl_options = NO_COMPRESSION
 
-# Outgoing TLS is more flexible because 1. not all receiving servers will
-# support TLS, 2. not all will have and up-to-date TLS stack.
-smtp_tls_security_level = may
+# By default, outgoing TLS is more flexible because
+# 1. not all receiving servers will support TLS,
+# 2. not all will have and up-to-date TLS stack.
+smtp_tls_security_level = {{ OUTBOUND_TLS_LEVEL|default('may') }}
 smtp_tls_mandatory_protocols = !SSLv2, !SSLv3
 smtp_tls_protocols =!SSLv2,!SSLv3
 smtp_tls_session_cache_database = btree:${data_directory}/smtp_scache
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -69,6 +69,10 @@ The ``RELAYHOST`` is an optional address of a mail server relaying all outgoing
 mail in following format: ``[HOST]:PORT``.
 ``RELAYUSER`` and ``RELAYPASSWORD`` can be used when authentication is needed.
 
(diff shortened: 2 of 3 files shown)


docs/configuration.rst
```diff
@@ -69,6 +69,10 @@ The ``RELAYHOST`` is an optional address of a mail server relaying all outgoing
 mail in following format: ``[HOST]:PORT``.
 ``RELAYUSER`` and ``RELAYPASSWORD`` can be used when authentication is needed.
 
+By default postfix uses "opportunistic TLS" for outbound mail. This can be changed
+by setting ``OUTBOUND_TLS_LEVEL`` to ``encrypt``. This setting is highly recommended
+if you are a relayhost that supports TLS.
```
muhlemmer, line 74: 1. Are `may` and `encrypt` the only options available? If there are other options, they need to be documented. Perhaps also make clear that the default value is set to `may`.

> This setting is highly recommended if you are a relayhost that supports TLS.

2. This wording is a bit strange. I would say "if you have a relayhost" or "if this is a relayhost".

core/postfix/conf/main.cf
```diff
@@ -52,9 +52,10 @@ tls_high_cipherlist = EDH+CAMELLIA:EDH+aRSA:EECDH+aRSA+AESGCM:EECDH+aRSA+SHA256:
 tls_preempt_cipherlist = yes
 tls_ssl_options = NO_COMPRESSION
 
-# Outgoing TLS is more flexible because 1. not all receiving servers will
-# support TLS, 2. not all will have and up-to-date TLS stack.
-smtp_tls_security_level = may
+# By default, outgoing TLS is more flexible because
+# 1. not all receiving servers will support TLS,
+# 2. not all will have and up-to-date TLS stack.
+smtp_tls_security_level = {{ OUTBOUND_TLS_LEVEL|default('may') }}
```
muhlemmer, line 58: Personally I like this syntax with `default()`. However, I don't remember seeing usage of this in Mailu. So I have some doubts about consistency. Other opinions welcome.

muhlemmer: Disregard this comments, I like it and I'm going to use it myself :smile:

muhlemmer:
Note to self: tested Ok :heavy_check_mark:

docs/configuration.rst
micw: Right, should have been "if you are using a ralayhost" ;)

ofthesun9: >     1. Are `may` and `encrypt` the only options available? If there are other options, they need to be documented. Perhaps also make clear that the default value is set to `may`.
> 
Another options would be DANE, when #707  is fixed

muhlemmer: Okay, we should take care of documenting that issue when the time is there.

Nebukadneza: Additional options are documented on http://www.postfix.org/postconf.5.html#smtp_tls_security_level  … however, i think that `may` or `encrypt` (and later dane) make most sense for a typical simplistic mailu deployment. Thus, i would not bloat documentation here by listing all of them. For me, it’s fine like this.

Issue #1419, opened 2020-03-25
Disk running out of space — Clamav fills up tmp directory


deze333:
After a few weeks of running, I noticed that the machine was left with 0 free  disk space and webmail stopped working.

It turned out that the directory `/mailu/filter/tmp` was filled with **36Gb** worth of temp subdirs.

```bash
me@mail:/mailu/filter/tmp$ ls
clamav-0018c4535a19139a285faa4214c50140.tmp  clamav-4668d952f865be239b6eaaac306ba65a.tmp  clamav-90cf5d29bbf6909f8d1e1e0ec4334a0c.tmp  clamav-c1b53d9a3671bb3a5eae26b1de3cb858.tmp
clamav-01c82ec5360e5c92fbe5333a44960f48.tmp  clamav-4c21930e870693ba2546412f57a83b8b.tmp  clamav-916db5c82e8edae16f4793dc368ec066.tmp  clamav-c6e99945b8217f608290a34a51d1e1b5.tmp
clamav-036365503c0d16eeaffb7c8412c0aeff.tmp  clamav-4cbca85252a4d0b1823f5c2e304b8e8e.tmp  clamav-916f64c79c047e33685d795526437fb6.tmp  clamav-caf5c0ce146a669133ed14e4195b25fb.tmp
clamav-0fd8bfdfe87628ecc26ad7aeea027815.tmp  clamav-4e66dce56d150d22c80bc2bfe0f33aee.tmp  clamav-92786c21fcc5dd03687962efeef6eadc.tmp  clamav-cbb918cf5dedb74319a9ba57655b1bfb.tmp
clamav-102341556d6f9215d641d47b84256f1c.tmp  clamav-4ee18219c935e0156c61d13c0c5de6a3.tmp  clamav-93718f7086b04d73e0a817d8585bd3d6.tmp  clamav-cdc372fcc5916a8befccfbdabf94b270.tmp
clamav-112134187c5e0c286bc851001af3e39e.tmp  clamav-501d7d8db64c04d2bd1a0ab815b9cf31.tmp  clamav-93b5c4ba88d7dce437b2271dcb2714e5.tmp  clamav-d0ef6c2dd7aa16f30a66bff2b714a2d0.tmp
[...]

Issue #470, opened 2018-05-06
ClamAV using all my disk space


jbonnett92:
Hi,
ClamAV keep's using all my disk space, and there is no option in the .env to limit it.
![](https://i.imgur.com/3ouD93y.png)
It seems to be the temp files e.g. `/home/core/mailu/filters/clamav-0033b19aa8a68887f57a9126f168010d.tmp`

Anyone know how to limit the disk usage? Seen as though I have 1gb memory.

Thank you,
Jamie

viridian44:
The same problem is also occuring on my system.

kaiyou:
You could limit disk space in the compose configuration using ulimits I guess. The real issue here is that files should be cleaned up though. I'll investigate as I don't see the same issue.

flayks:
Same issue, I have tons of .tmp files from Clamav filters

>   264.1 MiB [##########] /clamav-a017b6824d619633b1d46a9026b3877a.tmp
>   259.3 MiB [######### ] /clamav-dbcb20187f29dfad673bc40081783f00.tmp
>   256.2 MiB [######### ] /clamav-a28a124a19ffd73780504eeb193c3a8d.tmp
>   247.9 MiB [######### ] /clamav-ca1b7ee4796f5b1cf08b4b433d569106.tmp
>   244.6 MiB [######### ] /clamav-fe6e51fb2c79a1e7400a6d1b69777b4d.tmp
>   242.5 MiB [######### ] /clamav-8e7e2d1f022516ee43531bfe3f5e7050.tmp
>   235.0 MiB [########  ] /clamav-3ad3e0f7da4783d6d4253fdba4a700ce.tmp
>   229.7 MiB [########  ] /clamav-87ac01f595385864e6e47aa5fd8744b8.tmp
>   229.4 MiB [########  ] /clamav-76961f4640c915c928648f87ccb405c5.tmp
>   225.0 MiB [########  ] /clamav-c278bd97ec95053012998202f334373c.tmp
>   223.2 MiB [########  ] /clamav-83af7e9227387273843bd51b57eb41df.tmp
>   220.4 MiB [########  ] /clamav-5259ad996415d545a9c96cb28a02c79a.tmp
>   220.0 MiB [########  ] /clamav-d6d49de527715fcb4045f7449f6167ec.tmp
>   218.4 MiB [########  ] /clamav-c10101aecf10d05dcd501588a928a811.tmp
[...]

PR #1482, opened 2020-05-03
ClamAV Debian based build

muhlemmer:
## What type of PR?

Hopefully a bugfix

## What does this PR do?

For two years, on and off, users have been plagued by disk fills by ClamAV / fresh-clam. A clear root cause was never found, but suspicions have been raised towards Alpine's Libc implementation.

This PR features ClamAV based on a Debian image.

I'm currently not affected by this bug and I'm unable to reproduce, like others on the Team. However, this is an experiment for the affected users. Once this PR is build by Bors / travis the images are available on Dockerhub. Users willing to test need to change the image name for the `anitvirus` service to:

````
  # Optional services
  antivirus:
    #image: ${DOCKER_ORG:-mailu}/${DOCKER_PREFIX:-}clamav:${MAILU_VERSION:-master}
    image: mailutest/clamav:pr-1482

````

Please report back to this PR if this solved the issues.

Note that only feedback from users with sufficient RAM are considered. Full Mailu stack + ClamAV needs 2GB. Space for the OS needs to be considered. So typically minimal 3GB total RAM. Or 2GB with sufficient [swap](https://wiki.archlinux.org/index.php/Swap) (will perform poorly, but prevents OOM kill).

### Related issue(s)
- #470 and #1419 

## Prerequistes

**Will be done if experiment succeeds and Team agrees move to Debian**

### DIFF
--- a/optional/clamav/Dockerfile
+++ b/optional/clamav/Dockerfile
@@ -1,11 +1,9 @@
-ARG DISTRO=alpine:3.11
+ARG DISTRO=debian:buster-slim
 FROM $DISTRO
-# python3 shared with most images
-RUN apk add --no-cache \
-    python3 py3-pip bash \
-  && pip3 install --upgrade pip
-# Image specific layers under this line
-RUN apk add --no-cache clamav rsyslog wget clamav-libunrar
+
+RUN apt-get update && apt-get install -y \
+  python3 clamav-daemon rsyslog netcat \
+  && rm -rf /var/lib/apt/lists
 
 COPY conf /etc/clamav
 COPY start.py /start.py


muhlemmer:
bors try

Teoxoy:
I've tried this image on a Debian 9 server with 2GB RAM with Mailu 1.7 and the issue still persists.

muhlemmer:
> I've tried this image on a Debian 9 server with 2GB RAM with Mailu 1.7 and the issue still persists.

Thanks for your feedback. As mentioned, 2GB of ram does not prove to be enough anymore. 3GB or more is advised.

If your server is only low traffic, you can also create a swap file of **at least** 1GB:

````
root@tester:~# dd if=/dev/zero of=/swapfile bs=1M count=1024
1024+0 records in
1024+0 records out
1073741824 bytes (1.1 GB, 1.0 GiB) copied, 1.32853 s, 808 MB/s
root@tester:~# chmod 0600 /swapfile
root@tester:~# mkswap /swapfile
Setting up swapspace version 1, size = 1024 MiB (1073737728 bytes)
no label, UUID=335ca5aa-19fb-4a93-89e8-45a24d6460f2
root@tester:~# swapon /swapfile
````
And in `/etc/fstab` add:

````
/swapfile       swap            swap    defaults 0 0
````

I'm running this setup on my test server without any issues. Note that swap space is not performant and is used by the kernel to offload pages in case of low memory conditions. In this case it helps to prevent Out Of Memory.

Teoxoy:
@muhlemmer I will give it a try. Thanks!

Teoxoy:
After I gave the os some swap space the mail server has been running just fine for the past week
@muhlemmer thanks for the tip!

Maybe this is worth mentioning somewhere in the docs.

Diman0:
We have another PR for updating the docs. So far we have evidence that this issue could occur when
- the host machine did not have at least 3G/4G memory and
- the host machine has no swap. We recommend at least 1GB swap.
- Clamav from 2019 or older is used. (Mailu 1.7 or older). 
- We only have reported issues on Mailu 1.7 and older.

We suspect the issue has been addressed since 2019 in Clamav and will close this PR. The documentation will be updated with the higher memory requirements.

PR #1488, opened 2020-05-04
Modify postfix Dockerfile to use debian as base image

ofthesun9:
## What type of PR?
To check feasibility to use debian-slim as base image

## What does this PR do?
Modify Dockerfile to use debian as base image for the postfix service

### Related issue(s)
- Would allow to have DANE working, see #707 #1478 

I do not consider 707 as closed yet because missing fragment and documentation for the variable to be added as per 1478. If we go toward this debian image, I will add those pieces


ofthesun9:
bors try

PR #1491, opened 2020-05-05
Modify dovecot Dockerfile to use debian as base image

ofthesun9:
## What type of PR?
To check feasibility to use debian-slim as base image

## What does this PR do?
Modify Dockerfile to use debian as base image for the imap service

### Related issue(s)
-None, this change would ensure (layer) communality with other images


ofthesun9:
Travis having IO ERROR with alpine repository
bors try

ofthesun9:
bors try

PR #1501, opened 2020-05-12
In setup/flavor, change DMARC RUA and RUF email default settings

### LINKED ISSUE #1463: When changing POSTMASTER in setup.mailu.io, DMARC_RUA and DMARC_RUF are not updated
For example, if in the web ui for generating mailu.env I change the POSTMASTER from 'admin' to 'jay', I get:

POSTMASTER=jay
DMARC_RUA=admin
DMARC_RUF=admin

I expected to get:

POSTMASTER=jay
DMARC_RUA=jay
DMARC_RUF=jay

ofthesun9:
## What type of PR?
bug-fix

## What does this PR do?
This PR changes the default value used to set DMARC_RUA and DMARC_RUF:
DMARC_RUA and DMARC_RUF defaults will reuse the value defined for POSTMASTER,
instead of 'admin' as previously.
Please note that the setup tool doesn't allow (yet?) to define dmarc_rua nor dmarc_ruf, so the default value is indeed used for the time being.

### Related issue(s)
closes #1463 

## Prerequistes

### DIFF
--- a/setup/flavors/compose/mailu.env
+++ b/setup/flavors/compose/mailu.env
@@ -76,8 +76,8 @@ FETCHMAIL_DELAY={{ fetchmail_delay or '600' }}
 RECIPIENT_DELIMITER={{ recipient_delimiter or '+' }}
 
 # DMARC rua and ruf email
-DMARC_RUA={{ dmarc_rua or 'admin' }}
-DMARC_RUF={{ dmarc_ruf or 'admin' }}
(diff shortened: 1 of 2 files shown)


liquidat:
Makes a lot of sense. At least in the docker compose installation otherwise you will get dmarc reports to admin@.... which are then not accepted because there is no admin mail address.

PR #1511, opened 2020-05-17
Setup, compose: add note about reverse DNS

liquidat:
## What type of PR?

documentation

## What does this PR do?

If the [mailu setup utility](https://setup.mailu.io/master/) is used with the Compose flavor, in step 4 a list of public host names needs to be provided:
Screenshot from 2020-05-17 01-54-06
The [first entry in this list is configured as hostname by postfix](https://github.com/Mailu/Mailu/blob/ddac2672fcdb7b47b468d314daeb8e2ee3db8a85/core/postfix/conf/main.cf#L10). That is crucial: if the reverse DNS entry for that host is not the same as this first entry in this list, it will be a mismatch between rDNS and HELO, which will lead to bad scoring in many spam filters.

This commit clarifies that point.
It also fixes a missing parenthesis.

### Related issue(s)

- (none)

## Prerequistes

- (none applicable)


Nebukadneza:
bors try

PR #1530, opened 2020-06-06
Update nginx.conf

m121212:
Moving overrides/*.conf in the http section and adding server_name to mail server.

What type of PR?
Enhancement

What does this PR do?
Since "front" is de facto a proxy server, by simply moving the overrides .conf in the http section, the developer has much more flexibility in the integration in custom web servers. Furthermore, the main logic listen exlusively on server_name mail

Related issue(s)
Mailu#1528


m121212:
Let me know if you fan update the docker image with this pull, in order to use the image in production.

PR #1538, opened 2020-06-14
Introduce environment variable to control dovecot full-text-search

tremlin:
## What type of PR?

Enhancement

## What does this PR do?

In #1320 a full-text-search feature was enabled in Dovecot by default. Since this can have a big impact on performance, I think it's preferable to offer an option to disable the feature if it is not needed. This PR doesn't change the default behavior (FTS on).

### Related issue(s)
- #1320

## Prerequistes

- [ ] In case of feature or enhancement: documentation updated accordinagly

### DIFF
--- a/core/dovecot/conf/dovecot.conf
+++ b/core/dovecot/conf/dovecot.conf
@@ -21,7 +21,10 @@ mail_access_groups = mail
 maildir_stat_dirs = yes
 mailbox_list_index = yes
 mail_vsize_bg_after_count = 100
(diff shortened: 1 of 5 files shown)


ofthesun9:
Thanks for your PR.
Could you please amend it to add a description in the advanced settings in the various .env templates:
- docs/compose/.env
- setup/flavors/compose/mailu.env

tremlin:
Thanks for your suggestion. I tried to add some basic documentation.

PR #1541, opened 2020-06-16
alpine 3.12 as base image

ofthesun9:
## What type of PR?
Keep in sync with alpine devel
alpine 3.12 was [released](https://alpinelinux.org/posts/Alpine-3.12.0-released.html) on  29 May 2020.
Update for the main services:
nginx 1.18 (vs 1.16)
dovecot 2.3.10 (unchanged)
postfix 3.5.2 (vs 3.4.7)
rspamd 2.5 (vs 1.9.4)
clamav 0.102/3 (vs 0.101.5)
fetchmail 6.4.5
 
## What does this PR do?
Mostly Dockerfile modification

### DIFF
--- a/optional/postgresql/Dockerfile
+++ b/optional/postgresql/Dockerfile
@@ -1,4 +1,4 @@
-ARG DISTRO=alpine:3.10
+ARG DISTRO=alpine:3.12
 FROM $DISTRO
 # python3 shared with most images
 RUN apk add --no-cache \
@@ -6,7 +6,7 @@ RUN apk add --no-cache \
   && pip3 install --upgrade pip
 
 # Shared layer between nginx, dovecot, postfix, postgresql, rspamd, unbound, rainloop, roundcube
-RUN pip3 install socrate
+RUN pip3 install socrate==0.2.0
 
 # Image specific layers under this line
 RUN apk add --no-cache \
--- a/core/admin/Dockerfile
+++ b/core/admin/Dockerfile
@@ -1,5 +1,5 @@
 # First stage to build assets
-ARG DISTRO=alpine:3.10
+ARG DISTRO=alpine:3.12
 ARG ARCH=""
 FROM ${ARCH}node:8 as assets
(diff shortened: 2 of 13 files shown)


optional/postgresql/Dockerfile
```diff
@@ -1,12 +1,12 @@
-ARG DISTRO=alpine:3.10
+ARG DISTRO=alpine:3.12
```
dhoppe, line 1: The upgrade from Alpine 3.10 to 3.12 resulted in an upgrade from PostgreSQL 11.7 to 12.3: 
* https://alpine.pkgs.org/3.10/alpine-main-x86_64/postgresql-11.7-r0.apk.html
* https://alpine.pkgs.org/3.12/alpine-main-x86_64/postgresql-12.3-r2.apk.html

Existing installations require a manual upgrade of the PostgreSQL instance: 
* https://www.postgresql.org/docs/current/pgupgrade.html

The following project has helped me a lot:
* https://github.com/tianon/docker-postgres-upgrade
* https://github.com/tianon/docker-postgres-upgrade/tree/master/11-to-12

For example, I have executed the following commands:
```
docker-compose -p mailu down
mv /data/mailu/psql_db /data/mailu/psql_tmp
mkdir /data/mailu/psql_db
chown 70:70 /data/mailu/psql_db
docker run --rm \
	-v /data/mailu/psql_tmp:/var/lib/postgresql/11/data \
	-v /data/mailu/psql_db:/var/lib/postgresql/12/data \
	tianon/postgres-upgrade:11-to-12
docker-compose -p mailu up -d
rm -rf /data/mailu/psql_tmp
```

Maybe we should document this process somewhere?

/cc @liquidat

HorayNarea: If I remember correctly, last time we upgraded the alpine base image, we discussed this to and documented that `administrator action is needed` in the release notes of the next version (which was 1.7 probably? :D), we could copy that to the release notes of 1.8 when it's done

PR #1553, opened 2020-06-26
Added FAQ about SMTP Banner

### LINKED ISSUE #1368: SMTP Banner missing
Hi i have two question reagrding the `main.cf` file.

- How can i add the `smtpd_banner = $myhostname ESMTP $mail_name (Debian/GNU)` 
- When i edit the `HOSTNAME` in `mailu.env` the entry `myhostname` is still the old one. Can i trigger the creation of that file. [1] I tried `docker container prune` and this files is also not in `/mailu` 

[1] 
```
foo:/mailu$ docker-compose exec smtp bash -c "cat /etc/postfix/main.cf | grep myhost"
myhostname = mail.foo.de
foo:/mailu$ cat mailu.env | grep HOS
HOSTNAMES=mail.foo.de,foo-de.foo.de
```

arother:
## What type of PR?
documentation

## What does this PR do?
Adding FAQ about SMTP Banner

### Related issue(s)
- closes #1368 

## Prerequistes

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -253,7 +253,7 @@ correct syntax. The following file names will be taken as override configuration
 - `Nginx`_ - All ``*.conf`` files in the ``nginx`` sub-directory;
 - `Rspamd`_ - All files in the ``rspamd`` sub-directory.
 
-*Issue reference:* `206`_.
+*Issue reference:* `206`_, `1368`_.
 
 I want to integrate Nextcloud 15 (and newer) with Mailu
 ```````````````````````````````````````````````````````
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -353,6 +353,15 @@ down and up again. A container restart is not sufficient.
 
 *Issue reference:* `615`_.
 
+SMTP Banner from overrides/postfix.cf is ignored
+````````````````````````````````````````````````
+
+Any mail related connection is proxied by nginx. Therefore the SMTP Banner is also set by nginx. Overwriting in overrides/postfix.cf does not apply. If you want to change the hostname in the SMTP Banner greeting, you must change the first entry of HOSTNAMES in mailu.env.
```
HorayNarea, line 359:
```suggestion
Any mail related connection is proxied by nginx. Therefore the SMTP Banner is also set by nginx. Overwriting in overrides/postfix.cf does not apply.
```

I think it would be better not to include anything about changing hostnames here, since that could lead to more errors when people try weird things without asking for help first :D

PR #1555, opened 2020-06-27
Fix 1552 configurable folder names

### LINKED ISSUE #1552: Is it possible to change the name of the Junk folder?
In https://github.com/Mailu/Mailu/pull/1470, it states that Mailu will learn messages as spam if they are moved to the `Junk` folder, and learn them as ham if they are moved out of it.

However, I do not call the folder `Junk` — I call it `@Spam`, as I don't like having a system folder appearing in the middle of the folder list. Is there a way to tell Mailu to use this folder instead?

unikitty37:
## What type of PR?

Enhancement

## What does this PR do?

Allows the default IMAP folder names to be configured for users who prefer different names, or who are using software which has fixed ideas about what those names should be.

As most people will be able to go with the defaults, there doesn't seem to be much need to complicate the setup generator, so this PR just adds four unset variables to `mailu.env`. Setting one of them will override the values placed in the generated `dovecot.conf`, both when assigning the special folder names and, when the Junk folder is overridden, in the monitoring rules generated.

I'm not 100% happy about the `user.spam_folder` as it's really a server setting — if anybody can suggest a better place to put that, I can move it.

### Related issue(s)
closes #1552 

## Prerequisites

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -25,3 +25,4 @@ Other contributors:
  - [Tim Mohlmann](https://github.com/muhlemmer) - [Contributions](https://github.com/Mailu/Mailu/commits?author=muhlemmer)
  - [Ionut Filip](https://github.com/ionutfilip) - [Contributions](https://github.com/Mailu/Mailu/commits?author=ionutfilip)
  - [Ichikawa Yuriko](https://github.com/IchikawaYukko) - [Contributions](https://github.com/Mailu/Mailu/commits?author=IchikawaYukko) Japanese translation
+ - [John Yeates](https://github.com/unikitty37) - [Contributions](https://github.com/Mailu/Mailu/commits?author=unikitty37)
--- a/core/admin/mailu/internal/templates/default.sieve
+++ b/core/admin/mailu/internal/templates/default.sieve
@@ -22,7 +22,7 @@ if header :index 2 :matches "Received" "from * by * for <*>; *"
 if spamtest :percent :value "gt" :comparator "i;ascii-numeric"  "{{ user.spam_threshold }}"
 {
   setflag "\\seen";
-  fileinto :create "Junk";
+  fileinto :create "{{ user.junk_folder }}";
   stop;
 }
 {% endif %}
--- a/core/admin/mailu/models.py
+++ b/core/admin/mailu/models.py
@@ -334,6 +334,7 @@ class User(Base, Email):
     displayed_name = db.Column(db.String(160), nullable=False, default="")
     spam_enabled = db.Column(db.Boolean(), nullable=False, default=True)
     spam_threshold = db.Column(db.Integer(), nullable=False, default=80)
+    spam_folder = os.environ.get('JUNK_FOLDER', 'Junk')
 
     # Flask-login attributes
     is_authenticated = True
--- a/core/dovecot/conf/dovecot.conf
+++ b/core/dovecot/conf/dovecot.conf
@@ -26,12 +26,23 @@ default_vsz_limit = 2GB
 
 namespace inbox {
   inbox = yes
-  {% for mailbox in ("Trash", "Drafts", "Sent", "Junk") %}
-  mailbox {{ mailbox }} {
+  mailbox "{{ TRASH_FOLDER | default('Trash') }}" {
     auto = subscribe
-    special_use = \{{ mailbox }}
+    special_use = \Trash
   }
-  {% endfor %}
+  mailbox "{{ DRAFTS_FOLDER | default('Drafts') }}" {
+    auto = subscribe
+    special_use = \Drafts
+  }
+  mailbox "{{ SENT_FOLDER | default('Sent') }}" {
+    auto = subscribe
+    special_use = \Sent
+  }
+}
+  mailbox "{{ JUNK_FOLDER | default('Junk') }}" {
+    auto = subscribe
+    special_use = \Junk
+  }
 }
 
 plugin {
@@ -155,11 +166,11 @@ plugin {
   sieve_spamtest_max_value = 15
 
   # Learn from spam
-  imapsieve_mailbox1_name = Junk
+  imapsieve_mailbox1_name = {{ JUNK_FOLDER | default('Junk') }}
   imapsieve_mailbox1_causes = COPY
   imapsieve_mailbox1_before = file:/conf/report-spam.sieve
   imapsieve_mailbox2_name = *
-  imapsieve_mailbox2_from = Junk
+  imapsieve_mailbox2_from = {{ JUNK_FOLDER | default('Junk') }}
   imapsieve_mailbox2_causes = COPY
   imapsieve_mailbox2_before = file:/conf/report-ham.sieve
 }
--- a/core/dovecot/conf/dovecot.conf
+++ b/core/dovecot/conf/dovecot.conf
@@ -38,7 +38,6 @@ namespace inbox {
     auto = subscribe
     special_use = \Sent
   }
-}
   mailbox "{{ JUNK_FOLDER | default('Junk') }}" {
     auto = subscribe
     special_use = \Junk
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -83,6 +83,10 @@ later classify incoming mail based on the custom part.
 The ``DMARC_RUA`` and ``DMARC_RUF`` are DMARC protocol specific values. They hold
 the localpart for DMARC rua and ruf email addresses.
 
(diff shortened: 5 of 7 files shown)


Nebukadneza:
bors try

unikitty37:
bors try

Nebukadneza:
bors ping

Nebukadneza:
bors try-

Nebukadneza:
bors ping

Nebukadneza:
bors try

Nebukadneza:
It seems the build failed because some python dependency could not be downloaded from pypi … trying again *sigh*

bors try

Nebukadneza:
Hi again,

we’ve just discussed this feature, and found the change a little bit too invasive for the small steps we’re currently trying to take.

Also, we came up with another idea:
Looking at this documentation:
https://doc.dovecot.org/configuration_manual/config_file/config_file_syntax/
you can merge multiple inbox-namespaces into one. This should work when one of the namespaces is specified in the included override. Since you can also specify multiple user-visible names for the same special-use-folder, as mailcow does here:
https://github.com/mailcow/mailcow-dockerized/blob/master/data/conf/dovecot/dovecot.conf#L86
… this would mean that technically your use-case should be possible already today … right?

 If you can find the time to try this out, that would be great. If it turns out to really work like we envisioned currently … how would you feel about adding that to the documentation, instead?

Thanks a lot for your contribution, and sorry for the slow response~

unikitty37:
Thanks — unfortunately, this is what I'm already doing, and it doesn't work fully. There are two problem areas:

The first is https://github.com/Mailu/Mailu/blob/3a89c1e91096b06f05c3919d8496bc20b1e81982/core/admin/mailu/internal/templates/default.sieve#L25 which has the folder name `Junk` hardcoded, and uses the `:create` flag, so there is still a system folder listed in the middle of the regular folders. This rule runs before any user-defined rules, so the only way to filter spam is rather unintuitive: you must *disable* spam filtering in the admin area, then create a custom rule based on headers added by rspamd. This loses the ability to set a threshold from the admin area; you must now change the level at which rspamd changes from classifying as "Spam: No" to "Spam: Yes", and this is not a per-user change.

The second is https://github.com/Mailu/Mailu/blob/535b95bca75ea34190a036a34a9832805eec405d/core/dovecot/conf/dovecot.conf#L176 which runs `overrides/dovecot.conf` as an include, rather than a complete replacement for the system-provided ones. This means that the existing Junk folder is still created by https://github.com/Mailu/Mailu/blob/535b95bca75ea34190a036a34a9832805eec405d/core/dovecot/conf/dovecot.conf#L32 and is still given its special behaviour by https://github.com/Mailu/Mailu/blob/535b95bca75ea34190a036a34a9832805eec405d/core/dovecot/conf/dovecot.conf#L163.

These are the criteria I'm going by, and how they work with the overrides:

* ❌   A top-level folder called `@Spam`is created by Mailu, and new users are auto-subscribed to it.
* ❌   There is not a top-level folder called `Junk`. I can create it without it conflicting with predefined folders.
* ✔️  When rspamd classifies a mail as spam, it is delivered to the `@Spam` folder.
* ✔️  When I move an email that wasn't classified as spam to the `@Spam` folder, it is classified as spam.
* ✔️  When I move an email that was classified as spam to the `@Ham` folder, it is classified as ham.
* ❌   Moving messages to the user-created `Junk` folder doesn't change their spam classification.
* ❌   Moving messages from the user-created `Junk` folder doesn't change their spam classification.
* ❌   Moving messages from the `@Spam` folder to somewhere other than `@Ham` doesn't change their spam classification.
  
Here's the override file I used to get this:

```
namespace inbox {
  inbox = yes
  mailbox "Trash" {
    auto = subscribe
    special_use = \Trash
  }
  mailbox "Drafts" {
    auto = subscribe
    special_use = \Drafts
  }
  mailbox "Sent" {
    auto = subscribe
    special_use = \Sent
  }
  mailbox "@Spam" {
    auto = subscribe
    special_use = \Junk
  }
}

plugin {
  sieve = file:~/sieve;active=~/.dovecot.sieve
  sieve_before = dict:proxy:/tmp/podop.socket:sieve
  sieve_plugins = sieve_imapsieve sieve_extprograms
  sieve_extensions = +spamtest +spamtestplus +editheader
  sieve_global_extensions = +vnd.dovecot.execute

  # Sieve execute
  sieve_execute_bin_dir = /conf/bin

  # Send vacation replies even for aliases
  # See the Pigeonhole documentation about warnings: http://wiki2.dovecot.org/Pigeonhole/Sieve/Extensions/Vacation
  # It appears that our implemntation of mail delivery meets criteria of section 4.5
  # from RFC 5230 and that disabling the recipient checks is not an issue here.
  sieve_vacation_dont_check_recipient = yes

  # Include the recipient in vacation replies so that DKIM applies
  sieve_vacation_send_from_recipient = yes

  # extract spam score from headers
  sieve_spamtest_status_type = strlen
  sieve_spamtest_status_header = X-Spam-Level
  sieve_spamtest_max_value = 15

  # Learn from spam
  imapsieve_mailbox1_name = @Spam
  imapsieve_mailbox1_causes = COPY
  imapsieve_mailbox1_before = file:/conf/report-spam.sieve
  imapsieve_mailbox2_name = *
  # Learn ham
  imapsieve_mailbox3_name = @Ham
  imapsieve_mailbox3_from = *
  imapsieve_mailbox3_causes = COPY
  imapsieve_mailbox3_before = file:/conf/report-ham.sieve
}
```

I can't see how to get around the points of failure using an override, given that the override file can't undo some of the things that has been done by the built-in one.

I realise you're trying to only make small changes at the moment, but this feels like it is a small change to me — it doesn't change default behaviour, and only adds four environment variables to the config.

Admittedly, I still don't know if it causes any tests to fail, as I've never managed to get bors to complete a run 😐

HorayNarea:
> The first is
> 
> https://github.com/Mailu/Mailu/blob/3a89c1e91096b06f05c3919d8496bc20b1e81982/core/admin/mailu/internal/templates/default.sieve#L25
> which has the folder name Junk hardcoded, and uses the :create flag, so there is still a system folder listed in the middle of the regular folders.

Hm interesting… now I wonder why a very similar setup works in mailcow as they also use the name "Junk" in their global sieve filter: https://github.com/mailcow/mailcow-dockerized/blob/master/data/conf/dovecot/global_sieve_after#L11 but without `:create`, So I'm not sure how this folder get's created initially… and if it would work in mailu if we just remove `:create`?

> This rule runs before any user-defined rules, so the only way to filter spam is rather unintuitive: …

Would it help if we allowed an override before our hardcoded sieve-rules?

unikitty37:
> Hm interesting… now I wonder why a very similar setup works in mailcow as they also use the name "Junk" in their global sieve filter: https://github.com/mailcow/mailcow-dockerized/blob/master/data/conf/dovecot/global_sieve_after#L11 but without `:create`, So I'm not sure how this folder get's created initially… and if it would work in mailu if we just remove `:create`?

I think the `dovecot.conf` entry causes it to get created anyway, because I've got `user.spam_enabled` set to off so I can use my own custom rule, which means that rule never gets called — and yet, if I delete the Junk folder, it reappears when Mailu is restarted.

I'm not sure what would happen if that rule got triggered when the folder didn't exist, though — would it cause the message to get dropped, or just passed on to the next rule?

I also tried to create a personal copy of the default rule but couldn't see a way to get over the fact that `user.spam_threshold` is inserted using the templating engine here https://github.com/Mailu/Mailu/blob/575f6b169133a925c0d51048e2f2d2d76fe74814/core/admin/mailu/internal/views/dovecot.py#L48

I'm assuming that rebuilds the user's default Sieve rules each time something's changed in that UI. But it means you can't use that threshold in your own rule, because that won't get rewritten — you have to configure spam rules yourself and mentally slap a "Do Not Touch" label on that section of the admin interface.
 
> > This rule runs before any user-defined rules, so the only way to filter spam is rather unintuitive: …
> 
> Would it help if we allowed an override before our hardcoded sieve-rules?

As long as the user-defined rule had a `stop` in it, I guess so. It would definitely be worth mentioning that in the docs 😀 

However, the inaccessibility of `user.spam_threshold` means you would still lose the use of the spam threshold setting.

nextgens:
IMHO this is the wrong approach: aliases in an override would be much better/simpler.

https://github.com/Mailu/Mailu/issues/1552#issuecomment-965676196

Diman0:
After discussing this PR during the last mailu meeting we have decided to close it. It  adds too much complexity with not enough gain for the project. This can be achieved via overrides for dovecot as mentioned by Nextgens.

Issue #1167, opened 2019-09-11
Documentation request: Spam filter


boldt:
The whole documentation on https://mailu.io/ does not describe, how the spam filter works or can be used. Maybe it's because I uses SA the last decade.

* Are there any (false positive) mails dropped into the Junk folder by rspamd?
* Does rspamd lean from mails I put into the Junk folder?

My observation: After some days in action, the rspam backend shows only  `no action` or `reject`.

kaiyou:
Here are some items, feel free to create a PR that adds some of them to the documentation if you have the time. Most of the lack of documentation is caused by a lack of time.

- rspamd rejects non-compliant messages and viruses (the latter being under discussion) only
 - rspamd adds antispam headers to messages
 - dovecot is then responsible of classifying messages to the Junk folder based on user preferences, especially the spam tolerance which is basically a threshold on the spam score
 - the logic in dovecot is implemented via sieve scripts loaded from the admin API.

aksdb:
So it would make sense (for every user) to add a sieve rule to move depending on the number of "*" in the `X-Spam-Level` header, right?

However: I would think the obvious target for these messages would be the "Junk" folder. But wouldn't that immediately retrigger the learning and therefore constantly reinforcing spam learnings, thereby increasing the spam likelyhood everytime you receive a already spam-classified mail?

liquidat:
@kaiyou Thanks for the details. I am happy to do a documentation PR to add this to the documentation!
Only one question left for me to fully understand this:

> the logic in dovecot is implemented via sieve scripts loaded from the admin API

How do we modify those Sieve scripts? Via an imap client connecting to dovecot, or from within the admin interface? I haven't seen an option to do that in the admin interface, but might be too blind to spot it.

Diman0:
> @kaiyou Thanks for the details. I am happy to do a documentation PR to add this to the documentation!
> Only one question left for me to fully understand this:
> 
> > the logic in dovecot is implemented via sieve scripts loaded from the admin API
> 
> How do we modify those Sieve scripts? Via an imap client connecting to dovecot, or from within the admin interface? I haven't seen an option to do that in the admin interface, but might be too blind to spot it.

The sieve script is hardcoded in  Mailu/core/admin/mailu/internal/templates/default.sieve (in the git project). 
The relevant sieve part is:
`if spamtest :percent :value "gt" :comparator "i;ascii-numeric"  "{{ user.spam_threshold }}"`
[...]

Issue #1566, opened 2020-07-11
Help please blocking spam


svaldivia2:
Hello, I'm new using rspamd and want to know if anyone knows how to block some domain.

There is a domain (let it be "spamdomain.com") that is spamming constantly, and I don't know how to stop it.

I'm using WebUI, and I was reading in your docs https://rspamd.com/doc/modules/multimap.html that multimap and prefilters would help with this, but I'm super noob and don't know where to put it, neither the syntax (I tried to understand the docs, but it was confusing to me). I suspect that the filter that I need is "email:domain", right?

Anyone can give me a hand with this?

Many thanks in advance

Diman0:
This can be achieved by adding additional symbols to the multimap filter. In this symbol you could indeed reject email using a prefilter.

Step 1 is creating the config file to add the symbol to the multimap filter.
Create the file /mailu/overrides/rspamd/multimap.conf  with contents:
```
#override.d/multimap.conf
local_bl_domain {
  type = "from";
  filter = "email:domain";
  map = "/etc/rspamd/override.d/blacklist.inc";
  score = 14;
  description = "Senders domain part is on the local blacklist";
  group = "local_bl";
  action = "reject";
}
```
Note that the prefilter "  action = "reject";" instructs rspamd to reject the email. The sender will receive a failed delivery notification email that this email was rejected. Another option is too use discord (accept the email and discard it). If you remove "  action = "reject";", then the email will get score 14 and will be moved to the spam folder depending on the spam score threshold setting of the user.

Step 2 is to create the blacklist file that will contain the domains to be blocked. You don't have to add the domains yet. This can also be done via the webgui.
Create the file  /mailu/overrides/rspamd/blacklist.inc with contents
```
 #blacklisted domains. ALL emails from these domains are blacklisted!!
 #this file is LIVE reloaded by rspamd. Any changes are EFFECTIVE IMMEDIATELY.
spamdomain.com


```

Step 3 is to restart rspamd to pick up the new multimap.conf file. If you use docker-compose:
```
docker-compose scale antispam=0
docker-compose scale antispam=1
```

Step 4 is too access the webui of rspamd and go to configuration. You should listed here: 
`"/etc/rspamd/override.d/blacklist.inc	Senders domain part is on the local blacklist". `
You can click this file to add new domains. Changes are persistent and effective immediately. When you edit the inc files in the overrides folder, the changes will also be effective immediately.

I hope this helps. I plan to create a PR to add this to the FAQ section

[...]

PR #1575, opened 2020-07-24
Added FAQ on how spam filtering works and how to create a local black…

Diman0:
…list for sender domain.

## What type of PR?

documentation

## What does this PR do?

Adds FAQ section 'Spam filtering' that describes how spam filtering works. It also describes how you can create your own local black list. This enables the user to set up a map that can be live-updated via Rspamd webgui for blocking emails from listed sender domains.

### Related issue(s)
- #1167 
- #1566 
- Closes issue #1167 

## Prerequistes

- [ N/A] In case of feature or enhancement: documentation updated accordingly
- [ Done] Unless it's docs or a minor change: add [changelog](https://mailu.io/master/contributors/guide.html#changelog) entry file.

### DIFF
--- a/core/admin/Dockerfile
+++ b/core/admin/Dockerfile
@@ -1,5 +1,5 @@
 # First stage to build assets
-ARG DISTRO=alpine:3.10
+ARG DISTRO=alpine:3.12
 ARG ARCH=""
 FROM ${ARCH}node:8 as assets
 COPY --from=balenalib/rpi-alpine:3.10 /usr/bin/qemu-arm-static /usr/bin/qemu-arm-static
--- a/core/admin/requirements-prod.txt
+++ b/core/admin/requirements-prod.txt
@@ -38,7 +38,8 @@ python-editor==1.0.4
 pytz==2019.1
 PyYAML==5.1
 redis==3.2.1
-six==1.12.0
+#alpine3:12 provides six==1.15.0
+#six==1.12.0
 socrate==0.1.1
 SQLAlchemy==1.3.3
 srslib==0.1.4
@@ -46,6 +47,6 @@ tabulate==0.8.3
 tenacity==5.0.4
 validators==0.12.6
 visitor==0.1.3
-Werkzeug==0.15.3
+Werkzeug==0.15.5
 WTForms==2.2.1
 WTForms-Components==0.10.4
--- a/core/dovecot/Dockerfile
+++ b/core/dovecot/Dockerfile
@@ -1,4 +1,4 @@
-ARG DISTRO=alpine:3.10
+ARG DISTRO=alpine:3.12
 FROM $DISTRO as builder
 WORKDIR /tmp
 RUN apk add git build-base automake autoconf libtool dovecot-dev xapian-core-dev icu-dev
--- a/core/nginx/Dockerfile
+++ b/core/nginx/Dockerfile
@@ -1,4 +1,4 @@
-ARG DISTRO=alpine:3.10
+ARG DISTRO=alpine:3.12
 FROM $DISTRO
 # python3 shared with most images
 RUN apk add --no-cache \
--- a/core/none/Dockerfile
+++ b/core/none/Dockerfile
@@ -1,6 +1,6 @@
 # This is an idle image to dynamically replace any component if disabled.
 
-ARG DISTRO=alpine:3.10
+ARG DISTRO=alpine:3.12
 FROM $DISTRO
 
 CMD sleep 1000000d
--- a/core/postfix/Dockerfile
+++ b/core/postfix/Dockerfile
@@ -1,4 +1,4 @@
-ARG DISTRO=alpine:3.10
+ARG DISTRO=alpine:3.12
 FROM $DISTRO
 # python3 shared with most images
 RUN apk add --no-cache \
--- a/core/rspamd/Dockerfile
+++ b/core/rspamd/Dockerfile
@@ -1,4 +1,4 @@
-ARG DISTRO=alpine:3.10
+ARG DISTRO=alpine:3.12
 FROM $DISTRO
 # python3 shared with most images
 RUN apk add --no-cache \
--- a/optional/clamav/Dockerfile
+++ b/optional/clamav/Dockerfile
@@ -1,4 +1,4 @@
-ARG DISTRO=alpine:3.11
+ARG DISTRO=alpine:3.12
 FROM $DISTRO
 # python3 shared with most images
 RUN apk add --no-cache \
--- a/optional/fetchmail/Dockerfile
+++ b/optional/fetchmail/Dockerfile
@@ -1,17 +1,4 @@
-# First stage: Build
-ARG DISTRO=alpine:3.10
-FROM $DISTRO as builder
-
-# build dependencies
-RUN apk add --no-cache curl tar xz autoconf git gettext build-base openssl openssl-dev
-
-RUN curl -L 'https://sourceforge.net/projects/fetchmail/files/branch_7-alpha/fetchmail-7.0.0-alpha6.tar.xz/download' | tar xJ
-RUN cd fetchmail-7.0.0-alpha6 && \
-    sed -i -e 's/SSLv3_client_method/SSLv23_client_method/' socket.c && \
-    ./configure --with-ssl --prefix /usr/local --disable-nls && \
-    make
-
-ARG DISTRO=alpine:3.10
+ARG DISTRO=alpine:3.12
 FROM $DISTRO
 
 # python3 shared with most images
@@ -20,13 +7,11 @@ RUN apk add --no-cache \
   && pip3 install --upgrade pip
 
 # Image specific layers under this line
-RUN apk add --no-cache ca-certificates openssl \
+RUN apk add --no-cache fetchmail ca-certificates openssl \
  && pip3 install requests
 
-COPY --from=builder /fetchmail-7.0.0-alpha6/fetchmail /usr/local/bin
 COPY fetchmail.py /fetchmail.py
 
-RUN adduser -D fetchmail
 USER fetchmail
 
 CMD ["/fetchmail.py"]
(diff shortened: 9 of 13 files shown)


Diman0:
This is the first time I have contributed to an open source project. I'm not very experienced with / pull request procedure. If I made any obvious mistakes, please enlighten me.

Nebukadneza:
Sorry for the late reply. This is great! Marking it for inclusion to 1.8-rc … but I need find some time to actually proof-read it. 

bors try

Nebukadneza:
Hi,

thanks again for your contribution and great explanations. We just discussed this PR, and we really want that part of the documentation included, but fear that it bloats the FAQ a bit.

How would you feel about introducing a new page under the "Administration" section in the documentation, instead of the FAQ. For this, you would have to create a new file, `antispam.rst`, and include that in the `toctree` under "Administration" in `index.rst`.

That change to the PR would be greatly appeciated.

Thanks a lot!

Diman0:
Hi,

I understand it is a bit large for the FAQ section. I'm fine with moving this to a new page under the Administration section. I'll look into this later this week and will update the pull request.

docs/faq.rst
```diff
@@ -334,6 +334,145 @@ After successfull login the domain part will be striped and the rest used as use
...
+
+* Email messages with a score of 6 or higher will receive a header with the spam score.
+
+* Email messages with a score of 4 or higher will be greylisted.
+
+* Email messages with a score of 3 or lower will be accepted.
+
+1. When an email message is received or send by Postfix, it is scanned by Rspamd. If the email message receives a spam score between 6 and 15, a header is added to the email message with the spam score. This is an example of an email header for a spam score of 14:
+
+  ``X-Spam-Level: **************``
```
liquidat, line 357: So when I check my mails, the headers look like this:

```
X-Spamd-Bar: ++++++
X-Spam-Level: ******
Authentication-Results: lisa.bayz.de;
	dkim=pass header.d=example.com header.s=mailing header.b=ABCDE;
	dkim=pass header.d=example.com header.s=mailing header.b=ABCDE;
	dmarc=pass (policy=none) header.from=eventim.de;
	spf=pass (lisa.bayz.de: domain of return@example.com designates 11.22.33.44 as permitted sender) smtp.mailfrom=return@example.com
X-Spam: Yes
```

How about you include them as well in your FAQ?

Diman0: That is a good idea. I did not originally  include the full header because I didn't want to bloat the FAQ entry to much. That is not a problem anymore since I'm moving everything to a new page.

docs/faq.rst
```diff
@@ -334,6 +334,145 @@ After successfull login the domain part will be striped and the rest used as use
...
+     
+     If <spam_score in percent> is greater than <spam filter tolerance>, then move the email message to the spam folder and mark the email message as read.
+   
+   For example if the user defined spam filter tolerance is set to 80%(default) and the spam score of an email message is 10:
+      
+   66% (10/15) is less than 80%, so the email is classified as ham. This email message will go to the inbox folder. If the user wants email messages with a score of 10 (66%) to be classified as spam, then the user defined spam filter tolerance can be lowered to 65% in the administration web interface.
+   
+.. image:: assets/screenshots/SpamFiltering.png
+
+The location in the administration web interface where the spam filter and spam filter tolerance can be configured.
```
liquidat, line 375: If we keep this as a simple FAQ, maybe that is too much?
If you want to include it, please provide either a link to more documentation which describes how this can be configured, or shortly summarize it here.

docs/faq.rst
```diff
@@ -334,6 +334,145 @@ After successfull login the domain part will be striped and the rest used as use
...
+
+*Issue reference:* `1167`_.
+
+Can I learn ham/spam messages from an already existing mailbox?
+```````````````````````````````````````````````````````````````
+
+Mailu is supporting automatic spam learning for messages moved to the Junk mailbox. Any email moved from the Junk Folder will be learnt as ham.
```
liquidat, line 382: Rephrase and add like this: "Any email moved to the Junk Folder will be re-learned as Spam, any email moved from the Junk Folder to any other folder will be re-learned as ham."

docs/faq.rst
```diff
@@ -334,6 +334,145 @@ After successfull login the domain part will be striped and the rest used as use
...
+*Issue reference:* `1438`_.
+
+How can I block emails from a domain?
+`````````````````````````````````````
+
+Via the multimap filter it is possible to block emails from a sender domain. See the `official rspamd documentation`_ for more information. A local blacklist that contains the domains to be blocked can be configured via the Rspamd overrides folder. 
+
+The following steps have to be taken to configure an additional symbol (rule) that uses the multimap filter to block emails from sender domain.
+
+1. In the overrides folder create a configuration file for the multimap filter. This configuration is included by Rspamd in the main multimap configuration file. This means you do not have to use the "multimap {}" element. Files in the /mailu/overrides/rspamd/ folder folder are mapped to /etc/rspamd/override.d.
```
liquidat, line 407: "folder folder"

liquidat: Also, please link to the mailu overrides documentation to make more clear what this is about.

docs/faq.rst
```diff
@@ -334,6 +334,145 @@ After successfull login the domain part will be striped and the rest used as use
...
+      description = "Senders domain part is on the local blacklist";
+      group = "local_bl";
+      action = "reject";
+    }
+
+   Note the "action = "reject";" line. This is a so-called pre-filter. If you omit this line, then the configured score will be added to the total score of the email message. Depending on the end-score after processing all rules, a verdict is made. To override this, you can add the action line. When this symbol (rule) is fired, then this action is immediately taken and no further processing occurs. You can use the following actions:
+  
+   * discard: drop an email message, but return success for sender (should be used merely in special cases)
+
+   * reject: reject the email message
```
liquidat, line 428: Add something here like ", this enables the actual blocking of mails from the domain like we intended here"

docs/faq.rst
```diff
@@ -334,6 +334,145 @@ After successfull login the domain part will be striped and the rest used as use
...
+
+   * add header: add email header to indicate spam
+
+   * no action: allow message
```
liquidat, line 432: Question: does this still add the score? If so, please clarify this here.

docs/faq.rst
```diff
@@ -334,6 +334,145 @@ After successfull login the domain part will be striped and the rest used as use
...
+      action = "reject";
+    }
+
+   Note the "action = "reject";" line. This is a so-called pre-filter. If you omit this line, then the configured score will be added to the total score of the email message. Depending on the end-score after processing all rules, a verdict is made. To override this, you can add the action line. When this symbol (rule) is fired, then this action is immediately taken and no further processing occurs. You can use the following actions:
+  
+   * discard: drop an email message, but return success for sender (should be used merely in special cases)
+
+   * reject: reject the email message
+
+   * add header: add email header to indicate spam
```
liquidat, line 430: How does this work? Does this enforce a header no matter what the final score is?
Or don't we get a header without this at all?
I thought that headers are added according to a final score after all rules.

docs/faq.rst
```diff
@@ -334,6 +334,145 @@ After successfull login the domain part will be striped and the rest used as use
...
+
+     #Blacklisted domains. All emails from these domains are blacklisted and will be rejected.
+     #This file is LIVE reloaded by rspamd. Any changes are EFFECTIVE IMMEDIATELY.
+     dummy.com
+
+3. Reload rspamd
+
+   .. code-block:: bash
+
+     docker-compose scale antispam=0
```
liquidat, line 453: This works only for the docker compose setup. Other users might use swarm or kubernetes, so I wonder how to write this in a more "general" way...

docs/faq.rst
```diff
@@ -334,6 +334,145 @@ After successfull login the domain part will be striped and the rest used as use
...
 
+Spam filtering
+--------------
+
+How does spam filtering work in Mailu?
+``````````````````````````````````````
+
+Mailu uses Rspamd for spam filtering. Rspamd is a Fast, free and open-source spam filtering system.
+
+Rspamd rejects non-compliant email messages and email messages that contain virusses. In Mailu Rspamd uses a scoring scale from 0 to 15:
```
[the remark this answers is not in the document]
Diman0: I'll add this. However this will lead to the situation where the configuration mismatches with score configuration in action.conf.

docs/faq.rst
Diman0: I'm keeping this because I'm moving everything to a separate page as suggested by  Nebukadneza.

docs/faq.rst
Diman0: This part was not originally written by me. For this reason I did't modify it. But the rephrased text reads much better. I'll rephrase it.

docs/faq.rst
Diman0: I have re-read all text 2 times before committing this. This shows it really helps to have someone else proofread your text. Iĺl add the link to the overrides documentation.

docs/faq.rst
Diman0: I have added this for more clarity. However this is also stated in the paragraph followed by this list.

docs/faq.rst
Diman0: Yes. These are pre-filters. This will add an header no matter what the final score is. Using a pre-filter also means that no further rules are processed. 

This is already explained in the paragraph before this list:
_Note the "action = "reject";" line. This is a so-called pre-filter. If you omit this line, then the configured score will be added to the total score of the email message. Depending on the end-score after processing all rules, a verdict is made. To override this, you can add the action line. When this symbol (rule) is fired, then this action is **immediately taken and no further processing occurs.** You can use the following actions:_

docs/faq.rst
Diman0: I have no knowledge of docker swarm and kubernetes. I'll rephrase the step to something general and mention the example is for docker-compose setup.
3. Reload rspamd by stopping the Rspamd container and starting the Rspamd container again. Example for docker-compose setup:

   .. code-block:: bash

     docker-compose scale antispam=0
     docker-compose scale antispam=1

PR #1586, opened 2020-08-11
Add documentation in regard to the spam filter in Mailu. Added all su…

Diman0:
…ggestions from Liquidat and Nebukadneza..

I have added the suggestions mentioned in pull request #1575 

## What type of PR?

Adds FAQ section 'Spam filtering' that describes how spam filtering works. It also describes how you can create your own local black list. This enables the user to set up a map that can be live-updated via Rspamd webgui for blocking emails from listed sender domains.

### Related issue(s)
    #1167
    #1566
    Closes issue #1167

## Prerequistes

    [ N/A] In case of feature or enhancement: documentation updated accordingly
    [ Done] Unless it's docs or a minor change: add changelog entry file.


Nebukadneza:
rebuild now that the docker-API fix is merged:

bors try

PR #1607, opened 2020-08-29
_FILE variables for Docker swarm secrets

lub:
## What type of PR?

enhancement

## What does this PR do?

This PR enables usage of DB_PW_FILE and SECRET_KEY_FILE instead of DB_PW and SECRET_KEY to load these values from files instead of supplying them directly. That way it's possible to use Docker secrets.

## Prerequistes


Diman0 (approved):
Looks good to me

Issue #1328, opened 2020-01-27
Incoming mail does not use TLS


elandorr:
How the headers look:

```
Received: from redacted.com (redacted.com [5.6.7.8])
	by my.redacted.server (Postfix) with ESMTP id randomid
	for <redacted@redacted.com>; Wed, 11 Dec 2019 11:03:20 +0000 (UTC)
```

How the headers **should** look:

```
Received: redacted.com (redacted.com [9.10.11.12]) (using TLSv1.2 with
 cipher ECDHE-RSA-AES256-GCM-SHA384 (256/256 bits)) (No client certificate requested) by
redacted (Postfix) with ESMTPS id randomid for <redacted>;
```

Default settings except `TLS_FLAVOR=cert`. TLS works fine in the webinterface. Outgoing emails do show that TLS has been used on the other side as well.

Only incoming emails do not.

Please help! Not having proper TLS support makes Mailu's entire efforts futile.

micw:
@elandorr TLS termination is done at NGINX (which acts as mail proxy and does authentication). Nginx is properly configured for TLS. Unfortunately, the downstream protocol allows to set the remote IP but not the SSL information (http://www.postfix.org/XCLIENT_README.html). So postfix has no information that it could add to the header because it sees only an unencrypted connection.
Seems that there's no easy way to prove that TLS actually works - in doubt try to sniff between the external smtp server and nginx with tcpdump or wireshark.

elandorr:
@micw Thank you for the explanation! That is very unfortunate. I've been using this for years: https://addons.thunderbird.net/en-US/thunderbird/addon/paranoia/
Why is NGINX used for that? I see no benefits. Is there any chance this may be changed in the future?

Mailu has been a pleasure to use so that's a bummer.

micw:
Main benefit is that nginx can delegate authentication to a http backend in a standardized way

kaiyou:
This has nothing to do with nginx afaik, nginx exposes starttls over port 25, and senders should use opportunistic starttls when connecting.

I will have to check that starttls is not broken on port 25, because that would indeed be a serious regression.

kaiyou:
Oh. I am now understanding that you are facing this error when sending mail. Are you using SMTPS?

elandorr:
@kaiyou Only incoming mails are affected. I checked the headers on the other end and they're fine. Check the header of a github e-mail for instance - they definitely do use TLS. Mailu does not show it.

[...]

PR #1610, opened 2020-09-01
add option to enforce inbound starttls

lub:
## What type of PR?

Feature

## What does this PR do?
It implements a check in the auth_http handler to check for Auth-SSL == on and otherwise returns a 530 starttls error.
If INBOUND_TLS_ENFORCE is not set the behaviour is still the same as before, so existing installations should be unaffected.

Although there is a small difference to e.g. smtpd_tls_security_level of Postfix.

Postfix already throws a 530 after mail from, but this solution only throws it after rcpt to. auth_http is only the request after rcpt to, so it's not possible to do it earlier.

### Related issue(s)
#1328 is kinda related, although this PR doesn't solve the issue that the headers will still display ESMTP instead of ESMTPS

## Prerequistes


Diman0 (dismissed):
Looks good to me.

Diman0 (approved):

PR #1611, opened 2020-09-03
Adds own server on port 80 for letsencrypt and redirect

### LINKED ISSUE #1564: nginx.conf with TLS_FLAVOR=letsencrypt does not allow certbot renew succeed
My certficate was not automatically renewed in mailu 1.7, after some digging I found, that the http->https redirect rule in [nginx.conf:66](https://github.com/Mailu/Mailu/blob/master/core/nginx/conf/nginx.conf#L66) is executed before [nginx.conf:83](https://github.com/Mailu/Mailu/blob/master/core/nginx/conf/nginx.conf#L83).

Therefore the letsencrypt http01 validation receives a 301 redirect:

```
front_1     | 18.196.96.172 - - [07/Jul/2020:22:23:49 +0000] "GET /.well-known/acme-challenge/gwzkNiW5AIPJ39JyYmfDzmvBb9mYcQioKI65HvZHdxI HTTP/1.1" 301 162 "-" "Mozilla/5.0 (compatible; Let's Encrypt validation server; +https://www.letsencrypt.org)"
front_1     | 52.15.254.228 - - [07/Jul/2020:22:23:49 +0000] "GET /.well-known/acme-challenge/gwzkNiW5AIPJ39JyYmfDzmvBb9mYcQioKI65HvZHdxI HTTP/1.1" 301 162 "-" "Mozilla/5.0 (compatible; Let's Encrypt validation server; +https://www.letsencrypt.org)"
```

And certbot terminates the process with:
```
 - The following errors were reported by the server:

   Domain: xxxx
   Type:   connection
   Detail: Fetching
   https://xxxxx/.well-known/acme-challenge/8wDiiX_o76-fHzOeGpa6j0O2wqLQin9nfb0MaWcXY-E:
   Timeout during connect (likely firewall problem)
```

Maybe a second server entry only for the http-https redirect and http-01 validation would be better in `nginx.conf`:
```
{% if KUBERNETES_INGRESS != 'true' and TLS_FLAVOR in [ 'letsencrypt', 'mail-letsencrypt' ] %}
server {
      root /static;
      listen 80;
      listen [::]:80;
      # In any case, enable the proxy for certbot if the flavor is letsencrypt and not on kubernetes
      location ^~ /.well-known/acme-challenge/ {
          proxy_pass http://127.0.0.1:8008;
      }
      location / { 
          return 301 https://$host$request_uri;
      } 
}
{% endif %}

# previous stuff 
{% if KUBERNETES_INGRESS != 'true' and TLS and not TLS_ERROR %}
server {
      listen 443 ssl http2;
      listen [::]:443 ssl http2;
...
{% endif %}
```

Temporary fix: `docker-compose exec front bash`, edit `/etc/nginx/nginx.conf` and comment out the above mentioned if-statement, reload nginx config by `kill -HUP 1`. Then run the `/letsencrypt.py` script. After successful renew: undo the change in  `/etc/nginx/nginx.conf` and reload nginx again.

elektro-wolle:
## What type of PR?

Bugfix

## What does this PR do?

Handle letsencrypt route to `.well-known` by own server configuration within nginx.

### Related issue(s)
closes #1564

## Prerequistes

### DIFF
--- a/core/nginx/conf/nginx.conf
+++ b/core/nginx/conf/nginx.conf
@@ -34,6 +34,25 @@ http {
       ''      $scheme;
     }
 
+    {% if KUBERNETES_INGRESS != 'true' and TLS_FLAVOR in [ 'letsencrypt', 'cert' ] and not TLS_ERROR %}
+    # Enable the proxy for certbot if the flavor is letsencrypt and not on kubernetes
+    # 
+    server {
+      # Listen over HTTP
+      listen 80;
+      listen [::]:80;
+      {% if TLS_FLAVOR == 'letsencrypt' %}      
+      location ^~ /.well-known/acme-challenge/ {
+          proxy_pass http://127.0.0.1:8008;
+      }
+      {% endif %}
+      # redirect to https
+      location / {
+        return 301 https://$host$request_uri;
+      }
+    }
+    {% endif %}
+
     # Main HTTP server
     server {
       # Favicon stuff
@@ -48,9 +67,11 @@ http {
       set $webdav {{ WEBDAV_ADDRESS }};
       {% endif %}
 
-      # Always listen over HTTP
+      # Listen on HTTP only in kubernetes or behind reverse proxy
+      {% if KUBERNETES_INGRESS == 'true' or TLS_FLAVOR in [ 'mail-letsencrypt', 'notls',  'mail' ] or TLS_ERROR %}
       listen 80;
       listen [::]:80;
+      {% endif %}
 
       # Only enable HTTPS if TLS is enabled with no error and not on kubernetes
       {% if KUBERNETES_INGRESS != 'true' and TLS and not TLS_ERROR %}
@@ -78,8 +99,7 @@ http {
       add_header X-XSS-Protection '1; mode=block';
       add_header Referrer-Policy 'same-origin';
 
-      # In any case, enable the proxy for certbot if the flavor is letsencrypt and not on kubernetes
-      {% if KUBERNETES_INGRESS != 'true' and TLS_FLAVOR in [ 'letsencrypt', 'mail-letsencrypt' ] %}
+      {% if TLS_FLAVOR == 'mail-letsencrypt' %}      
       location ^~ /.well-known/acme-challenge/ {
           proxy_pass http://127.0.0.1:8008;
       }
--- a/core/nginx/conf/nginx.conf
+++ b/core/nginx/conf/nginx.conf
@@ -34,7 +34,7 @@ http {
       ''      $scheme;
     }
 
-    {% if KUBERNETES_INGRESS != 'true' and TLS_FLAVOR in [ 'letsencrypt', 'cert' ] and not TLS_ERROR %}
+    {% if KUBERNETES_INGRESS != 'true' and TLS_FLAVOR in [ 'letsencrypt', 'cert' ] %}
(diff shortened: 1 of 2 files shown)


Diman0 (dismissed):
Looks good to me. Thank you for the PR.

core/nginx/conf/nginx.conf
```diff
@@ -34,6 +34,25 @@ http {
       ''      $scheme;
     }
 
+    {% if KUBERNETES_INGRESS != 'true' and TLS_FLAVOR in [ 'letsencrypt', 'cert' ] and not TLS_ERROR %}
```
lub, line 37: Why is `TLS_ERROR` excluded?

If I understand this change correctly, there is no way to recover from `TLS_ERROR`, when `TLS_FLAVOR=letsencrypt`.
This is due to the only other place with a acme-challenge proxy being reserved for `TLS_FLAVOR=mail-letsencrypt` in L102.

I suspect the `TLS_ERROR` checks in this line and in L71 could be deleted.

This would prevent http access while `TLS_ERROR=yes`, because of the `return 301`, but I'd honestly consider it a feature to prevent unencrypted access when TLS is actually intended.

lub (changes requested):

Nebukadneza:
@Diman0 and me just reviewed and tested this in pair-programming. We could not reproduce the behaviour on current `mailu/nginx:master` unfortunately.

We both believe that the control-flow logic in todays `nginx.conf` is unclear.

For this PR, I think it would be beneficial to first establish clearly understanding of which "cases" of tls-configuration need to be handled, and what the right behavior of these would be. Maybe we should try to write that down first, and review with this.

We both need a clear / non-tired head first ^_^. Please be patient a little bit, and thanks a LOT for taking this on!

Diman0:
> @Diman0 and me just reviewed and tested this in pair-programming. We could not reproduce the behaviour on current `mailu/nginx:master` unfortunately.
> 
> We both believe that the control-flow logic in todays `nginx.conf` is unclear.
> 
> For this PR, I think it would be beneficial to first establish clearly understanding of which "cases" of tls-configuration need to be handled, and what the right behavior of these would be. Maybe we should try to write that down first, and review with this.
> 
> We both need a clear / non-tired head first ^_^. Please be patient a little bit, and thanks a LOT for taking this on!

We think it is fine, but need to go through it one more time. Go through the config one more time where we will check what happens when a specific TLS flavour is selected.
Hint for myself: In the second server block, we only need to listen on port 80 if TLS flavour [cert, letsencrypt] is **not** used.

elektro-wolle:
Maybe for clarification: The initial certbot setup for letsencrypt works fine. Only the update after 2 month is failing.

Diman0 (approved):
After reviewing it again with a more clear head, all changes look good to me again. Thank you for providing the fix.

LGTM

Nebukadneza (approved):
Indeed, after having a proper look on a weekend, this makes sense to me.
One thing I didn’t understand yet is that in some very rare cases I manage to get this:
`front_1     | 2020/09/26 05:40:48 [warn] 1#1: conflicting server name "" on 0.0.0.0:80, ignored`
Immediately after a (simulated) renewal, and I can’t follow how 2 `:80`-listeners can come to live with your templating.

However, as the behavior seems correct even in these cases, I’d say we’re good to take this in.

Thanks for your time on this!

LGTM

PR #1360, opened 2020-02-09
Ratelimit outgoing emails per user

kaiyou:
## What type of PR?

Feature

## What does this PR do?

Using the refactored rate limiter in #1359, this provides a configuration variable (for now, waiting for user and domain settings) for rate limiting outgoing traffic.

### Related issue(s)

- Closes: #1031 

## Prerequistes
Before we can consider review and merge, please make sure the following list is done and checked.
If an entry in not applicable, you can check it or remove it from the list.

- [x] In case of feature or enhancement: documentation updated accordingly
- [x] Unless it's docs or a minor change: add [changelog](https://mailu.io/master/contributors/guide.html#changelog) entry file.

PR #1612, opened 2020-09-03
Add option to disable rate limiting via AUTH_RATELIMIT=False

lub:
## What type of PR?

Feature, enhancement

### Related issue(s)
- maybe this conflicts with the changes in #1360, although I think #1360 would need to be rebased anyway

## Prerequistes

### DIFF
--- a/core/admin/mailu/internal/views/auth.py
+++ b/core/admin/mailu/internal/views/auth.py
@@ -12,24 +12,26 @@ import ipaddress
 def nginx_authentication():
     """ Main authentication endpoint for Nginx email server
     """
-    limiter = utils.limiter.get_limiter(app.config["AUTH_RATELIMIT"], "auth-ip")
-    client_ip = flask.request.headers["Client-Ip"]
-    if not limiter.test(client_ip):
-        response = flask.Response()
-        response.headers['Auth-Status'] = 'Authentication rate limit from one source exceeded'
-        response.headers['Auth-Error-Code'] = '451 4.3.2'
-        if int(flask.request.headers['Auth-Login-Attempt']) < 10:
-            response.headers['Auth-Wait'] = '3'
-        return response
     headers = nginx.handle_authentication(flask.request.headers)
     response = flask.Response()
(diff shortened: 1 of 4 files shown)


core/admin/mailu/internal/views/auth.py
```diff
@@ -12,24 +12,26 @@
...
-    if not limiter.test(client_ip):
-        response = flask.Response()
-        response.headers['Auth-Status'] = 'Authentication rate limit from one source exceeded'
-        response.headers['Auth-Error-Code'] = '451 4.3.2'
-        if int(flask.request.headers['Auth-Login-Attempt']) < 10:
-            response.headers['Auth-Wait'] = '3'
-        return response
     headers = nginx.handle_authentication(flask.request.headers)
     response = flask.Response()
+    ratelimit = app.config["AUTH_RATELIMIT"]
```
nextgens, line 17: Rate-limiting after the actual password check is a big no-no, I've proposed an alternative implementation in #1745

nextgens (changes requested):
The expensive call is nginx.handle_authentication(flask.request.headers) ... this shouldn't run post rate limiting

docs/configuration.rst
```diff
@@ -40,6 +40,7 @@ address.
 The ``AUTH_RATELIMIT`` holds a security setting for fighting attackers that
 try to guess user passwords. The value is the limit of failed authentication attempts
 that a single IP address can perform against IMAP, POP and SMTP authentication endpoints.
+Setting this this to ``False`` disables all rate limiting.
```
Diman0, line 43: You want to remove the duplicate "this" in this sentence.

Diman0 (changes requested):
Correct the typo in the doc. I think we can approve the PR after this. 
At a later moment we can look at the changes suggested by Nextgens.

nextgens:
IMHO this has very little added value over putting a very high limit. If you want it disabled, set  it to 10000/second or something.

nextgens:
This has been deprecated by #1910; more work will happen on rate limiting in next releases

PR #1613, opened 2020-09-05
Add warnings about Dockers IPv6 solution

### LINKED ISSUE #1578: Open relay when using ipv6
Okay. I used the setup config generator from mailu, then commented all ipv4 ports (changed the http for my reverse proxy on an other machine) and then exposed the ipv6 ports on `::` as advised in the documentation for running accessible for the internet.
This causes the the front to identify all incoming emails as internal (from a ipv4 (?!?) address within the private subnet) and relays them without further checking!

Yeah I know there are some problems at the moment with ipv6 - but this should be **CLEARLY STATED** inside the documentation!

**Workaround**: Comment `SUBNET6` and the subnet part for ipv6 (and `enable_ipv6`) from docker-compose. Also allow all from the internet accessible ports only to listen on `0.0.0.0` (by modifying the port expose statement).

Simonmicro:
## What type of PR?

enhancement

## What does this PR do?
Added some warnings and explanations about the current Docker ipv6 situation as suggested in https://github.com/Mailu/Mailu/issues/1578.

### Related issue(s)
closes #1578

## Prerequistes

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -170,6 +170,13 @@ Lets start with quoting everything that's wrong:
       (`docker/libnetwork#1099 <https://github.com/docker/libnetwork/issues/1099>`_).
   
   -- `Robbert Klarenbeek <https://github.com/robbertkl>`_ (docker-ipv6nat author)
+  
+Okay, but I still want to use IPv6! Can I just use the installers IPv6 checkbox? **NO, YOU SHOULD NOT DO THAT!** Why you ask?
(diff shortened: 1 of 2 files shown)


Diman0 (approved):
Looks good to me.

Nebukadneza (approved):
Thank you so much for this notice. This kind of big warning was long overdue, and you will help a lot of users with this in the future … Thanks!

LGTM

PR #1615, opened 2020-09-07
Create ISSUE_TEMPLATE.md

### LINKED ISSUE #1331: Discussion: Create issue templates for github
We currently have many unstructured issue report, mixed with support requests and feature requests. For many we don't even know the affected Mailu version. We should create issue templates for the type of issues we'd like to support (bugs and feature requests at least) and define mandatory information we require.
We should also decide if we accept support requests in the issue tracker (IMO we should point to the chat here and close support request issues).

Diman0:
Add bug template to be used for reporting issues.

## What type of PR?

Feature?

## What does this PR do?
Add bug template which we will use to get better quality bug reports.

### Related issue(s)
- Auto close an issue like: closes #1331

## Prerequistes

- [N/A] In case of feature or enhancement: documentation updated accordingly
- [N/A] Unless it's docs or a minor change: add [changelog](https://mailu.io/master/contributors/guide.html#changelog) entry file.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,36 @@
+Thank you for opening an issue with Mailu. Please understand that issues are meant for bugs and enhancement-requests.  
+For **user-support questions**, reach out to us  on [matrix](https://matrix.to/#/#mailu:tedomum.net).
+
+To be able to help you best, we need some more information.
+
+## Before you open your issue
+- [ ] Check if no issue or pull-request for this already exists.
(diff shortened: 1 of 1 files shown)


Nebukadneza (approved):
Now I feel kinda bad approving this after having tinkered around in it myself. But i agree with @Diman0 and @liquidat that we should get it in rather sooner than later.
So — Thanks a lot for taking the first step here!
LGTM

Diman0 (commented):
Looks good.

Nebukadneza:
Since this would help us a lot with sorting and filtering inbound issues, I’m going to manually merge it … even though i don’t quite like doing that. It’s worth it here, though, i belive …

bors r+

PR #1617, opened 2020-09-10
relax TLS settings on port 25

lub:
Because basically every MTA out there uses opportunistic TLS _in
the best case_, it's actually counter productive to use such strict
settings.

The alternative to a handshake error is often an unencrypted submission,
which is basically the opposite of what strict ssl_protocols and
ssl_ciphers tries to achieve.

Even big and established providers like Amazon SES are incompatible with the current
settings.

This reverts commit 2ddf46ad2b5054988ea5b270dcdb88ea80e6067e.

## What type of PR?

bug-fix

### Related issue(s)
- this reverts #1398
- the settings in this PR were initially introduced in #1321 and intentionally divert from tls.conf

cc @bladeswords as they introduced the current config


Diman0 (approved):
Looks good to me

Nebukadneza (approved):
Since we talked about this in the video-meeting … LGTM&Thanks ^_^

LGTM

PR #1592, opened 2020-08-18
Add documentation for the web administration gui.

Diman0:
## What type of PR?

Documentation

## What does this PR do?

This PR adds the section Web Administration Interface to the documentation site which completely documents all available settings in the web administration interface.

### Related issue(s)
- Closes issue #1590 (please close this issue for me)

## Prerequistes
Before we can consider review and merge, please make sure the following list is done and checked.
If an entry in not applicable, you can check it or remove it from the list.

- [ Done ] In case of feature or enhancement: documentation updated accordingly
- [ Done ] Unless it's docs or a minor change: add [changelog](https://mailu.io/master/contributors/guide.html#changelog) entry file.

Issue #783, opened 2018-12-30
Redesign the authentication systems


muhlemmer:
This issue is to enhance the security of authentication within Mailu. It is based on a small brain-storming session me and @kaiyou had.

In order to properly secure and manageable 2fa system, we will need to centralize all web logins to a central point (eg admin). This system will forward an authentication token to the final app (webmail / admin). From this system we can extend to 2fa. Additionally, we want to investigate using a more advanced authentication system for IMAP:

> [...] there are a couple technos available, OAuth, SAML, and also the good all redirect-and-post-the-login-form-for-me technique. We have to experiment 

- [x] Write a central authentication handler in admin
- [x] Implement remote authentication for Roundcube
- [x] Implement remote authentication for Rainloop
- [ ] Advanced auth system for IMAP
- [ ] Implement 2fa in the central handler

This issue replaces #67, #264 and #685

briantopping:
Hi @pierreozoux! I am not a member of the project, just a user with similar interests such as 2FA and Kubernetes. I opened #605, but after digging quite deep into PAM, I am not sure it is sufficient for a great user experience. Much of this experience has come from generating a XAUTH IPSEC VPN deployment for k8s. (It works against FreeIPA, but not yet with 2FA due to Kerberos FAST and preauth issues.)

The problem with mail and 2FA as I have come to realize is the users tend to be less forgiving to re-enter a 2FA password every time the IMAP session resets or they send an email. You can imagine the situation, every time the front end tries to validate against the authN, the TOTP will have rotated and the password is no longer valid. So not only does the user have to retype their password, they have to find their 2FA token. A good discussion of use cases is in https://docs.pagure.org/SSSD.sssd/design_pages/pam_conversation_for_otp.html, which is the design doc for FreeIPA OTP. 

One of the considerations the author there makes is that if the first and second factor is split out, the second factor request (the TOTP token, typically) can be skipped as a convenience. You'll see as you read it that it is a slightly different use case though. In our case, the IMAP/SMTP client needs the capability to have a long lasting strong authentication method that can span sessions. For instance, if I close my laptop, take it to another location and re-open it, I typically do not expect to sign back in to each email account. 

One means is to use Kerberos tickets, but I've found while tickets work on OSX, there is no way to configure the KDC on iOS, so there is no way to get mail on the phone.

[...]

PR #1619, opened 2020-09-13
Add password plugin

harveyconnor:
## What type of PR?

enhancement

## What does this PR do?
Adds password plugin so users can change their password within roundcube.
By going to Settings > Password.

## Prerequistes

### DIFF
--- a/webmails/roundcube/config.inc.php
+++ b/webmails/roundcube/config.inc.php
@@ -17,7 +17,8 @@ $config['plugins'] = array(
     'markasjunk',
     'managesieve',
     'enigma',
-    'carddav'
+    'carddav',
+    'password'
 );
 
 $front = getenv('FRONT_ADDRESS') ? getenv('FRONT_ADDRESS') : 'front';


lub:
Does this actually work? As far as I can see you'd also have to supply the password plugin settings to set the password directly in the database in the same format as the admin gui.

https://github.com/roundcube/roundcubemail/blob/master/plugins/password/config.inc.php.dist

Also I don't know if it makes sense to support two ways to change the password, because e.g. when 2fa gets implemented due to #783 you maybe want to verify the second factor during password changes.
Or when you login with an authentication token you shouldn't be able to change your password with it, because that would defeat the purpose of the authentication token. I'm not sure if roundcube could differentiate that.

Diman0:
As lub already stated, this will not work. This plugin requires additional configuration. The password is stored in the database.  

The admin web interface is the main interface for changing user settings including the password. I'm not certain if we want to allow other services than admin to change user passwords.

harveyconnor:
okay makes sense. I think it would be a good option for the user to have some control over that password.

lub:
Users should be able to change the password themselves via admin, when you actually expose the admin interface to them.

Diman0:
The web administration interface is not only meant for administrators. Each user can change user preferences via the web administration interface. Users are also able to change their password via the web administration interface.

Unfortunately the web administration interface is not documented. This includes what settings can be configured by normal users (such as your password). The PR to add this to the documentation is pending (https://github.com/Mailu/Mailu/pull/1592). You can read more here.

harveyconnor:
ahh okay well that makes sense then. I'll let my users know that they can do that.

PR #1624, opened 2020-09-20
Fix link to installation guide for Cert Manager

Sobuno:
## What type of PR?
Bugfix in Documentation

## What does this PR do?
Fixes link to Cert Manager installation in documentation. Previous link was returning 404.


lub:
This is already fixed in master, but not in the current stable. So looks good to me.

lub (approved):

Diman0 (approved):
Normally we only accepts PRs on master. If necessary we backport it to an earlier branch. But since it is a one-liner fix for a link and is already fixed in master, I'm also fine with this.

PR #1627, opened 2020-09-21
1.7

blainvillem:


lub:
@blainvillem What are you trying to accomplish? The commits for the stable branches (in this case 1.7) are normally originating from master, so no point in merging the branch back.

See https://mailu.io/master/contributors/workflow.html#pr-workflow for more details.

PR #1629, opened 2020-09-21
Add mergify team-members

Nebukadneza:
## What type of PR?

enhancement

## What does this PR do?

Sync the newly added team-members to the mergify team-regex so that they will benefit from the "team needs only one approve" rule

## Prerequistes

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -27,7 +27,7 @@ pull_request_rules:
 
   - name: Trusted author and 1 approved review; trigger bors r+
     conditions:
-      - author~=^(kaiyou|muhlemmer|mildred|HorayNarea|adi90x|hoellen|ofthesun9|Nebukadneza|micw)$
+      - author~=^(kaiyou|muhlemmer|mildred|HorayNarea|adi90x|hoellen|ofthesun9|Nebukadneza|micw|lub|Diman0)$
       - -title~=(WIP|wip)
       - -label~=^(status/wip|status/blocked|review/need2)$
       - "#approved-reviews-by>=1"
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -11,9 +11,9 @@ pull_request_rules:
         message: |
           Thanks for submitting this pull request.
           Bors-ng will now build test images. When it succeeds, we will continue to review and test your PR.
-
```
lub, line 14: Aren't these indentations required by the yaml spec? Afaik | continues until indentation changes.

Diman0: Yes. These indentations are required to continue the multiline string.

.mergify.yml
```diff
@@ -11,9 +11,9 @@ pull_request_rules:
         message: |
           Thanks for submitting this pull request.
           Bors-ng will now build test images. When it succeeds, we will continue to review and test your PR.
-          
+
           bors try
-
```
Diman0, line 16: These indentations are required to continue the multiline string. Please use space characters for the indendation.

Diman0 (changes requested):
so messages from lub and me. You removed indendations from the yaml file which are required to continue the multiline text.

.mergify.yml
Nebukadneza: Argh, sorry for not having paid attention. Thanks for the pointer!

Diman0 (approved):
Looks good to me

Nebukadneza:
In accordance with @Diman0 we believe this is now fixed, and since mergify won’t do it’s job on PRs containing mergify config, let’s do this by hand:

bors r+

Issue #1465, opened 2020-04-27
Sending via an alias leaks mailbox address/login


githtz:
When a mail is sent by an alias, the mail header contains information about the original address/login.
```
ARC-Authentication-Results: i=1;
	example.com;
	auth=pass smtp.auth=user@example.com smtp.mailfrom=alias@example.com
```

Is there anything that can be done about that?

Nebukadneza:
Hi There,

The `Mailu`-Project is currently in a bit of a bind! We are short on man-power, and we need to judge if it is possible for us to put in some work on this issue.

To help with that, we are currently trying to find out which issues are actively keeping users from using `Mailu`, which issues have someone who want to work on them — and which issues may be less important. These a less important ones could be discarded for the time being, until the project is in a more stable and regular state once again.

In order for us to better assess this, it would be helpful if you could put a **reaction on this post** (use the :smiley: icon to the top-right).
- 👍️ if you **need this** to be able to use Mailu. Ideally, you’d also be able to test this on your installation, and provide feedback …
- 🎉 if you find it a **nice bonus**, but no deal-breaker
- 🚀 if you want to **work on it yourself**!
We want to keep this **voting open for 2 weeks** from now, so please help out!

githtz:
I have found a solution for the problem here by adding an override.

**milter_headers.conf** mounted into the rspamd-container
```bash
authenticated_headers = ["authentication-results"];
skip_local = false;
skip_authenticated = false;

use = ["x-spamd-bar", "x-spam-level", "x-virus", "authentication-results"];

routines {
  authentication-results {
    add_smtp_user = false;
  }
  x-virus {
    symbols = ["CLAM_VIRUS", "FPROT_VIRUS", "JUST_EICAR"];
  }
}
```

Diman0:
Thank you for sharing the workaround. I will have a look at this. I understand you do not want to leak the main email address when sending via an alias.

PR #1638, opened 2020-09-24
Remove the username from the milter_headers

githtz:
Rspamd adds the name of the authenticated user by default. Setting add_smtp_user to false prevents the login to be leaked.

## What type of PR?
Enhancement

## What does this PR do?
This PR prevents the user login to be leaked in sent emails (for example using an alias)

### Related issue(s)
Closes https://github.com/Mailu/Mailu/issues/1465

## Prerequistes


Diman0 (approved):
Looks good to me.

nextgens (approved):

PR #1640, opened 2020-09-25
Update stale.yml

Diman0:
Add label backlog to exempt labels in stale.yml. We will use label backlog to collect bugs/features that we want to consider for a future release. Thus we do not want to auto-close those issues.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -15,6 +15,7 @@ exemptLabels:
   - priority/p0
   - priority/p1
   - priority/p2
+  - backlog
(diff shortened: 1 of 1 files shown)


lub (approved):
lgtm

lub:
more discussion on this happened here:
https://matrix.to/#/!RJFCFtixHgPhzacdhW:tedomum.net/$160102509115Ilobh:huisman.xyz?via=matrix.org&via=tedomum.net&via=converser.eu

PR #1645, opened 2020-09-27
Update stale.yml

Diman0:
as discussed in chat https://matrix.to/#/!RJFCFtixHgPhzacdhW:tedomum.net/$160120578037UHNkM:huisman.xyz?via=ghostdub.de&via=matrix.org&via=tedomum.net
stalebot should only touch
- user support (issue with no label or issue with label type/question).
- issues we explicitly mark with a response_needed label (whatever the name will be of this label).
To give more time to respond when we mark issue with more info needed, we increase the daysUntilClose to 14


Nebukadneza (approved):
Thanks ^_^

PR #1654, opened 2020-10-04
Ensure that the rendered file ends with newline in order to make `pos…

### LINKED ISSUE #1580: postfix.cf is not read (again)
The issue described in #941 is present again in master branch as the fix #945 was removed from the code (https://github.com/Mailu/Mailu/commit/bd69b7a4915cb1841adf700003018b07f7a260ad#diff-a3dd17dd36aa26b6fcf013616ab2b0c0). I think that adding an uncommented blank line is likely to break again.
Can we use `keep_trailing_newline` as suggested in the original issue or change the code in `start.py` in another way that enforces a new line at the end?

tremlin:
## What type of PR?

Bugfix

## What does this PR do?

This fixes #1580 

### Related issue(s)
- closes #1580

## Prerequistes

### DIFF
--- a/core/postfix/conf/main.cf
+++ b/core/postfix/conf/main.cf
@@ -126,3 +126,5 @@ milter_default_action = tempfail
 ###############
 # Extra Settings
 ###############
+{# Ensure that the rendered file ends with newline in order to make `postconf` work correctly #}
+{{- "\n" }}


ofthesun9:
Thanks !

ofthesun9 (approved):

lub (approved):

PR #1655, opened 2020-10-05
Postfix main.cf needs a EOL on last line

dkruyt:
When using a override file, a EOL is needed for the main.cf on the last line, or else the first line of the overrides file ends up after the #### a quick workaround is to place a dummy option on the first line of your override file.

```
###############
# Extra Settings
###############dummy = monkey
postscreen_upstream_proxy_protocol = haproxy
```

## What type of PR?

bug-fix

## What does this PR do?

Fixes the override option in postfix main.cf

## Prerequistes

- [] In case of feature or enhancement: documentation updated accordingly

### DIFF
--- a/core/postfix/conf/main.cf
+++ b/core/postfix/conf/main.cf
@@ -126,3 +126,4 @@ milter_default_action = tempfail
 ###############
 # Extra Settings
 ###############
+


dkruyt:
Just see there is already a duplicated #1654 created a couple of hours ago...

PR #1656, opened 2020-10-05
Add ability to set no WEBROOT_REDIRECT to Nginx

DavidFair:
## What type of PR?

Enhancement / Documentation

## What does this PR do?

From commit:

---

Adds a 'none' env option to WEBROOT_REDIRECT so that no `location /`
configuration is written to nginx.conf.

This is useful for setting up Mailu and Mailman where we override the
root to proxy to the mailing list server instead. Without this change
the nginx container will not start, or for 1.7 users can set their
WEBMAIL_PATH to / with no webmail to get the same results.

This fix means that future users don't have to choose between webmail
and a root override and makes the configuration intention clear.

---

I've also added bullet points to break up a long flowing sentence in `configuration.rst` - it should be a bit easier to read now

### Related issue(s)
No Related Issue - I just jumped to a PR

## Prerequistes

@ Maintainers - Is this worthy of the changelog, it's useful to know about but I imagine the number of people it affects is equally minimal?

### DIFF
--- a/core/nginx/conf/nginx.conf
+++ b/core/nginx/conf/nginx.conf
@@ -115,7 +115,7 @@ http {
       include /overrides/*.conf;
 
       # Actual logic
-      {% if WEB_WEBMAIL != '/' %}
+      {% if WEB_WEBMAIL != '/' and WEBROOT_REDIRECT != 'none' %}
       location / {
       {% if WEBROOT_REDIRECT %}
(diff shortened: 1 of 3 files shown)


Diman0 (approved):
Looks good to me. Thank you for your contribution.

nextgens (approved):

PR #1669, opened 2020-10-23
Fix extract_host_port port separation

cbachert:
Regex quantifier should be lazy to make port separation work.

## What type of PR?
bug-fix

## What does this PR do?
The "extract_host_port" function in admin/mailu/internal/nginx.py and optional/fetchmail/fetchmail.py is not actually separating host and port due to the `(.*)` part of the regex being too generous. Lazy quantifier `(.*?)` allows the other capturing groups to match.

### Related issue(s)
- No issue raised for this

## Prerequistes
- [x] Documentation updated accordingly: N/A, bug-fix
- [x] Add [changelog] entry file: Added towncrier newsfragment with second commit

### DIFF
--- a/core/admin/mailu/internal/nginx.py
+++ b/core/admin/mailu/internal/nginx.py
@@ -84,7 +84,7 @@ def get_status(protocol, status):
     return status, codes[protocol]
 
 def extract_host_port(host_and_port, default_port):
-    host, _, port = re.match('^(.*)(:([0-9]*))?$', host_and_port).groups()
+    host, _, port = re.match('^(.*?)(:([0-9]*))?$', host_and_port).groups()
     return host, int(port) if port else default_port
 
 def get_server(protocol, authenticated=False):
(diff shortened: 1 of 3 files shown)


lub (dismissed):
lgtm, like discussed in chat https://matrix.to/#/!MINuyJjJSrfowljYCK:tedomum.net/$16033217572389nMzHT:imninja.net?via=tedomum.net&via=matrix.org&via=ghostdub.de

> hm, seems to make sense. I was wondering why it didn't affect me because I tried to set all _ADDRESS vars manually in my setup, but the ones where extract_host_port is used don't have a custom port in my config. that would explain why I didn't notice it.

lub:
Although.. could you maybe add an newsfragment file as short changelog?

https://mailu.io/master/contributors/workflow.html#changelog

cbachert:
> Although.. could you maybe add an newsfragment file as short changelog?

Definitely, just did that now.

lub (approved):

ofthesun9 (approved):

PR #1672, opened 2020-10-27
mark radio buttons in setup utility as required

lub:
## What type of PR?
bug-fix

## What does this PR do?
mark radio buttons in setup utility as required

Otherwise it's possible to submit the form without selecting e.g. any
flavor, which would need additional handling on the server side.

## Prerequistes


lub:
In firefox it looks like this after the change:

image
image

And like this before the change:
image

ofthesun9 (approved):

PR #1710, opened 2020-12-20
fix typo in faq.rst

tomwojcik:


lub:
Hi, thanks for the fix :)

Would you mind opening a PR against master to fix it there? We should be able to backport it then via labels to the stable branch.

tomwojcik:
Hey, sure. I pushed changes via web browser and for some reason 1.7 was set as a default branch.

I see there's some diff from 1.7 to master so to make things easy I will create a new PR from master.

#1711

PR #1712, opened 2020-12-23
Add details for postfix-overrides

### LINKED ISSUE #1628: MASTER.CF
Hello,

We would like to know how to change the content of the configuration file "master.cf" for postfix.

The main.cf file is changed by creating a file called postfix.cf in the folder named "overrides".

But what about master.cf?

Thanks in advance.

sholl:
## What type of PR?

Documentation clarification

### Related issue(s)

this clarifies the FAQ about overrides and fixes #1628

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -258,9 +258,9 @@ Postfix, Dovecot, Nginx and Rspamd support overriding configuration files. Overr
 correct syntax. The following file names will be taken as override configuration:
 
 - `Postfix`_ :
-   - ``postfix.cf`` as ``/overrides/postfix.cf``
-   - ``master.cf`` as ``/overrides/postfix.master``
-   - All ``/overrides/*.map`` files
+   - ``main.cf`` as ``$ROOT/overrides/postfix/postfix.cf``
+   - ``master.cf`` as ``$ROOT/overrides/postfix/postfix.master``
+   - All ``$ROOT/overrides/postfix/*.map`` files
 - `Dovecot`_ - ``dovecot.conf`` in dovecot sub-directory;
 - `Nginx`_ - All ``*.conf`` files in the ``nginx`` sub-directory;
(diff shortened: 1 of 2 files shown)


nextgens (approved):

docs/faq.rst
```diff
@@ -257,7 +257,10 @@ Postfix, Dovecot, Nginx and Rspamd support overriding configuration files. Overr
 ``$ROOT/overrides``. Please refer to the official documentation of those programs for the
 correct syntax. The following file names will be taken as override configuration:
 
-- `Postfix`_ - ``postfix.cf`` in postfix sub-directory;
+- `Postfix`_ :
+   - ``postfix.cf`` as ``/overrides/postfix.cf``
+   - ``master.cf`` as ``/overrides/postfix.master``
+   - All ``/overrides/*.map`` files
```
lub, line 263:
```suggestion
   - ``main.cf`` as ``$ROOT/overrides/postfix/postfix.cf``
   - ``master.cf`` as ``$ROOT/overrides/postfix/postfix.master``
   - All ``$ROOT/overrides/postfix/*.map`` files
```

I know that these are the paths mentioned in `start.py`, but from the perspective of the compose file and volumes the postfix overrides have their own subdirectory. It just gets mounted to `/overrides` within the resulting container.

Also the default name for `postfix.cf` within a postfix setup is `main.cf`, so we should probably call it that although the path is a different.

In the long tail it probably makes sense to rename these files and prevent further confusion, but documenting the current behavior as-is already helps very much :)

lub (dismissed):

docs/faq.rst
sholl: hi @lub, I am fine with your suggested changes and will resolve this soon.

lub (approved):
thank you very much :)

Diman0 (approved):
Thank you. Looks good to me.

Issue #1430, opened 2020-04-03
mailu/nginx:1.7 does not start


numericalatina:
This is a refrase of #1152 since the real proble is the container being unable to start
This happens also in the *notls* flavour

Checking processes, it seems stuck at **config.py**
```
    1 root      0:00 {start.py} /usr/bin/python3 /start.py
    6 root      0:00 {config.py} /usr/bin/python3 /config.py
   14 root      0:00 bash
   29 root      0:00 ps aux
```
Indeed the configuration has not taken place at **/etc/nginx** which has only the default *nginx.conf*

Trying to start *nginx* manually with the default conf throws an error
```
>>> os.execv("/usr/sbin/nginx", ["nginx", "-g", "daemon off;"])
nginx: [emerg] open() "/run/nginx/nginx.pid" failed (2: No such file or directory)
```

ofthesun9:
From within the front container, can you ping:
- admin
- antispam
Are those 2 services up and running ?

Same for webmail and webdav if you have selected those services.

j0x00:
I have the same issue. Containers like admin are trying to resolve the ip for the front/nginx container and because of that, they are stuck in the start up loop. This is happening due to dockers dns system which resolves only for container which are healthy.

Additional info: I use docker swarm

ofthesun9:
@j0x00 Good catch !
I made a quick test with healtcheck disabled for various services (front, admin, imap, smtp, webmail) on master, and I could launch the swarm

smoebody:
i just experienced the same. I had to disable all healthchecks, otherwise the containers would not be connected to the network. it seems that at least for docker swarm a positive healthcheck is mandatory for connecting the container to the network.

j0x00:
Basically this is intended by docker swarms dns. Only when a container reaches the state "healthy" the domain name will be routed to the containers/service ip. I have currently no clue why nginx/front has issues getting to the correct state, but the reason the admin and several other containers are stuck in the start up loop it because they can not resolve front as domain name.

smoebody:
@j0x00 Yes, you are right. 

Regarding nginx not starting: I know that nginx refuses to start if an upstream is not available. Since webmail, antispam, admin .... are all upstreams, resolved via service-name nginx would not be start until dns is working. 
The question is, how did it work in the first place ....

CARB2:
How did you solve the issue?

j0x00:
[...]

Issue #1013, opened 2019-05-07
Refactor the way our containers communicate


kaiyou:
Now that the admin container is using mailustart, it does all kinds of DNS resolving at startup. I believe this is wrong, since it does not account for any container changing IP during the life of the admin container, and we know this happens. A lot.

Currently the admin container take a couple minutes to start in a dev environment since not all hosts exist and resolve. Also, once an IP changes, all things break. Also, this basically prevents any scaling and load balancing (even though I agree it is an edge case for Mailu).

I would rather have a simple wrapper for live resolving that supports retrying and caching, and do live resolving on the fly.

kaiyou:
I am starting to think something a lot more radical needs be done. The admin container is fine since resolving can be performed live, but nginx and Postfix do not support this for instance. Many ideas, I'll dig some of them, including live config reloading and proxying.

kaiyou:
After some thought, I  renamed the issue to be more general because I think we need to think fresh and avoid any old preconceived idea.

Basically, most container environments rely on either DNS or distributed configuration for resolving the addresses of other containers. We need to support these. I would says we start with DNS (what we have been doing so far) and slowly build upon it.

Now, how should we reconcile dynamically resolving another containers address and services that do not provide live address resolution? I believe we have a couple possibilities :

- hacking the software in question (nginx -- which we already did :smile: , Postfix -- which we mostly already did in Podop) to use only live queries;
- switch to software that support live queries or stop relying on them  altogether (hard, I believe);
- reload the configuration at runtime when the destination of a query should change.

I am thinking a lot about these options, and will be reviewing where exactly we use address resolution and what the alternatives are in each case, reporting here.

kaiyou:
Here are the places containers currently reference each other : 
 - the admin container provides references to nginx and podop for various use cases, it is inside the API so perfectly able to do live querying
 - nginx has references to the admin, antispam, webdav and webmail, it is nginx so we know how to hack si it does live DNS querying
 - Postfix has references to imap and antispam, those are static; they could probably replaced with podop calls
 - Fetchmail has a reference to the smtp call, but it is able to do live resolution
- Rspamd does live resolution already.

So it is conceivable to keep going with DNS. What we need though is to get back to live resolution for all these addresses. I can work on using Podop for the remaining ones in Postfix. Then I think we should write a simply retrying wrapper with cache for DNS resolution. And finally I say we move back to dynamic resolution.

Nebukadneza:
Hi There,

[...]

PR #1716, opened 2020-12-31
Remove startup resolution of HOST_* variables to IP addresses in *_ADDRESS variables

### LINKED ISSUE #1341: HOST_* and *_ADDRESS resolution errors for docker-compose deployments
I have encountered the following error when trying to access the admin UI whenever some  containers in the stack are recreated or restarted in a docker-compose deployment.
**Steps to reproduce:**
1. Start Mailu using docker-compose, everything works fine.
2. Update .env or perform an action that will recreate or restart some containers.
**Result**
A service that relies on HOST_* or/and *_ADDRESS functionality if not restarted/recreated with the rest, will break.
Note that this is intermittent.

`docker-compose logs -f front`
```
front_1       | 2020/02/05 16:36:42 [info] 11#11: *13 client 192.168.203.11:38536 connected to 0.0.0.0:10143
front_1       | 2020/02/05 16:36:42 [error] 11#11: *15 connect() failed (111: Connection refused) while connecting to upstream, client: 127.0.0.1, server: , request: "GET /auth/email HTTP/1.0", upstream: "http://192.168.203.2:80/internal/auth/email", host: "127.0.0.1"
front_1       | 127.0.0.1 - - [05/Feb/2020:16:36:42 +0000] "GET /auth/email HTTP/1.0" 502 150 "-" "-"
front_1       | 2020/02/05 16:36:42 [error] 11#11: *13 auth http server 127.0.0.1:8000 did not send server or port while in http auth state, client: 192.168.203.11, server: 0.0.0.0:10143, login: "user@mail"
```
`docker network inspect mailu_default`
```
----Snip---
 "e0d90201de96842f31391fc71a13324f353ead73fe085bc2fa3fc8b11d6ddf8c": {
                "Name": "mailu_antivirus_1",
                "EndpointID": "a0b581d759b534bbe3b93c3259845d2fa4422ab5a852cc27b8a2ce1d5a28b1df",
                "MacAddress": "02:42:c0:a8:cb:02",
                "IPv4Address": "192.168.203.2/24",
                "IPv6Address": ""
            },
            "e8ebf91bcda51637c6b92eeb299434a301d6bc00335b390b67ed70ef9af6f299": {
                "Name": "mailu_admin_1",
                "EndpointID": "e470f9f4220a9330223950db2acba4a9320493b2845c3d3f067e71e0064f35f0",
                "MacAddress": "02:42:c0:a8:cb:06",
                "IPv4Address": "192.168.203.6/24",
                "IPv6Address": ""
            },
----Snip---
```
It seems that containers were assigned new IP addresses when restarted but the front container is still using the IP it had saved when it first started.
In my example `front` is trying to connect to `admin` using an IP address that now belongs to the `antivirus` container.

The [documentation ](https://mailu.io/master/configuration.html#infrastructure-settings)for the master branch indicates that the startup scripts will resolve HOST_* to their IP addresses and store the result in *_ADDRESS for further use.

To resolve this "drift" I have to stop and start all containers so the IP address can be refreshed, subsequent restarts for some containers result in the same issue.
This behaviour should at least be documented for users that are upgrading from 1.7 and definitely before the *_ADDRESS updates are backported to 1.7.

tvelocity:
## What type of PR?

The *_ADDRESS variables that are initialized at container startup cannot be relied upon, since any IP addresses in them are not guaranteed to remain valid (e.g. when restarting a container). This directly causes issues mentioned in #1341 and is also related with #1430 

The complexity of having two separate addressing mechanisms for Mailu containers is also mentioned in #1133 and #1013 

To avoid any issues with stale addresses, the strategy of resolving host addresses at startup needs to be dropped. Mailu containers can use the HOST_* environment variables directly to reach other containers regardless of their current IP address.

This also means that users with a broken DNS setup, will need to deal with the problem at the source instead of relying on a workaround that makes Mailu unreliable for everyone else.

### Related issue(s)
- Closes #1341 
- Related with #1430 
- Removes complexity mentioned in #1133 and #1013 
- Reverts startup address resolution from #886 / #827

## Prerequistes


micw:
Hello,
I do not exactly understand the purpose of this PR. If you want dynamic address resolution (which absolutely makes sense), just set *_ADDRESS raher than HOST_* - that's it. See the helm chart as an example that actually does this.

tvelocity:
> Hello,
> I do not exactly understand the purpose of this PR. If you want dynamic address resolution (which absolutely makes sense), just set `*_ADDRESS` raher than `HOST_*` - that's it. See the helm chart as an example that actually does this.

Hello,

Thanks for your feedback. I think it is ok to disagree, however it is not clear to me what the misunderstanding is either. Is there something more specific that is unclear?

The current default behavior directly causes #1341, and I don't seem to be the only one that finds it very very problematic as there are further discussions about this point in #1430 #1133 and #1013. I also don't think that the existence of a workaround makes this any less of an issue, since without it Mailu is broken out of the box.

Docker assumes dynamic addresses for containers, so #1341 likely affects 99% of Mailu setups, and it is only a matter of time and chance until any of them stops working after a restart, reconfiguration or update. This is true for vanilla docker, docker-compose and Kubernetes.

This leaves the user either restarting or recreating his Mailu containers repeatedly until they somehow come up again, or to try to figure out the current behavior, understand the difference between HOST\_ and \_ADDRESS, and finally use \_ADDRESS to override what Mailu does by default.

I think those are valid reasons to remove the startup address resolution. On the other hand, as much as I have tried, I have been unable to find a good reason to keep it. As mentioned in #1341, the startup address resolution should cause issues even in the scenarios that it was supposed to help with (IPv6 DNS lookup bugs in Docker).

**TL; DR:** The \_ADDRESS override's whole purpose is to work around the startup address resolution from HOST\_ to \_ADDRESS. If you are not using it, then you are hit by #1341. If you haven't experienced this issue, you just haven't restarted Mailu enough times yet.

For these reasons, I think it is valid to question whether the startup resolution should even exist in the first place. In my opinion it shouldn't. It adds a layer of complexity, and causes issues without really fixing the issues it was supposed to fix.

micw:
Hello,

you are right, the HOST_* brings a lot of issues (that's why I added *_ADDRESS and use it in the k8s helm chart). There had been reasons to keep it (but I do not remember what those reasons were and if they still are valid). I'd discuss it in #1013 (which is hopefully the right place) and if the others agree, we can test/merge this.

Kind regards,
Michael.

micw:
bors try

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ The Dovecot and Postfix overrides are moved in their own sub-directory.
 If there are local override files, they will need to be moved from overrides/ to overrides/dovecot and overrides/postfix/.
 See https://mailu.io/1.8/faq.html#how-can-i-override-settings for all the mappings.
 
+Resolution of HOST_* environment variables at startup, and the corresponding *_ADDRESS variables have been removed.
+Use the HOST_* variables directly instead.
+
```
lub, line 16: This change would be in an eventual 1.9, not 1.8. Not sure if it makes sense to clear the entire section when releasing to add new entries like this or alternatively to change it in a single commit when doing the release. Either way currently it's mixed.

docs/releases.rst
```diff
@@ -37,6 +37,12 @@ Please not that the shipped image for PostgreSQL database is deprecated.
 The shipped image for PostgreSQL is not maintained anymore from release 1.8.
 We recommend switching to an external database as soon as possible.
 
+Removal of HOST_* address resolution
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+The startup scripts will no longer resolve ``HOST_*`` to their IP addresses.
+
+Previously, ``*_ADDRESS`` variables could directly be set by the user to override the startup address resolution. Since container addresses are not resolved at startup anymore, this override has been removed and ``HOST_*`` variables can be used directly.
+
```
lub, line 45: Kinda similarly to CHANGELOG.md this is in the 1.8 section, although 1.8 already has branched off.

towncrier/newsfragments/1341.removal
```diff
@@ -0,0 +1 @@
+Remove startup resolution of HOST_* variables to *_ADDRESS
```
lub, line 1: Normally the file name points to the PR (#1716), not to the initial issue. That way you can click on it in releases.rst and jump directly to the PR that introduced the described change.

lub (changes requested):
Using this PR a couple of days now and the changes seem sensible. As stated in https://github.com/Mailu/Mailu/issues/1341#issuecomment-711086481 I've also bypassed it via the 'old ' `*_ADDRESS` variables and couldn't find any issues with it.

But I've commented on some things regarding the documentation for the changelog(s). 1.8 is already branched off, so further changes should imo be introduced in a separate 1.9 section or omitted for now and added when doing the release.

friedrich:
@tvelocity are you going to address the comments by @lub? It would be great if this could be merged.

Diman0:
@tvelocity Are you planning to update this PR?

Diman0 (approved):
@lub  I have applied your suggestions. I made the newsfragment verbose and added the instructions to it that we should not forget to update changelog.md/releases.rst. Can you review this PR again?

micw:
I have read through the PR. What bothers me is that this is a breaking change that would instantly make a lot of installations fail as soon as a new image is pulled. That's especially true for installations that use `*_ADDRESS` to already to dynamic host resolution (e.g. helm chart deployments). I think we should have some intermediate solution to support both syntax.

Another thing I saw is that HOST_* variables sometimes contain ports, sometimes not. When we do a breaking change, we should maybe use the chance to sort this out.

And last, I remember that there's one place where we still need host-to-ip resolution. Nginx as mail proxy asks the backend (admin api) for an MTA to proxy to. This _must_ return ip+port and not a hostname, so resolution must be done within admin service. I have not verified wether this is untouched by this PR or not.

Diman0:
If this is changed, then it would be a change in a new release. It will not touch existing deployments for people who use mailu 1.7 or 1.8. 
This only has an effect for an upgrade scenario where you are warned of the change in the release notes or a brand new deployment. 

Anyway, this PR does exactly what you want. Only implement dynamic dns resolution. The _ADDRESS workaround is not required anymore. It is now the default behaviour. It also simplifies our installation a lot. We only have one method of name resolution.

The current solution does not work (static IP assignments). Each time when you restart a container it is possible that mailu does not come up (due to ip addresses changing). In such a situation an administrator must manually log on the server and recreate all containers until it finally works again. I have experienced this multiple times already. I rather have dynamic ip resolution that always returns the correct IP address.

>Another thing I saw is that HOST_* variables sometimes contain ports, sometimes not. When we do a breaking change, we should maybe use the chance to sort this out.

What do you mean with this?Can perhaps highlight the line you are referring to?

>And last, I remember that there's one place where we still need host-to-ip resolution. Nginx as mail proxy asks the backend (admin api) for an MTA to proxy to. This must return ip+port and not a hostname, so resolution must be done within admin service. I have not verified wether this is untouched by this PR or not.

Is this situation not covered by the core and filter test? Both send an email to mailu. The tests also connect to read the email. Bors try succeeded. @lub also mentioned that he has run with this PR for multiple days without any issues. @lub can you confirm you also received email when testing the PR and thus can confirm that the nginx mail proxy works?

lub:
@Diman0 I use this fork on my production system: https://github.com/lub/Mailu/commits/master

So yes, I do receive emails with it. But it needs a rebase before merging iirc.

Diman0:
@lub would you be able to rebase it? I do not have a lot of experience with this.

micw:
In discussion we found another objection. At the moment, hostnames are resolved at system level, so service names and full qualified hostnames are both OK. With the change, we always require full qualified hostnames, especially for the "front" container. We need to investigate here and make it work with service names (because full name is not always known).

Diman0:
After the Mailu 1.9 release, we will have a look at implementing this PR. 
We should also have a look at Socrates at the same time. The part that handles DNS resolution in socrates.

Diman0:
PR #2460 has been merged which switches Mailu to dynamic DNS resolution. I'm closing this PR.

Issue #1720, opened 2021-01-06
Mailu affected by dovecot CVE-2020-25275, CVE-2020-24386, CVE-2020-12100


sbias:
CVE-2020-25275
Dovecot before 2.3.13 has Improper Input Validation in lda, lmtp, and imap, leading to an application crash via a crafted email message with certain choices for ten thousand MIME parts.

CVE-2020-24386
An issue was discovered in Dovecot before 2.3.13. By using IMAP IDLE, an authenticated attacker can trigger unhibernation via attacker-controlled parameters, leading to access to other users' email messages (and path disclosure).

by checking the dovecot-version in mailu imap container (master-tag) i get version 2.3.11.3 , so i'm not sure if that fix has been backported or not.

```
bash-5.0# dovecot --version
2.3.11.3 (502c39af9)
```

peter-englmaier:
in 1.7 is still 2.3.10.1. I hope there will be an update soon. Are you running 1.8 or master?

peter-englmaier:
According to this: https://repology.org/project/dovecot/versions
the upstream alpine has not released a version with the fix (only release Edge has dovecot 2.3.13).
However, 2.3.11.3 is in alpine 3.12 which is used in master. We will have to wait for next alpine release or build our own dovecot based on alpine edge. Not sure if it won't break everything in mailu 1.17.

mawilmsen:
Voting for a fix in 1.8, which is running on alpine 3.12 afaik!

micw:
I created an issue in the alpine bug tracker: https://gitlab.alpinelinux.org/alpine/aports/-/issues/12286
Hope they get it fixed very soon, otherwise we need to find a workaround.

peter-englmaier:
Thanks micw, that was a good idea. I could build the container with 'alpine:edge' for version 1.7, but that doesn't mean it will work. I do not want to lose email, so I keep waiting for an official fix from the mailu developers.

peter-englmaier:
Ok, I jumped the line and tried it out:

```
--- a/core/dovecot/Dockerfile
+++ b/core/dovecot/Dockerfile
@@ -1,4 +1,4 @@
-ARG DISTRO=alpine:3.12
+ARG DISTRO=alpine:edge

```
then, I have build with (git branch master)
`docker build . -t dovecot:myversion
`
and modified my docker-compose to use image `dovecot:myversion`.

This worked, I can still receive email and my dovecot version is now:
`2.3.13 (89f716dc2)
[...]

PR #1736, opened 2021-01-16
Fix dovecot cve

micw:
## What type of PR?

bug-fix

## What does this PR do?

Upgrade dovecot alpine to 3.13 to fix CVEs in dovecot

### Related issue(s)

- #1720

## Prerequistes


micw:
Oh, did not work as I expected...

PR #1743, opened 2021-01-21
Doc fixes

rbarazzutti:
## Documentations fixes

- 82b5920 typos
- 444529b PostgreSQL deprecation (reword)

### DIFF
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -33,7 +33,7 @@ Upgrading
 Upgrade should run fine as long as you generate a new compose or stack
 configuration and upgrade your mailu.env.
 
-Please not that the shipped image for PostgreSQL database is deprecated.
+Please note that the shipped image for PostgreSQL database is deprecated.
 The shipped image for PostgreSQL is not maintained anymore from release 1.8.
 We recommend switching to an external database as soon as possible.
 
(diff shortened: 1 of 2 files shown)


docs/releases.rst
```diff
@@ -33,9 +33,9 @@ Upgrading
 Upgrade should run fine as long as you generate a new compose or stack
 configuration and upgrade your mailu.env.
 
-Please not that the shipped image for PostgreSQL database is deprecated.
+Please note that the shipped image for PostgreSQL database is deprecated.
 The shipped image for PostgreSQL is not maintained anymore from release 1.8.
-We recommend switching to an external database as soon as possible.
+We recommend switching to another database as soon as possible (SQLite or MySQL).
```
lub, line 38:
```suggestion
We recommend switching to an external PostgreSQL database as soon as possible.
```

PostgreSQL will be supported in future, just the Mailu PostgreSQL image will be dropped. No need to switch to SQLite or MySQL.

lub (changes requested):
Thanks for the fixes, but I think there is a misunderstanding regarding the PostgreSQL image. I hope I could make it clearer :)

rbarazzutti:
Thanks for the feedback @lub! Good to learn that PostgreSQL will still have support in the future.

lub (approved):

Diman0 (approved):

PR #1745, opened 2021-01-24
Make authentication fast

### LINKED ISSUE #1411: Cache password verification in-memory
Currently our users must decide between strong password hashes and good system performance. The reason is that password hashes needs to be verified on every single request (e.g. on every click on webmail). With strong hashes, a single user can easily max-out multiple cpu cores.

Proposal:
- on the admin service, do a verification of passwords against the strong hash in the database.
- on successfull auth, cache the email + strong hash along with a weak/fast hash of the password in memory
- on each auth attempt, 1st check if there's a weak hash cached with the strong hash for that email address. If so, validate the password against the weak/fast hash instead.

With this change, password probe attempts are still slowed down. Passwords are still stored secure in database. Weak hashes are only cached in-memory. Auth for valid users will be speed up a lot.

Comments are welcome!

nextgens:
## What type of PR?

It's an enhancement that adds a credential cache and optimize the logic to speedup authentication requests.

## What does this PR do?

- it fixes rate-limiting and token authentication on /webdav/
- it enhances the logic to ensure that no CPU cycles are wasted when valid credentials are found
- it lowers the number of rounds used for token storage (on the basis that they are high-entropy: not bruteforceable and speed matters)
- it implements an auth-cache as described in #1194 and #1411. As it stands that auth-cache is per-guicorn and not global

### Related issue(s)
- optimize the authentication logic to ensure that no further attempt are made when valid authentication credentials are found (see #1194 and  close #1411)

## Prerequistes

### DIFF
--- a/core/admin/mailu/__init__.py
+++ b/core/admin/mailu/__init__.py
@@ -1,5 +1,6 @@
 import flask
 import flask_bootstrap
+from flask_caching import Cache
 
 from mailu import utils, debug, models, manage, configuration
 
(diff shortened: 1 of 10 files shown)


nextgens:
bors retry

nextgens:
I've split this up in other PRs (#1753 #1754 #1755 #1756). Rate limiting related changes are still undergoing discussion

PR #1748, opened 2021-01-31
For travis Ci: Need to docker login before pulling images

ofthesun9:
To avoid triggering the Download rate limit from Docker Hub

## What type of PR?
enhancement

## What does this PR do?
This PR add a docker login cmd before launching the build script


ofthesun9:
bors try

Diman0 (approved):
Good catch. Lgtm

PR #1753, opened 2021-02-07
Better password storage

### LINKED ISSUE #1194: Make PBKDF2 the default password scheme

### LINKED ISSUE #1662: Support more password schemes from passlib
## Before you open your issue

### Description
Mailu uses the passlib python library for handling password hashes. The password hashes that can be used must be passed to the constructor of passlib. These are currently the [supported hashes](https://github.com/Mailu/Mailu/blob/90984c4d2bd8b0f82559f26d9837a7454626438a/core/admin/mailu/models.py#L365).

```
    scheme_dict = {'PBKDF2': "pbkdf2_sha512",
                   'BLF-CRYPT': "bcrypt",
                   'SHA512-CRYPT': "sha512_crypt",
                   'SHA256-CRYPT': "sha256_crypt",
                   'MD5-CRYPT': "md5_crypt",
                   'CRYPT': "des_crypt"}
```

In the mailu chat someone asked for support SSHA512 (salted ssh512).  It would be nice if we could support more hashing algorithms. This enables users to export/import users from other systems that use these hash algorithms.

#### What must happen for implementing this enhancement request?
1. modify requirements.txt to use a newer version of passlib to have support for newer hashes. Version 1.7.1 of passlib is currently used. Version 1.7.3 also has support for salted sha512.
2. modify models.py with additional hash algorithms. eg.
```
    scheme_dict = {'PBKDF2': "pbkdf2_sha512",
                   'BLF-CRYPT': "bcrypt",
                   'SHA512-CRYPT': "sha512_crypt",
                   'SHA256-CRYPT': "sha256_crypt",
                   'MD5-CRYPT': "md5_crypt",
                   'CRYPT': "des_crypt",
                   'SSHA256': "ldap_salted_sha256"
                   'SSHA512': "ldap_salted_sha512"  }
```

The proper solution is probably to make it possible to configure the scheme_dict via the Mailu.env file. This way you can still add support for other hashes via the mailu.env file.

nextgens:
## What type of PR?

Enhancement: optimization of the logic to speedup authentication requests, support the import of most hashes passlib supports.

## What does this PR do?

- it changes the default password cold-storage format to sha256+bcrypt
- it enhances the logic to ensure that no CPU cycles are wasted when valid credentials are found
- it fixes token authentication on /webdav/
- it lowers the number of rounds used for token storage (on the basis that they are high-entropy: not bruteforceable and speed matters)
- it introduces a new setting to set the number of rounds used by the password hashing function (CREDENTIAL_ROUNDS). The setting can be adjusted as required and existing hashes will be migrated to the new cost-factor.
- it updates the version of passlib in use and enables all supported hash types (that will be converted to the current settings on first use)
- it removes the PASSWORD_SCHEME setting

### Related issue(s)
- close #1194
- close #1662
- close #1706

## Prerequistes

### DIFF
--- a/core/admin/mailu/configuration.py
+++ b/core/admin/mailu/configuration.py
@@ -33,7 +33,6 @@ DEFAULT_CONFIG = {
     'TLS_FLAVOR': 'cert',
     'AUTH_RATELIMIT': '10/minute;1000/hour',
     'AUTH_RATELIMIT_SUBNET': True,
-    'CREDENTIAL_ROUNDS': 12,
     'DISABLE_STATISTICS': False,
     # Mail settings
     'DMARC_RUA': None,
@@ -55,6 +54,7 @@ DEFAULT_CONFIG = {
     # Advanced settings
     'LOG_LEVEL': 'WARNING',
     'SESSION_COOKIE_SECURE': True,
+    'CREDENTIAL_ROUNDS': 12,
     # Host settings
     'HOST_IMAP': 'imap',
     'HOST_LMTP': 'imap:2525',
--- a/core/admin/mailu/configuration.py
+++ b/core/admin/mailu/configuration.py
@@ -33,6 +33,7 @@ DEFAULT_CONFIG = {
     'TLS_FLAVOR': 'cert',
     'AUTH_RATELIMIT': '10/minute;1000/hour',
     'AUTH_RATELIMIT_SUBNET': True,
+    'CREDENTIAL_ROUNDS': 12,
     'DISABLE_STATISTICS': False,
     # Mail settings
     'DMARC_RUA': None,
@@ -52,7 +53,6 @@ DEFAULT_CONFIG = {
     'RECAPTCHA_PUBLIC_KEY': '',
     'RECAPTCHA_PRIVATE_KEY': '',
     # Advanced settings
-    'PASSWORD_SCHEME': 'PBKDF2',
     'LOG_LEVEL': 'WARNING',
     'SESSION_COOKIE_SECURE': True,
     # Host settings
--- a/core/admin/mailu/internal/nginx.py
+++ b/core/admin/mailu/internal/nginx.py
@@ -49,11 +49,14 @@ def handle_authentication(headers):
         user = models.User.query.get(user_email)
         status = False
         if user:
-            for token in user.tokens:
-                if (token.check_password(password) and
-                    (not token.ip or token.ip == ip)):
-                        status = True
-            if user.check_password(password):
+            # All tokens are 32 characters hex lowercase
+            if len(password) == 32:
+                for token in user.tokens:
+                    if (token.check_password(password) and
+                        (not token.ip or token.ip == ip)):
+                            status = True
+                            break
+            if not status and user.check_password(password):
                 status = True
             if status:
                 if protocol == "imap" and not user.enable_imap:
--- a/core/admin/mailu/internal/nginx.py
+++ b/core/admin/mailu/internal/nginx.py
@@ -19,6 +19,20 @@ STATUSES = {
     }),
 }
 
+def check_credentials(user, password, ip, protocol=None):
+    if not user or not user.enabled or (protocol == "imap" and not user.enable_imap) or (protocol == "pop3" and not user.enable_pop):
+        return False
+    is_ok = False
+    # All tokens are 32 characters hex lowercase
+    if len(password) == 32:
+        for token in user.tokens:
+            if (token.check_password(password) and
+                (not token.ip or token.ip == ip)):
+                    is_ok = True
+                    break
+    if not is_ok and user.check_password(password):
+        is_ok = True
+    return is_ok
 
 def handle_authentication(headers):
     """ Handle an HTTP nginx authentication request
@@ -47,23 +61,7 @@ def handle_authentication(headers):
         password = raw_password.encode("iso8859-1").decode("utf8")
         ip = urllib.parse.unquote(headers["Client-Ip"])
         user = models.User.query.get(user_email)
-        status = False
-        if user:
-            # All tokens are 32 characters hex lowercase
-            if len(password) == 32:
-                for token in user.tokens:
-                    if (token.check_password(password) and
-                        (not token.ip or token.ip == ip)):
-                            status = True
-                            break
-            if not status and user.check_password(password):
-                status = True
-            if status:
-                if protocol == "imap" and not user.enable_imap:
-                    status = False
-                elif protocol == "pop3" and not user.enable_pop:
-                    status = False
-        if status and user.enabled:
+        if check_credentials(user, password, ip, protocol):
             return {
                 "Auth-Status": "OK",
                 "Auth-Server": server,
--- a/core/admin/mailu/internal/views/auth.py
+++ b/core/admin/mailu/internal/views/auth.py
@@ -53,22 +53,10 @@ def basic_authentication():
         encoded = authorization.replace("Basic ", "")
         user_email, password = base64.b64decode(encoded).split(b":")
         user = models.User.query.get(user_email.decode("utf8"))
-        if user and user.enabled:
-            password = password.decode('utf-8')
-            status = False
-            # All tokens are 32 characters hex lowercase
-            if len(password) == 32:
-                for token in user.tokens:
-                    if (token.check_password(password) and
-                        (not token.ip or token.ip == ip)):
-                            status = True
-                            break
-            if not status and user.check_password(password):
-                status = True
-            if status:
-                response = flask.Response()
-                response.headers["X-User"] = user.email
-                return response
+        if nginx.check_credentials(user, password.decode('utf-8'), flask.request.remote_addr, "web"):
+            response = flask.Response()
+            response.headers["X-User"] = user.email
+            return response
     response = flask.Response(status=401)
     response.headers["WWW-Authenticate"] = 'Basic realm="Login Required"'
     return response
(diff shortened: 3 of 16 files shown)


nextgens:
bors retry

core/admin/mailu/models.py
```diff
@@ -493,10 +507,18 @@ class Token(Base):
...
+                db.session.add(self)
+                db.session.commit()
+                return True
+            return False
+        return hash.pbkdf2_sha256.verify(password, self.password)
 
     def set_password(self, password):
-        self.password = hash.sha256_crypt.using(rounds=1000).hash(password)
+        # tokens have 128bits of entropy, they are not bruteforceable
+        self.password = hash.pbkdf2_sha256.using(rounds=1).hash(password)
```
lub, line 521: Why should the tokens use pbkdf2_sha256 with 1 round instead of e.g. bcrypt with 1 round?

core/admin/mailu/models.py
```diff
@@ -493,10 +507,18 @@ class Token(Base):
...
     def check_password(self, password):
-        return hash.sha256_crypt.verify(password, self.password)
+        if self.password.startswith("$5$"):
+            if hash.sha256_crypt.verify(password, self.password):
+                self.set_password(password)
+                db.session.add(self)
+                db.session.commit()
+                return True
+            return False
+        return hash.pbkdf2_sha256.verify(password, self.password)
```
lub, line 517: Is there a particular reason why the tokens have a hash detection on their own instead of relying on passlib?

docs/configuration.rst
```diff
@@ -138,9 +138,7 @@ Depending on your particular deployment you most probably will want to change th
 Advanced settings
 -----------------
 
-The ``PASSWORD_SCHEME`` is the password encryption scheme. You should use the
-default value, unless you are importing password from a separate system and
-want to keep using the old password encryption scheme.
+The ``CREDENTIAL_ROUNDS`` (default: 12) is the number of rounds used by the password hashing scheme. You should use the default value.
```
lub, line 141:
```suggestion
The ``CREDENTIAL_ROUNDS`` (default: 12) is the number of rounds used by the password hashing scheme. The number of rounds can be reduced in case faster authentication is needed or increased when more paranoid hashes are desired.
It normally shouldn't be changed, as the default value is explicitly chosen to be secure while still being fast enough on most systems.
```

I think this should be a little more expressive considering we discussed this very value in matrix :)

towncrier/newsfragments/1194.misc
```diff
@@ -0,0 +1 @@
+Switch to bcrypt_sha256, remove PASSWORD_SCHEME and replace it with CREDENTIAL_ROUNDS
```
lub, line 1:
```suggestion
Switch to bcrypt_sha256, replace PASSWORD_SCHEME with CREDENTIAL_ROUNDS and update existing hashes on first login
```

This PR fixes multiple issues, so it's a little bit difficult to express it properly in one newsfragment and I think it sometimes makes sense to also link to other issues like you did. But imo at least one of them should link to this pr, so users can easily find the actual change.

I'd suggest to link this line to #1753, because in case someone has problems with changes introduced by this PR after upgrading they'll probably suspect this change to be the reason for it after grepping for `password` or `login`.

towncrier/newsfragments/1662.feature
```diff
@@ -0,0 +1 @@
+Enable support of all hash types passlib supports. Convert them to the default scheme on first use.
```
lub, line 1:
```suggestion
Enable support of all hash types passlib supports.
```

In accordance with the other suggestion.

lub (changes requested):
Thank you very much, although the review took me longer than anticipated. But that's not your fault :)

I've left some thoughts and suggestions. Would be great if you could have a look at them, although I suspect you already have more experience with that part of the code than me. So maybe I missed some things.

I haven't tested the code on my system(s) yet, in case you want to incorporate some of the notes. But from reading it I didn't find any bugs.

core/admin/mailu/models.py
nextgens: Because it's a lot faster and there is no need for any strengthening here (128 bit tokens aren't brute-forceable regardless of the hash function)

core/admin/mailu/models.py
nextgens: Simplicity. It's a lot more verbose to use a context... and unlike for passwords we are 100% sure of what we are working with here.

If you feel strongly about it I'll change it, but IMHO it's fine and explicit like this

towncrier/newsfragments/1662.feature
nextgens: That's 537bb3c

towncrier/newsfragments/1194.misc
nextgens: That's 537bb3c

docs/configuration.rst
nextgens: That's f67c9a9

core/admin/mailu/internal/views/auth.py
```diff
@@ -53,10 +53,22 @@ def basic_authentication():
...
+            # All tokens are 32 characters hex lowercase
+            if len(password) == 32:
+                for token in user.tokens:
+                    if (token.check_password(password) and
+                        (not token.ip or token.ip == ip)):
+                            status = True
+                            break
+            if not status and user.check_password(password):
+                status = True
+            if status:
```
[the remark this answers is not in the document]
nextgens: Good idea, that's d9aa226

core/admin/mailu/internal/nginx.py
```diff
@@ -49,11 +49,14 @@ def handle_authentication(headers):
         user = models.User.query.get(user_email)
         status = False
         if user:
-            for token in user.tokens:
-                if (token.check_password(password) and
-                    (not token.ip or token.ip == ip)):
-                        status = True
-            if user.check_password(password):
+            # All tokens are 32 characters hex lowercase
+            if len(password) == 32:
```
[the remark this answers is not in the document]
nextgens: That's 2fa97c9

core/admin/mailu/configuration.py
```diff
@@ -33,6 +33,7 @@
     'TLS_FLAVOR': 'cert',
     'AUTH_RATELIMIT': '10/minute;1000/hour',
     'AUTH_RATELIMIT_SUBNET': True,
+    'CREDENTIAL_ROUNDS': 12,
```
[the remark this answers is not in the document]
nextgens: That's a2d68ba

core/admin/mailu/models.py
```diff
@@ -370,17 +370,34 @@ def reply_active(self):
...
-        result = context.verify(password, reference)
-        if result and context.identify(reference) != context.default_scheme():
-            self.set_password(password)
+        reference = self.password
+        # strip {scheme} if that's something mailu has added
+        # passlib will identify *crypt based hashes just fine
+        # on its own
+        if self.password.startswith("{"):
+            scheme = self.password.split('}')[0][1:]
+            if scheme in self.scheme_dict:
```
ghostwheel42, line 395: What about moving the scheme_dict into the check_password method, as it is only for detection of legacy hashes and should not be used outside the model anymore:
`if scheme in {'PBKDF2', 'BLF-CRYPT', 'SHA512-CRYPT', 'SHA256-CRYPT', 'MD5-CRYPT', 'CRYPT'}:`

nextgens: Good idea; that's abd099d

core/admin/mailu/models.py
ghostwheel42: I'm implementing a config import/export (json/yaml) using marshmallow and I would need to verify the imported hash is using the correct algorithm if this is implemented without passlib's autodetection.
So my first thought is: check the token like the user password. 
But there are other problems arising when implementing tokens in this fixed way. For example I am unable to make sure the hashed token has the correct length.
Perhaps this just has to be documented somewhere so anyone mass-creating and importing tokens via config-import does not use tokens with length 64...
Alex

nextgens: Just to clarify: the code it replaces wasn't doing any validation nor auto-detection either... 

This is mixing two different problems: import/export is fine... The question is whether mailu should add complexity to allow 3rd party tools to modify what has been imported/exported (and/or standardize what the blob is made of). 

IMHO the answer to that question is "no". 3rd party tools should use the API to get mailu to generate token it likes/can work with... This enables for simplifications and optimizations to be made (like avoiding the token checking code altogether if the tokens aren't in the desired format).

ghostwheel42: You are correct - the user_import function for example is not really checking its input now.
I also think anyone fiddling around with the exported yaml/json should know what he/she is doing.
As you said: when using the admin interface or api everything will be fine (as mailu is controlling the creation).

It's okay for me to leave the model the way it is (not too much verification) and verify user input on first touch (in multiple places: cmdline, api, ...) - I'll try not to duplicate too much code.

core/admin/mailu/models.py
lub: After benchmarking it I honestly am surprised, that it's that much faster (espeically since the minimum bcrypt rounds in passlib is 4). So nevermind then.

core/admin/mailu/models.py
lub: Seems it's not only less verbose, but constructing a context every time also takes ~three times as long on my notebook. Considering that there will be many tokens, but only one proper password I think it makes sense to diverge here.

If it's fine for you we can resolve this conversation.

lub (dismissed):

lub (approved):
thank you for also adding context caching :)

nextgens:
@ghostwheel42 any chance you can update your review to get it merged?

ghostwheel42 (approved):
@nextgens sorry, I somehow overlooked this.
But I think there's just a bors try missing, isn't it?

Diman0 (dismissed):
Looks good to me. Thank you for this PR. It is a great improvement for mailu.

Diman0:
bors retry

Diman0:
bors retry

PR #1755, opened 2021-02-07
Implement a credential cache (alternative to #1758)

### LINKED ISSUE #1194: Make PBKDF2 the default password scheme

### LINKED ISSUE #1411: Cache password verification in-memory

nextgens:
## What type of PR?

Feature: it implements a credential cache to speedup authentication requests.

## What does this PR do?

Credentials are stored in cold-storage using a slow, salted/iterated hash function to prevent offline bruteforce attacks. This creates a performance bottleneck for no valid reason (see the
rationale/long version on https://github.com/Mailu/Mailu/issues/1194#issuecomment-762115549).

The new credential cache makes things fast again.

Keep in mind that the cache is local to each gunicorn.

There is an argument to be said about making it less configurable (remove the new knobs, no new dependency) ... that is at #1758

### Related issue(s)
- close #1411
- close #1194 
- close #1758

## Prerequistes

### DIFF
--- a/core/admin/mailu/__init__.py
+++ b/core/admin/mailu/__init__.py
@@ -1,5 +1,6 @@
 import flask
 import flask_bootstrap
+from flask_caching import Cache
 
 from mailu import utils, debug, models, manage, configuration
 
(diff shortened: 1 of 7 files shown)


nextgens:
@micw Can you review this PR please? Are you okay with flask_caching as a new dependency?

nextgens:
#1758 is better

PR #1756, opened 2021-02-07
Ensure that flashed messages are displayed

nextgens:
## What type of PR?

Bugfix: display flashed messages

## What does this PR do?

It ensure that we display the flashed messages somewhere. Without it basic things like authentication failures give no feedback!

### DIFF
--- a/core/admin/mailu/ui/templates/base.html
+++ b/core/admin/mailu/ui/templates/base.html
@@ -38,6 +38,15 @@
 
         <section class="content">
           {% block content %}{% endblock %}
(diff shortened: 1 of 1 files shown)


lub:
Hi,

I've digged a little bit in the git log, because I was curious why flash() is called while it's not displayed anywhere. Turns out it was displayed some time ago via bootstrap utils, but somehow got removed while moving to webpack?

I've created #1763 to re-introduce the previous behavior, which imo results in prettier and more obvious messages.

image

nextgens:
#1763 is better

PR #1760, opened 2021-02-11
Security updates to postgresql

WebSpider:
## What type of PR?

Security update

## What does this PR do?

It fixes vulnerabilities in the sudo package in the postgresql optional container documented in
CVE-2021-23240, CVE-2021-3156 and CVE-2021-23239

### Related issue(s)

None

## Prerequistes

### DIFF
--- a/optional/postgresql/Dockerfile
+++ b/optional/postgresql/Dockerfile
@@ -3,6 +3,7 @@ FROM $DISTRO
 # python3 shared with most images
 RUN apk add --no-cache \
     python3 py3-pip bash py3-multidict \
+  && apk add --upgrade sudo \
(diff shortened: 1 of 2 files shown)


nextgens:
Wouldn't upgrading to a patched version of alpine be a better solution?

WebSpider:
It would be, but given the state of this container (in https://github.com/Mailu/Mailu/projects/3#card-27739555) I figured let's keep the changes minimal

Diman0 (approved):
Looks good to me. The PostgreSQL image is indeed deprecated. That is no reason to not accept patches for it.

WebSpider:
bors try

nextgens (approved):

PR #1761, opened 2021-02-12
check for `ipv6_enabled` in the compose template

lub:
Checking only `ipv6` isn't sufficient, because it has a default value.

## What type of PR?

bug-fix

### Related issue(s)
-

## Prerequistes

- [it's a minor change] Unless it's docs or a minor change: add [changelog](https://mailu.io/master/contributors/guide.html#changelog) entry file.


Diman0 (approved):
Lgtm

Diman0:
bors retry

PR #1763, opened 2021-02-13
show flash messages again

lub:
## What type of PR?

bug-fix

## What does this PR do?
This basically restores the behaviour, that got removed in
ecdf0c25b3d9bbaa028bdc46cb721d2fb406dde2 during refactoring.

### Related issue(s)
- noticed it while reviewing #1756

## Prerequistes

- [-] In case of feature or enhancement: documentation updated accordingly
- [-] Unless it's docs or a minor change: add [changelog](https://mailu.io/master/contributors/guide.html#changelog) entry file.

### DIFF
--- a/core/admin/mailu/ui/templates/base.html
+++ b/core/admin/mailu/ui/templates/base.html
@@ -1,4 +1,5 @@
 {% import "macros.html" as macros %}
+{% import "bootstrap/utils.html" as utils %}
 <!doctype html>
 <html>
   <head>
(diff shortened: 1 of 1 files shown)


nextgens (approved):

nextgens:
Another good reason to have this backported is that the flashed messages are stored in the session (currently in a client-side cookie) that will never be drained... it will grow in size up to the point where performance will suck (it gets uploaded with every single request) and things will break (4k max according to rfc2109)

Diman0 (approved):
lgtm

Issue #1740, opened 2021-01-19
Gunicorn in the admin container does not pass the X-Forwarded-Proto header to Flask


verdel:
### Environment
 - [x] docker-compose
 - [ ] kubernetes
 - [ ] docker swarm

### Versions
I use version 1.7

## Description
I use a scheme with Nginx that handles SSL connections. This Nginx stands in the schema above the Mailu stack. The `TLS_FLAVOR` environment variable has the value `mail`.

When I access endpoint `/admin` using the `https` scheme, I end up getting a redirect to `/admin/ui/`, but with the `http` scheme specified.

This is because gunicorn doesn't allow to handle set secure headers by [default](https://docs.gunicorn.org/en/stable/settings.html#forwarded-allow-ips).
Secure headers are:
 - X-Forwarded-Protocol
 - X-Forwarded-Proto
 - X-Forwarded-SSL

## Options for solving the problem
 - **Less Secure**: Add `--forwarded-allow-ips='*'` to gunicorn execution command [here](https://github.com/Mailu/Mailu/blob/6518ef19af92730fd99b9c148d279a971b856cd8/core/admin/start.py#L21)
 - **More Secure**: Add `--forwarded-allow-ips=` but with `front` container IP. You can get it using `socrate`, as in `config.py` in the `front` container

Diman0:
It is very common to have a reverse proxy in front of mailu.
I cannot replicate your issue. How did you configure your nginx reverse proxy?
I used as reverse proxy the swag image of [linuxserver.io](https://docs.linuxserver.io/images/docker-swag).

This is the site config I used. Maybe this can help you configuring nginx:
```
# redirect all traffic to https
server {
        listen 80 default_server;
        listen [::]:80 default_server;
        server_name _;
        return 301 https://$host$request_uri;
        add_header Strict-Transport-Security "max-age=15552000; includeSubDomains" always;
}

#drop traffice to mydomain.com
server {
    listen 443 ssl;
    listen [::]:443 ssl;

    server_name _;
    return      444;

    include /config/nginx/ssl.conf;
}

#mailu
server {
    listen 443 ssl;
    server_name mydomain.com;
    add_header Strict-Transport-Security "max-age=15552000; includeSubDomains" always;
    include /config/nginx/ssl.conf;
    
    client_max_body_size 0;

    location /webmail {
        include /config/nginx/proxy.conf;
        resolver 127.0.0.11 valid=30s;
        proxy_pass https://front:443/webmail;
[...]

PR #1765, opened 2021-02-18
Set sensible cookie flags on the admin app

nextgens:
## What type of PR?

Bugfix

## What does this PR do?

It sets the right flags on the session cookie issued by the admin app.
This should probably be backported as the lack of secure flag on TLS-enabled setup is a high risk vulnerability.

SameSite is hardening / helps against CSRF on modern browsers
HTTPOnly is hardening / helps reduce the impact of XSS

### DIFF
--- a/core/admin/mailu/configuration.py
+++ b/core/admin/mailu/configuration.py
@@ -54,6 +54,7 @@ DEFAULT_CONFIG = {
     # Advanced settings
     'PASSWORD_SCHEME': 'PBKDF2',
     'LOG_LEVEL': 'WARNING',
+    'SESSION_COOKIE_SECURE': True,
     # Host settings
     'HOST_IMAP': 'imap',
     'HOST_LMTP': 'imap:2525',
@@ -125,7 +126,6 @@ class ConfigManager(dict):
         self.config['QUOTA_STORAGE_URL'] = 'redis://{0}/1'.format(self.config['REDIS_ADDRESS'])
         self.config['SESSION_COOKIE_SAMESITE'] = 'Strict'
         self.config['SESSION_COOKIE_HTTPONLY'] = True
-        self.config['SESSION_COOKIE_SECURE'] = self.config['TLS_FLAVOR'] != 'notls'
         # update the app config itself
         app.config = self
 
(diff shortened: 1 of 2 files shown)


core/admin/mailu/configuration.py
```diff
@@ -123,6 +123,9 @@ def init_app(self, app):
 
         self.config['RATELIMIT_STORAGE_URL'] = 'redis://{0}/2'.format(self.config['REDIS_ADDRESS'])
         self.config['QUOTA_STORAGE_URL'] = 'redis://{0}/1'.format(self.config['REDIS_ADDRESS'])
+        self.config['SESSION_COOKIE_SAMESITE'] = 'Strict'
+        self.config['SESSION_COOKIE_HTTPONLY'] = True
+        self.config['SESSION_COOKIE_SECURE'] = self.config['TLS_FLAVOR'] != 'notls'
```
lub, line 128: I don't think that check is correct. For example if someone uses mail or mail-letsencrypt they don't necessarily access admin via https. The other way around when someone uses cert or letsencrypt they could still access admin unencrypted by exposing it directly (iirc setup does this by default).

Not sure what the proper solution would be. Adding x-forwarded-proto to nginx and checking for that? But that has other caveats, e.g. we'd have to check if the request originated from nginx.

nextgens: I guess this is related to #1740
I'm not sure what the proper solution is either.

Maybe we should go further and delegate all of this to something like flask-talisman ?

nextgens: One of the complications is that we don't have a flask request context at configuration time...

I'm not convinced that auto-detecting the right value at runtime is the way forward (hard to cover all the usecases, will be difficult to debug by users, ...). What about just documenting it and allowing the default to be overridden in the configuration?

nextgens: Meh. The complexity for handling this is just too high.

https://mailu.io/1.7/general.html#philosophy
"Mailu has a central front container that routes all HTTP and mail traffic."

Can't we just stick with that? :)
And if we do, would you rather I get nginx to set the cookie flags? (set_cookie_flag https://docs.nginx.com/nginx/admin-guide/dynamic-modules/cookie-flag/)

nextgens: I've implemented what was agreed on matrix in 0dcc059

lub (approved):

Diman0 (approved):
Lgtm. Since this is a security fix, i agree we should backport it.

Diman0:
bors retry

PR #1767, opened 2021-02-20
Remove "CHUNKING" capability from nginx-smtp

### LINKED ISSUE #1766: nginx advertises SMTP CHUNKING capability which it cannot parse — possible user mail-loss
### Environment
 - [x] docker-compose
 - [ ] kubernetes
 - [ ] docker swarm

### Versions
`1.8`

## Description
With `CHUNKING`set as a capability, nginx advertises this capability to clients at a stage where the SMTP dialog does not seem to be forwarded to the proxy-target (postfix) yet. Nginx' SMTP parser itself does not support the `BDAT` command issued as part of a chunke-d dialog. This makes Nginx respond with a `250 2.0.0 OK` and close the connection, after the mail-data got sent by the client — without forwarding this to the proxy-target.
    
With this, users mail can be lost.
    
Furthermore, when a user uses a sieve filter to forward mail, dovecot sometimes chunks the forwarded mail when sending it through `front`. These forwards then fail.
    
Removing `CHUNKING` from the capabilities fixes this behavior.

For what the Nginx parser supports, see:
https://github.com/nginx/nginx/blob/1e92a0a4cef98902aed35d7b402a6a402951aba4/src/mail/ngx_mail.h#L276
And the uses of these definitions in the `src/mail` folder.

## Replication Steps
* Create a sieve filter in a users mailbox, forwarding mail to another target (e.g. mailinator)
* Send mail into this users mailbox, that matches the filter
* Vary the size of the mail sent
* Observe sieve-forward errors in `imap` containers logs
* When tcpdumping (in `front`, filtered from `imap`), observe a well-looking SMTP dialog with `front` responding `250 2.0.0 OK` before closing the socket
* When tcpdumping (in `smtp` from `front`) no attempt to proxy the connection is made by nginx
* Observe no mail actually got forwarded

## Expected behaviour
* Mail does not get silently dropped.

… PR incoming.

Nebukadneza:
## What type of PR?
bug-fix

## What does this PR do?
With `CHUNKING`set as a capability, nginx advertises this capability to
clients at a stage where the SMTP dialog does not seem to be forwarded
to the proxy-target (postfix) yet. Nginx' SMTP parser itself does not
support the `BDAT` command issued as part of a chunke-d dialog. This makes
Nginx respond with a `250 2.0.0 OK` and close the connection, after the
mail-data got sent by the client — without forwarding this to the
proxy-target.

With this, users mail can be lost.

Furthermore, when a user uses a sieve filter to forward mail, dovecot
sometimes chunks the forwarded mail when sending it through `front`.
These forwards then fail.

Removing `CHUNKING` from the capabilities fixes this behavior.

### Related issue(s)
closes #1766

### DIFF
--- a/core/nginx/conf/nginx.conf
+++ b/core/nginx/conf/nginx.conf
@@ -215,7 +215,7 @@ mail {
     {% endif %}
 
     # Advertise real capabilites of backends (postfix/dovecot)
-    smtp_capabilities PIPELINING SIZE {{ MESSAGE_SIZE_LIMIT }} ETRN ENHANCEDSTATUSCODES 8BITMIME DSN CHUNKING;
+    smtp_capabilities PIPELINING SIZE {{ MESSAGE_SIZE_LIMIT }} ETRN ENHANCEDSTATUSCODES 8BITMIME DSN;
     pop3_capabilities TOP UIDL RESP-CODES PIPELINING AUTH-RESP-CODE USER;
     imap_capabilities IMAP4 IMAP4rev1 UIDPLUS SASL-IR LOGIN-REFERRALS ID ENABLE IDLE LITERAL+;
 


nextgens (approved):

ofthesun9 (approved):

ofthesun9:
We need to back port #1748 to 1.8 (#1771 )

PR #1783, opened 2021-03-09
Switch to server-side sessions

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

It simplifies session management.
- it ensures that sessions will eventually expire (*)
- it implements some mitigation against session-fixation attacks
- it switches from client-side to server-side sessions (in Redis)

It doesn't prevent us from (re)-implementing a "remember_me" type of feature if that's considered useful by some.

### DIFF
--- a/core/admin/mailu/__init__.py
+++ b/core/admin/mailu/__init__.py
@@ -1,5 +1,8 @@
 import flask
 import flask_bootstrap
+import redis
+from flask_kvsession import KVSessionExtension
+from simplekv.memory.redisstore import RedisStore
 
 from mailu import utils, debug, models, manage, configuration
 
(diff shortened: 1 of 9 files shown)


lub (changes requested):
It's missing a newsfragment (to later generate the changelog), otherwise seems fine.

nextgens:
> It's missing a newsfragment (to later generate the changelog), otherwise seems fine.

I'm not sure how to phrase it... that's why I've left it out. Feel free to improve upon it :)
We may want to mention the change to session lifetime... but I'm not sure how the towncrier output will look like

Diman0 (approved):
It looks good to me. It is nice we can prevent replay attacks with this. This will make mailu more secure. 
If this functionality does not work properly, we must not forget to revert the backport. 
Fortunately test.mailu.io will be automatically updated later today. So we can already test via test.mailu.io. 
However we should also test mailu 1.8.

PR #1790, opened 2021-03-11
Prevent clamav from failing with HTTP429 when testing

### LINKED ISSUE #1784: BORS / Travis fails building due to sending too many requests to clamav
## Description
See the [travis logs](https://travis-ci.org/github/Mailu/Mailu/builds/762136308?utm_source=github_status&utm_medium=notification).
```
Tue Mar  9 13:30:29 2021 -> daily database available for download (remote version: 26103)
Tue Mar  9 13:30:32 2021 -> !The daily.cvd database downloaded from https://database.clamav.net is more than one version older than the version advertised in the DNS TXT record.
Tue Mar  9 13:30:32 2021 -> Trying again in 5 secs...
Tue Mar  9 13:30:37 2021 -> daily database available for download (remote version: 26103)
Tue Mar  9 13:30:37 2021 -> !downloadFile: Unexpected response (429) from https://database.clamav.net/daily.cvd
```
One of the tests that involve clamav requires clamav to download the latest virus definitions. Apparently we send too many requests to clamav since the clamav website responds with HTTP CODE 429: Too many requests.

We can resolve this by configuring clamav **not** to download the latest virus definitions. We could modify the test to use a static database file for clamav and not to check for updates. This way we do not send update requests to the https://database.clamav.net/daily.cvd site and we have eliminated one common cause for travis failing.

Diman0:
## What type of PR?

enhancement

## What does this PR do?

See #1784 . We often have that builds fail due to hammering the clamav update software. When this happens, it is basically not possible anymore to perform more builds. This is a temporary solution to use a static database file when a test is run. This is done via an environment variable that is only set when testing. In normal situations the static DB file is removed on startup.

Ideally the CI/CD environment caches the clamav file. We cannot investigate this possibility until we have switched from Travis to the new environment. 

### Related issue(s)
- closes #1784

## Prerequistes

### DIFF
--- a/optional/clamav/Dockerfile
+++ b/optional/clamav/Dockerfile
@@ -10,6 +10,14 @@ RUN apk add --no-cache clamav rsyslog wget clamav-libunrar
 COPY conf /etc/clamav
 COPY start.py /start.py
 COPY health.sh /health.sh
+COPY testing.cvd.taraa /data/testing.cvd.taraa
+COPY testing.cvd.tarab /data/testing.cvd.tarab
+COPY testing.cvd.tarac /data/testing.cvd.tarac
+
+RUN cat /data/testing.cvd.taraa /data/testing.cvd.tarab \ 
+/data/testing.cvd.tarac > /data/testing.cvd.tar
+RUN tar -xvf /data/testing.cvd.tar -C /data/
+RUN rm -v /data/testing.cvd.tar
 
 EXPOSE 3310/tcp
 VOLUME ["/data"]
--- a/optional/clamav/start.py
+++ b/optional/clamav/start.py
@@ -5,16 +5,28 @@ import logging as log
 import sys
 
 log.basicConfig(stream=sys.stderr, level=os.environ.get("LOG_LEVEL", "WARNING"))
-logger=log.getLogger(__name__)
+logger = log.getLogger(__name__)
+
+# Clamav is run for testing. Use a static database.
+if os.environ.get("TESTING") and os.path.isfile("/data/testing.cvd"):
+    os.system("mv /data/testing.cvd /data/main.cvd")
+    log.info("TESTING mode, use static test database.")
+
+# Clamav is not run for testing. Clean up test database.
+elif os.path.isfile("/data/testing.cvd"):
+    os.system("rm /data/testing.cvd")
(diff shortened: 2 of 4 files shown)


Diman0:
I still have to create the towncrier announcement.

Diman0:
Discussed with @nextgens it makes much more sense to
- set a environment variable in travis that we are testing. This could be added to travis.yml so that it is explicitly clear that our tests depend on this environment variable. E.g. in the before_script: pass testing=true to the docker-compose file
``` - testing=true docker-compose -f tests/build.yml build```
- In the Docker-compose file pass the env variable as an argument to the build script
```
  antivirus:
    image: ${DOCKER_ORG:-mailu}/${DOCKER_PREFIX:-}clamav:${MAILU_VERSION:-local}
    build: ../optional/clamav
      args:
        TESTING: $testing
```
- In the Dockerfile, only copy the test definitions if ARG TESTING equals true. 
- In the Dockerfile pass the value of arg TESTING as env variable to the image
- In the image (start.py) skip freshclam if env var TESTING=true.

nextgens:
> Discussed with @nextgens it makes much more sense to
> 
>     * set a environment variable in travis that we are testing. This could be added to travis.yml so that it is explicitly clear that our tests depend on this environment variable. E.g. in the before_script: pass testing=true to the docker-compose file
>       ` - testing=true docker-compose -f tests/build.yml build`
> 
>     * In the Docker-compose file pass the env variable as an argument to the build script
> 
> 
> ```
>   antivirus:
>     image: ${DOCKER_ORG:-mailu}/${DOCKER_PREFIX:-}clamav:${MAILU_VERSION:-local}
>     build: ../optional/clamav
>       args:
>         TESTING: $testing
> ```
> 
>     * In the Dockerfile, only copy the test definitions if ARG TESTING equals true.
> 
>     * In the Dockerfile pass the value of arg TESTING as env variable to the image
> 
>     * In the image (start.py) skip freshclam if env var TESTING=true.

Yes, I was thinking to use one of Travis' built-in env variables instead of TESTING (say CI instead)
https://docs-staging.travis-ci.com/user/environment-variables/#default-environment-variables

Diman0:
I will close this PR since we are moving to a different CI/CD. If required, we can create a new PR for this issue.

PR #1796, opened 2021-03-26
Add SnappyMail webmail

the-djmaze:
https://snappymail.eu/

## What type of PR?

Feature

## What does this PR do?

Add SnappyMail as webmail option

## Prerequistes


parisni:
SnappyMail can be backed by postgres AFAIK ? That would be nice to optionally provide postgres then for it (as admin and roundcube do)
otherwise I failed to make it working, 500 error and the container logs :

```
webmail_1    | [Fri Jun 04 19:37:41.627648 2021] [core:alert] [pid 23] [client 127.0.0.1:59524] /var/www/html/.htaccess: Invalid command 'RewriteEngine', perhaps misspelled or defined by a module not included in the server configuration
```

tests/compose/snappymail/docker-compose.yml
```diff
@@ -0,0 +1,97 @@
...
+
+
+
+  # Webmail
+  webmail:
+    image: ${DOCKER_ORG:-mailu}/${DOCKER_PREFIX:-}snappymail:${MAILU_VERSION:-master}
+    restart: always
+    env_file: mailu.env
+    volumes:
+      - "/mailu/webmail:/data"
```
parisni, line 86: maybe mount something like `/mailu/webmail/snappymail` to avoid collision with other webmails ?

the-djmaze:
> SnappyMail can be backed by postgres AFAIK ? That would be nice to optionally provide postgres then for it (as admin and roundcube do)
PDO is only used when activating the Addressbook feature.
Since it is PDO you can use: mysql, pgsql or sqlite


> otherwise I failed to make it working, 500 error and the container logs :
> 
> ```
> webmail_1    | [Fri Jun 04 19:37:41.627648 2021] [core:alert] [pid 23] [client 127.0.0.1:59524] /var/www/html/.htaccess: Invalid command 'RewriteEngine', perhaps misspelled or defined by a module not included in the server configuration
> ```

Thanks, i solved that issue

tests/compose/snappymail/docker-compose.yml
the-djmaze: > maybe mount something like `/mailu/webmail/snappymail` to avoid collision with other webmails ?

Would be nice if i knew how.
Seems that the others (RoundCube and RainLoop) don't have that either.

Diman0:
After internal discussion we have decided not to accept this PR. Two webmails are sufficient for the project. The Mailu project always had a very low amount of active contributors. We cannot use another webmail that adds more cost of ownership. We would have to maintain this and update it for every future release. Any webmail is basically an IMAP client. You can simply run your own image of SnappyMail and connect via IMAP to Mailu.

PR #1802, opened 2021-05-03
make admin app listen on ipv4 and ipv6

### LINKED ISSUE #1789: gunicorn not listening on IPv6
### Environment
 - [x] docker-compose
 - [ ] kubernetes
 - [ ] docker swarm

### Versions
1.8

## Description

## Replication Steps
Enable IPv6.

## Expected behaviour

## Logs

````markdown
```

```
```

kahrpatrick:
## What type of PR?

bug-fix

## What does this PR do?

Make gunicorn in admin container listen on ipv4 and ipv6.

### Related issue(s)
Fixes #1789 

## Prerequistes

### DIFF
--- a/core/admin/start.py
+++ b/core/admin/start.py
@@ -19,7 +19,7 @@ if account is not None and domain is not None and password is not None:
     os.system("flask mailu admin %s %s '%s' --mode %s" % (account, domain, password, mode))
 
 start_command="".join([
-    "gunicorn -w 4 -b :80 ",
+    "gunicorn -w 4 -b [::]:80 ",
     "--access-logfile - " if (log.root.level<=log.INFO) else "",
     "--error-logfile - ",
     "--preload ",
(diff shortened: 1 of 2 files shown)


core/admin/start.py
```diff
@@ -19,7 +19,7 @@
     os.system("flask mailu admin %s %s '%s' --mode %s" % (account, domain, password, mode))
 
 start_command="".join([
-    "gunicorn -w 4 -b :80 ",
+    "gunicorn -w 4 -b [::]:80 ",
```
nextgens, line 22: it should be made conditional on SUBNET6 ... it will break on setups where ipv6 is disabled.

Getting ipv6 to work properly is more complicated than it looks: https://mailu.io/1.8/faq.html#how-to-make-ipv6-work

It may be simpler to force the resolution of IPv4 addresses as suggested in https://github.com/Mailu/Mailu/issues/1789#issuecomment-796538821

nextgens (changes requested):

Issue #1370, opened 2020-02-19
OpenRelay enabled by default ?


VFourneau:
Hey,
I setup mailu through the helm chart  for my kubernetes cluster. Everything works so far but after going through my log aggregator I found an issue.

```txt
Feb 19 17:24:02 mail postfix/smtp[27484]: 99419D237E7: to=<RANDMOM>@gmail.com>, relay=alt1.gmail-smtp-in.l.google.com[142.250.4.27]:25, delay=100681, delays=100669/1.5/7/3.1, dsn=4.7.0, status=deferred (host alt1.gmail-smtp-in.l.google.com[142.250.4.27] said: 421-4.7.0 [<MY_IP>      15] Our system has detected that this message is 421-4.7.0 suspicious due to the very low reputation of the sending IP address. 421-4.7.0 To protect our users from spam, mail sent from your IP address has 421-4.7.0 been temporarily rate limited. Please visit 421 4.7.0  https://support.google.com/mail/answer/188131 for more information. b18si301567pge.389 - gsmtp (in reply to end of DATA command))\n
```

And there is a lot. 
I setup mailu last week and I'm already blacklisted with some providers.

I used the default config.

How can I disable /block smtp relaying ?

rouja:
Hi,

Is-it a new IP ?
Did you check the reputation of your IP before to install mailu ?

To check if you are an open relay, you can do this test (from a linux terminal and from outside your email server) : 

```
nc -v $YOUR_PUBLIC_SERVER_IP 25
Ncat: Version 7.80 ( https://nmap.org/ncat )
Ncat: Connected to $YOUR_PUBLIC_SERVER_IP:25.
220 $YOUR_MAIL_SERVER_DOMAIN ESMTP ready
HELO $YOUR_MAIL_SERVER_DOMAIN
250 $YOUR_MAIL_SERVER_DOMAIN
MAIL FROM: $A_MAIL_NOT_MANAGED_BY_YOUR_SERVER
250 2.0.0 OK
RCPT TO: $ANOTHER_MAIL_NOT_MANAGED_BY_YOUR_SERVER
554 5.7.1 <$ANOTHER_MAIL_NOT_MANAGED_BY_YOUR_SERVER>: Relay access denied
```
If the last response is **250 2.1.5 Ok** that means that you are an open relay and there is something wrong in your configuration but if it's not, the problem is probably you IP reputation.

Heziode:
@rouja
I have the same issue.
[...]

PR #1823, opened 2021-06-04
Update manifests for recent K8s

eyusupov:
## What type of PR?

Bug-fix

## What does this PR do?

This makes example manifests work on recent K8s (tested on 1.20)

### Related issue(s)

None

## Prerequistes

(minor change)

### DIFF
--- a/docs/kubernetes/mailu/admin.yaml
+++ b/docs/kubernetes/mailu/admin.yaml
@@ -4,6 +4,11 @@ metadata:
   name: mailu-admin
   namespace: mailu-mailserver
 spec:
+  selector:
+    matchLabels:
+      app: mailu-admin
+      role: mail
+      tier: backend
   replicas: 1
   template:
     metadata:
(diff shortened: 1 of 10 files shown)


nextgens:
Can you set "externalTrafficPolicy: Local" as suggested in https://github.com/Mailu/Mailu/issues/1370 too please?

eyusupov:
@nextgens what service does that need to be set on?

nextgens:
probably all except "front"

eyusupov:
I'm not sure I understand how that's supposed to work for non-LoadBalancer type services, which are used in the example configuration. I also don't think I have an issue from the linked PR. I use an ingress configuration similar to one in the docs that I run with traefik. The poster could have meant the loadbalancer service for the ingress controller,  but it's not part of the example manifests, and I don't have it in my current configuration to test it.

Diman0:
@micw @HaleyACS can you review this?

HaleyACS:
With traefik this may be right. I however don't use traefik (nginx-ingress here) with system ingress. Means I only need to tell which ingress-instance I am using. So can't comment on that.
I suppose we'll need to have that separated and documented somehow.

PS: I don't have my test-cluster up yet. So no possibility to test. With a little luck, I'll have it running end of next week.

PR #1836, opened 2021-06-28
Test ci parallel

### LINKED ISSUE #1830: Increasing the speed of our CI/CD setup
To break off one part of https://github.com/Mailu/Mailu/issues/1828 into a separate ticket:

> 2: Future topics for CI/CD
> Last project meeting someone mentioned these topics. I think these should be discussed and tackled once we have switched to a new CI/CD provider. Possibly we could already do this when switching, but I think it takes time to make a proper decision.
> 
> 2.1: Testing, can this be reduced? Solve long build times.
> Currently a build takes takes ~30 minutes on travis. The main reason is that there are many tests. We need to thoroughly check if we really want to perform all the tests that are done currently. For this we must document what component/functionality each test tests. After that we can make an informed decision if and what can be changed. Of course it is vital that all components are still tested.
> 
> Personally I think we should first check how long a build takes on the new CI/CD pipeline and then decide if this is too long. For me 20-30 minutes is already acceptable. What max build time acceptable for you?

I think a good step forward would be for us to try and run these steps locally and understand what is taking a long time. We can use this ticket to gather our tips & tricks for what could be done to speed things up. Typically this is some parts what you are trying to achieve and other parts knowing how to hack your CI system to cache/reduce build time/etc. So, we still need to crack https://github.com/Mailu/Mailu/issues/1828.

Diman0:
## What type of PR?

enhancement

## What does this PR do?

Changes CI workflow to run all tests in parellel.  After performing some tests (see #1830  ), I determined that using actions/cache to only cache a tar.gz. file with all build images and use this for all parallel tests is the fasted solution.

With Travis builds took ~30 minutes. Now each build runs for a maximum of 20 minutes (bors test and merge on master).
Bors r+ runs take about ~16/17 minutes.

### Related issue(s)
- Auto close an issue like: closes #1830

## Prerequistes

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -292,6 +292,7 @@ jobs:
           path: /images
           key: ${{ env.BRANCH }}-${{ github.run_id }}-${{ github.run_number }}
       - name: Load docker images
+        if: ${{ env.BRANCH != 'staging' }}
         run: docker load -i /images/images.tar.gz
       - name: Deploy built docker images
         env:
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -85,6 +85,8 @@ jobs:
         with:
           path: /images
           key: ${{ env.BRANCH }}-${{ github.run_id }}-${{ github.run_number }}
+      - name: Load docker images
+        run: docker load -i /images/images.tar.gz
       - name: Install python packages
         run: python3 -m pip install -r tests/requirements.txt
       - name: Copy all certs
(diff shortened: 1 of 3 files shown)


Diman0:
Bors retry

decentral1se (approved):
There is a lot of duplication of tasks but :woman_shrugging: if that can be avoided?

Solid work nonetheless!

Diman0:
> There is a lot of duplication of tasks but woman_shrugging if that can be avoided?
> 
> Solid work nonetheless!
That is required. Each job within a workflow runs in a separate clean VM. Therefore it is required for each separate job (clean VM) to checkout the repo, set the env var, retrieve and load the docker images etc. 

bors r+

PR #1851, opened 2021-07-05
Upgrade alpine

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Upgrade all the images to alpine 3.14 and switch from libressl to openssl on the admin container to work around a bug in alpine

## Prerequistes

### DIFF
--- a/core/admin/Dockerfile
+++ b/core/admin/Dockerfile
@@ -2,7 +2,7 @@
 ARG DISTRO=alpine:3.14
 ARG ARCH=""
 FROM ${ARCH}node:8 as assets
-COPY --from=balenalib/rpi-alpine:3.10 /usr/bin/qemu-arm-static /usr/bin/qemu-arm-static
+COPY --from=balenalib/rpi-alpine:3.14 /usr/bin/qemu-arm-static /usr/bin/qemu-arm-static
 
 COPY package.json ./
 RUN npm install
--- a/core/admin/Dockerfile
+++ b/core/admin/Dockerfile
@@ -1,5 +1,5 @@
 # First stage to build assets
-ARG DISTRO=alpine:3.12
+ARG DISTRO=alpine:3.14
 ARG ARCH=""
 FROM ${ARCH}node:8 as assets
 COPY --from=balenalib/rpi-alpine:3.10 /usr/bin/qemu-arm-static /usr/bin/qemu-arm-static
--- a/core/admin/Dockerfile
+++ b/core/admin/Dockerfile
@@ -24,9 +24,9 @@ RUN mkdir -p /app
 WORKDIR /app
 
 COPY requirements-prod.txt requirements.txt
-RUN apk add --no-cache libressl curl postgresql-libs mariadb-connector-c \
+RUN apk add --no-cache openssl curl postgresql-libs mariadb-connector-c \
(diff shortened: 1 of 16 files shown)


Diman0:
I agree that removing cyrus-sasl-plain is the easiest approach for now. If people really want the functionality back, we could look into building cyrus-sasl-plain manually.

Diman0:
bors try

Diman0:
New error for building admin image. 
```
 (1/10) Installing curl (7.77.0-r1)
(2/10) Installing libressl3.3-libcrypto (3.3.3-r0)
(3/10) Installing libressl3.3-libssl (3.3.3-r0)
(4/10) Installing libressl3.3-libtls (3.3.3-r0)
ERROR: libressl3.3-libtls-3.3.3-r0: trying to overwrite usr/lib/libtls.so.20 owned by libretls-3.3.3-r0.
ERROR: libressl3.3-libtls-3.3.3-r0: trying to overwrite usr/lib/libtls.so.20.0.3 owned by libretls-3.3.3-r0.
(5/10) Installing libressl (3.3.3-r0)
```

Diman0:
This is a defect which is reported at alpine:
https://gitlab.alpinelinux.org/alpine/aports/-/issues/12763#note_163217

They plan to include a fix in 3.14.1 which is scheduled in 8 days (according to the milestone)
https://gitlab.alpinelinux.org/alpine/aports/-/milestones/178

We could wait until 3.14.1 is released. Or perhaps it is possible to use an older version of the package?

nextgens:
Let's wait; I'll revisit it in a week or so

Erriez:
New release date is `Aug 1, 2021`.

nextgens:
>  build ARM images.

IMHO ARM images should be build using the same base image as the rest... unless there is a good reason for them to be different.

We will use 3.14.1 when it's available

Diman0:
The alpine 3.14.1 milestone has been violated again. It appears to show that no work has happened at all. All issues still have the same status. It might take months before the next alpine version is released.

As a workaround we could try using openssl instead of libressl.

nextgens:
bors try

nextgens:
bors try

nextgens:
bors try

Erriez:
> The alpine 3.14.1 milestone has been violated again. It appears to show that no work has happened at all. All issues still have the same status. It might take months before the next alpine version is released.

I'm testing with `python:3.8-alpine3.14` locally for almost all images. Reason is that Python and PIP are already installed on alpine which saves Docker build time. They are also officially supported and maintained by Python Docker community.

> As a workaround we could try using openssl instead of libressl.

Build successful locally. Need some testing.

Is there a good reason to upgrade PIP every time, even when using a recent image?

nextgens:
> The download / PIP3 upgrade takes a lot of time in total, even when using an optimized cached build. Is there a good reason to upgrade PIP every time, even when using a recent Python image?

No. Since this PR is meant to be backported I went for the least intrusive changes I could... but there is no reason for it.

nextgens:
@Erriez thinking about it again, rpi-alpine:3.14 won't include pip ... so this will most likely break arm builds... no?

Diman0 (approved):
lgtm

nextgens:
@Mergifyio backport 1.8

Issue #1829, opened 2021-06-17
Discussion of v1.8.x Roadmap


decentral1se:
As proposed in the last project meeting (https://github.com/Mailu/Mailu/issues/1582#issuecomment-862642385), we are experimenting with a discussion driven model for determining the roadmap of the next release. This follows the existing practice of the succesful community run project Gitea (e.g https://github.com/go-gitea/gitea/issues/14477).

Starting point for the roadmap is:

- [x] master version postgresql database roundcube webmail 502 Bad Gateway https://github.com/Mailu/Mailu/issues/1648
- [ ] HOST_* and *_ADDRESS resolution errors for docker-compose deployments https://github.com/Mailu/Mailu/issues/1341
- [x] https://travis-ci.org/ ceased. We need to switch to a new CI/CD pipeline https://github.com/Mailu/Mailu/issues/1828
 
These have been derived from the existing milestones: [1.8-rc](https://github.com/Mailu/Mailu/issues?q=is%3Aopen+is%3Aissue+milestone%3A1.8-rc) and [1.8](https://github.com/Mailu/Mailu/milestone/11) and the discussion from the last meeting.

This release is 7 months overdue following the original milestone deadline (for `$reasons`) and it was discussed a little bit that we should try to minimise what we include in this roadmap so as to be able to make a release sooner rather than later. Please also bear in mind, the project has very little capacity right now and none of the new would-be maintainers have access to actually get things merged. We're on the way though.

@nextgens I am not sure which our your current PRs were suggested to also be up for inclusion in https://github.com/Mailu/Mailu/pulls/nextgens re: pressing issues with regard to security? Please let us know.

nextgens:
I was talking about #1787 (already merged in master but the backport is blocked) that became #1791.

We should also consider updating the various webmails (#1739 and its successor).

We need a solution for #1612 ; My vote goes towards "documenting the problem" (explain the problem and a workaround: setting a stupidly high limit).

Diman0:
1.8 (current rc) has indeed been waiting for 6 months to be released already. 1.7 (current stable) was released about 1.5 years ago. So people on stable (1.7) have been waiting for 1.5 years already for an update. I think we should focus on a 1.8 release within a couple of months.

I suggest to first address
- #1828
- #1791
- Update all involved libraries and components.

and release 1.8 as is. #1648 already has a workaround. #1739 has a workaround as mentioned by nextgens. We can document this.

[...]

PR #1853, opened 2021-07-05
Update dependencies

ghostwheel42:
## What type of PR?

security update python dependencies

## What does this PR do?

updates cryptography, Jinja2, pyOpenSSL and  PyYAML to current versions

### Related issue(s)

security updates as discussed in #1829

### DIFF
--- a/core/admin/Dockerfile
+++ b/core/admin/Dockerfile
@@ -26,7 +26,7 @@ WORKDIR /app
 COPY requirements-prod.txt requirements.txt
 RUN apk add --no-cache libressl curl postgresql-libs mariadb-connector-c \
   && apk add --no-cache --virtual build-dep \
-  libressl-dev libffi-dev python3-dev build-base postgresql-dev mariadb-connector-c-dev \
+  libressl-dev libffi-dev python3-dev build-base postgresql-dev mariadb-connector-c-dev cargo \
   && pip3 install -r requirements.txt \
   && apk del --no-cache build-dep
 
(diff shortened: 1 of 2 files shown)


nextgens (approved):

core/admin/Dockerfile
```diff
@@ -26,7 +26,7 @@ WORKDIR /app
 COPY requirements-prod.txt requirements.txt
 RUN apk add --no-cache libressl curl postgresql-libs mariadb-connector-c \
   && apk add --no-cache --virtual build-dep \
-  libressl-dev libffi-dev python3-dev build-base postgresql-dev mariadb-connector-c-dev \
+  libressl-dev libffi-dev python3-dev build-base postgresql-dev mariadb-connector-c-dev cargo \
```
Diman0, line 29: Is there a specific reason why are you adding the cargo module?
Is this a new dependency for the updated modules in requirements-prod.txt?

nextgens: Not my code but it's to ensure that the rust wheels get built properly; without there was a warning (and either the wheels wouldn't be built or it would be downloaded at build-time)

ghostwheel42: Yes. It seems to be possible to disable the rust part via an environment variable, but only for cryptography 3.4:
"You can disable Rust integration in 3.4.x (see FAQ). Starting with 3.5 cryptography will have mandatory Rust code."
Enabling it now seemed to be more future proof.

nextgens: It's a bad idea to not have the wheels; an attacker will have an optimized runtime, so should we :)

This is useful for (fast) iterated password hashing

PR #1856, opened 2021-07-06
update asset builder dependencies

ghostwheel42:
## What type of PR?

update asset builder dependencies

## What does this PR do?

only include needed dependencies to build mailu assets with nodejs v8

### Related issue(s)

update dependencies as discussed in #1829


Diman0:
I haven't tested yet, but looking at this old PR it appears that JQuery is actually used:
https://github.com/Mailu/Mailu/pull/1048#issuecomment-507068607
At that time it was used when activating the checkbox for "enable automatic reply". I will try this during my testing.

If JQuery is really not used, then it should  probably also be removed from:
core/admin/webpack.config.js
core/admin/assets/vendor.js
core/admin/assets/app.js

Once I have tested later today I will report results here.

ghostwheel42:
> I haven't tested yet, but looking at this old PR it appears that JQuery is actually used:

Sorry for not being clear on this. I never meant to say jquery is not used at all.
It is used by admin-lte and is automatically included as dependency.

I've included the newest possible versions of jquery and select2 in package.json (again).

Diman0:
Ok. It looks good to me. I will do some smoke tests after the bors try test.

bors try

smercan89 (approved):

nextgens (approved):

PR #1857, opened 2021-07-07
disable startdate when autoreply is disabled

ghostwheel42:
## What type of PR?

bug-fix

## What does this PR do?

disable the reply startdate field when autoreply is disabled


Diman0 (approved):
Looks good to me!

PR #1858, opened 2021-07-08
Docs: Replace hardcoded journald logpath with systemd backend

networkException:
The file at /var/log/messages is not universal for every
distribution. Fail2ban can access journald logs directly
by using the systemd backend.

## What type of PR?

documentation

## What does this PR do?

The path /var/log/messages does not apply for Ubuntu 20.04 for example, because of that I have looked
at alternative ways to access journald in fail2ban. The proper way seems to be to use the systemd
backend, this patch updates the documentation accordingly.

### Related issue(s)
*none*

## Prerequistes


networkException:
bors retry

nextgens (approved):

nextgens:
bors retry

Diman0 (approved):
Lgtm. Thank you for your contribution.

Diman0:
bors retry

PR #1862, opened 2021-07-14
fix #1861: allow colons in passwords: fix the parsing of http-basic auth headers

### LINKED ISSUE #1861: Radicale password check fails (length limited?)
So far, the webdav radicale is protected with basic auth (apparently htaccess or similar).
If user password > 8 then it fails to connect. 
we should remove this limitation to let use stong passwords

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Fix a bug preventing colons from being used in passwords when using radicale/webdav.
Thank you to @parisni for reporting it and @ghostwheel42 for spotting it.

### Related issue(s)
- close #1861

## Prerequistes


ghostwheel42 (approved):
lgtm

Issue #466, opened 2018-04-30
Received: from authenticated-user (PRIMARY_HOSTNAME [PUBLIC_IP])


jkaberg:
Inspecting the mail headers reveals an oversight,

> Received: from authenticated-user (**PRIMARY_HOSTNAME [PUBLIC_IP]**)
> by mail.domain.tld (Postfix) with ESMTPA id 63XXXXXXX2
> for <user@domain.tld>; Mon, 30 Apr 2018 07:17:10 +0000 (UTC)

I traced this down to;

https://github.com/Mailu/Mailu/blob/edbea372e9e2254d9dff155083f0b118b4d2679b/core/postfix/conf/outclean_header_filter.cf#L7

This should be fixed asap 😄

jkaberg:
Right, never mind me: https://github.com/Mailu/Mailu/issues/191

PR #1871, opened 2021-07-18
Replace PUBLIC_HOSTNAME/IP with real data in Received headers

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

This will ensure that we don't get spam points for not respecting the RFC (gmail is notorious for not liking it)

### Related issue(s)
- close #1448
- #466
- #191

## Prerequistes


nextgens:
bors try

Diman0 (approved):
lgtm

PR #1888, opened 2021-07-24
Use threads in gunicorn rather than workers/processes

nextgens:
## What type of PR?

enhancement

## What does this PR do?

This ensures that we share the auth-cache... will enable memory savings
and may improve performances when a higher number of cores is available

"smarter default"


nextgens:
Not sure if it's worth backporting this to 1.8 ...

ghostwheel42 (approved):
lgtm

PR #1894, opened 2021-07-24
Migrate to PHP8-fpm

### LINKED ISSUE #1521: Build webmails on php:fpm-alpine
Mailu's webmails are both built on debian/apache/php.

AFAIK all other mailu components are based on alpine, and we already have an nginx frontend.

IMHO it would make sense to rebase webmails on php:fpm-alpine and use the same nginx frontend for the static parts of the webmails.

If we want independent webmail containers (which is also nice), we should at least build the images with alpine/nginx/php-fpm built-in.

nextgens:
## What type of PR?

enhancement

## What does this PR do?

It migrates all the PHP code to PHP8-FPM

### Related issue(s)
- closes #1521

## Prerequistes


nextgens:
Right now it's just rainloop... but the rest could come

nextgens:
As discussed on #mailu-dev, what we want is unclear and needs to be discussed; #1521 will be where it happens

Issue #116, opened 2016-11-13
Brute force protection


Ndrou:
It could be interesting to integrate Fail2ban to protect from brute force attack on SMTP/POP3/IMAP protocols.

Ideally, a log viewer could be added on admin web interface.

kaiyou:
Given the current architecture, this is very difficult to implement: not all users have the same Docker logging configuration, etc.

We could distribute a ``docker-compose.yml`` to force logging to specific files then parse these files, but I doubt this would please many users, most of which are used to their Docker logging tools and maybe remote logging systems (Gelf, etc.)

So in the current state of things, it is not Mailu's responsibility to handle logs and display them, even if I would like to be able to detect some events in the future and produce statistics (see #86). You are more than welcome to add some Wiki documentation about setting up ELK logging or any other fancy system to display logs in a Web interface.


Regarding lockout and bruteforce protection, I agree as well: we need something. Account lockout is a must-have for many regulations even if I dislike it. Rate limiting and IP-based lockouts are better and we will have to implement them.

Currently, your only option is to have Fail2ban installed on your host and parsing Docker logs. Some Docker images are using the host network stack to run Fail2ban inside a container, but I would definitely not recommend this (https://hub.docker.com/r/superitman/fail2ban/). Again, any contribution to the documentation about setting this up is welcome.

Eventually, we should be able to concentrate critical operations like authentication and authorized client maps in a single container that could perform rate limiting with a Redis backend for instance.

kaiyou:
This will be related to #272 and could actually be implemented then if we succeed.

PR #1914, opened 2021-08-07
Prevent all form of authentication on port 25

nextgens:
## What type of PR?

Feature

## What does this PR do?

It prevents all form of authentication on port 25. Well-behaved clients shouldn't attempt it anyway as AUTH isn't advertised there...

### Related issue(s)
- #116

## Prerequistes


nextgens:
This is a band-aid... I'm not sure it makes sense at all and will discuss it on #1915

nextgens:
bad idea. let's discuss it on #1915

PR #1922, opened 2021-08-09
Harden postfix's configuration

### LINKED ISSUE #1804: smtp plain text authentication permitted

## Before you open your issue

### Environment
 - [X] docker-compose
 - [ ] kubernetes
 - [ ] docker swarm

### Versions
To find your version, get the image name of a mailu container and read  the version from the tag (example for version 1.7).
```
$> docker ps -a | grep mailu
140b09d4b09c    mailu/roundcube:1.7    "docker-php-entrypoi…"    2 weeks ago    Up 2 days (healthy)    80/tcp
$> grep MAILU_VERSION docker-compose.yml mailu.env

d0d3826162f4        mailu/roundcube:master   "docker-php-entrypoi…"   6 days ago   Up 6 days (healthy)   80/tcp

docker-compose.yml:    image: ${DOCKER_ORG:-mailu}/${DOCKER_PREFIX:-}nginx:${MAILU_VERSION:-master}
docker-compose.yml:    image: ${DOCKER_ORG:-mailu}/${DOCKER_PREFIX:-}unbound:${MAILU_VERSION:-master}
docker-compose.yml:    image: ${DOCKER_ORG:-mailu}/${DOCKER_PREFIX:-}admin:${MAILU_VERSION:-master}
docker-compose.yml:    image: ${DOCKER_ORG:-mailu}/${DOCKER_PREFIX:-}dovecot:${MAILU_VERSION:-master}
docker-compose.yml:    image: ${DOCKER_ORG:-mailu}/${DOCKER_PREFIX:-}postfix:${MAILU_VERSION:-master}
docker-compose.yml:    image: ${DOCKER_ORG:-mailu}/${DOCKER_PREFIX:-}rspamd:${MAILU_VERSION:-master}
docker-compose.yml:    image: ${DOCKER_ORG:-mailu}/${DOCKER_PREFIX:-}clamav:${MAILU_VERSION:-master}
docker-compose.yml:    image: ${DOCKER_ORG:-mailu}/${DOCKER_PREFIX:-}radicale:${MAILU_VERSION:-master}
docker-compose.yml:    image: ${DOCKER_ORG:-mailu}/${DOCKER_PREFIX:-}fetchmail:${MAILU_VERSION:-master}
docker-compose.yml:    image: ${DOCKER_ORG:-mailu}/${DOCKER_PREFIX:-}roundcube:${MAILU_VERSION:-master}

```

## Description
Further explain the bug in a few words. It should be clear what the unexpected behaviour is.  Share it in an easy-to-understand language.

I've been working through the results of our first penetration test and have one item I need some assistance on.

I have a low priority result "SMTP Service Cleartext Login Permitted" received.  The IP address this is tagged to is our email's public IP address.  Unfortunately, the report does not specify which port this is being tagged on. 

Description: The remote host is running an SMTP server that advertises that it allows cleartext logins over unencrypted connections. An attacker may be able to uncover user names and passwords by sniffing traffic to the server if a less secure authentication mechanism (i.e. LOGIN or PLAIN) is used.

Solution:  Configure the service to support less secure authentication mechanisms only over an encrypted channel.

I'm not really sure what I need to do.  I have an SSL on the mail server and the password stuff should all be encrypted.  What might I be missing?  Any guidance would be appreciated.

## Replication Steps
Steps for replicating your issue

I sought help
at https://matrix.to/#/#mailu:tedomum.net and I was instructed to view the dovecot documentation (https://doc.dovecot.org/configuration_manual/authentication/).
It is already with this configuration made in the environment but it still presents this flaw in my email environment (mailu).

## Expected behaviour
Explain what results you expected - be as specific as possible. Just saying "it doesn’t work as expected" is not useful. It's also helpful to describe what you actually experienced.

I expected smtp plain text authentication permitted to be blocked or just safe to be allowed.

## Logs
Often it is very useful to include log fragments of the involved component. You can get the logs via `docker logs <container name> --tail 1000`. For example for the admin container:
`docker logs mailu_admin_1 --tail 1000`
or using docker-compose `docker-compose -f /mailu/docker-compose.yml logs --tail 1000 admin`
If you can find the relevant section, please share only the parts that seem relevant. If you have any logs, please enclose them in code tags, like so:
````markdown
```
Your logs here!

admin_1_c1b8d3c6db94 | INFO  [alembic.runtime.migration] Context impl SQLiteImpl.
admin_1_c1b8d3c6db94 | INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
admin_1_c1b8d3c6db94 | [2021-04-29 15:29:03 +0000] [11] [INFO] Starting gunicorn 19.9.0
admin_1_c1b8d3c6db94 | [2021-04-29 15:29:03 +0000] [11] [INFO] Listening at: http://0.0.0.0:80 (11)
admin_1_c1b8d3c6db94 | [2021-04-29 15:29:03 +0000] [11] [INFO] Using worker: sync
admin_1_c1b8d3c6db94 | /usr/lib/python3.8/os.py:1023: RuntimeWarning: line buffering (buffering=1) isn't supported in binary mode, the default buffer size will be used
admin_1_c1b8d3c6db94 |   return io.open(fd, *args, **kwargs)
admin_1_c1b8d3c6db94 | [2021-04-29 15:29:04 +0000] [13] [INFO] Booting worker with pid: 13
admin_1_c1b8d3c6db94 | /usr/lib/python3.8/os.py:1023: RuntimeWarning: line buffering (buffering=1) isn't supported in binary mode, the default buffer size will be used
admin_1_c1b8d3c6db94 |   return io.open(fd, *args, **kwargs)
admin_1_c1b8d3c6db94 | [2021-04-29 15:29:04 +0000] [14] [INFO] Booting worker with pid: 14
admin_1_c1b8d3c6db94 | /usr/lib/python3.8/os.py:1023: RuntimeWarning: line buffering (buffering=1) isn't supported in binary mode, the default buffer size will be used
admin_1_c1b8d3c6db94 |   return io.open(fd, *args, **kwargs)
admin_1_c1b8d3c6db94 | [2021-04-29 15:29:04 +0000] [15] [INFO] Booting worker with pid: 15
admin_1_c1b8d3c6db94 | /usr/lib/python3.8/os.py:1023: RuntimeWarning: line buffering (buffering=1) isn't supported in binary mode, the default buffer size will be used
admin_1_c1b8d3c6db94 |   return io.open(fd, *args, **kwargs)
admin_1_c1b8d3c6db94 | [2021-04-29 15:29:04 +0000] [16] [INFO] Booting worker with pid: 16
admin_1_c1b8d3c6db94 | [2021-04-30 16:10:02,316] ERROR in app: Exception on /internal/auth/email [GET]
admin_1_c1b8d3c6db94 | Traceback (most recent call last):
admin_1_c1b8d3c6db94 |   File "/usr/lib/python3.8/site-packages/flask/app.py", line 2292, in wsgi_app
admin_1_c1b8d3c6db94 |     response = self.full_dispatch_request()
admin_1_c1b8d3c6db94 |   File "/usr/lib/python3.8/site-packages/flask/app.py", line 1815, in full_dispatch_request
admin_1_c1b8d3c6db94 |     rv = self.handle_user_exception(e)
admin_1_c1b8d3c6db94 |   File "/usr/lib/python3.8/site-packages/flask/app.py", line 1718, in handle_user_exception
admin_1_c1b8d3c6db94 |     reraise(exc_type, exc_value, tb)
admin_1_c1b8d3c6db94 |   File "/usr/lib/python3.8/site-packages/flask/_compat.py", line 35, in reraise
admin_1_c1b8d3c6db94 |     raise value
admin_1_c1b8d3c6db94 |   File "/usr/lib/python3.8/site-packages/flask/app.py", line 1813, in full_dispatch_request
admin_1_c1b8d3c6db94 |     rv = self.dispatch_request()
admin_1_c1b8d3c6db94 |   File "/usr/lib/python3.8/site-packages/flask/app.py", line 1799, in dispatch_request
admin_1_c1b8d3c6db94 |     return self.view_functions[rule.endpoint](**req.view_args)
admin_1_c1b8d3c6db94 |   File "/app/mailu/internal/views/auth.py", line 24, in nginx_authentication
admin_1_c1b8d3c6db94 |     headers = nginx.handle_authentication(flask.request.headers)
admin_1_c1b8d3c6db94 |   File "/app/mailu/internal/nginx.py", line 45, in handle_authentication
admin_1_c1b8d3c6db94 |     user_email = raw_user_email.encode("iso8859-1").decode("utf8")
admin_1_c1b8d3c6db94 | UnicodeDecodeError: 'utf-8' codec can't decode byte 0x99 in position 0: invalid start byte

```
````

### PR DESCRIPTION
**Pull request by nextgens**
## What type of PR?

enhancement

## What does this PR do?

It hardens the default configuration:
- disable AUTH commands on port 25 (nginx was not advertising the capability: normal clients wouldn't attempt it)
- fix Forward Secrecy by ensuring that we don't use session tickets and don't cache on forensically carveable mediums
- prevent clear-text credentials from being sent while authenticating to remote relays (this may break things if the relay doesn't support challenge-based authentication NOR STARTTLS - unlikely).
- switch to default RSA keysizes (2048 bits and they get rekeyed every 3 months -modern clients will do ECC)
- enable ECC certificates (much smaller than RSA keys, faster for better security margin)
- configure nginx so that it doesn't send the legacy/root CA (clients that require it are unlikely to do TLS1.2 any ways)

I don't think that any of those changes is impactful enough to warrant being documented.

### Related issue(s)
- close #1804

## Prerequistes

### CODE REVIEW

**Comment by nextgens** [comment]
bors retry

**Comment by nextgens** [comment]
bors retry

**Comment by nextgens** [comment]
bors retry

**Comment by nextgens** [comment]
bors retry

**Comment by nextgens** [comment]
bors retry

File: core/admin/mailu/internal/nginx.py [hunk 1]
```diff
@@ -81,6 +81,13 @@ def handle_authentication(headers):
         raw_password = urllib.parse.unquote(headers["Auth-Pass"])
         password = raw_password.encode("iso8859-1").decode("utf8")
         ip = urllib.parse.unquote(headers["Client-Ip"])
+        port = int(urllib.parse.unquote(headers["Auth-Port"]))
```
Diman0, line 84: I think this causes the test to fail. Here you overwrite the port (see line 74) of the destination postfix server. 
Port 10025 is overwritten with port 587 (the port from the header Auth-Port.)

This results in the connection refused error that nginx receives when it tries to proxy the connection to port 587 instead of 10025.

I suggest to rename the var port to something else and of course also rename the var on line 85 (if port == 25).

Diman0 (changes requested):
See my in line comment. You overwrite the port variable in nginx.py on line 84. Use a different variable name.

core/admin/mailu/internal/nginx.py
nextgens: Doh.
Thanks!

nextgens:
bors retry

Diman0:
After the test works again, I think it is handy to re-apply the commit 7285c6bfd98fed56addf8d1b736f10feb6cb8048 where you remove LOGIN as authentication method.
The authentication script (admin) only supports authentication method PLAIN.

nextgens:
I've found out using tcpdump that nginx is transparently transforming LOGIN requests to PLAIN ... so it's not as bad as I initially thought

nextgens:
Removing the extra certificate will break compatibility with android handsets before Nougat/7.1 ...  do we want to introduce a knob? and if so, how far back should the knob bring compatibility back?

Diman0:
> I've found out using tcpdump that nginx is transparently transforming LOGIN requests to PLAIN ... so it's not as bad as I initially thought

It does? 
That is pretty neat. In that case I agree we should not change this to increase compatibility with smtp clients.

Diman0:
> Removing the extra certificate will break compatibility with android handsets before Nougat/7.1 ... do we want to introduce a knob? and if so, how far back should the knob bring compatibility back?

Will it really break all possible email clients on android 7.0 and earlier? Or is it possible to built in support in your application (on android <7.1) for handling certificates without root ca? 
I will also check If I can find the answer to this question. but maybe you already know this.

nextgens:
> > Removing the extra certificate will break compatibility with android handsets before Nougat/7.1 ... do we want to introduce a knob? and if so, how far back should the knob bring compatibility back?
> 
> Will it really break all possible email clients on android 7.0 and earlier? Or is it possible to built in support in your application (on android <7.1) for handling certificates without root ca?
> I will also check If I can find the answer to this question. but maybe you already know this.

You can definitely import/trust additional CAs in your application or your OS... You'll have to "trust on first use" (users will get a scary warning) instead of relying on what the OS can check, that's all.

Diman0:
So we have two questions for the knob (correct me if wrong)
1. Old and new chain as discussed here https://community.letsencrypt.org/t/production-chain-changes/150739
The longer chain ensures compatibility with android <7.1.1

2. Use smaller ECDSA certificates. This requires android 4.1.1+

As of September 2020 https://www.xda-developers.com/android-version-distribution-statistics-android-studio/:
In September last year about ~25 percent of all users were on Android <7.1.1. That is a significant amount. 
About 0.8% was on android <4.2 and about 0.2% was on android <4.1. 

25% is significant. Less than 1% is insignificant. I'd say we create the knob only for option 1 and not for option 2. So switch to ECDSA and provide a possibility to use the longer chain or smaller chain.

nextgens:
That's correct.

There's also a 3rd (related) question:
- Are we happy with dropping the security level from RSA4096 to P256/RSA2048? If not there's an additional android-specific caveat that we need to be aware of: versions before 7.0 have a bug and do not support other curves (see https://community.letsencrypt.org/t/any-info-on-the-upcoming-ecdsa-certificate-chain-with-android-compatibility/149613).

nextgens:
@Diman0 Here's the proposed implementation... The default behaviour is "do like before" and it's currently an opt-in.

Letsencrypt will eventually switch to a full ECC (shorter) chain... we can revisit this when we're closer to next release

Diman0:
Yes that makes sense. I will test this later this week.

Diman0:
bors try

I will try testing it today. There is only one thing that looks a bit confusing to me. When I check tls.conf I see that in case of letsencrypt both the old (RSA) and new certificate type (ECDSA) are used by nginx. As far as I know this results in nginx negotiating with the client what certificate will be used. If the client does not support ECDSA it will fallback to the RSA certificate.

Doesn't this mean that old clients (android <4.1.1) should still work?

nextgens:
> Doesn't this mean that old clients (android <4.1.1) should still work?

No. The requirement for TLS1.2 (that went in 1.8 - 4f973f63e) broke support for android < 5... Some handsets on 4.x+ will work but realistically you can't count on it. If the target is to support "just android", we could drop RSA certificates altogether

Diman0:
So ECDSA and TLS 1.2 are not supported in most android < 4.4.2 phones. Thank you for the clarification.

Checking your links:
Android 4.4.2+ supports tls 1.2 and ecdsa.
Windows 7+ supports tls 1.2 and ecdsa.
macos 10.6 (last version released in 2011) supports tls 1.2 and ecdsa
All internet browsers supports tls 1.2 and ecdsa.

I agree that dropping support for RSA is viable. If there are clients that do not support ecdsa, then this client most likely does not support TLS 1.2 as well.

nextgens:
Ok. Do you want me to tweak the PR to that effect?

Diman0:
Yes (if you don't mind). There is no real advantage for using both. This will make the code less complex and more easy to maintain. There will only be one command function now in letsencrypt.py.

Initial tests look good. I can send/receive email and connect to mailu without any issues.
I also nicely receive the message "5.5.1 AUTH not supported" when I try to use AUTH LOGIN on port 25 via telnet.

nextgens:
bors try

nextgens:
@Diman0 now I remember why I did it that way: if we remove RSA certificates we also impact what can be negotiated on port 25 (where we want the maximum compat as any crypto beats no crypto)... I'll force-push back to where we were

nextgens:
bors try

nextgens:
bors cancel

nextgens:
bors try

Diman0:
Your earlier link for ecdsa mentions ecdsa support was added in openssl 0.9.8 released in ~2008. 
TLS 1.2 support was not present in OpenSSL 0.9.8y
TLS 1.2 support is present in OpenSSL 1.0.1h (released in ~2014)
I'm not sure how trustworthy those reports are, but Id say if the client supports TLS 1.2 then it also supports ECDSA .

nextgens:
On port 25 we support TLS1.0 ... see https://github.com/Mailu/Mailu/blob/966b9cb91858f33d9800b75cbdf43e9056bf4e0a/core/nginx/conf/nginx.conf#L267

nextgens:
@Diman0 let me know how it currently looks for you; IMHO it's good enough to make it to master and receive some testing

Diman0:
Ok. I missed this. So HTTPS is TLS 1.2 and higher and SMTP (on port 25) is TLSv1 and higher. In that case I'm wrong and I agree with you.

I will approve the PR after the last test (BORS try) has succeeded.

Diman0 (approved):
Lgtm

PR #1925, opened 2021-08-10
Optimize Rainloop: Change to NGINX

### LINKED ISSUE #1924: Optimize Rainloop image
## Before you open your issue

### Environment
 - [X] docker-compose
 - [X] kubernetes
 - [X] docker swarm

### Versions
Master
```
$> docker images
mailu/rainloop               master            2ad8d1c29ff3   45 hours ago     607MB
```
Optimized version using NGINX:
```
user/rainloop                master            7de9dee9286d   2 seconds ago   116MB
```

## Description
In order to resolve issues #1830 and #1200, it is required to reduce build time, complexity and image size of images before multi-architecture builds can be supported. The current Rainloop image size is 607MB and can be optimized when changing from Apache to NGINX with officially supported Docker images. This can also increase overall web performance.

Erriez:
## What type of PR?

- Reduce build time.
- Reduce image size.
- Faster user response using CGI.

### Related issue(s)
- Mention an issue like: #1830, #1200 and #1924
- Auto close an issue like: closes #1924

## Prerequistes

Documentation updates TBD (requires some guidance):

## Technical details

- Image from `php:7.4-apache` to `nginx:1.21-alpine` followed by PHP7 installation.
- Move `.ini` files to directory `defaults/`.
- Move files `sso.php and include.php` to directory `login/`.
- NGINX configuration:
  - `access_log off;` as this is handled by front.
  - `error_log /dev/stderr err; configured to reduce output. The logging in `start.py` is useless.
  - PHP configuration optimized for CGI usage and requires a config file `config/php-rainloop.conf` -> `/etc/php7/php-fpm.d/rainloop.conf`.
  - `.ini` files are parsed / substituted by `socrate` Python module.

Further optimization is possible by completely removing Python. This is only used to parse the `.ini` files and can be done via Bash scripts. This saves more build time and image size can be reduced to 112MB.

## Reviewing

This PR requires multiple reviewers and extensive testing before merging into master. Data/settings are compatible with previous images.


webmails/rainloop/config/nginx-rainloop.conf
```diff
@@ -0,0 +1,38 @@
...
+
+    # /dev/stderr (Default), <path>, debug, info, notice, warn, error, crit, alert, emerg
+    error_log /dev/stderr error;
+
+    index index.php;
+
+    location / {
+        try_files $uri $uri/ /index.php?$query_string;
+    }
+
```
nextgens, line 18: location ~* \.(?:css|js)$ {
      expires 7d;
      add_header Pragma public;
      add_header Cache-Control "public";
    }

    location ~* \.(?:gif|jpe?g|png|ico|otf|eot|svg|ttf|woff|woff2)$ {
      expires 30d;
      add_header Pragma public;
      add_header Cache-Control "public";
    }

webmails/rainloop/config/nginx-rainloop.conf
```diff
@@ -0,0 +1,38 @@
...
+    location ~ \.php$ {
+        fastcgi_split_path_info ^(.+\.php)(/.*)$;
+        
+        fastcgi_intercept_errors on;
+        fastcgi_index  index.php;
+
```
nextgens, line 24: fastcgi_param REMOTE_ADDR $http_x_real_ip; 

Should have something of the sort, no? What about the protocol so that it knows whether it's behind HTTP/HTTPS and sets the right cookie flags?

webmails/rainloop/config/php-rainloop.conf
```diff
@@ -0,0 +1,83 @@
...
+;                                    of 'idle' processes is greater than this
+;                                    number then some children will be killed.
+;  ondemand - no children are created at startup. Children will be forked when
+;             new requests will connect. The following parameter are used:
+;             pm.max_children           - the maximum number of children that
+;                                         can be alive at the same time.
+;             pm.process_idle_timeout   - The number of seconds after which
+;                                         an idle process will be killed.
+; Note: This value is mandatory.
+pm = dynamic
```
nextgens, line 57: Is this the right setting considering it's already behind another nginx instance? I'd try to minimize stanby RAM usage if possible

webmails/rainloop/Dockerfile
```diff
@@ -1,51 +1,65 @@
-ARG ARCH=""
-ARG QEMU=other
+ARG DISTRO=nginx:1.21-alpine
```
nextgens, line 1: It makes me nervous when we don't pin a specific alpine version... are we sure that the tag isn't going to "move" ?

Erriez: I agree to pin to specific version. 

The official Docker nginx images do not have frozen version such as ~~nginx:1.21-alpine3.14~~. See https://hub.docker.com/_/nginx?tab=tags&page=1&ordering=last_updated&name=alpine

This is the Dockerfile source:
https://github.com/nginxinc/docker-nginx/blob/master/Dockerfile-alpine.template

My proposal is to use `alpine:3.14` and install required PHP packages + Python3.9 for `socrate`. Then all packages are have a frozen version. I'll investigate this further.

webmails/rainloop/config/nginx-rainloop.conf
Erriez: The Rainloop installation does not mention this: https://www.rainloop.net/docs/installation/

Do you have a use case where this is required? Then I'll add it.

webmails/rainloop/config/nginx-rainloop.conf
Erriez: Well, I've tested with two reverse proxies: First one is my [Swag reverse proxy](https://docs.linuxserver.io/general/swag) pointing to port 80 and 443 and handles SSL certificates. The front container running nginx forwards webmail traffic to Rainloop at HTTP port 80. That's no difference with previous Apache setup.

With and without `fastcgi_param REMOTE_ADDR $http_x_real_ip;` works in my setup. 

I can add this line when we're sure.

webmails/rainloop/config/php-rainloop.conf
Erriez: > I'd try to minimize stanby RAM usage if possible

Good point. I did not benchmark this yet. It depends on the use case: lots of idle time or a few..100+  simultaneous users makes sense. It requires some further investigation and maybe input from more experienced users to define the default preferred values.

Interesting article: https://haydenjames.io/php-fpm-tuning-using-pm-static-max-performance/

webmails/rainloop/Dockerfile
```diff
@@ -1,51 +1,65 @@
...
+FROM $DISTRO
 
-# NOTE: only add file if building for arm
-FROM ${ARCH}php:7.4-apache as build_arm
-ONBUILD COPY --from=balenalib/rpi-alpine:3.14 /usr/bin/qemu-arm-static /usr/bin/qemu-arm-static
```
Erriez, line 6: The `arm_build.sh` script is broken. I'll fix it.

webmails/rainloop/Dockerfile
```diff
@@ -1,51 +1,65 @@
...
+# Install Rainloop from source
 ENV RAINLOOP_URL https://github.com/RainLoop/rainloop-webmail/releases/download/v1.16.0/rainloop-community-1.16.0.zip
 
-RUN apt-get update && apt-get install -y \
-      unzip python3-jinja2 \
- && rm -rf /var/www/html/ \
- && mkdir /var/www/html \
- && cd /var/www/html \
+RUN apk add --no-cache \
+      unzip py3-jinja2 \
```
Erriez, line 48: py3-jinja2 not used and can be removed.

webmails/rainloop/Dockerfile
```diff
@@ -1,51 +1,65 @@
...
+ && mkdir -p /var/www/rainloop \
```
Erriez, line 49: Move line `&& mkdir -p /var/www/rainloop \` before the `COPY` using this path.

webmails/rainloop/Dockerfile
```diff
@@ -1,51 +1,65 @@
...
- && rm -rf /var/lib/apt/lists
-
-COPY include.php /var/www/html/include.php
-COPY sso.php /var/www/html/sso.php
-COPY php.ini /php.ini
-
-COPY application.ini /application.ini
-COPY default.ini /default.ini
+ && chown -R nginx:nginx /var/www/rainloop \
+ && apk del unzip
```
Erriez, line 58: Remove `curl` afterwards as it is not used. 

Note: `HEALTHCHECK` check has been removed as several users complained to see the output in the logs.

webmails/rainloop/config/nginx-rainloop.conf
```diff
@@ -0,0 +1,38 @@
...
+    # /dev/stdout (Default), <path>, off
+    access_log off;
+
+    # /dev/stderr (Default), <path>, debug, info, notice, warn, error, crit, alert, emerg
+    error_log /dev/stderr error;
+
+    index index.php;
+
+    location / {
+        try_files $uri $uri/ /index.php?$query_string;
```
Erriez, line 16: Received feedback to remove `$uri/` which is required when installing `nginx` on `alpine:3.14`.

Related issue: https://stackoverflow.com/questions/19285355/nginx-403-error-directory-index-of-folder-is-forbidden

webmails/rainloop/Dockerfile
```diff
@@ -1,51 +1,65 @@
...
+#     dom             php7-dom
+#     openssl         php7-openssl
+#     DateTime        Builtin
+#     PCRE            Builtin
+#     SPL             Builtin
+#  Recommended:
+#     php7-fpm        FastCGI Process Manager
+#  Optional PHP extension (for contacts):
+#     php7-pdo        Accessing databases in PHP
+#     php7-pdo_sqlite Access to SQLite 3 databases
```
Erriez, line 25: Does anyone know if `php7-pdo` and `php7-pdo_sqlite` are used?

webmails/rainloop/Dockerfile
```diff
@@ -1,51 +1,65 @@
...
-# NOTE: only add file if building for arm
-FROM ${ARCH}php:7.4-apache as build_arm
-ONBUILD COPY --from=balenalib/rpi-alpine:3.14 /usr/bin/qemu-arm-static /usr/bin/qemu-arm-static
+# Shared later between dovecot postfix nginx rspamd rainloop and roundloop
+RUN apk add --no-cache \
+    python3 py3-pip \
+ && pip3 install socrate==0.2.0
 
-FROM ${ARCH}php:7.4-apache as build_other
+# Shared layer between rainloop and roundcube
```
Erriez, line 9: Layer is not shared with Roundcube (yet).

webmails/rainloop/config/nginx-rainloop.conf
nextgens: It's not required it's just to enhance performance (caching headers limit round-trips)

Erriez: Ok, I'll add it.

webmails/rainloop/config/nginx-rainloop.conf
Erriez: No need to add `fastcgi_param REMOTE_ADDR $http_x_real_ip;` to `nginx-rainloop.conf`. Evidence is shown by running `nginx -T` command (test configuration, dump it and exit) in the rainloop container which shows all enabled settings:

```
...
fastcgi_param  REMOTE_ADDR        $remote_addr;
fastcgi_param  REMOTE_PORT        $remote_port;
fastcgi_param  SERVER_ADDR        $server_addr;
fastcgi_param  SERVER_PORT        $server_port;
fastcgi_param  SERVER_NAME        $server_name;

# PHP only, required if PHP was built with --enable-force-cgi-redirect
fastcgi_param  REDIRECT_STATUS    200;
```

Configuration file stack:
```
/etc/nginx/http.d/rainloop.conf
    include /etc/nginx/fastcgi_params; 
         fastcgi_param  REMOTE_ADDR        $remote_addr;
```

Erriez:
@nextgens All feedback processed. Thanks! Do you want me to squash the commits?

nextgens:
No, that's fine; we usually don't

Diman0:
You removed the HEALTHCHECK in the dockerfile. Please add it back. The CI workflow failed to start, I will retry

BORS try

Diman0:
It is obvious you didn´t try to build all images yourself. It is good we have a CI/CD workflow. The built failed because you removed files for roundcube. Can you re-add these files?
The relevant files:
webmails/roundcube/config.inc.php 
webmails/roundcube/mailu.php 

Nextgens and I also overlooked this while reviewing this PR.

Erriez:
> You removed the HEALTHCHECK in the dockerfile. Please add it back. The CI workflow failed to start, I will retry
> 
> BORS try

No problem, I'll put it back in. Someone mentioned recently:
> We should rework the "healthchecks". The ones working on Docker can cause heavoc (High load) on kubernetes side. 

My suggestion is to perform a `wget or curl` check from the test without the HEALTHCHECK to verify if the container is running correctly. Something for a new issue.


> The built failed because fles for roundcube were removed in commit 0fd9712. Can you re-add these files?
> The relevant files:
> webmails/roundcube/config.inc.php
> webmails/roundcube/mailu.php

I accidentally overlooked this and discovered in my IDE that a file move adds it automatically to the staging area. I'll check if it is possible to disable this unwanted behavior. I've tested send/receive emails with Rainloop manually with local changes in the checkout and did not see this problem earlier.

I'll continue with setting-up an identical CI on a local machine _before_ a push instead of running the CI _after_ the push. Documentation is pretty limited to setup Github Actions on a forked project: https://mailu.io/master/contributors/workflow.html?highlight=github%20actions#pr-workflow.

I apologize for this build failure and will fix it soon.

Diman0:
It doesn't matter if it fails building. The purpose of the ci workflow is to catch this. Of course it is nice if it immediately builts well.

To test building you can simply use the docker-compose file as discussed here https://mailu.io/master/contributors/environment.html#building-images

If you do not set env vars, the images will be built with org mailu and tag local.

Erriez:
Thanks @Diman0! I made my own advanced build scripts to build and test with docker-compose, but should do this with a fresh checkout in the future.

I've enabled Github Actions in my forked project now, but the build time is 21 minutes _after_ a single push, so this does not help so much. I'm now setting up [Act](https://github.com/nektos/act) to build/test the exact same CI locally. Hopefully this is a nice workflow for other developers as well: Test locally before publishing. Is someone already using this?

I noticed that when I make a mistake commit in my forked branch which is already provided as PR, there is no way to rewrite the history without squashing as `git push --force` is listed as a [Bad habit](https://mailu.io/master/contributors/environment.html#bad-habits) in the documentation. Squashing is mandatory in my organization to keep `git logs` clean, so I'm confused about good and wrong.

Is there a better way to restore accidental commits, or is this automatically done during the merge? Sorry for asking these basic questions, I have to align to this project which has a different workflow I'm not familiar with.

Erriez:
BORS try

nextgens:
bors retry

Erriez:
Curl needs to be restored. I'll fix it later today.

Diman0:
> I noticed that when I make a mistake commit in my forked branch which is already provided as PR, there is no way to rewrite the history without squashing as `git push --force` is listed as a [Bad habit](https://mailu.io/master/contributors/environment.html#bad-habits) in the documentation. Squashing is mandatory in my organization to keep `git logs` clean, so I'm confused about good and wrong.
> 
> Is there a better way to restore accidental commits, or is this automatically done during the merge? Sorry for asking these basic questions, I have to align to this project which has a different workflow I'm not familiar with.

We never rewrite history for submitted PRs. If you make a mistake in a commit (for an existing PR), then we normally create an additional commit where you undo this mistake. It is fine to have multiple small commits.

Erriez:
Cannot run tests locally (yet) and cannot start CI.

nextgens:
bors try

Diman0:
Interesting. I have no clue why yet. But building the images took longer now. In stead of ~10 minutes, it took ~12 minutes.

Maybe we should alter the workflow file to allow Manuel trigger? This would help using the workflow on your fork. Of course this would entail that we must instruct all contributors to never use this on the main project.

Erriez:
> But building the images took longer now. In stead of ~10 minutes, it took ~12 minutes.

I was not aware of this. Maybe it is faster the next time when packages are cached. I'm still struggling on an [issue](https://github.com/nektos/act/issues/329) to get `act` up and running. Then I can test the build+test CI flow on my development machine.

> Maybe we should alter the workflow file to allow Manual trigger?

I'm thinking about adding [pull_request](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags) to `.github/workflows/CI.yml`.

nextgens:
Let's test the cache theory.

bors retry

Diman0:
> > But building the images took longer now. In stead of ~10 minutes, it took ~12 minutes.
> 
> I was not aware of this. Maybe it is faster the next time when packages are cached. I'm still struggling on an [issue](https://github.com/nektos/act/issues/329) to get `act` up and running. Then I can test the build+test CI flow on my development machine.
> 
Packages are never cached. Every github action workflow run occurs on a completely new VM. I'm very surprised if the workflow run does show that packages were loaded from cache.
What might be possible, is
1. copying the packages cache folder (I assume this exists?) to the github action cache, and
2. load this cache before the action step for building all images.

> > Maybe we should alter the workflow file to allow Manual trigger?
> 
> I'm thinking about adding [pull_request](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#on) to `.github/workflows/CI.yml`.

I do not agree. For running locally in your own forked repo, we could add the [workflow_dispatch event](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_dispatch).
The CI.yml in context of a PR must be run in context of bors try. This ensures the pr-xxxx images are updated.
We could update the CI.yml (or create a new file), to add a ```bors try``` message when a commit is pushed to the PR. I checked the mergify documentation, but could not find the answer if mergify can do this as well.

Erriez:
Tests on my development machine:

**Build Rainloop on `master`:**
```
$ git status
    On branch master
    Your branch is up to date with 'origin/master'.
    nothing to commit, working tree clean

$ git log
    commit 966b9cb91858f33d9800b75cbdf43e9056bf4e0a (HEAD -> master, origin/staging, origin/master, origin/HEAD)
    Date:   Sat Aug 14 21:59:37 2021 +0000

$ docker system prune -f
$ cd tests
$ time docker-compose -f build.yml build rainloop
    ...
    Successfully tagged mailu/rainloop:local
    
    real	0m56,513s
    user	0m0,378s
    sys	0m0,023s

$ docker images
    REPOSITORY             TAG              IMAGE ID       CREATED          SIZE
    mailu/rainloop         local            4e084bf6c72e   20 seconds ago   609MB
```

**Build Rainloop on branch `feat-rainloop-nginx`**

```
$ git status
    On branch feat-rainloop-nginx
    Your branch is up to date with 'origin/feat-rainloop-nginx'.
    nothing to commit, working tree clean

$ git log
    commit 5adc4f08f632e1fd4e66488a070ecb382b95d8d1 (HEAD -> feat-rainloop-nginx, origin/feat-rainloop-nginx)
    Date:   Mon Aug 16 17:41:36 2021 +0200

$ docker system prune -f
$ cd tests
$ time docker-compose -f build.yml build rainloop
    ...
    real	0m14,018s
    user	0m0,334s
    sys	0m0,047s

$ docker images
    REPOSITORY             TAG              IMAGE ID       CREATED         SIZE
    mailu/rainloop         local            034e8b4a9a5d   5 minutes ago   116MB
```

Build speed: 56.6 seconds reduced to 14.0 seconds without caching: **75% faster build**.
Image size: 605MB to 116MB: **80% smaller image**.

Diman0:
That looks very promising. Very nice numbers!

Erriez:
> That looks very promising. Very nice numbers!

Yes. I cannot explain why the Github Action CI is slower compared with my development machine.

HaleyACS:
Just for comment on the build process.
I saw that the build process uses cached components (tested with: docker-compose -f tests/build.yml build rainloop) . Of course this makes it fast, but it uses the old build components.
I hope that our "real" build system does not use cached components. This would prevent the build system to integrated new fixes etc.

Diman0:
The PR itself looks sound to me. I confirmed that the image works. I checked:
- logging in via SSO
- logging in via rainloop itself if admin is disabled
- sending email
- receiving email

The only thing I could not verify is the logging. 
https://github.com/Mailu/Mailu/blob/966b9cb91858f33d9800b75cbdf43e9056bf4e0a/webmails/rainloop/start.py#L9
Even when I change the log level to INFO and restart the container, then I see no logging in the output of the container. 
I would at least expect to see the logging of the HEALTHCHECK (the CURL to localhost) with log level INFO

Upon further looking I see you set the log level to ERROR on line 11 of nginx-rainloop.conf . 
This should be set to ```warn``` to match start.py.

To catch php errors, I think the following must be added to the php.ini file. This will log the php errors to nginx:
catch_workers_output = 1
display_errors = On

webmails/rainloop/config/nginx-rainloop.conf
```diff
@@ -0,0 +1,38 @@
...
+    listen 80 default_server;
+    listen [::]:80 default_server;
+
+    root /var/www/rainloop;
+    
+    # /dev/stdout (Default), <path>, off
+    access_log off;
+
+    # /dev/stderr (Default), <path>, debug, info, notice, warn, error, crit, alert, emerg
+    error_log /dev/stderr error;
```
Diman0, line 11: can you set this to level warn to match start.py?

Diman0 (dismissed):
See inline comment.

For php.ini look into adding "catch_workers_output = 1" and "display_errors = On" to enable logging of php errors.

Erriez:
> Just for comment on the build process.
> I saw that the build process uses cached components (tested with: docker-compose -f tests/build.yml build rainloop) . Of course this makes it fast, but it uses the old build components.
> I hope that our "real" build system does not use cached components. This would prevent the build system to integrated new fixes etc.

`docker system prune -f` was executed before the build to remove all caches. Otherwise the build was done within 1 second.

> can you set this to level warn to match start.py?

Thanks for your feedback. Yes, I'll fix and test this evening.

HaleyACS:
> > Just for comment on the build process.
> > I saw that the build process uses cached components (tested with: docker-compose -f tests/build.yml build rainloop) . Of course this makes it fast, but it uses the old build components.
> > I hope that our "real" build system does not use cached components. This would prevent the build system to integrated new fixes etc.
> 
> `docker system prune -f` was executed before the build to remove all caches. Otherwise the build was done within 1 second.
> 
> > can you set this to level warn to match start.py?
> 
> Thanks for your feedback. Yes, I'll fix and test this evening.

That was my assumption too though ... but:

`$ docker system prune && time docker-compose -f tests/build.yml build rainloop
WARNING! This will remove:
  - all stopped containers
  - all networks not used by at least one container
  - all dangling images
  - all dangling build cache

Are you sure you want to continue? [y/N] y
Deleted Images:
deleted: sha256:e5b0696e576c069f5b76d8ac63bbe235309317cf1ff56f23ff3aa55d4e749800

Total reclaimed space: 0B
Building rainloop
Step 1/20 : ARG ARCH=""
Step 2/20 : ARG QEMU=other
Step 3/20 : FROM ${ARCH}php:7.4-apache as build_arm
 ---> 93e55f680811
Step 4/20 : ONBUILD COPY --from=balenalib/rpi-alpine:3.14 /usr/bin/qemu-arm-static /usr/bin/qemu-arm-static
 ---> Running in 4aa6b9eb3aaa
Removing intermediate container 4aa6b9eb3aaa
 ---> e3b1fb34ebd4

Step 5/20 : FROM ${ARCH}php:7.4-apache as build_other
 ---> 93e55f680811

Step 6/20 : FROM build_${QEMU}
 ---> 93e55f680811
Step 7/20 : RUN apt-get update && apt-get install -y   python3 curl python3-pip git python3-multidict   && rm -rf /var/lib/apt/lists   && echo "ServerSignature Off" >> /etc/apache2/apache2.conf
 ---> Using cache
 ---> e2d1f22bf2a1
Step 8/20 : RUN pip3 install socrate
 ---> Using cache
 ---> 102726d6632d
Step 9/20 : ENV RAINLOOP_URL https://github.com/RainLoop/rainloop-webmail/releases/download/v1.16.0/rainloop-community-1.16.0.zip
 ---> Using cache
 ---> aa6b9be1336e
Step 10/20 : RUN apt-get update && apt-get install -y       unzip python3-jinja2  && rm -rf /var/www/html/  && mkdir /var/www/html  && cd /var/www/html  && curl -L -O ${RAINLOOP_URL}  && unzip -q *.zip  && rm -f *.zip  && rm -rf data/  && find . -type d -exec chmod 755 {} \;  && find . -type f -exec chmod 644 {} \;  && chown -R www-data: *  && apt-get purge -y unzip  && rm -rf /var/lib/apt/lists
 ---> Using cache
 ---> 8f48fcaa1c37
Step 11/20 : COPY include.php /var/www/html/include.php
 ---> Using cache
 ---> 12ca059e0e54
Step 12/20 : COPY sso.php /var/www/html/sso.php
 ---> Using cache
 ---> ef6fcb4db993
Step 13/20 : COPY php.ini /php.ini
 ---> Using cache
 ---> faaf39dc0d11
Step 14/20 : COPY application.ini /application.ini
 ---> Using cache
 ---> 187ee92463e1
Step 15/20 : COPY default.ini /default.ini
 ---> Using cache
 ---> 09726ccb2219
Step 16/20 : COPY start.py /start.py
 ---> Using cache
 ---> b9c82f5c85c1
Step 17/20 : EXPOSE 80/tcp
 ---> Using cache
 ---> 3fdeec815495
Step 18/20 : VOLUME ["/data"]
 ---> Using cache
 ---> 1d65da9d57e9
Step 19/20 : CMD /start.py
 ---> Using cache
 ---> 3f4712a9691a
Step 20/20 : HEALTHCHECK CMD curl -f -L http://localhost/ || exit 1
 ---> Using cache
 ---> 33e43e6311fe

Successfully built 33e43e6311fe
Successfully tagged mailu/rainloop:local

real    0m0,628s
user    0m0,430s
sys     0m0,051s
`

Diman0:
> > Just for comment on the build process.
> > I saw that the build process uses cached components (tested with: docker-compose -f tests/build.yml build rainloop) . Of course this makes it fast, but it uses the old build components.
> > I hope that our "real" build system does not use cached components. This would prevent the build system to integrated new fixes etc.
> 
> `docker system prune -f` was executed before the build to remove all caches. Otherwise the build was done within 1 second.
> 
> > can you set this to level warn to match start.py?
> 
> Thanks for your feedback. Yes, I'll fix and test this evening.

Great. It might still be difficult to test if the logging works since I don't know how to get a warning logged. Hopefully nginx or php already have warning messages.

BTW, there will be no caching issues. Every workflow run happens on a clean new environment. GH actions creates a new VM instance for every workflow run. We will only have caching for shared steps among the Dockerfiles.

On your own environment you can use ```docker-compose -f tests/build.yml --no-cache build rainloop```  to build without cache.

HaleyACS:
Thanks. That's what I'm doing.
Currently I still do that on my Workstation. I'll however do the migration this week-end, means also that I'll finally add one of the NUC's to act as a satellite for kubernetes dev & testing.

Erriez:
> On your own environment you can use docker-compose -f tests/build.yml --no-cache build rainloop to build without cache.

`--no-cache` is an argument of `build` and should be moved after `build`. This should result in similar timing around 14 sec:

`time docker-compose -f build.yml build --no-cache rainloop`

webmails/rainloop/config/nginx-rainloop.conf
Erriez: Fixed.

Diman0:
I think we are only missing ```catch_workers_output = 1``` and ```display_errors = On``` in webmails/rainloop/config/php-rainloop.conf to enable logging of php warnings and errors and then I'd say the PR is ready.

Erriez:
I found documentation here:
* [catch_workers_output](https://www.php.net/manual/en/install.fpm.configuration.php).
* [display_errors](https://www.php.net/manual/en/errorfunc.configuration.php).

Do you prefer to configure it in `php.ini` as mentioned in the documentation, or `php-rainloop.conf`?

Diman0:
As far as I can see catch_workers_output can only be configured in the pool. Let's configure this is php-rainloop.conf within the rainloop pool (below line 4).

Please ignore display_errors, I did not correctly understand this setting. We do not need it. See https://www.php.net/manual/en/errorfunc.configuration.php#ini.display-errors. They actually advise to disable this for production. We only need catch_workers_output to report warnings/errors to nginx. So nginx can throw it to standard error again.

Diman0:
> I did not see any difference in output. Maybe there are no warnings.

I guess this means you did a good job and there are simply no warnings :+1:

Erriez:
Configured as requested:

```ini
[rainloop]

; Redirect worker stdout and stderr into main error log. If not set, stdout and 
; stderr will be redirected to /dev/null according to FastCGI specs. 
; Default value: no. 
catch_workers_output = 1
```

> I guess this means you did a good job and there are simply no warnings +1

Thanks! There was a lot of tweaking needed to get this result.

I forgot to mention the memory usage which automatically drops on idle, for example:

```
CONTAINER ID   NAME                        CPU %     MEM USAGE / LIMIT    MEM %     NET I/O           BLOCK I/O         PIDS
2ab7020f201e   rainloop_webmail_1          0.00%     29.54MiB / 7.55GiB   0.38%     4.15kB / 0B       31MB / 8.19kB     18
```

Diman0 (approved):
Lgtm

webmails/rainloop/config/php-rainloop.conf
```diff
@@ -0,0 +1,101 @@
...
+; The desired maximum number of idle server processes.
+; Note: Used only when pm is set to 'dynamic'
+; Note: Mandatory when pm is set to 'dynamic'
+; pm.max_spare_servers = 3
+
+; This sets the maximum time in seconds a script is allowed to run before it is 
+; terminated by the parser. This helps prevent poorly written scripts from tying up 
+; the server. The default setting is 30s.
+; Note: Used only when pm is set to 'ondemand'
+pm.process_idle_timeout = 10s
```
nextgens, line 94: Is this long-enough? what was apache set to?

webmails/rainloop/config/nginx-rainloop.conf
nextgens: Have we given up on any form of caching? The cache could be even more aggressive than suggested above as there is the version in the path for static assets.

webmails/rainloop/config/php-rainloop.conf
Erriez: Apache used `php:7.4-apache` without fpm.

Erriez:
> Have we given up on any form of caching? The cache could be even more aggressive than suggested above as there is the version in the path for static assets.

It is recommended to create a new issue when static assets are required, because the goal of this PR is to reduce build time and decrease image/RAM size.

nextgens:
Ok, fair enough, I'll create another issue/PR at the back of it then

webmails/rainloop/config/php-rainloop.conf
nextgens: I have just tested it on my mailu 1.8, apache powered instance, with:
```php
<?php sleep(60000); ?>
```

... It takes 60s before getting 504 Gateway Time-out from nginx.
I'd feel a lot better if the value there was larger (at least 60s).

nextgens: that being said, it's not the only difference... so ok, let's try it and we'll see if it's good enough.

nextgens (approved):
LGTM. Thank you for working on this!

PR #1928, opened 2021-08-14
Change letsencrypt timer from 1h --> 1 day

jackmurray:
There's no need to be calling certbot so frequently. Letsencrypt certificates last for 90 days so polling every hour is just wasteful. Once per day should be more than sufficient to catch any certificates before they even get close to expiring.

## What type of PR?

Enhancement

## What does this PR do?

Reduces unnecessary load on the Letsencrypt ACME servers.

## Prerequistes

### DIFF
--- a/core/nginx/letsencrypt.py
+++ b/core/nginx/letsencrypt.py
@@ -22,8 +22,8 @@ command = [
 # Wait for nginx to start
 time.sleep(5)
 
-# Run certbot every hour
+# Run certbot every day
 while True:
     subprocess.call(command)
-    time.sleep(3600)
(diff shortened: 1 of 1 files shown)


nextgens (approved):
LGTM

jackmurray:
I couldn't find anything in the docs that explained the current behaviour - should this be documented somewhere to make it clear how the LE integration is set up?

nextgens:
IMHO that's an internal thing... it doesn't have to be documented.

I'm changing some of it in #1922 too

ghostwheel42 (approved):
Lgtm

Issue #1938, opened 2021-08-21
Some repeat DNS lookups are introducing TOCTOU problems


nextgens:
As discussed in the regular project meeting, I'm opening an issue for it: this will have to be addressed before we can consider merging #1716

Time Of Check Time of Use (or race conditions) are a serious problem: we want operations to be atomic (either succeed or fail but not have undetermined state).

#1716 is proposing to change the behaviour from "lookup once at container start-up" to "lookup each time we need it"

The codebase should be audited to ensure that should a lookup fail somewhere it won't lead to any operation being "half-completed".

One obvious occurrence of such problem is the sieve scripts:
- https://github.com/Mailu/Mailu/blob/master/core/dovecot/conf/ham.script
- https://github.com/Mailu/Mailu/blob/master/core/dovecot/conf/spam.script
where any failure in DNS resolution may lead to inconsistent state (that won't necessarily raise an error!).

lub:
How are race conditions special with regards to DNS lookups? Resolving some name during startup doesn't guarantee availability at the time of a request.

I think it's reasonable to assume the docker (or kubernets) internal DNS to be available on a healthy system and thus just accept is as dependency.

nextgens:
We've talked about it during the dev meeting: it's fine and we'll do dynamic resolution.

PR #1939, opened 2021-08-21
Ensure that we don't do multiple DNS lookups in the sieve script

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

It ensures that DNS lookups don't introduce inconsistent state. We may want to go further and actually check the return codes of rspamc too.

I haven't tested it but it should work.

### Related issue(s)
- #1938

### DIFF
--- a/core/dovecot/conf/ham.script
+++ b/core/dovecot/conf/ham.script
@@ -1,5 +1,12 @@
 #!/bin/bash
 
-tee >(rspamc -h {{ ANTISPAM_WEBUI_ADDRESS }} -P mailu learn_ham /dev/stdin) \
-	>(rspamc -h {{ ANTISPAM_WEBUI_ADDRESS }} -P mailu -f 11 fuzzy_del /dev/stdin) \
-	| rspamc -h {{ ANTISPAM_WEBUI_ADDRESS }} -P mailu -f 13 fuzzy_add /dev/stdin
+RSPAMD_HOST="$(getent hosts {{ ANTISPAM_WEBUI_ADDRESS }})"
+if [[ $? -ne 0 ]]
+then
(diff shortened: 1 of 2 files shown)


nextgens:
bors cancel

nextgens:
bors try

core/dovecot/conf/ham.script
```diff
@@ -1,5 +1,12 @@
 #!/bin/bash
 
-tee >(rspamc -h {{ ANTISPAM_WEBUI_ADDRESS }} -P mailu learn_ham /dev/stdin) \
-	>(rspamc -h {{ ANTISPAM_WEBUI_ADDRESS }} -P mailu -f 11 fuzzy_del /dev/stdin) \
-	| rspamc -h {{ ANTISPAM_WEBUI_ADDRESS }} -P mailu -f 13 fuzzy_add /dev/stdin
+RSPAMD_HOST="$(getent hosts {{ ANTISPAM_WEBUI_ADDRESS }})"
```
ghostwheel42, line 3: I this will not work. `getent hosts` returns multiple values:

```
bash-5.1# getent hosts antispam
fd09:d816:d504:beef::9  antispam  antispam
```
you need to use cut, sed or a bash array:
- `RSPAMD_HOSTS="$(getent hosts antispam | cut -d" " -f1)"`
- `RSPAMD_HOSTS="$(getent hosts antispam | sed 's/\s.*//')"`
- `RSPAMD_HOSTS=($(getent hosts antispam))`

ghostwheel42 (changes requested):

core/dovecot/conf/ham.script
nextgens: I don't know what I was thinking...

ghostwheel42 (approved):
LGTM

PR #1941, opened 2021-08-23
Fix a bug whereby adding new HOSTNAMEs won't necessarily lead to certificates being renewed.

### LINKED ISSUE #1270: Letsencrypt Force Renewal
Is there a limit on the Subject Alt Name entries?

I have updated my /mailu/mailu.env "HOSTNAMES" variable, but when I restart Mailu it doesn't update the Subject Alt Names on the mailu cert.

Previously it has worked, so I am guessing that I need to force Letsencrypt to refresh as it isnt within the renewal window. But there is no guidance for the new letsencrypt certbot.

I am using the latest Mailu version (1.7) and this is the command I am using to restart mailu '/mailu/docker-compose -p mailu up -d'

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Fix a bug whereby adding new HOSTNAMEs won't necessarily lead to certificates being renewed.

certbot's defaut behaviour has changed when --renew-with-new-domains was introduced

### Related issue(s)
- close #1270

### DIFF
--- a/core/nginx/letsencrypt.py
+++ b/core/nginx/letsencrypt.py
@@ -13,6 +13,7 @@ command = [
     "--cert-name", "mailu",
     "--preferred-challenges", "http", "--http-01-port", "8008",
     "--keep-until-expiring",
+    "--renew-with-new-domains",
     "--config-dir", "/certs/letsencrypt",
     "--post-hook", "/config.py"
 ]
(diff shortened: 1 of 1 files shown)


ghostwheel42 (approved):
LGTM

Issue #1940, opened 2021-08-23
[webmail] endless sso.php redirect when trying to open webmail


gdanov:
I have been using mailu for a while, without webmail client. Decided to add roundcube today. 

Added this snippet (found in the roundcube sub-project) to my docker-compose.yml:
```
  # Webmail
  webmail:
    image: ${DOCKER_ORG:-mailu}/${DOCKER_PREFIX:-}roundcube:${MAILU_VERSION:-master}
    restart: always
    env_file: mailu.env
    volumes:
      - "/mailu/webmail:/data"
    depends_on:
      - imap
```
Then updated the WEBMAIL and WEB_WEBMAIL vars in mailu.env
Pulled all images to be latest, restarted.

At boot I see this:
```
webmail_1    | 127.0.0.1 - - [23/Aug/2021:13:00:14 +0000] "GET / HTTP/1.1" 302 577 "-" "curl/7.74.0"
webmail_1    | 127.0.0.1 - - [23/Aug/2021:13:00:14 +0000] "GET /sso.php HTTP/1.1" 302 577 "-" "curl/7.74.0"
webmail_1    | 127.0.0.1 - - [23/Aug/2021:13:00:14 +0000] "GET /sso.php HTTP/1.1" 302 577 "-" "curl/7.74.0"
webmail_1    | 127.0.0.1 - - [23/Aug/2021:13:00:14 +0000] "GET /sso.php HTTP/1.1" 302 577 "-" "curl/7.74.0"
```
then when I try to use the webmail in the browser effect is the same — endless redirect to sso.php

## Environment & Versions
### Environment
 - [x] docker-compose
 - [ ] kubernetes
 - [ ] docker swarm

### Versions
To find your version, get the image name of a mailu container and read  the version from the tag (example for version 1.7).
```
root@danov:/mailu# docker ps -a
CONTAINER ID   IMAGE                    COMMAND                  CREATED         STATUS                     PORTS                                                                                                                                                                                                                                                                                                                      NAMES
f258b358bc73   mailu/roundcube:master   "docker-php-entrypoi…"   8 minutes ago   Up 8 minutes (unhealthy)   80/tcp                                                                                                                                                                                                                                                                                                                     mailu_webmail_1
[...]

PR #1946, opened 2021-08-24
Use fixed version of roundcube

AndreKoepke:
#1940 
Need this because of endless redirect error in roundcube:master

## What type of PR?

bug-fix

## What does this PR do?

Pin version of roundcube to latest working (1.8).

### Related issue(s)
- #1940 

## Prerequistes


nextgens:
We have discussed it on Matrix and the consensus is that we should drop the k8s templates as no one is willing to maintain them. The recommended way to set Mailu up in k8s is through helm charts.

#1947 will make this obvious

PR #1959, opened 2021-08-28
Ensure that we don't trust client headers

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Document how REAL_IP_FROM and REAL_IP_HEADER should be used. Ensure that we strip True-Client-IP and X-Forwarded-For if neither are set.

We should also update the documentation on reverse-proxies... but that's #1958

### Related issue(s)
- #1958

## Prerequistes

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -169,6 +169,7 @@ The ``LETSENCRYPT_SHORTCHAIN`` (default: False) setting controls whether we send
 
 .. _`android handsets older than 7.1.1`: https://community.letsencrypt.org/t/production-chain-changes/150739
 
+The ``REAL_IP_HEADER`` (default: unset) and ``REAL_IP_FROM`` (default: unset) settings controls whether HTTP headers such as ``X-Forwarded-For`` or ``X-Real-IP`` should be trusted. The former should be the name of the HTTP header to extract the client IP address from and the later a comma separated list of IP addresses designing which proxies to trust. If you are using Mailu behind a reverse proxy, you should set both. Setting the former without the later introduces a security vulnerability allowing a potential attacker to spoof his source address.
 
 Antivirus settings
 ------------------
--- a/core/nginx/conf/proxy.conf
+++ b/core/nginx/conf/proxy.conf
@@ -1,7 +1,12 @@
 # Default proxy setup
 proxy_set_header Host $http_host;
 proxy_set_header X-Real-IP $remote_addr;
+proxy_set_header True-Client-IP $remote_addr;
 proxy_set_header Forwarded "";
-proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
 proxy_set_header X-Forwarded-Proto $proxy_x_forwarded_proto;
+{% if REAL_IP_HEADER and REAL_IP_FROM %}
+proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+{% else %}
+proxy_set_header X-Forwarded-For $remote_addr;
+{% endif %}
 proxy_http_version 1.1;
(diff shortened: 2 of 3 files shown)


docs/configuration.rst
```diff
@@ -169,6 +169,7 @@ The ``LETSENCRYPT_SHORTCHAIN`` (default: False) setting controls whether we send
 
 .. _`android handsets older than 7.1.1`: https://community.letsencrypt.org/t/production-chain-changes/150739
 
+The ``REAL_IP_HEADER`` (default: unset) and ``REAL_IP_FROM`` (default: unset) settings controls whether HTTP headers such as ``X-Forwarded-For`` or ``X-Real-IP`` should be trusted. The former should be the name of the HTTP header to extract the client IP address from and the later a comma separated list of IP addresses designing which proxies to trust. If you are using Mailu behind a reverse proxy, you should set both. Setting the former without the later introduces a security vulnerability allowing a potential attacker to spoof his source address.
```
Diman0, line 172: Since the real client ip is only used if both settings are set, we must say must instead of should.
> if you are using Mailu behind a reverse proxy, you **must** set both. Setting the former without the later introduces a security vulnerability allowing a potential attacker to spoof his source address. The source client ip address from the reverse proxy is only used if both settings (REAL_IP_HEADER and REAP_IP_FROM) are set in mailu.env.

Diman0 (changes requested):
It looks good to me. Only the documentation must need some clarification that you must set both settings. That not setting both settings result in the source client ip address not being used from the reverse proxy.

docs/configuration.rst
nextgens: Hmm, no; If you set REAL_IP_HEADER, the nginx module will replace $remote_addr (see https://github.com/Mailu/Mailu/blob/1d9850490c05828aa9e310f64578a189ceecfe8d/core/nginx/conf/nginx.conf#L23)... so the documentation is accurate; You can get away with setting only this one (but shouldn't for security reasons).

X-Forwarded-For will be mangled (sanitized as what's in it can't and shouldn't be trusted) but it will work.

Do you want me to actually enforce that both are set to actively prevent insecure configurations?

Diman0:
Am I correct that after this PR, the client IP is only passed on if at a minimum REAL_IP_HEADER is set in mailu.env?

In the new merged #1958, I do not see any headers being set. 
E.g. proxy_set_header X-Forwarded-For $remote_addr;
Or does proxy_pass already set headers? I could not find this information in the documenation.

If someone uses nginx (or traefik) as reverse proxy as documented in #1958, what setting should be set for REAP_IP_HEADER?
If  we could add this to the documentation what header to set and what IP to configure then I think the PR is fine.

nextgens:
> Am I correct that after this PR, the client IP is only passed on if at a minimum REAL_IP_HEADER is set in mailu.env?
> 

Yes

> In the new merged #1958, I do not see any headers being set.
> E.g. proxy_set_header X-Forwarded-For $remote_addr;
> Or does proxy_pass already set headers? I could not find this information in the documenation.
> 

You're right, the documentation on reverse proxies needs to be clarified further. #1958 was mostly cosmetic (removing what won't work and fixing links)

> If someone uses nginx (or traefik) as reverse proxy as documented in #1958, what setting should be set for REAP_IP_HEADER?
> If we could add this to the documentation what header to set and what IP to configure then I think the PR is fine.

REAP_IP_HEADER=X-Real-IP
REAP_IP_FROM=x.x.x.x,y.y.y.y.y

I propose to do the documentation update elsewhere though; IMHO we shouldn't support that many options (as discussed on #mailu-dev).

Diman0 (approved):
Lgtm. 

We must create an additional PR to update the reverse proxy documentation that the X-Real-IP header should be set as well in the nginx reverse proxy configuration. 
And then if you follow the example in the reverse proxy configuration, then that you can set in mailu.env:
REAP_IP_HEADER=X-Real-IP
REAP_IP_FROM=x.x.x.x,y.y.y.y.y

PR #1964, opened 2021-09-01
Alpine3.14.2

nextgens:
Upgrade to alpine 3.14.2, retry upgrading unbound & switch back to libressl


ghostwheel42 (approved):
lgtm

HaleyACS (commented):
You sure this is a rollback? I looks to me that we go forward to 3.14 here.

PR #1918, opened 2021-08-08
Alpine has removed support for btree and hash from postfix

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

fix the following errors:
Aug 08 16:52:03 ocloud postfix/smtp[376]: error: unsupported dictionary type: hash
Aug 08 16:52:03 ocloud postfix/tlsmgr[377]: error: unsupported dictionary type: btree
Aug 08 16:52:03 ocloud postfix/tlsmgr[377]: warning: btree:/var/lib/postfix/smtp_scache is unavailable. unsupported dictionary type: btree

Without it Mailu is unusable with a relay.

PR #1902, opened 2021-08-01
Make smtp_tls_policy_maps easily configurable

nextgens:
## What type of PR?

Feature

## What does this PR do?

- Make smtp_tls_policy_maps easily configurable. This is useful to force TLS verification of specific destinations (or relays).
We should probably discuss what's on the list by default. I have found a top100 list online, ran it through a script to check all the records and found 90 destinations we could use.
- disable TLS session tickets (this reduces the PFS window from 1day to 1h)
- enable system CAs by default (to allow for OUTBOUND_TLS_LEVEL above encrypt without additional overrides)

### Related issue(s)
- closes #1558 
- #707

## Prerequistes
Before we can consider review and merge, please make sure the following list is done and checked.
If an entry in not applicable, you can check it or remove it from the list.

- [x] In case of feature or enhancement: documentation updated accordingly
- [x] Unless it's docs or a minor change: add [changelog](https://mailu.io/master/contributors/workflow.html#changelog) entry file.

PR #1961, opened 2021-08-29
Implement MTA-STS and DANE validation

nextgens:
## What type of PR?

Feature

## What does this PR do?

Implement MTA-STS: the tls_policy_map will now be auto-configured based on the policies published by the various domains. A FAQ entry has been added to document how to publish a policy using Mailu.

As configured by default there is no persistence. If we want persistence we can have either sqlite3 (with a db in the mailqueue) or redis...

This also introduces a DEFER_ON_TLS_ERROR (default: True) setting that will harden policy enforcement and defer emails that shouldn't be delivered. Turn it off if you never want to set an override.

### Related issue(s)
- closes #1798
- closes #707 

## Prerequistes
Before we can consider review and merge, please make sure the following list is done and checked.
If an entry in not applicable, you can check it or remove it from the list.

- [x] In case of feature or enhancement: documentation updated accordingly
- [x] Unless it's docs or a minor change: add [changelog](https://mailu.io/master/contributors/workflow.html#changelog) entry file.

PR #1965, opened 2021-09-01
postfix/tls_policy: Use lmdb map instead of hash

tonobo:
## What type of PR?

bug-fix

### Related issue(s)

#1918

https://github.com/Mailu/Mailu/pull/1902/#issuecomment-902108080


nextgens (approved):
Doh. Thank you for this...

I have re-introduced the problem several times as "other branches" were started before the fix got in; It's also fixed by #1961

Diman0 (approved):

ghostwheel42 (approved):
LGTM

PR #1800, opened 2021-04-12
AdminLTE 3

DjVinnii:
## What type of PR?

Enhancement

## What does this PR do?

This PR implements AdminLTE 3 for the admin interface. It also includes the implementation of DataTables and a language selector.

### Related issue(s)
- closes: #1567
- closes: #1764 

## Prerequistes

- [x] Unless it's docs or a minor change: add [changelog](https://mailu.io/master/contributors/guide.html#changelog) entry file.

PR #1966, opened 2021-09-02
AdminLTE3 optimizations & compression and caching

### LINKED ISSUE #1905: Is possible to customize the admin UI?
My company logo has primary color is RED. So I wanna change the color, does it possible and how to do that?
* I'm not familiar with python so please help me!

ghostwheel42:
## What type of PR?

enhancement, bugfix

## What does this PR do?

Optimization and cleanup of styles and javascript code for AdminLTE 3
Adds caching headers, gzip and robots.txt to nginx.

### Related issue(s)

Makes #1800 even better. Thanks to @DjVinnii and @Diman0 for the good work.
Closes #1905

## Prerequistes


core/admin/mailu/ui/views/languages.py
```diff
@@ -2,8 +2,8 @@
 
 import flask
 
-
-@ui.route('/language/<language>', methods=['GET'])
+@ui.route('/language/<language>', methods=['POST'])
 def set_language(language=None):
     flask.session['language'] = language
-    return flask.redirect(flask.url_for('.user_settings'))
```
nextgens, line 9: Hmm, you should keep a redirect here; the idiom is POST+redirect (302) ... so that if the user refreshes the page he doesn't get shown the scary "do you want to resubmit the form" dialog.

core/admin/mailu/ui/templates/base.html
```diff
@@ -1,65 +1,83 @@
...
+              {%- endfor %}
             </div>
           </li>
         </ul>
       </nav>
-      <aside class="main-sidebar sidebar-dark-primary">
-        <a href="{{ config["WEB_ADMIN"] }}" class="brand-link">
-         <span class="brand-text font-weight-light">{{ config["SITENAME"] }}</span>
+      <aside class="main-sidebar sidebar-dark-primary nav-compact elevation-4">
+        <a href="{{ url_for('.domain_list') }}" class="brand-link bg-primary">
```
Diman0, line 48: This should only direct to domain_list if you have access to it. Maybe change it to
```
{%- if current_user.manager_of or current_user.global_admin %}
<a href="{{ url_for('.domain_list') }}" class="brand-link bg-primary">
{%- elif current_user.is_authenticated %}
<a href="{{ url_for('.user_settings') }}" class="brand-link bg-primary">
{%- else %}
<a class="brand-link bg-primary">
{%- endif %}

**Review by Diman0 — CHANGES_REQUESTED** [review]

File: core/nginx/conf/nginx.conf [hunk 3]
```diff
@@ -33,6 +33,17 @@ http {
       default $http_x_forwarded_proto;
       ''      $scheme;
     }
+    map $uri $expires {
```
nextgens, line 36: Let's keep caching out of this PR maybe? it goes hand-in-hand with @Diman0's work on moving /static out ...

and as is, I am not convinced that it won't break things (.js may be dynamic on webmails)

core/admin/mailu/configuration.py
```diff
@@ -143,6 +143,7 @@ def init_app(self, app):
         self.config['SESSION_COOKIE_SAMESITE'] = 'Strict'
         self.config['SESSION_COOKIE_HTTPONLY'] = True
         self.config['PERMANENT_SESSION_LIFETIME'] = timedelta(hours=int(self.config['SESSION_LIFETIME']))
+        self.config['HOSTNAME'] = self.config['HOSTNAMES'].split(',', 1)[0].strip()
```
nextgens, line 146: Maybe we should normalize HOSTNAMES instead (to avoid having to strip elsewhere in the code too)

core/admin/mailu/ui/templates/base.html
ghostwheel42: yes. sure. I've updated the PR.

core/admin/mailu/ui/views/languages.py
ghostwheel42: The language "POST" ist not called via a form submit but by an ajax call.
So the user won't see resubmit dialogs etc - the call only needs a "200 OK" and triggers a reload of the current page (see app.js).
`
$.post({
            url: $(this).attr('href'),
            success: function() {
                location.reload();
            },
        });
`

But we could discuss about an error handler or a warning about losing form data when changing the language while configuring something.

Diman0:
bors try

core/nginx/conf/nginx.conf
ghostwheel42: I'd like to set the caching header in one place, but maybe it's better to set the headers in the backend containers (admin, webmail, ...). This way users of "other" proxy setups will have headers too.
I'll triple-check to see if proxied content is handled correctly (ie. upstreams caching and vary headers are used).
And I'll try to add the headers via gunicorn/flask/static handler (and maybe even use your snipped for serving pre-compressed content)

core/admin/mailu/configuration.py
ghostwheel42: Yes, I'll do this. But I like to keep the separate HOSTAME variable in the config, so I don' need to spit everywhere.
But I don't know a way to fix/check the environment before used by docker-compose.
After all the splitting is done in every container in start.py or config.py

core/admin/mailu/ui/views/languages.py
nextgens: OK the important thing is that there is a GET following... I guess you could replace the success: by a .done() afterwards instead

core/nginx/conf/nginx.conf
ghostwheel42: @nextgens Okay, I've modified the caching to work only on mailu static content (admin and front container).
And I've enabled caching headers in the roundcube container (enabled apache modules necessary).

Diman0 (changes requested):
You added two new env variables (LOGO_BACKGROUND, LOGO_URL). These need to be documented. Do you want me to add a commit for this?

ghostwheel42:
> Do you want me to add a commit for this?

Yes, please. That would be nice.

Diman0:
bors try

core/admin/mailu/configuration.py
ghostwheel42: I am normalizing the HOSTNAMES now.
But we should - as discussed - move this to a python library in start.py

ghostwheel42:
bors try

ghostwheel42:
bors try

core/admin/mailu/models.py
```diff
@@ -285,7 +312,7 @@ def has_email(self, localpart):
     def check_mx(self):
         """ checks if MX record for domain points to mailu host """
         try:
-            hostnames = set(app.config['HOSTNAMES'].split(','))
+            hostnames = set(fqdn.strip() for fqdn in app.config['HOSTNAMES'].split(','))
```
nextgens, line 315: That's not required anymore

ghostwheel42: Yes. I've removed the change.

Diman0:
I think the PR is fine now? 
Can you confirm you have made all desired changes now?
I will do one more quick test and will approve it after your confirmation.

ghostwheel42:
@Diman0 yes. this works as intended now.
Maybe we could open a new Issue for the "Move startup functions (normalization, etc) to a python library to be used in start.py".

Diman0 (approved):
Yes. That is something we must tackle in the future. I think this should go together with update documentation for developing for Mailu. 

LGTM. Thank you again for your contribution!

PR #1967, opened 2021-09-03
fix 1789: ensure that nginx resolves ipv4 addresses

### LINKED ISSUE #1789: gunicorn not listening on IPv6

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

This fixes ipv6 enabled setup by disabling it. If you were using SUBNET6 in your configuration, odds are it's broken since gunicorn isn't bound on an on an ipv6 enabled socket.

Should we backport this?

### Related issue(s)
- close #1789
- close #1802

### DIFF
--- a/core/nginx/conf/nginx.conf
+++ b/core/nginx/conf/nginx.conf
@@ -18,7 +18,7 @@ http {
     keepalive_timeout  65;
     server_tokens off;
     absolute_redirect off;
-    resolver {{ RESOLVER }} valid=30s;
+    resolver {{ RESOLVER }} ipv6=off valid=30s;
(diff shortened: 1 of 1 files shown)


ghostwheel42 (approved):
Pragmatic solution which can easily be backported. 
Maybe we’ll switch to fully enabled IPv6 in the internal network later.

PR #1980, opened 2021-09-16
fix freshclam startup

### LINKED ISSUE #1632: freshclam log is not on stdout
`freshclam` is started with `-d` and thus forks into the background. That way there are no log messages regarding successful/failed/skipped updates.

In bash it would probably be something like this

```sh
freshclam -c 6 &
clamav
```

ghostwheel42:
## What type of PR?

bug-fix

## What does this PR do?

make freshclam start and run
- create pid file in existing folder /run
- let freshclam log to stdout
- remove deprecated SafeBrowsing

### Related issue(s)

closes #1632 

## Prerequisites

### DIFF
--- a/optional/clamav/conf/freshclam.conf
+++ b/optional/clamav/conf/freshclam.conf
@@ -3,9 +3,9 @@
 ###############
 
 DatabaseDirectory /data
-LogSyslog yes
+UpdateLogFile /dev/stdout
 LogTime yes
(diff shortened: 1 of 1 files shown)


Diman0 (approved):
lgtm

teadur:
By any chance woult it be possible to include suggested fixes from https://github.com/Mailu/Mailu/issues/470#issuecomment-760964079 here aswell ?

Issue #224, opened 2017-07-25
Provide auto-configuration files


nalt:
Modern mail clients attempt auto-configuration to greatly simplify the setup process. Mainly they use XML files requested via http.

Can this be included in mailu? It shouldn't be much effort - an XML template, SSI and the nginx config might be enough.

kaiyou:
What do you mean SSI? Anyway I think this could indeed be handled in the current version using `302` to the admin interface and generating the cert there. However, how does the MUA query the XML for a given domain if that domain only exposes a `MX` record?

nalt:
Thunderbird checks e.g. autoconfig.<domain>. If the name cannot be resolved, autoconfig fails. Needs to be checked if a 302 is ok. The process is described in https://www.ullright.org/ullWiki/show/providing-email-client-autoconfiguration-information-from-moens-ch .

SSI is like a simple template language handled by the server. But it's very old.

There is also a project, automx.org, but that's probably overkill.

kaiyou:
Well, this sounds a bit more complex than I originally expected. First, we have to handle certificates properly: we can't have domain certificate for every `autoconfig.<domain>` generated automatically, so we should check if a `CNAME` on autoconfig will do the trick or if maybe Thunderbird can rely on the MX record or any other SRV record.

Also, implementation for Outlook seems completely different.

kaiyou:
Could this be partially implemented if we integrate with z-push for Outlook integration?

HorayNarea:
I'm looking currently looking into this, at least Mozilla-compatible auto-config seems very easy because:
- CNAMEs are fine
- HTTP-redirects are fine
- even non-HTTPS is fine (but i hope we all agree that it should not be done this way…)
- it's basically just a static XML-file (real-life example in production: https://autoconfig.lynx.uberspace.de/mail/config-v1.1.xml?emailaddress=myname@example.com)

kaiyou:
So, a simple CNAME for `autoconfig.`` on the domain (and zone configuration provided on the details page), and a ``/mail/config*`` endpoint on both the frontend and admin container could actually do the trick. All configurations would be identical I guess, so it's pretty stateless and simple to implement.

kaiyou:
If anyone want to give it a go, please PR. Otherwise, I will start working on this after ``1.5.1`` is properly released :)

HorayNarea:
[...]

PR #1986, opened 2021-09-23
Document how to setup autoconfig

nextgens:
## What type of PR?

enhancement

## What does this PR do?

Document how to setup autoconfig. This works with most open-source MUAs (thunderbird, k9mail, evolution, ...)

We could go further than that by providing dynamic configuration (issue an auth token for each MUA request)... but it won't work unless a new DNS entry (and matching certificate) is created.

### Related issue(s)
- #224

## Prerequisites


Diman0:
I think this is simple enough for now. I will try the PR this weekend.

PR #1992, opened 2021-09-25
Improve unbound configuration

nextgens:
## What type of PR?

enhancement

## What does this PR do?

Improve the unbound configuration:
- remove settings that were default
- don't setup IPv6 if it's not configured
- enable qname minimization (for privacy)
- add a minimum TTL of 5m (prevent excessive resource usage and naive rebinding attacks)

## Prerequisites

### DIFF
--- a/optional/unbound/unbound.conf
+++ b/optional/unbound/unbound.conf
@@ -1,19 +1,21 @@
 server:
   verbosity: 1
   interface: 0.0.0.0
-  interface: ::0
+  {{ 'interface: ::0' if SUBNET6 }}
   logfile: ""
   do-ip4: yes
-  do-ip6: yes
+  do-ip6: {{ 'yes' if SUBNET6 else 'no' }}
   do-udp: yes
   do-tcp: yes
   do-daemonize: no
   access-control: {{ SUBNET }} allow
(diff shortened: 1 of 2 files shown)


optional/unbound/unbound.conf
```diff
@@ -1,19 +1,21 @@
...
   directory: "/etc/unbound"
   username: unbound
   auto-trust-anchor-file: trusted-key.key
   root-hints: "/etc/unbound/root.hints"
   hide-identity: yes
   hide-version: yes
-  max-udp-size: 4096
-  msg-buffer-size: 65552
+  cache-min-ttl: 300
+  qname-minimisation: yes
```
HorayNarea, line 20: this is not necessary since qname-minimisation is enabled by default - quoting from the unbound.conf manpage:

```
qname-minimisation: <yes or no>
       Send minimum amount of information to upstream servers to
       enhance privacy.  Only send minimum required labels of the QNAME
       and set QTYPE to A when possible. Best effort approach; full
       QNAME and original QTYPE will be sent when upstream replies with
       a RCODE other than NOERROR, except when receiving NXDOMAIN from
       a DNSSEC signed zone. Default is yes.
```

nextgens: Good catch, my config hasn't been updated since qname-minimisation-strict has been introduced :$

PR #1993, opened 2021-09-25
Corrected typo in the fail2ban documentation

marioja:
Once this has been merged, will this update the mailu.io website?

## What type of PR?

documentation

## Prerequisites

none

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -551,7 +551,8 @@ follow these steps:
   [Definition]
   failregex = .* client login failed: .+ client:\ <HOST>
   ignoreregex =
-  journalmatch = CONTAINER_TAG=mailu-front
+  journalmatch =
+  CONTAINER_TAG=mailu-front
 
 3. Add the /etc/fail2ban/jail.d/bad-auth.conf
 


ghostwheel42:
Hi @marioja
This is not a type. It's a correct journald filter on the tag "mailu-front" set before.
Maybe there's an error in your docker/journal configuration.

marioja:
> Hi @marioja This is not a type. It's a correct journald filter on the tag "mailu-front" set before. Maybe there's an error in your docker/journal configuration.

When you say that this is a correct journald filter, you mean that the line:
`journalmatch = CONTAINER_TAG=mailu-front`
is correctly formatted, i.e. it should be on a single line.  I can accept that.  However, I do not understand your comment about docker/journal configuration.  Remember that fail2ban is not running in a container but rather on the host.  So how can a docker journal configuration affect fail2ban running on the host.

One thing that puzzles me even more is that when I installed fail2ban on the host (Ubuntu 20.04) the journalmatch line as refered to above (on a single line) was causing fail2ban to fail.  However, today upon your last comment, I reverted the change and fail2ban now starts ok on the host. I will try to dig the old log files with the error I was getting.

ghostwheel42:
>  However, I do not understand your comment about docker/journal configuration

One word before: I'm using fail2ban, but not with mailu and I also don't use journald logging.

With `docker / journal configuration` I was referring to the change to the compose file needed for this to work:
```
    logging:
      driver: journald
      options:
        tag: mailu-front
```

I changed my test-setup to do this journald-logging and I could see log entries when running `journalctl CONTAINER_TAG=mailu-front`. I also saw the lines the ban-regex is looking for:
```
2021/10/14 10:18:15 [info] 11#11: *44 client login failed: "Authentication credentials invalid" while in http auth state, client: ...
```

But I didn't test if fail2ban matches this. Maybe the regex in the docs is broken?
You could try `failregex = ^.*client login failed:.*client: <HOST>,`

You can debug fail2ban using `fail2ban-regex`. The following command should show you failed attempts (not tested):
`fail2ban-regex systemd-journal bad-auth.conf --print-all-matched`

PR #1998, opened 2021-09-28
Fix healthcheck

ghostwheel42:
## What type of PR?

bug-fix

## What does this PR do?

- fixes health check of roundcube container (introduced with sso / endless redirect)
- silence logging of health checks
- write php error log directly to /dev/stderr

## Prerequisites


nextgens (approved):
lgtm

PR #2002, opened 2021-10-01
Misc improvements to the postfix configuration

nextgens:
## What type of PR?

enhancement

## What does this PR do?

- It removes a useless "Received" header
- It changes the SRS key from SECRET_KEY to a subkey (this will affect "in flight" bounces)

### Related issue(s)
- #466

## Prerequisites


core/postfix/conf/outclean_header_filter.cf
```diff
@@ -1,10 +1,7 @@
 # This configuration was copied from Mailinabox. The original version is available at:
 # https://raw.githubusercontent.com/mail-in-a-box/mailinabox/master/conf/postfix_outgoing_mail_header_filters
 
-# Remove the first line of the Received: header. Note that we cannot fully remove the Received: header
-# because OpenDKIM requires that a header be present when signing outbound mail. The first line is
-# where the user's home IP address would be.
-/^\s*Received:[^\n]*(.*)/         REPLACE Received: from authenticated-user ({{OUTCLEAN}} [{{OUTCLEAN_ADDRESS}}])$1
+/^\s*Received:[^\n]*(.*)/	IGNORE
```
ghostwheel42, line 4: This should be `/^\s*Received:/ IGNORE` to ignore the received header.

But doesn't the comment say "We can't remove it because OpenDKIM requires that a header is present"? Is it not necessary anymore?

nextgens: My understanding is that we use rspamd for DKIM signing... so it's unlikely to have ever been relevant.

It was introduced in 40b9883c8 without any further justification; googling it I stumbled upon

https://bofhskull.wordpress.com/2014/03/25/postfix-opendkim-and-missing-from-header/

I'm running it with IGNORE here locally and so far no ill effect has been spotted

ghostwheel42: I understand. Why did you leave `[^\n]*` inside the regex? Just use `/^\s*Received:/ IGNORE`
We can even condense the whole block (line 4-11) to
```
# Remove other typically private information.
/^\s*(Received|User-Agent|X-(Enigmail|Mailer|Originating-IP|Pgp-Agent)):/        IGNORE
```

nextgens: Agreed; Since we have a regexp engine we may as well use it...

ghostwheel42 (approved):
lgtm

PR #2005, opened 2021-10-08
SESSION_COOKIE_SECURE and HTTP won't work; make it explicit by displaying an alert

### LINKED ISSUE #1996: Admin login not working
I installed mailu using docker-compose, following all the steps. I can access /webmail, send emails and so on. But I can't access /admin at all, if I put the wrong password it gives an error, if I put the correct one it simply doesn't enter and doesn't return an error.

nextgens:
## What type of PR?

enhancement

## What does this PR do?

Display an explicit alert and disable the login form (using javascript) when we have SESSION_COOKIE_SECURE=True and clients using HTTP to access it.

### Related issue(s)
- close #1996

## Prerequisites

### DIFF
--- a/core/admin/mailu/ui/templates/login.html
+++ b/core/admin/mailu/ui/templates/login.html
@@ -8,11 +8,11 @@
 {% trans %}to access the administration tools{% endtrans %}
 {%- endblock %}
 
-{% if config["SESSION_COOKIE_SECURE"] %}
 {%- block content %}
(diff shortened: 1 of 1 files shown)


ghostwheel42:
@nextgens What do you think about this?
We could (or should?) set a content-security-policy without 'unsafe-inline' - but that's for another PR.

nextgens:
bors cancel

PR #2012, opened 2021-10-11
Feature/dkim on redis

abs0lom:
## What type of PR?

Feature for helm-chart contribution

## What does this PR do?

Many K8s distrib doesn't support sharing persistent volumes between pods. The only trully shared folder between containers is containing the dkim keys.

rspamd support storage of the dkim keys into redis and it's the best solution for share dkim between the admin and rspamd imo.

This PR achieve this.

### Related issue(s)
- none, I believe so.

## Prerequisites

### DIFF
--- a/core/admin/mailu/models.py
+++ b/core/admin/mailu/models.py
@@ -192,6 +192,13 @@ class Domain(Base):
     _dkim_key = None
     _dkim_key_on_redis = None
 
+    def _dkim_file(self):
+        """ return filename for active DKIM key """
+        return app.config['DKIM_PATH'].format(
+            domain=self.name,
+            selector=app.config['DKIM_SELECTOR']
+        )
+
     def save_dkim_key(self):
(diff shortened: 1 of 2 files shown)


ghostwheel42:
I've created a proof of concept for using dkim keys from the admin container via the hashicorp vault api v1 supported by rspamd. This can be upgraded later to deliver multiple keys with multiple selectors from the mailu configuration database.
Here is what I've changed: https://github.com/Mailu/Mailu/compare/master...ghostwheel42:enhance_dkim

ghostwheel42:
> Hashicorp allow to add ed25519, safe key rotation, and permit to break the shared volume.

I only implemented the "GET" part of the vault api. You will not be able to use the `rspamadm` commands for key management.
Key management is solely made by mailu and currently supports only one key per domain (rsa 2048 bits).

But as nextgens said: it will allow us to improve the key management part in the admin container.

PR #2029, opened 2021-10-31
The Rate-limiter should run after the deny

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

It ensures that we *always* prevent login attempts on port 25; without it clients eventually get a temporary error instead of a permanent one.


Diman0 (approved):
Looks good to me.

PR #2030, opened 2021-10-31
Fix RELAYNETS

### LINKED ISSUE #360: RELAYNETS behavior
Hello

I'ld like to know if the comma separated list is valid

    RELAYNETS=172.16.0.0/12,10.0.0.0/8,192.168.0.0/16

By the way, this example list is RFC1918, aka "all local network". Maybe this should be the default. Fact is, my Docker Swarm's overlay network is in the 10 range, not the 172.

And very nice jobs putting this mailu together, I'll definitively put it to works.

Regards,

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

RELAYNETS should be comma separated like everything else; rspamd should also be aware of what is considered "trusted".

I am not sure whether ```local_networks``` is the right configuration option for it though

- close #360

### DIFF
--- a/core/postfix/conf/main.cf
+++ b/core/postfix/conf/main.cf
@@ -17,7 +17,7 @@ queue_directory = /queue
 message_size_limit = {{ MESSAGE_SIZE_LIMIT }}
 
 # Relayed networks
-mynetworks = 127.0.0.1/32 [::1]/128 {{ SUBNET }} {{ RELAYNETS }}
+mynetworks = 127.0.0.1/32 [::1]/128 {{ SUBNET }} {{ RELAYNETS.split(",") }}
 
 # Empty alias list to override the configuration variable and disable NIS
 alias_maps =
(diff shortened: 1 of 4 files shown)


Diman0:
Can you also update the RELAYNETS description in configuration.rst? It does not state that this is a comma delimited list.
E.g. change
```The RELAYNETS are network addresses for which mail is relayed for free with no authentication required. ```
to
```The RELAYNETS is a comma delimited list of network addresses for which mail is relayed for free with no authentication required. ```

The rspamd documentation states for local_networks: ```map or list of IP networks used as local, so certain checks are skipped for them (e.g. SPF checks)```. 
After searching I found this page (https://rspamd.com/doc/tutorials/scanning_outbound.html)  that states what will be disabled. It states that some checks for outbound email are disabled for '_authenticated users_' and IP addresses from local_networks.

I think you found the correct setting. If you want to, I can update that line in the documentation. Then everything should be fine for this PR.

Diman0 (approved):
Looks good to me.

PR #2033, opened 2021-11-01
Misc

### LINKED ISSUE #1705: Every mail is tagged by RSPAMD as RCVD_NO_TLS_LAST

## Before you open your issue

### Environment
 - [x] docker-compose
 - [ ] kubernetes
 - [ ] docker swarm

### Versions
My version is 1.7

## Description
Every mail received also from GMAIL is marked by RSPAMD as RCVD_NO_TLS_LAST.
Is normal because I think that mail from famous provider use TLS 

## Replication Steps
Simply send an email from gmail anch check RSPAMD history

## Expected behaviour
I expected to recieve an email with TLS in each step.

## Thanks
Thank you very much for this wonderful project that makes it possible for everyone to build a mail server

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

- Remove /dkim from the rspamd container in setup as that's not necessary anymore;
@micw helms-chart should probably be updated too. There is an open question on whether we want to keep it on admin. We can either have DKIM keys auto-imported to the DB on upgrade (if they exist) or provide a facility to manually import them (so that the directory can be removed altogether). Having the former doesn't allow for removing /dkim from admin. @Diman0 any opinion on what we want for 1.9?

- Remove spam points from RCVD_NO_TLS_LAST as we don't detect whether TLS was used or not.

### Related issue(s)
- close #1705

## Prerequisites

### DIFF
--- a/core/rspamd/conf/settings.conf
+++ b/core/rspamd/conf/settings.conf
@@ -1,0 +1,4 @@
+apply {
+	# see https://github.com/Mailu/Mailu/issues/1705
+	RCVD_NO_TLS_LAST = 0;
+}
--- a/setup/flavors/compose/docker-compose.yml
+++ b/setup/flavors/compose/docker-compose.yml
@@ -90,7 +90,6 @@ services:
     env_file: {{ env }}
     volumes:
       - "{{ root }}/filter:/var/lib/rspamd"
-      - "{{ root }}/dkim:/dkim:ro"
       - "{{ root }}/overrides/rspamd:/etc/rspamd/override.d:ro"
     depends_on:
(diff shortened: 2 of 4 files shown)


Diman0:
@nextgens  (DKIM) If it does not exist yet, someone should create an issue for discussing this. Before any decision is taken, I'd like to have discussed this. We could also discuss this at the next meeting.

/dkim is indeed not used. It was used in the past. When we switched to retrieving these keys via admin, I guess the person forgot to remove this line in the Dockerfile.

Diman0 (approved):

PR #2034, opened 2021-11-02
Add timezone to containers

### LINKED ISSUE #1154: Timezone in containers
Hello,

I have seen that timezones are not set for containers. This causes logs to have wrong timestamps and email "received" headers have wrong timezones.

It should be possible to set the timezone for all containers. The setting should be exposed and described in the .env file.

DjVinnii:
## What type of PR?

Enhancement

## What does this PR do?
This PR adds the tzdata package so that the environment variable `TZ` can be used to set the timezone of containers.

### Related issue(s)
- closes #1154 

## Prerequisites

### DIFF
--- a/webmails/roundcube/Dockerfile
+++ b/webmails/roundcube/Dockerfile
@@ -9,7 +9,7 @@ FROM ${ARCH}php:7.4-apache as build_other
 FROM build_${QEMU}
 #Shared layer between rainloop and roundcube
 RUN apt-get update && apt-get install -y \
-  python3 curl python3-pip git python3-multidict \
+  python3 curl python3-pip git python3-multidict tzdata \
   && rm -rf /var/lib/apt/lists \
   && echo "ServerSignature Off\nServerName roundcube" >> /etc/apache2/apache2.conf \
   && sed -i 's,CustomLog.*combined$,\0 "'"expr=!(%{HTTP_USER_AGENT}=='health'\&\&(-R '127.0.0.1/8' || -R '::1'))"'",' /etc/apache2/sites-available/000-default.conf
@@ -21,12 +21,14 @@ ENV ROUNDCUBE_URL https://github.com/roundcube/roundcubemail/releases/download/1
 
 ENV CARDDAV_URL https://github.com/mstilkerich/rcmcarddav/releases/download/v4.1.2/carddav-v4.1.2.tar.gz
 
+ENV TZ Etc/UTC
+
 RUN apt-get update && apt-get install -y \
       zlib1g-dev libzip4 libzip-dev libpq-dev \
       python3-jinja2 \
       gpg \
  && docker-php-ext-install zip pdo_mysql pdo_pgsql \
- && echo date.timezone=UTC > /usr/local/etc/php/conf.d/timezone.ini \
+ && echo date.timezone=${TZ} > /usr/local/etc/php/conf.d/timezone.ini \
  && rm -rf /var/www/html/ \
  && cd /var/www \
  && curl -sL ${ROUNDCUBE_URL} | tar xz \
--- a/webmails/roundcube/Dockerfile
+++ b/webmails/roundcube/Dockerfile
@@ -7,6 +7,9 @@ ONBUILD COPY --from=balenalib/rpi-alpine:3.14 /usr/bin/qemu-arm-static /usr/bin/
 FROM ${ARCH}php:7.4-apache as build_other
 
 FROM build_${QEMU}
+
+ENV TZ Etc/UTC
+
 #Shared layer between rainloop and roundcube
 RUN apt-get update && apt-get install -y \
   python3 curl python3-pip git python3-multidict tzdata \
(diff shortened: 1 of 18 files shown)


Diman0:
Did you forget adapting the Dockerfiles of rainloop and roundcube?
https://github.com/Mailu/Mailu/blob/master/webmails/rainloop/Dockerfile
https://github.com/Mailu/Mailu/blob/master/webmails/roundcube/Dockerfile

For the documentation, please mention that the value for TZ can be found on https://en.wikipedia.org/wiki/List_of_tz_database_time_zones. The value  "TZ database name" can be used from the table on that page.

DjVinnii:
@Diman0 My bad. I added the tzdata to the webmails as well.

ghostwheel42:
LGTM

I'm wondering if creating /etc/localtime and /etc/timezone would be useful?
```
cp /usr/share/zoneinfo/${TZ} /etc/localtime
echo "${TZ}" >  /etc/timezone
```
I think tzdata could be uninstalled after this (smaller image).

DjVinnii:
@ghostwheel42 Tzdata has to be present to make use of /etc/localtime and /etc/timezone.

Diman0 (dismissed):
Since we introduce a new ENV variable, it should be added to the config manager with a default value. I'm referring to this file:
https://github.com/Mailu/Mailu/blob/5e212ea46d869dfc97fc22a71ee705b5c6a46901/core/admin/mailu/configuration.py#L9

You could add the line
``'TZ': 'Etc/UTC',``

It would help users a lot, if we added the new setting to the mailu.env file (with a comment above it). This is another hint for users that they can change the timezone. To alter the mailu.env file for compose/stack you can alter:
https://github.com/Mailu/Mailu/blob/master/setup/flavors/compose/mailu.env
It makes sense to me to add it in the section "# Common configuration variables".
E.g.
```
# Timezone for the Mailu containers. See this link for all possible values https://en.wikipedia.org/wiki/List_of_tz_database_time_zones
TZ=Etc/UTC
```

ghostwheel42:
@DjVinnii 
> Tzdata has to be present to make use of /etc/localtime and /etc/timezone

That's not exactly correct. It has to be present when copying, but after this the whole set of timezones is not necessary anymore. See https://wiki.alpinelinux.org/wiki/Setting_the_timezone

This is: install - copy/echo - remove

The nginx Dockerfile would look like this:

```
ENV TZ=Etc/UTC

RUN set -eo \
 && apk add --no-cache certbot nginx nginx-mod-mail openssl curl tzdata \
 && pip3 install watchdog \
 && cp /usr/share/zoneinfo/${TZ} /etc/localtime \
 && echo "${TZ}" >/etc/timezone \
 && apk del tzdata

```

DjVinnii:
@ghostwheel42 That's weird. I tried this, however the timezone reverted back to UTC after removing tzdata. I will take a look at it this afternoon.

ghostwheel42:
Yes. that's weird. I've tested it using only the nginx container and it worked as expected.
Maybe the `ENV TZ=Etc/UTC` makes the difference.

DjVinnii:
It just came to mind that your way of setting the timezone wouldn't allow users to set the timezone without building the image themselves.

ghostwheel42:
> It just came to mind that your way of setting the timezone wouldn't allow users to set the timezone without building the image themselves.

Yes, you are correct. I didn't see this - my bad. tzdata needs to remain inside the image.
I still do not know if setting `/etc/localtime` or `/etc/timezone` should be done too.
Currently I don't see any need, but some (obscure) software added later could rely on this.

ghostwheel42:
If it's necessary to change the /etc files it can be done in a later PR in the entrypoint.

webmails/roundcube/Dockerfile
```diff
@@ -21,12 +21,14 @@ ENV ROUNDCUBE_URL https://github.com/roundcube/roundcubemail/releases/download/1
 
 ENV CARDDAV_URL https://github.com/mstilkerich/rcmcarddav/releases/download/v4.1.2/carddav-v4.1.2.tar.gz
 
+ENV TZ Etc/UTC
```
ghostwheel42, line 24: Please add this line to all Dockerfiles to set a default. 

The 'echo TZ' needs to be moved to entrypoint (jinja template php.ini)

ghostwheel42 (dismissed):

webmails/roundcube/Dockerfile
DjVinnii: When setting the timezone in roundcube.ini (jinja), does the timezone still has to be set in timezone.ini?

ghostwheel42:
@DjVinnii Thanks for changing this. 
Now there are some discussions in the matrix dev channel about using any other timezone except UTC making Mail debugging really painful.
It's true: a real MTA should always use UTC to make the logs consistent. 
And the best solution for the different timestamps in the logs would be to change the hosts timezone to UTC.

I think this PR can still be merged, but maybe you could add another comment to the env file and/or docs like so: "warning: if you are observing different timestamps in your log files you should change your hosts timezone to UTC instead of changing TZ to your local timezone. Using UTC allows easy log correlation with remote MTAs."

ghostwheel42:
bors try

nextgens:
Please move it to the "Advanced settings" section too; it's definitely not "Common configuration" and not something we should encourage people to change.

Daylight Saving Time makes all of this very difficult to correlate: DST isn't the same everywhere (https://en.wikipedia.org/wiki/Daylight_saving_time_by_country). The solution is trivial: stick to GMT/Zulu time everywhere.

ghostwheel42:
bors try

ghostwheel42 (approved):
local test successful. LGTM

Diman0 (approved):
LGTM. Thank you for your contribution.

PR #2037, opened 2021-11-02
update python dependencies of admin container

ghostwheel42:
## What type of PR?

updates python dependencies of admin container

## Prerequisites


Diman0:
Can't you remove all fixed version numbers from requirements.txt, upgrade all packages and then run pip freeze? Or is this too simplistic?
I see that the requirements for example still contains SQLAlchemy=1.3.3 while the current version is 1.4.26.  I do not know if what uses SQLAlchemy actually supports version 1.4.26.

ghostwheel42:
Hi @Diman0 - this PR is tagged Work In Progress.
I'm doing this step by step to catch and resolve errors easier.

ghostwheel42:
bors try

core/admin/requirements.txt
```diff
@@ -18,10 +18,8 @@ PyYAML
 PyOpenSSL
 Pygments
 dnspython
-bcrypt
```
nextgens, line 21: Hmm? it's still in requirements-prod.txt

core/admin/mailu/utils.py
```diff
@@ -265,7 +268,7 @@ def save(self):
 
         # set uid from dict data
         if self._uid is None:
-            self._uid = self.app.session_config.gen_uid(self.get('user_id', ''))
+            self._uid = self.app.session_config.gen_uid(self.get('_user_id', ''))
```
nextgens, line 271: What's going on here?

core/admin/requirements.txt
ghostwheel42: Yes. Mailu doesn't use it directly but it's used by cryptography which is used by pyOpenSSL.

core/admin/mailu/utils.py
ghostwheel42: It's not set by us and Flask-login now prepends values with an underscore:

https://github.com/maxcountryman/flask-login/blob/c760c0ef7ccc95d49b4693200245a4f2b148d41b/CHANGES#L32

ghostwheel42:
bors try

core/admin/migrations/env.py
```diff
@@ -69,28 +63,35 @@ def process_revision_directives(context, revision, directives):
...
-                      process_revision_directives=process_revision_directives,
-                      **current_app.extensions['migrate'].configure_args)
-
-    try:
         with context.begin_transaction():
             context.run_migrations()
-    finally:
-        connection.close()
+
+    connection.close()
```
nextgens, line 94: It's probably not necessary. Are we sure that it doesn't throw if it's closed twice? The ```with``` above should have closed it already

nextgens (commented):
I think we should switch to a single requirements.txt (with pinned versions) and have an extra requirements-dev.txt

core/admin/migrations/env.py
ghostwheel42: No, afaik the with part is only a sql transaction using the connection. The connection is not closed by the transaction. 

But I don't know why I removed the try/finally. It probably doesn't matter, the connection is closed automatically at the end. 

I'd like to leave the close call as is.

ghostwheel42:
> I think we should switch to a single requirements.txt

I didn't invent but I like the use of a requirements.txt only listing our direct dependencies.

An extra requiremts-dev.txt is a good idea, it would also include the stuff mentioned in the "reformat code" PR (black, etc)

Diman0:
I also like we have a requirements file with only dependencies. I'd like to have at least one requirements file with pinned versions and one without pinned versions. This makes it easy to use the latest version of each package during development.

I had a look at the code and it looks sound to me. The SQLAlchemy part was a bit difficult since I'm not familiar with this and had to lookup in the SQLAlchemy docs that cache_ok and python_type are valid. 

I haven't tested the PR yet. I will try doing that later today.

core/admin/mailu/ui/templates/domain/list.html
```diff
@@ -46,8 +46,8 @@
     <td>{{ domain.users | count }} / {{ '∞' if domain.max_users == -1 else domain.max_users }}</td>
     <td>{{ domain.aliases | count }} / {{ '∞' if domain.max_aliases == -1 else domain.max_aliases }}</td>
     <td>{{ domain.comment or '' }}</td>
-    <td>{{ domain.created_at }}</td>
-    <td>{{ domain.updated_at or '' }}</td>
+    <td>{{ domain.created_at | format_date }}</td>
```
Diman0, line 49: Can you provide more information why you introduced this formatting? Did someone request this, or do you simply more like a more human readable date?
Can you also provide more information why you only format these two date columns, but do not apply this date formatting in the other locations in the admin interface where we use dates?

Diman0:
bors try

Diman0 (changes requested):
Adding a new fetch account (https://mydomain/admin/fetch/create/) is broken. You have no dropdown for imap/pop3 anymore. It shows two separate entries.
afbeelding


Add alias (for user) screen is also broken (https://Mydomain.com/admin/alias/create/MyDomain.com). It shows all email addresses as plain text. The drop down which is displayed when entering a target email address is broken.
afbeelding


Maybe your changes to macros.html causes this? 
I mean line 27
```
  {%- if field is iterable %}
          {%- for subfield in field %}
```

ghostwheel42:
@Diman0 thanks for spotting this, I've fixed it.

The Date-Selector we talked about `/admin/user/reply` does not seem to be localized. It always display the German date format here regardless of the selected language (that is DD.MM.YYYY) - I've not found a way to change this yet.

Diman0:
> @Diman0 thanks for spotting this, I've fixed it.
> 
> The Date-Selector we talked about `/admin/user/reply` does not seem to be localized. It always display the German date format here regardless of the selected language (that is DD.MM.YYYY) - I've not found a way to change this yet.

I did not notice this. This is also the Dutch format so it looks normal to me. This should be localized or use  the ISO-format (YYYY-MM-DD). 

The date fields are defined in forms.py (you probably already know this)
https://github.com/Mailu/Mailu/blob/d8c6a2d15edce892c92363d45f8127cdd89ec8a4/core/admin/mailu/ui/forms.py#L121
Maybe you could play with the field type?

Looking at the documentation you can provide formatting. Maybe you could set it to ISO format for now?
Funny thing is that according to the docs it should already use the ISO format!!
https://wtforms.readthedocs.io/en/2.3.x/fields/#wtforms.fields.html5.DateField
```
    reply_startdate = fields.html5.DateField(_('Start of vacation'), format='%Y-%m-%d')
    reply_enddate = fields.html5.DateField(_('End of vacation'), format='%Y-%m-%d')
```

ghostwheel42:
Yes, the date format already is ISO here (see the html source: `value="2999-12-31"`).
It's the html5 DateField which converts this to the browser's local date (your selected language in the browser's config).
What I didn't find yet is how to make the DateField use our selected language instead.
It's not a show-stopper and it probably is like this since version 1.5.

Diman0:
I concur it is not a show stopper. Does the non html5 control show the same behaviour?

ghostwheel42:
> I concur it is not a show stopper. Does the non html5 control show the same behaviour?

the non html5 selector is just an input field containing the iso date.
When testing I found a bug - the textarea is not "enabled". fixed it.

There also seems to be a problem when submitting invalid dates (not possible with the html5 selector) - there is no error message shown and the setting is not saved. We should stick to the html5 fields, so this is not a problem.

Diman0:
> > I concur it is not a show stopper. Does the non html5 control show the same behaviour?
> 
> the non html5 selector is just an input field containing the iso date. When testing I found a bug - the textarea is not "enabled". fixed it.
> 
> There also seems to be a problem when submitting invalid dates (not possible with the html5 selector) - there is no error message shown and the setting is not saved. We should stick to the html5 fields, so this is not a problem.

Yes. Let's stick with the HTML5 date controls. I will test one more time with the fixed field iteration and latest commit. Hopefully I will find nothing more and can approve the PR. 

bors try

Diman0 (approved):
It's looking good now :+1: 
All dates are localised now. Dropdown boxes are fixed. As a bonus on the login page, the submit buttons are  now aligned with the above entries (that was not the case before).

As usual, thank you for your efforts!

nextgens (approved):
LGTM

Issue #1774, opened 2021-02-25
Is MESSAGE_RATELIMIT used at all??


okossuth:
Hello

I found that in the master and version 1.8 docs appears the MESSAGE_RATELIMIT setting but in the source code it doesnt seem to be used at all. Only appear in the docs subdir:
```
# grep -r  MESSAGE_RATELIMIT *
docs/compose/.env:MESSAGE_RATELIMIT=100/day
docs/configuration.rst:The ``MESSAGE_RATELIMIT`` is the limit of messages a single user can send. This is
```
So what is the current status of this setting? Can i use it in mailu 1.7 or 1.8 or is it a future addition in the works?

Thanks

Diman0:
This issue can be closed as soon #1916 is merged.

Diman0:
MESSAGE_RATELIMIT is now available on master (well technically in about ~20 minutes once the CI/CD workflow has finished). For details see #1916 
I'm closing this issue.

PR #2042, opened 2021-11-06
Add MESSAGE_RATELIMIT_EXEMPTION

nextgens:
## What type of PR?

Enhancement

## What does this PR do?

Add a new knob called ```MESSAGE_RATELIMIT_EXEMPTION```.

### Related issue(s)
- #1774

## Prerequisites


ghostwheel42 (changes requested):
I think you should normalize this to a set in configuration.py and leave a simple `if sender in app.config['MESSAGE_RATELIMIT_EXEMPTION']:` here.

PR #2044, opened 2021-11-07
Vault/rspamd: don't return any key for relayed domains

nextgens:
## What type of PR?

enhancement

## What does this PR 

Don't return any key for relayed domains. We may want to revisit this (ARC signing)... but in the meantime it saves from a scary message in rspamd.
    
```signing failure: cannot request data from the vault url: /internal/rspamd/vault/v1/dkim/ ...```

### DIFF
--- a/core/admin/mailu/internal/views/rspamd.py
+++ b/core/admin/mailu/internal/views/rspamd.py
@@ -14,22 +14,14 @@ def vault_error(*messages, status=404):
 
 @internal.route("/rspamd/vault/v1/dkim/<domain_name>", methods=['GET'])
 def rspamd_dkim_key(domain_name):
-    models.Relay.query.get(domain_name) and return flask.jsonify({
-        'data': {
-            'selectors': []
-        }
-    })
-    domain = models.Domain.query.get(domain_name) or flask.abort(vault_error('unknown domain'))
-    key = domain.dkim_key or flask.abort(vault_error('no dkim key', status=400))
-    return flask.jsonify({
-        'data': {
-            'selectors': [
+    selectors = []
+    if domain := models.Domain.query.get(domain_name):
+        if key := domain.dkim_key:
+            selectors.append(
                 {
                     'domain'  : domain.name,
                     'key'     : key.decode('utf8'),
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by ghostwheel42 — COMMENTED** [review]
Shouldn't we always return an empty selector if the domain is not found or has no key?
When developing this rspamd only looked up the key for skim domains, so returning an error was correct.
I'd say we look up the domain and try to get the key and return the empty set of selectors if as default when looking up the key failed. What do you think?

**Comment by ghostwheel42** [comment]
Should we look up the key like this?
```
def rspamd_dkim_key(domain_name):
    selectors = []
    if domain := models.Domain.query.get(domain_name):
        if key := domain.dkim_key:
          selectors.append(
              {
                  'domain'  : domain.name,
                   key'     : key.decode('utf8'),
                  'selector': flask.current_app.config.get('DKIM_SELECTOR', 'dkim'),
              }
          )
    return flask.jsonify({'data': {'selectors': selectors}})
```

nextgens:
> Shouldn't we always return an empty selector if the domain is not found or has no key?

That's arguable. We need to separate the handling of DKIM and ARC:
For DKIM: that's arguable; IMHO it's a severe misconfiguration not to have DKIM enabled.
For ARC: it should always return a key. If none is set maybe the default (hostname's domain or something) or all keys should be returned.

> When developing this rspamd only looked up the key for skim domains, so returning an error was correct. I'd say we look up the domain and try to get the key and return the empty set of selectors if as default when looking up the key failed. What do you think?

I don't think that it's an option for ARC: there I do think that any key beats no key.

nextgens:
In fact rspamd doesn't differentiate and uses always the same query (regardless of whether it's DKIM or ARC)... so your approach is probably the way to go.

https://github.com/rspamd/rspamd/search?q=sign_using_vault

nextgens:
bors try

ghostwheel42:
We could add another route called /arc/ for this.
It's a separate configuration file in rspamd. 
As you proposed arc could use the main server's key (when configured).

ghostwheel42 (approved):
Lgtm

PR #2047, opened 2021-11-09
Do not call .split() on RELAYNETS if not specified

Grennith:
## What type of PR?

bug-fix

## What does this PR do?

The call to {{ RELAYNETS.split(",") | join(' ') }} when starting postfix breaks if RELAYNETS has not been specified using the environmental variables.

## Prerequisites


nextgens (approved):
LGTM

I don't understand why the tests didn't catch it

ghostwheel42 (approved):
Lgtm
This will also be handled by confit normalization later.

PR #2048, opened 2021-11-09
Add documentation for catch-all addresses

### LINKED ISSUE #2045: Request: Improve Documentation on "SQL Like" Aliases in the UI
Currently, the alias wildcard interface looks like this:

image

From here: https://github.com/Mailu/Mailu/blob/master/core/admin/mailu/ui/forms.py#L141

The wildcard config is well documented here (https://mailu.io/master/webadministration.html?highlight=sql%20like#aliases), but it's not obvious looking directly in the UI.

Would you be open to a UI improvement that would be a bit more self-documenting?

I see a few potential options, which I'll outline with UI examples:

## Option 1: Provide multi-char example inline

Since presumably most users would be using multi-char wildcard, perhaps an inline example of that makes the most sense.

image

## Option 2: Link to docs

Another option would be a link to the documentation directly.

image

## Option 3: Both multi and single char wildcard inline

This seems the clunkiest, but it's an option as well.
image

benyanke:
Resolves issue #2045 - see that thread for more information.

## What type of PR?

enhancement

### Related issue(s)
- closes #2045

### DIFF
--- a/core/admin/mailu/ui/forms.py
+++ b/core/admin/mailu/ui/forms.py
@@ -138,7 +138,7 @@ class TokenForm(flask_wtf.FlaskForm):
 class AliasForm(flask_wtf.FlaskForm):
     localpart = fields.StringField(_('Alias'), [validators.DataRequired(), validators.Regexp(LOCALPART_REGEX)])
     wildcard = fields.BooleanField(
-        _('Use SQL LIKE Syntax (e.g. for catch-all aliases)'))
+        _('Use SQL LIKE Syntax (e.g. for catch-all aliases - <a href="https://mailu.io/master/webadministration.html?highlight=sql%20like#aliases">documentation</a>)'))
     destination = DestinationField(_('Destination'))
     comment = fields.StringField(_('Comment'))
     submit = fields.SubmitField(_('Save'))


core/admin/mailu/ui/forms.py
```diff
@@ -138,7 +138,7 @@ class TokenForm(flask_wtf.FlaskForm):
 class AliasForm(flask_wtf.FlaskForm):
     localpart = fields.StringField(_('Alias'), [validators.DataRequired(), validators.Regexp(LOCALPART_REGEX)])
     wildcard = fields.BooleanField(
-        _('Use SQL LIKE Syntax (e.g. for catch-all aliases)'))
+        _('Use SQL LIKE Syntax (e.g. for catch-all aliases - <a href="https://mailu.io/master/webadministration.html?highlight=sql%20like#aliases">documentation</a>)'))
```
Diman0, line 141: Don't hard-link to the master documentation. The user should be on the current stable release. The documentation on Master can be different than the current stable version. Technically the link can also be different on master then the current version. It is the active development branch after all.

The link should use the same version as the current image. E.g. https://mailu.io/1.9/webadministration.html?highlight=sql%20like#aliases for mailu 1.9.

The only problem is that I do not know how to derive the version. I cannot find any information that this is stored as part of the Admin image. I suppose we could pass the image tag via an env variable (image tag is always the version)? Then you could use this env variable in the URL.

Another option is linking to https://mailu.io. This link always opens the documentation of the current stable release. Mention in the link description something like:  "See 'Web Administration Interface -> Mail domains -> Aliases' in the help."

Diman0 (changes requested):
See in-line comment.

core/admin/mailu/ui/forms.py
benyanke: I think getting the version number would absolutely be the cleanest, but I'm also not sure how to best do that. Let me play with it a bit.

benyanke: I didn't see any obvious ways to do so. I'm leaning back toward what was discussed in the original issue, and simply mentioning that % is a wildcard. would you be open to that?

Diman0: I'm open for that. Maybe you could make the hint multi-line?

```
Use SQL LIKE Syntax (e.g. for catch-all aliases. Syntax:
  % - The percent sign represents zero, one, or multiple characters
  _ - The underscore represents a single character)
```

Diman0: I did not notice it before, but this line change breaks the translations for all 17 other languages the mailu web interface has been translated to. That is not acceptable. I'm not very experienced with using pybabel. I do not know if this line can be constructed in such a way that the translations are retained.

Maybe it works if you add the url on the route py file?
That would be:
https://github.com/Mailu/Mailu/blob/20f00a36999efbe6b9d5c1d9279eec7d6c359eb6/core/admin/mailu/ui/views/aliases.py#L15
https://github.com/Mailu/Mailu/blob/20f00a36999efbe6b9d5c1d9279eec7d6c359eb6/core/admin/mailu/ui/views/aliases.py#L39

Of course this change can only occur if it does not break all the translations. You can test the translations by clicking on the language selector in the top-right of the mailu web administration page.

Diman0 (changes requested):

ghostwheel42:
Hi. "_" is the translation function. It's necessary to leave the old arg intact to not break the translations. So this would be fine:
```
_('Use SQL LIKE Syntax (e.g. for catch-all aliases)') + ' - <a href="https://mailu.io/master/webadministration.html?highlight=sql%20like#aliases">documentation</a>'
```

Diman0:
@benyanke  are you planning to continue this PR?

benyanke:
Probably not at this time, given the greater than expected complexity.

PR #2049, opened 2021-11-10
fix for issue 1223 (fetchmail persistence idfile)

### LINKED ISSUE #1223: Fetchmail: /var/lib/fetchmail needs persistence
According [fetchmail documentation](https://www.fetchmail.info/fetchmail-man.html#12), an `.idfile` is used to keep track of previously downloaded messages. Shouldn't that file persistent over container restarts?

I'm not a Fetchmail user, perhaps somebody can shine a light on how this currently works?

cc: @Nebukadneza, @hoellen, @kaiyou

Diman0:
## What type of PR?

bug-fix

## What does this PR do?
It introduces a new data folder (/mailu/fetchmail) that will hold the idfile. The file that is used by fetchmail to keep track of what messages where retrieved. Recreating the fetchmail container does not result in all messages being retrieved again. It also configurs fetchmail to actually create this file (--uidl).

It changes fetchmail to run as root. For now this is required, because the mailu data folder (/mailu) is owned by root. In the future we must change all images at the same time, to run without root and use a mailu folder that is not owned by root. That is out of scope for this PR. 

### Related issue(s)
- closes #1223

## Prerequisites

### DIFF
--- a/optional/fetchmail/fetchmail.py
+++ b/optional/fetchmail/fetchmail.py
@@ -13,6 +13,7 @@ import traceback
 
 FETCHMAIL = """
 fetchmail -N \
+    --idfile /data/.fetchids \
     --sslcertck --sslcertpath /etc/ssl/certs \
     -f {}
 """
--- a/optional/fetchmail/fetchmail.py
+++ b/optional/fetchmail/fetchmail.py
@@ -13,7 +13,7 @@ import traceback
 
 FETCHMAIL = """
 fetchmail -N \
-    --idfile /data/.fetchids \
+    --idfile /data/.fetchids --uidl \
     --sslcertck --sslcertpath /etc/ssl/certs \
     -f {}
 """
--- a/optional/fetchmail/fetchmail.py
+++ b/optional/fetchmail/fetchmail.py
@@ -13,7 +13,7 @@ import traceback
 
 FETCHMAIL = """
 fetchmail -N \
-    --idfile /data/.fetchids --uidl \
+    --idfile /data/fetchids --uidl \
     --sslcertck --sslcertpath /etc/ssl/certs \
     -f {}
 """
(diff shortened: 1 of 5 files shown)


Diman0:
@micw This PR introduces a bind mount (/$(data/fetchmail:/data) for fetchmail. Assuming this PR is merged, this means you must reflect the same changes in the helm-charts.

nextgens:
Why ```{{ root }}/fetchmail``` and not ```{{ root }}/data/fetchmail``` (à la #1224) ?

Diman0:
In the current structure we have unique folders for each type of component. 
But using `{{ root }}/data/fetchmail` is also fine with me. It might be a bit confusing, we will have
admin: `{{ root }}/data/*.*`
fetchmail ``{{ root }}/data/fetchmail/*.*``
However that is what that RFC is all about. Move all files according to a well thought out structure.

Diman0:
bors try

Diman0:
If you want to test the PR, then don't forget to add the bind mount to your docker-compose.yml. For example:
```
  fetchmail:
    image: ${DOCKER_ORG:-mailu}/${DOCKER_PREFIX:-}fetchmail:${MAILU_VERSION:-master}
    restart: always
    env_file: mailu.env
    volumes:
      - /mailu/data/fetchmail:/data
    depends_on:
      - resolver
    dns:
      - 192.168.203.254
```

optional/fetchmail/fetchmail.py
```diff
@@ -13,6 +13,7 @@
 
 FETCHMAIL = """
 fetchmail -N \
+    --idfile /data/.fetchids --uidl \
```
nextgens, line 16: I wouldn't use a hidden file here... why prefixing it with a dot? Some users may remove the folder thinking it's empty

Diman0: That is the suggested filename from the fetchmail documentation. I just followed their default file name. It indeed makes more sense to make the file not hidden (removing the leading dot).

nextgens (approved):
LGTM

PR #2055, opened 2021-11-16
Own redis Dockerfile

ghostwheel42:
## What type of PR?

minor change

## What does this PR do?

Introduces own redis Dockerfile to add health check and also set TZ
`redis-server` can be started via mailu entrypoint later.

### Related issue(s)

adds timezone like in #2034

## Prerequisites


nextgens:
Forgot to commit the entrypoint? 

I don't see how maintaining our own beats using upstream's. If upstream's lacks TZ and health-checks maybe it's there that it needs to be fixed?

ghostwheel42:
> Forgot to commit the entrypoint?

No, I wanted to add it later if necessary (it could for example change redis config via templates if we wish to)

> I don't see how maintaining our own beats using upstream's. If upstream's lacks TZ and health-checks maybe it's there that it needs to be fixed?

I don't think upstream would merge the TZ-stuff, as we all agree using anything other than UTC is bad.
Upstream maybe would add the health check.

I wouldn't say this Dockerfile really needs to be maintained as it only enhances the version maintained upstream.

We can flag this as "Work in progress" - I think we'll need it when switching to services not running as root in the future.

PR #2066, opened 2021-11-24
Upgrade rspamd

nextgens:
## What type of PR?

enhancement

## What does this PR do?

Upgrade rspamd to a version that hopefully won't segfault on arm

### Related issue(s)
- #1200

### DIFF
--- a/core/rspamd/Dockerfile
+++ b/core/rspamd/Dockerfile
@@ -1,4 +1,4 @@
-ARG DISTRO=alpine:3.14.2
+ARG DISTRO=alpine:3.15
 FROM $DISTRO
 
 ENV TZ Etc/UTC
--- a/core/rspamd/conf/headers_group.conf
+++ b/core/rspamd/conf/headers_group.conf
@@ -1,0 +1,7 @@
+symbols = {
+    "RCVD_NO_TLS_LAST" {
+	# see https://github.com/Mailu/Mailu/issues/1705
+        weight = 0.0;
+        description = "Last hop did not use encrypted transports";
+    }
+}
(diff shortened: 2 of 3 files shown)


Diman0:
@nextgens 
When testing receiving emails I see that RCVD_NO_TLS_LAST still has score 0.1. 
Can you check if you observe the same?

No matter what I try in settings.conf, it appears to be ignored :(
I have tried so far
```
disable_symbol {
	priority = high;
	apply {
		RCVD_NO_TLS_LAST = 0;
	}
}
``` 
and
```
disable_symbol {
	priority = high;
	rcpt = "*" ;
	apply {
		RCVD_NO_TLS_LAST = 0;
	}
}
``` 

I must confess I don't remember checking this thoroughly when you applied PR #2033. I assumed you tested this yourself and it looked sound to me.

Diman0 (changes requested):
Ok. I found the correct setting and config file. Pff. I always have the feeling they make the rspamd documentation as incomprehensible as possible. They do everything in their power to make you not understand the documentation.

Remove the file /conf/settings.conf
Create the file /conf/headers_group.conf with contents
```
symbols = {
    "RCVD_NO_TLS_LAST" {
        weight = 0.0;
        description = "Last hop did not use encrypted transports";
    }
}
```
With this I see that the symbol RCVD_NO_TLS_LAST now has score 0

nextgens:
Yeah, it's everything but clear :/

Diman0 (approved):
Looks good to me.

kirbylink:
Will this changes also merged into the 1.8 branch?

nextgens:
@kirbylink no. ARM isn't supported so there is no reason to backport it.

1.9 will be released soon and while we do expect ARM to work it won't be supported on that version either.

kirbylink:
@nextgens 
Thanks for the update/information. :)
Only one last question: Is there a release plan for 1.9?

PR #2067, opened 2021-11-25
Update messaging on imap to include webmail

benyanke:
Update messaging on imap to make clear that it also controls access to webmail as well.

## What type of PR?

Enhancement

## What does this PR do?

Improves the admin interface to be more self-documenting.

An admin disabling/not-enabling IMAP on an account might not immediately realize that this setting also disables webmail functionality on this account. This PR improves the admin experience in this area.

### Related issue(s)
none

## Prerequisites
none

### DIFF
--- a/core/admin/mailu/ui/forms.py
+++ b/core/admin/mailu/ui/forms.py
@@ -80,7 +80,7 @@ class UserForm(flask_wtf.FlaskForm):
     pw = fields.PasswordField(_('Password'))
     pw2 = fields.PasswordField(_('Confirm password'), [validators.EqualTo('pw')])
     quota_bytes = fields_.IntegerSliderField(_('Quota'), default=10**9)
-    enable_imap = fields.BooleanField(_('Allow IMAP access'), default=True)
+    enable_imap = fields.BooleanField(_('Allow IMAP/Webmail access'), default=True)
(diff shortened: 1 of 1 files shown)


nextgens (approved):
LGTM

Diman0 (changes requested):
Is this really that useful?
The mailu interface is fully translated in 18 different languages. That line change **breaks ALL translations**, but English. For a minor change.  This is not a user setting, but a setting that is changed by an administrator who is supposed to know what he is doing. A webmail uses imap. If you disable imap, then a webmail client cannot connect via imap. 

I'm fine with the documentation being updated. That would be the file https://github.com/Mailu/Mailu/blob/master/docs/webadministration.rst and the line:
`Allow IMAP access. When ticked, allows email retrieval via the IMAP protocol.`
This line could be enhanced that disabling IMAP disables webmail access.

I'm also fine with updating the line in the admin gui as long it does not break the translations for all other languages or if you update the translations.

The long term solution might be looking into a way to have webmail also work if a user has imap disabled (but the user itself is still enabled of course). I'm not sure if this is possible, but we could investigate this.

benyanke:
I'll probably close this for the time being, as I don't have a good way to work with the various translations. Thank you for the additional information!

PR #2072, opened 2021-12-06
use dovecot-fts-xapian from alpine package

willofr:
## What type of PR?

enhancement

## What does this PR do?
use dovecot-fts-xapian from alpine packages repository (newer) instead of compiling an older version from source
see https://pkgs.alpinelinux.org/package/edge/community/x86/dovecot-fts-xapian

### Related issue(s)
No

## Prerequisites

### DIFF
--- a/core/dovecot/Dockerfile
+++ b/core/dovecot/Dockerfile
@@ -1,14 +1,4 @@
-ARG DISTRO=alpine:3.14.2
-FROM $DISTRO as builder
-WORKDIR /tmp
-RUN apk add git build-base automake autoconf libtool dovecot-dev xapian-core-dev icu-dev
-RUN git clone https://github.com/grosjo/fts-xapian.git \
-  && cd fts-xapian \
-  && git checkout 1.2.7 \
-  && autoreconf -vi \
-  && PANDOC=false ./configure --with-dovecot=/usr/lib/dovecot \
-  && make \
-  && make install
+ARG DISTRO=alpine:3.14
 
 FROM $DISTRO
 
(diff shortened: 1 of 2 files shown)


core/dovecot/Dockerfile
```diff
@@ -1,14 +1,4 @@
...
-WORKDIR /tmp
-RUN apk add git build-base automake autoconf libtool dovecot-dev xapian-core-dev icu-dev
-RUN git clone https://github.com/grosjo/fts-xapian.git \
-  && cd fts-xapian \
-  && git checkout 1.2.7 \
-  && autoreconf -vi \
-  && PANDOC=false ./configure --with-dovecot=/usr/lib/dovecot \
-  && make \
-  && make install
+ARG DISTRO=alpine:3.14
```
Diman0, line 1: Why did you change this?
We use pinned versions. Please change this back to 3.14.2.

Diman0 (changes requested):
See inline comment. 
You must also add a [changelog entry](https://mailu.io/master/contributors/workflow.html#changelog). As the file name you can use the PR number.

core/dovecot/Dockerfile
willofr: I think it's more interesting to follow the major version (3.14) than a minor version, so the project follows the evolutions of the underlying distribution (bug fixes, no major evolution). The latest 3.14 version of alpine is 3.14.3.
I rollback and propose to open a discussion on the subject ?

willofr:
Revert back to alpine 3.14.2 as requested
Add changelog entry

Diman0:
It looks good to me :+1: 
I will do some testing later this week and will then approve it.

nextgens (approved):
LGTM; I think it's working here ... but I'm not 100% sure it's a server search that's started.

Diman0 (approved):

Issue #2073, opened 2021-12-08
DB_PORT environment setting is ignored


mojerro:
### Environment
 - [x] docker-compose
 - [ ] kubernetes
 - [ ] docker swarm

### Versions
mailu 1.8

## Description
As a reason that the default database port is not secure, and I hope to install mailu with other postgres port instead of 5432. But in the dockerfile and env file, I cannot find a place to change it , can someone help or this will be a new feature?

Diman0:
In Mailu it is not possible to use a custom port. I will mark this as an enhancement request.

mojerro:
I just found I can add `DB_PORT=xxxx` in `mailu.env` to assign a custom port, and it seems working!  This env is found in this [documentation](https://github.com/Mailu/Mailu/blob/20f00a36999efbe6b9d5c1d9279eec7d6c359eb6/docs/database.rst)

Diman0:
Ah yes. You are correct https://mailu.io/master/configuration.html#database-settings
However I think it is not used. See:
https://github.com/Mailu/Mailu/blob/1675399047be7fe6813ceb7db418be9010124400/core/admin/mailu/configuration.py#L97

Can you double check if the setting works? I cannot find any reference to the DB_PORT env variable in the project.

mojerro:
Well, sorry about what I said above, the real change is directly add port after the link. Such as:
`portgres.domain:1234`

mojerro:
I did many changes and one of them get successful. I also get another mistake from the database which puzzled me ...

Diman0:
Looking at the SQLAlchemy]

Diman0:
Note: After changing the settings you must **recreate** the containers. Restarting is not sufficient.

mojerro:
thanks a lot

> 

That's pretty clear, and my mailu server is now working well! thx!

PR #2074, opened 2021-12-08
The DB_PORT and ROUNDCUBE_DB_PORT env vars were not used

Diman0:
## What type of PR?

Bug fix 

## What does this PR do?
The DB_PORT and ROUNDCUBE_DB_PORT env vars can be used to configure a custom port for the database connection. These environment variables were documented, but not actually used. 

This fix addresses that. The syntax should be correct. I did not test it yet. I will do that later this week.

### Related issue(s)
- See #2073

## Prerequisites

- [ n/a ] In case of feature or enhancement: documentation updated accordingly

### DIFF
--- a/core/admin/mailu/configuration.py
+++ b/core/admin/mailu/configuration.py
@@ -21,6 +21,7 @@ DEFAULT_CONFIG = {
     'DB_USER': 'mailu',
     'DB_PW': None,
     'DB_HOST': 'database',
+    'DB_PORT': None,
     'DB_NAME': 'mailu',
     'SQLITE_DATABASE_FILE':'data/main.db',
     'SQLALCHEMY_DATABASE_URI': 'sqlite:////data/main.db',
@@ -144,6 +145,8 @@ class ConfigManager:
 
         # automatically set the sqlalchemy string
         if self.config['DB_FLAVOR']:
+            if self.config['DB_PORT']:
+                self.config['DB_HOST'] = f'{self.config["DB_HOST"]}:{str(self.config["DB_PORT"])}'
             template = self.DB_TEMPLATES[self.config['DB_FLAVOR']]
             self.config['SQLALCHEMY_DATABASE_URI'] = template.format(**self.config)
 
--- a/core/admin/mailu/configuration.py
+++ b/core/admin/mailu/configuration.py
@@ -145,7 +145,7 @@ class ConfigManager:
 
         # automatically set the sqlalchemy string
         if self.config['DB_FLAVOR']:
-            if self.config['DB_PORT']:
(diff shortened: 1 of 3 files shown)


Diman0:
If the changes in webmails/roundcube/start.py can be done much more elegantly, feel free to push this change to the PR branch.
I just wanted to fix this quickly.

Diman0:
bors try

t0mtaylor:
Usually you can add the port to the host to sql db - not sure if it needs to be separate?

I have it working on a different port without this PR just by setting it to the end of the host within the env file

```
DB_HOST=sql-server-address:12345
ROUNDCUBE_DB_HOST=sql-server-address:12345
```

Diman0:
I will remove the *_PORT environment variables and will document that you can specify the port via the DB_HOST/ROUNDCUBE_DB_HOST using colon as a separator
`HOSTNAME:PORT`

core/admin/mailu/configuration.py
```diff
@@ -144,6 +145,8 @@ def init_app(self, app):
 
         # automatically set the sqlalchemy string
         if self.config['DB_FLAVOR']:
+            if self.config['DB_PORT'] and self.config['DB_FLAVOR'] != 'sqlite':
+                self.config['DB_HOST'] = f'{self.config["DB_HOST"]}:{str(self.config["DB_PORT"])}'
```
ghostwheel42, line 149: you don't need to use `str()` inside the `{}`
This would be even shorter:
`self.config['DB_HOST'] += f':{self.config["DB_PORT"]}'`

webmails/roundcube/start.py
```diff
@@ -11,22 +11,41 @@
 os.environ["MAX_FILESIZE"] = str(int(int(os.environ.get("MESSAGE_SIZE_LIMIT")) * 0.66 / 1048576))
 
 db_flavor = os.environ.get("ROUNDCUBE_DB_FLAVOR", "sqlite")
+db_port = os.environ.get("ROUNDCUBE_DB_PORT")
```
ghostwheel42, line 14: What about this?
```
db_host = os.environ.get("ROUNDCUBE_DB_HOST", "database")
if db_port := os.environ.get("ROUNDCUBE_DB_PORT"):
  db_host += f':{db_port}'
```
and instead of your changes replace `os.environ.get("ROUNDCUBE_DB_HOST", "database")` with `db_host` in the original code.

ghostwheel42 (changes requested):

Diman0:
> If the changes in webmails/roundcube/start.py can be done much more elegantly, feel free to push this change to the PR branch. I just wanted to fix this quickly.

I'm closing this PR. I will submit a new one with the above changes.

PR #2076, opened 2021-12-08
fix the default for DEFER_ON_TLS_ERROR

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

The default wasn't set anywhere

### DIFF
--- a/core/postfix/start.py
+++ b/core/postfix/start.py
@@ -35,6 +35,7 @@ def is_valid_postconf_line(line):
             and not line == ''
 
 # Actual startup script
+os.environ['DEFER_ON_TLS_ERROR'] = os.environ['DEFER_ON_TLS_ERROR'] if 'DEFER_ON_TLS_ERROR' in os.environ else 'True'
 os.environ["FRONT_ADDRESS"] = system.get_host_address_from_environment("FRONT", "front")
 os.environ["ADMIN_ADDRESS"] = system.get_host_address_from_environment("ADMIN", "admin")
 os.environ["ANTISPAM_MILTER_ADDRESS"] = system.get_host_address_from_environment("ANTISPAM_MILTER", "antispam:11332")


Diman0:
@nextgens Am I doing something wrong?
The env variable appears not to be available. But that also applies to all other env variables set in start.py. E.g. try:
docker-compose exec smtp printenv | grep FRONT_ADDRESS
docker-compose exec smtp printenv | grep DEFER
docker-compose exec smtp printenv | grep OUTCLEAN
They all result in no output.

I also tried running bash on the postfix container, start python, then run the os.environ.get("DEFER_ON_TLS_ERROR"). This also result in no output.

nextgens:
You're not: when it's not defined it should be set to True

We could do os.environ.get("DEFER_ON_TLS_ERROR", True) if you think it's more clear

nextgens:
> They all result in no output.

cat /proc/1/environ

Diman0:
Ah wait https://docs.python.org/3/library/os.html?highlight=os%20environ#os.putenv

--------------

Set the environment variable named key to the string value. Such changes to the environment affect subprocesses started with os.system(), popen() or fork() and execv().

--------------

So all those set ENV variables are **ONLY** available for the subprocesses started from the python script. 
Are you familiar with any bash/linux commands we could use to list the env variables available to the postfix  subprocess?

The code itself is sound. When socrates is replaced we should maybe also change that upon startup all derived env variables are dumped to standard out?

Diman0:
Hmm cat /proc/8/environ gives me an access denied.
```
dimitri@mailutest:/mailu$ docker-compose exec smtp cat /etc/postfix/main.cf | grep dane
smtp_tls_dane_insecure_mx_policy = dane
```
This however confirms that the ENV variable is set. This setting is only set to 'dane' if DEFER_ON_TLS_ERROR is set to true.

I will accept the PR, but in the future we should really look into making the derived environment variables available.
Possibly we could write all derived env variables to a file, or perhaps log it to standard out on startup. But that is something to discuss in a different location.

Diman0 (approved):
LGTM

PR #2083, opened 2021-12-14
Fix Webmail token check. Fix Auth-Port for Webmail. #2079

### LINKED ISSUE #2079: [master] Webmail token check does not work if WEBMAIL_ADDRESS is set to a hostname
In master with the new token stuff, webmail is not working when WEBMAIL_ADDRESS is set to a hostname.

The issue is in https://github.com/Mailu/Mailu/blob/e3396b83ce0496231515cbefa88fca61f8355410/core/admin/mailu/internal/nginx.py#L35 where the hostname is compared to the IP adress.

See also https://github.com/Mailu/Mailu/blob/e52a3de1b04fa1f8968e0f596a04f2d17205ef9c/core/admin/mailu/configuration.py#L118

When HOST_WEBMAIL is set, it is statically resolved and it works (but with all drawbacks of statically resolved hosts, we don't want this anymore). When WEBMAIL_ADDRESS is set, it's used as it is and not resolved.

Possible solutions:
* check against $SUBNET
* set a special mark on temp tokens so that the check is not required
* set a special auth header on FRONT so that it's clear that the request comes from webmail (best solution IMO)

### LINKED ISSUE #2081: [master] Typo in front port
In https://github.com/Mailu/Mailu/blob/dbbfa44461c3b8a537e9fe6b56f4690417d3c8a2/core/nginx/conf/nginx.conf#L280 port number must be 10143, not 10043. Just cosmetical, will have no other effect.

Diman0:
## What type of PR?

Bug fix

## What does this PR do?
Fixes issues #2079 and #2081. 

### Related issue(s)
- closes #2079 
- closes #2081 

## Prerequisites

- [x] n/a In case of feature or enhancement: documentation updated accordingly

### DIFF
--- a/core/admin/mailu/internal/nginx.py
+++ b/core/admin/mailu/internal/nginx.py
@@ -27,12 +27,12 @@ STATUSES = {
     }),
 }
 
-def check_credentials(user, password, ip, protocol=None):
+def check_credentials(user, password, ip, protocol=None, auth_port=None):
     if not user or not user.enabled or (protocol == "imap" and not user.enable_imap) or (protocol == "pop3" and not user.enable_pop):
         return False
     is_ok = False
     # webmails
-    if len(password) == 64 and ip == app.config['WEBMAIL_ADDRESS']:
+    if len(password) == 64 and auth_port == '10143':
         if user.verify_temp_token(password):
             is_ok = True
     # All tokens are 32 characters hex lowercase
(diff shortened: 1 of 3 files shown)


core/admin/mailu/internal/nginx.py
```diff
@@ -27,12 +27,12 @@
...
 }
 
-def check_credentials(user, password, ip, protocol=None):
+def check_credentials(user, password, ip, protocol=None, auth_port=None):
     if not user or not user.enabled or (protocol == "imap" and not user.enable_imap) or (protocol == "pop3" and not user.enable_pop):
         return False
     is_ok = False
     # webmails
-    if len(password) == 64 and ip == app.config['WEBMAIL_ADDRESS']:
+    if len(password) == 64 and auth_port == '10143':
```
nextgens, line 35: SMTP is authenticated too for the webmail (port 10025 ; both should accept tokens)

nextgens (changes requested):

Diman0:
I checked I can login webmail, and that I can send email via webmail.

PR #2084, opened 2021-12-14
Fix #2078 (login to webmail did not work when WEB_WEBMAIL=/ was set)

### LINKED ISSUE #2078: [master] login into webmail does not work with WEB_WEBMAIL=/
On the master version the SSO login does not work when `WEB_WEBMAIL=/` is set using docker-compose.
After clicking "Sign in Webmail", you receive a HTTP 404 page.

I fixed it by:
https://github.com/Mailu/Mailu/blob/e52a3de1b04fa1f8968e0f596a04f2d17205ef9c/core/nginx/conf/nginx.conf#L138
Add a trailing slash here. Not sure if this affects `WEB_WEBMAIL=/webmail`. Need to test this.

https://github.com/Mailu/Mailu/blob/dbbfa44461c3b8a537e9fe6b56f4690417d3c8a2/core/nginx/conf/nginx.conf#L168
Has a double slash if `WEB_WEBMAIL=/` is set. I removed this double slash in the running container.

With both modifications the login to webmail works again.

Diman0:
## What type of PR?

bug-fix

## What does this PR do?
It fixes #2078. Login from SSO page to webmail did not work if WEB_WEBMAIL=/ was set in mailu.env.

I tested that it works with
- WEB_WEBMAIL=/webmail
- WEB_WEBMAIL=/

### Related issue(s)
- closes #2078 

## Prerequisites

- [x] n/a In case of feature or enhancement: documentation updated accordingly

### DIFF
--- a/core/nginx/conf/nginx.conf
+++ b/core/nginx/conf/nginx.conf
@@ -135,7 +135,7 @@ http {
 
       # Actual logic
       {% if ADMIN == 'true' or WEBMAIL != 'none' %}
-      location ~ ^/(sso|static) {
+      location ~ ^/(sso|static)/ {
         include /etc/nginx/proxy.conf;
         proxy_pass http://$admin;
       }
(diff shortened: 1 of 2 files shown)


nextgens (dismissed):
LGTM

PR #2087, opened 2021-12-16
[Feature] Added: Dependabot

GoliathLabs:
## What type of PR?
This is a Feature

## What does this PR do?
This PR adds dependabot. Dependabot will take care of package updates (e.g. python pip, npm (JavaScript), Dockerfiles (Docker) and the used GH-Actions

## Prerequisites

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,20 @@
+# Docs: <https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/customizing-dependency-updates>
+
+version: 2
+
+updates:
+  - package-ecosystem: pip
+    directory: /
+    schedule: {interval: monthly}
+
+  - package-ecosystem: github-actions
+    directory: /
+    schedule: {interval: monthly}
+
+  - package-ecosystem: docker
+    directory: /
+    schedule: {interval: monthly}
+
+  - package-ecosystem: npm
+    directory: /website
+    schedule: {interval: monthly}


GoliathLabs:
Should I add a changelog entry or something in the documentation?

GoliathLabs:
@Diman0 @ghostwheel42 Any opinions regarding this?

GoliathLabs:
bors try

Diman0:
It feels strange to me you submit this pull request without any discussion via an issue beforehand.  
Next time I recommend to create an issue before you create a PR so we can discuss it. 

Dependabot does not fit with our workflow and will most likely introduce more issues than it resolves. 
Our workflow is to carefully upgrade all dependencies before a new Mailu release and resolve any issues introduced by the updated dependencies. We do this on our own pace and will not have this determined by a bot. This ensures we only have to update all dependencies once before a new release.

The reason for this approach is that updating dependencies always introduces new issues. Either new bugs in the updated dependency or breaking changes in the updated dependency. It is always a painful manual process. That is why we do all this updating near the release of a new Mailu version. This way we only have to spend time once for testing all new dependencies. We only have to do this once for each release. With dependabot we would waste a lot of time testing in between Mailu releases, because dependabot will often submit new update PRs. We are a project with a very low amount of active contributors. We do not have the resources for this.

However we will discuss this at the next meeting. If we decide to use dependabot we will do it in a new pull request.

Issue #1869, opened 2021-07-18
Weblate instance is down


yarons:
I tried accessing the Weblate instance and potentially add another language but it looks down.

Diman0:
I didnt know what you were referring to earlier. In the future it helps to include a link for what you mean. I thought you meant setup.mailu.io.
You are referring to https://translate.tedomum.net/projects/mailu/admin/

@kaiyou are you hosting this site?

Diman0:
I got in touch with kaiyou. The entity that hosted the weblate instance decided to stop hosting it without informing us. There is weblate instance and the best way to do translations is to use a tool such as poedit and do the translations locally.

yarons:
@Diman0 Sorry, I had no idea there was another Weblate :)

Diman0:
I see I stated the opposite what I meant. I edited my previous comment. There is **no** weblate instance available anymore. Unfortunately the entity who made this available decided to stop hosting. We must reflect this in the documentation.

yarons:
Ohhhh now I get it, how about joining the main Hosted Weblate?

https://hosted.weblate.org

Diman0:
That might be interesting. I have no experience with weblate. I think we will first update the documenation that there is no weblate instance available informing the reader that he/she can use a tool such as poedit. 

At a later moment we can check using https://hosted.weblate.org.

Erriez:
Note: Documentation page https://mailu.io/master/contributors/localization.html refers to the dead link:

> The project : https://translate.tedomum.net/projects/mailu/admin/

WebSpider:
@nextgens it would seem this can be closed, docs have been updated in #2107

PR #2107, opened 2021-12-24
Remove weblate from documentation

Diman0:
## What type of PR?

documentation

## What does this PR do?
See #1869. The weblate instance is not available anymore. Therefore this not available weblate instance should not be mentioned in the documentation anymore.

This PR removes it from the documentation

### Related issue(s)
- #1869

## Prerequisites


nextgens (approved):
LGTM

PR #2116, opened 2021-12-30
fix 2114: redirect old path

### LINKED ISSUE #2114: Error 404 not found when opening admin after upgrade 1.8 to master
## Before you open your issue

### Environment
 - [X] docker-compose

### Versions

Before upgrade: Docker 1.8 images.
After upgrade: Docker master images (pulled 30 December 2021).

## Description

**Mailu 1.8** image redirects `/admin` to `/admin/ui`.

**Mailu master** image no longer redirects `/admin/ui` as the `ui` part in the URL has been removed according to [Tomcat 1929.enhacement](https://github.com/Mailu/Mailu/blob/master/towncrier/newsfragments/1929.enhancement):

> Removed the /admin/ prefix to reduce complexity of routing with Mailu. Admin is accessible directly via /admin instead of /admin/ui

After the upgrade from `1.8` to `master` and visiting the admin page, the browser still uses the cached URL `/admin/ui` and results in 404 not found.

## Replication Steps

1. Create 1.8 production environment on AMD64 platform using `mailu 1.8 Docker images`.
2. Make sure the Admin page works.
3. Remove docker containers (`docker-compose down`).
4. Recreate **all** containers at the same time using `mailu master Docker images`.
5. Open root mail domain. The browser uses the cached URL `admin/ui` and shows Error 404 not found.

Note:  Tested with `TLS_FLAVOR=letsencrypt`, admin and roundcube and Firefox.

## Expected behaviour

Backwards compatibility after Mailu 1.8 upgrade without the need of removing browser caches.

## Front log

```
front_1     | <IP> - - [30/Dec/2021:10:14:35 +0000] "GET /admin/ui/ HTTP/2.0" 404 198 "https://mail.mydomain.nl/sso/login" "Mozilla/5.0 (X11; Linux x86_64; rv:95.0) Gecko/20100101 Firefox/95.0"
```

## Bugfix

Proposal is to redirect `/admin/ui` always to `/admin` to prevent browser caching problems after the upgrade.

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Old paths may still be cached in browsers, it's easy enough to redirect them

### Related issue(s)
- close #2114

### DIFF
--- a/core/admin/mailu/ui/views/base.py
+++ b/core/admin/mailu/ui/views/base.py
@@ -11,6 +11,10 @@ import flask_login
 def index():
     return flask.redirect(flask.url_for('.user_settings'))
 
+@ui.route('/ui/')
+def redirect_old_path():
+    return flask.redirect(flask.url_for('.index'), code=301)
+
 @ui.route('/announcement', methods=['GET', 'POST'])
 @access.global_admin
 def announcement():


core/admin/mailu/ui/views/base.py
```diff
@@ -11,6 +11,10 @@
 def index():
     return flask.redirect(flask.url_for('.user_settings'))
 
+@ui.route('/ui/')
```
ghostwheel42, line 14: What about removing the trailing slash?

ghostwheel42 (changes requested):
LOoks good. 
Can the healing slash be removed?

core/admin/mailu/ui/views/base.py
nextgens: We can't remove it otherwise it would mean "file" and wouldn't act as a wildcard
See the section about trailing slashes on https://flask.palletsprojects.com/en/2.0.x/api/#url-route-registrations

ghostwheel42 (approved):
Lgtm

PR #2144, opened 2022-01-05
Enable unbound by default, warn if the DNS resolver doesn't work

### LINKED ISSUE #2135: Cannot send email after upgrade to 1.9 `non DNSSEC destination`
## Before you open your issue

### Environment
 - [x] docker-compose
 - [ ] kubernetes
 - [ ] docker swarm

### Versions
1.9

## Description
Thanks for all the work! 1.9 seems like an amazing release overall ❤️ 

I upgraded to 1.9, generated the new docker-compose file with the tool.
Copied the relevant env variables over and started everything.
I can still receive mails, and send internal one (from and to my own domain)
I cannot send mails to external providers.

after restoring from backup everything works in 1.8

```
Final-Recipient: rfc822; ******
Original-Recipient: rfc822;******
Action: delayed
Status: 4.7.5
Diagnostic-Code: X-Postfix; non DNSSEC destination
Will-Retry-Until: Fri,  7 Jan 2022 18:38:34 +0000 (UTC)
Return-Path: <*****>
From: Cloud <*******>
DKIM-Signature: v=1; a=rsa-sha256; c=relaxed/relaxed; d=nicco.io; s=dkim;
	t=1641148715;
	h=from:from:reply-to:subject:subject:date:date:message-id:message-id:
	 to:to:cc:mime-version:mime-version:content-type:content-type:
	 in-reply-to:in-reply-to:references:references;
	bh=Q0j3Ph9l8nLeBMIzdq6aOtNcsZOyiD8WuiQSGgc2oeY=;
	b=nF+9OZeRSSEDZ995inbj/6nDYgbTDMhPGc63Gab3FN1A984PxY7rDoHIhkO5nvh2wzNZG6
	5jIAmfCyYfHriJawNrtcKeUA4wBO5YgYPWag6CbGmWQ8sGSIty2fjIO0W4dWfWy+OjsksX
	sJ2TK8uft8Ax9F9QmQKMuZHBP3Myh/I=
Content-Type: multipart/mixed;
	boundary="Apple-Mail=_DD0DBE9A-59E8-455E-B4DD-1317DF1D24F9"
Mime-Version: 1.0
Subject: =*****
Date: Sun, 2 Jan 2022 19:38:33 +0100
References: <****>
<****>
<****>
<****>
To: ***
In-Reply-To: ****
Message-Id: ****
X-Spam-Level: *
X-Spamd-Bar: +
Authentication-Results: *****;
	auth=pass smtp.mailfrom=*****
```

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Enable unbound by default, warn if the DNS resolver doesn't work

### Related issue(s)
- close #2135

## Prerequisites

### DIFF
--- a/core/admin/start.py
+++ b/core/admin/start.py
@@ -18,6 +18,36 @@ if account is not None and domain is not None and password is not None:
     log.info("Creating initial admin accout %s@%s with mode %s",account,domain,mode)
     os.system("flask mailu admin %s %s '%s' --mode %s" % (account, domain, password, mode))
 
+def test_DNS():
+    import dns.resolver
+    import dns.exception
+    import dns.flags
+    import dns.rdtypes
+    import dns.rdatatype
+    import dns.rdataclass
+    import time
+    # DNS stub configured to do DNSSEC enabled queries
+    resolver = dns.resolver.Resolver()
+    resolver.use_edns(0, 0, 1232)
+    resolver.flags = dns.flags.AD | dns.flags.RD
+    nameservers = resolver.nameservers
+    for ns in nameservers:
+        resolver.nameservers=[ns]
+        error = True
+        while error:
+            try:
+                result = resolver.query('example.org', dns.rdatatype.A, dns.rdataclass.IN, lifetime=10)
+                if not result.response.flags & dns.flags.AD:
+                    log.critical("Your DNS resolver at %s isn't doing DNSSEC validation; Please install unbound.", ns)
+                else:
+                    error = False
+                    continue
+            except Exception as e:
+                log.critical("Your DNS resolver at %s is not working (%s). Please install unbound.", ns, e);
+            time.sleep(5)
+
+test_DNS()
+
 start_command="".join([
     "gunicorn --threads ", str(os.cpu_count()),
     " -b :80 ",
--- a/core/admin/start.py
+++ b/core/admin/start.py
@@ -33,17 +33,15 @@ def test_DNS():
     nameservers = resolver.nameservers
     for ns in nameservers:
         resolver.nameservers=[ns]
-        error = True
-        while error:
+        while True:
             try:
                 result = resolver.query('example.org', dns.rdatatype.A, dns.rdataclass.IN, lifetime=10)
-                if not result.response.flags & dns.flags.AD:
-                    log.critical("Your DNS resolver at %s isn't doing DNSSEC validation; Please use another resolver or enable unbound via https://setup.mailu.io.", ns)
-                else:
-                    error = False
-                    continue
             except Exception as e:
                 log.critical("Your DNS resolver at %s is not working (%s). Please use another resolver or enable unbound via https://setup.mailu.io.", ns, e);
+            elif result.response.flags & dns.flags.AD:
+                break
+            else:
+                log.critical("Your DNS resolver at %s isn't doing DNSSEC validation; Please use another resolver or enable unbound via https://setup.mailu.io.", ns)
             time.sleep(5)
 
 test_DNS()
--- a/core/admin/start.py
+++ b/core/admin/start.py
@@ -38,12 +38,12 @@ def test_DNS():
             try:
                 result = resolver.query('example.org', dns.rdatatype.A, dns.rdataclass.IN, lifetime=10)
                 if not result.response.flags & dns.flags.AD:
-                    log.critical("Your DNS resolver at %s isn't doing DNSSEC validation; Please install unbound.", ns)
+                    log.critical("Your DNS resolver at %s isn't doing DNSSEC validation; Please use another resolver or enable unbound via https://setup.mailu.io.", ns)
                 else:
                     error = False
                     continue
             except Exception as e:
-                log.critical("Your DNS resolver at %s is not working (%s). Please install unbound.", ns, e);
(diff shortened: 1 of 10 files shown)


nextgens:
bors try

nextgens:
bors try

nextgens:
Not sure what to do for swarm/helm though...

ghostwheel42:
> Not sure what to do for swarm/helm though...

I would remove the swarm flavor in favor or kubernetes. The helm chart needs to be updated, yes.

nextgens:
@micw how does this work on the k8s front? Is the default DNS resolver DNSSEC capable/enabled?

I see some code for it in https://github.com/uswitch/kubernetes-dns/blob/master/vendor/github.com/miekg/dns/dnssec.go but it doesn't look like it's in the main repo

core/admin/start.py
```diff
@@ -18,6 +18,36 @@
...
+        resolver.nameservers=[ns]
+        error = True
+        while error:
+            try:
+                result = resolver.query('example.org', dns.rdatatype.A, dns.rdataclass.IN, lifetime=10)
+                if not result.response.flags & dns.flags.AD:
+                    log.critical("Your DNS resolver at %s isn't doing DNSSEC validation; Please install unbound.", ns)
+                else:
+                    error = False
+                    continue
```
ghostwheel42, line 44: You could use `return` here and `while True:` above  instead of the variable `error`.
Like this:
```
while True:
    try:
        result = resolver.query('example.org', dns.rdatatype.A, dns.rdataclass.IN, lifetime=10)
    except Exception as e:
        log.critical("Your DNS resolver at %s is not working (%s). Please use another resolver or enable unbound via setup.mailu.io.", ns, e);
    else:
        if result.response.flags & dns.flags.AD:
            return
        log.critical("Your DNS resolver at %s isn't doing DNSSEC validation. Please use another resolver or enable unbound via setup.mailu.io.", ns)
    time.sleep(5)
```

ghostwheel42 (changes requested):

core/admin/start.py
nextgens: That would change the logic though, wouldn't it?

Here I was trying to ensure that **all** listed nameservers are DNSSEC capable... since we don't control which one will be queried

core/admin/start.py
```diff
@@ -18,6 +18,36 @@
...
+    resolver.flags = dns.flags.AD | dns.flags.RD
+    nameservers = resolver.nameservers
+    for ns in nameservers:
+        resolver.nameservers=[ns]
+        error = True
+        while error:
+            try:
+                result = resolver.query('example.org', dns.rdatatype.A, dns.rdataclass.IN, lifetime=10)
+                if not result.response.flags & dns.flags.AD:
+                    log.critical("Your DNS resolver at %s isn't doing DNSSEC validation; Please install unbound.", ns)
```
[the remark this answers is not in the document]
nextgens: That's b12616b93f6a4df23fe24d7f34ed7feb5b8ee4cb

core/admin/start.py
ghostwheel42: Yes sure - so it would be `break` instead of `return`.

HorayNarea: the `error` variable looks indeed unnecesary … we want to loop over all resolver, regardless of what happened in one of them before… if you have 4 nameservers configured an 2 are not DNSSEC-capable, just log them once they are reached in the loop and continue with the next one
Right?

HorayNarea: so more like:

```python
while True:
  try:
    result = resolver.query('example.org', dns.rdatatype.A, dns.rdataclass.IN, lifetime=10)
  except Exception as e:
    log.critical("Your DNS resolver at %s is not working (%s). Please use another resolver or enable unbound via setup.mailu.io.", ns, e)
  else:
    if not result.response.flags & dns.flags.AD:
      log.critical("Your DNS resolver at %s isn't doing DNSSEC validation. Please use another resolver or enable unbound via setup.mailu.io.", ns)
  time.sleep(5)
```

ghostwheel42: Nope. This would be infinite.
_all_ Nameservers need to be functioning and doing DNSSEC - so a break (ie "continue testing next server") when a server is "good" is the correct thing to do.

nextgens: That's a2f6243

nextgens:
bors try

core/admin/start.py
```diff
@@ -33,17 +33,15 @@ def test_DNS():
...
             try:
                 result = resolver.query('example.org', dns.rdatatype.A, dns.rdataclass.IN, lifetime=10)
-                if not result.response.flags & dns.flags.AD:
-                    log.critical("Your DNS resolver at %s isn't doing DNSSEC validation; Please use another resolver or enable unbound via https://setup.mailu.io.", ns)
-                else:
-                    error = False
-                    continue
             except Exception as e:
                 log.critical("Your DNS resolver at %s is not working (%s). Please use another resolver or enable unbound via https://setup.mailu.io.", ns, e);
+            elif result.response.flags & dns.flags.AD:
```
ghostwheel42, line 41: Sorry, you can't do `elif` here. It's `try - except - else` only.

nextgens: Doh, fixed in a9da0c084a4d40078f706772c420db15a57df993

nextgens:
bors try

ghostwheel42 (approved):
lgtm

PR #2177, opened 2022-01-21
Update start.py to fix the resolver test

shing6326:
Use method resolve instead of query to fix the resolver test

## What type of PR?

bug-fix

### Related issue(s)
- Mention an issue like: #001
- Auto close an issue like: closes #001

## Prerequisites
Before we can consider review and merge, please make sure the following list is done and checked.
If an entry in not applicable, you can check it or remove it from the list.

- [ ] In case of feature or enhancement: documentation updated accordingly
- [ ] Unless it's docs or a minor change: add [changelog](https://mailu.io/master/contributors/workflow.html#changelog) entry file.

PR #2179, opened 2022-01-21
Fix resolver

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Switch to the new API to solve a k8s specific problem

### Related issue(s)
- #2177


ghostwheel42:
what about this?
./core/admin/mailu/models.py:                for rset in dns.resolver.query(self.name, 'MX')

nextgens:
Good catch

ghostwheel42 (approved):
lgtm

PR #2183, opened 2022-01-22
resolve #1972 authentication using external proxy

### LINKED ISSUE #1972: Header authentication using external proxy
## Before you open your issue

### Environment
 - [X] docker-compose
 - [X] kubernetes
 - [X] docker swarm

### Versions
Not Relevant to the issue.
## Description
This is either an enhancement or documentationrequest, depending on the feedback.
I tried to find out if it is supported, but cannot find anything about it.

Header authentication.
Simply put, a proxy sends the username in a header to the backend server (which would in this case be the mailu login/auth system), which then logs in the user without requesting a password or login action.

This is limited to the web interfaces ofcoarse. This enables the use of SSO proxies like authelia in front of the webmail. While still relying on internal mailu passwords for everything going directly to mailu (such as the email itself.)

I've no idea if this is already possible and/or if it is in theory reasonably possible to be added.

## Replication Steps
Not Relevant to the issue.

## Expected behaviour
Expected/Requested behavior would be allowing upstream headers containing the username or email and/or group to login users.

## Logs
Not Relevant to the issue.

tnissen375:
## What type of PR?

enhancement

## What does this PR do?

This PR enables header authentication on frontends. A proxy sends the username in a header (X-Auth-Email) to the backend server, which logs in the user without requesting a password or login action.

Therefore 2 Headers need to be set by proxy:
X-Auth-Proxy-Secret
X-Auth-Email

I m using openresty so the lines would look like:
```
ngx.req.set_header("X-Auth-Proxy-Secret", "aa0111f948a1cb2a947137e56922cb68") 
ngx.req.set_header("X-Auth-Email", res.id_token.email)
ngx.req.set_header("X-Auth-Proxy-Target", "Admin")  # optional
```
The third header is optional, its possible to choose the target after successful proxy login. Start at admin interface or webmail.

You need to add to values in mailu.env
```
PROXY_SECRET=aa0111f948a1cb2a947137e56922cb68
PROXY_CREATE=True 
```
The secret is the same as used by proxy. It substitutes user credentials and has to be kept secret.
With the second value you can decide what happens if the user is not found. User can be created automatically if it does not exist (The username must specify a user form a domain used by mailu instance) 

Now there is a new route `/sso/proxy`. If called and the proxy credentials are correct the user gets logged in. Its possible to set `WEBROOT_REDIRECT=/sso/proxy`  to enable auto login.

Thats it.
 
### Related issue(s)
- Auto close an issue like: closes #1972 

## Prerequisites

### DIFF
--- a/core/admin/mailu/configuration.py
+++ b/core/admin/mailu/configuration.py
@@ -68,6 +68,8 @@ DEFAULT_CONFIG = {
     'LOGO_URL': None,
     'LOGO_BACKGROUND': None,
     # Advanced settings
+    'PROXY_CREATE': False,
+    'PROXY_SECRET': None,
     'LOG_LEVEL': 'WARNING',
     'SESSION_KEY_BITS': 128,
     'SESSION_TIMEOUT': 3600,
--- a/core/admin/mailu/sso/views/base.py
+++ b/core/admin/mailu/sso/views/base.py
@@ -92,7 +92,7 @@ def proxy():
         flask.flash('Too many attempts for this user (rate-limit)', 'error')
         return flask.render_template('login.html', form=form, fields=fields)
 
-    if not (app.config['PROXY_SECRET'] in (None, '') or not app.config['PROXY_SECRET']) and app.config['PROXY_SECRET'] == secret and username != "":
+    if username and len(secret)>16 and app.config.get('PROXY_SECRET', "") == secret:
         user = models.User.get(flask.request.headers.get('X-Auth-Email'))
         if user:
             flask.session.regenerate()
@@ -119,9 +119,8 @@ def proxy():
                     flask.current_app.logger.info(f'Login succeeded by proxy created user: {username} from {client_ip}.')
                     return flask.render_template('login.html', form=form, fields=fields), payload_dict
             else:
-                utils.limiter.rate_limit_user(username, client_ip, device_cookie, device_cookie_username) if models.User.get(username) else utils.limiter.rate_limit_ip(client_ip)
+                utils.limiter.rate_limit_ip(client_ip)
                 flask.current_app.logger.warn(f'Login failed by proxy for {username} from {client_ip}.')
-                flask.flash('Auth by proxy failed - try to login by credentials', 'error')
     return flask.redirect(flask.url_for('.login'))
 
 @sso.route('/logout', methods=['GET'])
--- a/core/admin/mailu/sso/views/base.py
+++ b/core/admin/mailu/sso/views/base.py
@@ -6,6 +6,7 @@ from mailu.ui import access
 from flask import current_app as app
 import flask
 import flask_login
+import secrets
 
 @sso.route('/login', methods=['GET', 'POST'])
 def login():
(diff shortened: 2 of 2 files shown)


nextgens:

I am afraid that this isn't mergeable as-is; There are a few things that would deserve to be re-considered:
- GET-only authentication: is this something we want? The usual way of doing this is to form-fill (force a POST maybe with a dynamically generated temporary token) and redirect in all cases. Even if we do decide that we are OK with GET-only, I strongly suggest that we do not duplicate the code and redirect to the login form if login through headers failed.
- rate-limiting: does it make sense in this context? What about device cookies?
- configuration: should ``REAL_IP_FROM`` be used to apply additional restrictions on where the route is allowed from? Should the header names be configurable?
- security: PROXY_SECRET should enforce at least a specific length (to prevent users from shooting themselves in the foot)

core/admin/mailu/sso/views/base.py
```diff
@@ -48,6 +49,81 @@ def login():
...
+                    models.db.session.commit()
+                    user.send_welcome()
+                    flask.session.regenerate()
+                    flask_login.login_user(user)
+                    flask.current_app.logger.info(f'Login succeeded by proxy created user: {username} from {client_ip}.')
+                    return flask.render_template('login.html', form=form, fields=fields), payload_dict
+            else:
+                utils.limiter.rate_limit_user(username, client_ip, device_cookie, device_cookie_username) if models.User.get(username) else utils.limiter.rate_limit_ip(client_ip)
+                flask.current_app.logger.warn(f'Login failed by proxy for {username} from {client_ip}.')
+                flask.flash('Auth by proxy failed - try to login by credentials', 'error')
```
nextgens, line 124: This doesn't make sense to me: if the PROXY_SECRET token matches it's not a malicious attempt, merely a misconfiguration... and rate-limiting non-existing users just won't work.

core/admin/mailu/sso/views/base.py
```diff
@@ -48,6 +49,81 @@ def login():
...
+
+    device_cookie, device_cookie_username = utils.limiter.parse_device_cookie(flask.request.cookies.get('rate_limit'))
+    if username != device_cookie_username and utils.limiter.should_rate_limit_ip(client_ip):
+        flask.flash('Too many attempts from your IP (rate-limit)', 'error')
+        return flask.render_template('login.html', form=form, fields=fields)
+    if utils.limiter.should_rate_limit_user(username, client_ip, device_cookie, device_cookie_username):
+        flask.flash('Too many attempts for this user (rate-limit)', 'error')
+        return flask.render_template('login.html', form=form, fields=fields)
+
+    if not (app.config['PROXY_SECRET'] in (None, '') or not app.config['PROXY_SECRET']) and app.config['PROXY_SECRET'] == secret and username != "":
```
nextgens, line 95: This is too complicated :)

What about:

```
if username and len(secret)>XX and app.config.get('PROXY_SECRET', "") == secret:
```

core/admin/mailu/configuration.py
```diff
@@ -68,6 +68,8 @@
     'LOGO_URL': None,
     'LOGO_BACKGROUND': None,
     # Advanced settings
+    'PROXY_CREATE': False,
+    'PROXY_SECRET': None,
```
nextgens, line 72: We need to ensure that PROXY_SECRET is not brute-forceable (check that it's at least 16 chars or something)

nextgens (changes requested):

core/admin/mailu/sso/views/base.py
tnissen375: Removed

core/admin/mailu/sso/views/base.py
tnissen375: Adopted ;)

tnissen375:
> GET-only authentication: is this something we want?
Dunno, it doesnt matter in this context, cause the method is not used to submit any data. (no vars used from post or get, so no security risk at all) 
All data is extracted from header. One method has to be choosen and get is the default method, even when none is explicit specified.
>Should the header names be configurable?
Not necessary in my opinion.

The proxy auth is okay for some kinds of setups.
Personally i would prefer OpenID, seems not too hard to implement with mailu stack...

nextgens:
> > GET-only authentication: is this something we want?
> 
> Dunno, it doesnt matter in this context, cause the method is not used to submit any data. (no vars used from post or get, so no security risk at all) All data is extracted from header. One method has to be choosen and get is the default method, even when none is explicit specified.
> 

It's about ensuring that a crawler going through the proxy doesn't accidentally get logged in... and that "browser tabs" remaining open in browsers don't create a myriad of sessions by accident.

nextgens:
> > > > GET-only authentication: is this something we want?
> > > 
> > > 
> > > Dunno, it doesnt matter in this context, cause the method is not used to submit any data. (no vars used from post or get, so no security risk at all) All data is extracted from header. One method has to be choosen and get is the default method, even when none is explicit specified.
> > 
> > 
> > It's about ensuring that a crawler going through the proxy doesn't accidentally get logged in... and that "browser tabs" remaining open in browsers don't create a myriad of sessions by accident.
> 
> Shame the creator of this PR didn't respond. Lets explain and get this back on track: Header auth should be toggleable with a **very** clear warning and (preferably) a sourceip whitelist.
> 
> It's often used with reverse proxies and always requires the header to be block/filtered on the proxy-level.
> 
> However: This should follow standardised best-practices: Just the email/name and or group of the user. Not all sorts of semi-custom headers added on top for a minimal level of added security. because that level of added 'security' means users would need to make all sorts of customisations to their reverse proxy setups that are not common, which will only increase the failure rate and leads to mistakes being made.

So a source-ip whitelist is preferable to a shared key as proposed here? We can definitely drop X-Auth-Proxy-Target and pass it as a parameter instead.

PR #2195, opened 2022-01-28
roundcube: Add /overrides directory in include

mnival:
Added the /overrides directory in the roundcube config.inc.php file

## What type of PR?

bug-fix

### Related issue(s)
none


Diman0:
bors try

Diman0 (approved):
Looks good to me. 
I also pushed an updated entry for the overrides faq entry. It is handy that we also document that you can add overrides for roundcube.

@nextgens  @ghostwheel42  I also added the backport label, because without this PR you cannot use overrides in roundcube. This is vital if you want to configure the plugins via *.inc files in the overrides folder.

nextgens (approved):

PR #2196, opened 2022-01-30
roundcube-carddav : Use des_key for pwstore_scheme 

### LINKED ISSUE #2230: Roundcube fails to decrypt carddav password
### Environment
 - [x] docker-compose
 - [ ] kubernetes
 - [ ] docker swarm

### Versions

1.9

## Description

Roundcube is thankfully shipped with the [CardDav plugin}(https://github.com/mstilkerich/rcmcarddav) for syncing contacts enabled.  In its default configuration, it stores the passwords to the external service encrypted with the IMAP password. However, this does not seem to work with the new SSO scheme used by mailu - I assume similar to the other methods mentioned in the [documentation](https://github.com/mstilkerich/rcmcarddav/blob/master/doc/ADMIN-SETTINGS.md). It successfully syncs the folder once, but after logging out of webmail and in again, it generates authentication errors until the password is entered again.

I successfully fixed this by setting another password scheme within the webmail container by adding `/var/www/html/plugins/carddav/config.inc.php` with the following content:

```
<?php $prefs['_Global']['pwstore_scheme'] = 'base64';
```

I guess, with mailu's SSO something similar should be the default setting since it is impossible (at least for me) to change this setting in a persistent way without manually rebuilding the container.

mnival:
roundcube-carddav: Configuring pwstore_scheme in carddav plugin with des_key because Mailu is incompatible with encrypted

https://github.com/mstilkerich/rcmcarddav/blob/master/doc/ADMIN-SETTINGS.md#password-storing-scheme

## What type of PR?

bug-fix

### Related issue(s)
- closes #2230

### DIFF
--- a/towncrier/newsfragments/2196.bugfix
+++ b/towncrier/newsfragments/2196.bugfix
@@ -1,0 +1,2 @@
+Configuring pwstore_scheme in carddav plugin with des_key because Mailu is incompatible with encrypted
+https://github.com/mstilkerich/rcmcarddav/blob/master/doc/ADMIN-SETTINGS.md#password-storing-scheme
--- a/webmails/roundcube/Dockerfile
+++ b/webmails/roundcube/Dockerfile
@@ -66,6 +66,7 @@ COPY mailu.php /var/www/html/plugins/mailu/mailu.php
 COPY php.ini /
 COPY config.inc.php /
 COPY start.py /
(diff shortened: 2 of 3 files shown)


nextgens (approved):
LGTM; not sure if we should backport it

Diman0 (approved):
Lgtm. Since the change is small in scope and affects multiple users we could backport it.

PR #2201, opened 2022-01-31
Add Password Complexity feature

### LINKED ISSUE #287: Password Complexity Requirements

0pc0deFR:
## What type of PR?

Feature

## What does this PR do?

This PR add the password complexity function (requested in #287). This option can be enabled by domain and it's applied to all users created on this domain.

### Related issue(s)
- Mention an issue like: #287
- Auto close an issue like: closes #287

## Prerequisites

### DIFF
--- a/core/admin/mailu/models.py
+++ b/core/admin/mailu/models.py
@@ -4,6 +4,7 @@
 import os
 import smtplib
 import json
+import string
 
 from datetime import date
 from email.mime import text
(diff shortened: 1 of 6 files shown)


nextgens:
Hmm, this is unreadable; can you fix the merge so that only relevant changes appear please?

I think that we should enforce a 8 char limit in any case... as for the rest of the complexity options, I'm not convinced that they are useful. Recent guidance from people who know better think that it's counter productive.

0pc0deFR:
I close this PR because I create #2208

PR #2205, opened 2022-01-31
Update nginx.conf

### LINKED ISSUE #2204: Setting TLS_FLAVOR=notls does not prevent front from redirecting to https
## Before you open your issue

### Environment
 - [x] docker-compose
 - [ ] kubernetes
 - [ ] docker swarm

### Versions
1.9

## Description
setting TLS_FLAVOR=notls seems to not works as front nginx keeps redirecting to https and also gives http 403

## Replication Steps
set TLS_FLAVOR=notls in mailu.env

## Expected behaviour
external proxy handling certificates correctly redirect to mailu http and mailu logic correctly works without certificates

## Logs
n/a

audioscavenger:
`notls` was missing from the exclusions to redirect 403 or https

## What type of PR?

bug-fix

### Related issue(s)
- closes #2204

## Prerequisites

### DIFF
--- a/core/nginx/conf/nginx.conf
+++ b/core/nginx/conf/nginx.conf
@@ -103,7 +103,7 @@ http {
       ssl_session_cache shared:SSLHTTP:50m;
       add_header Strict-Transport-Security 'max-age=31536000';
 
-      {% if not TLS_FLAVOR in [ 'mail', 'mail-letsencrypt' ] %}
+      {% if not TLS_FLAVOR in [ 'mail', 'mail-letsencrypt', 'notls' ] %}
       if ($proxy_x_forwarded_proto = http) {
         return 301 https://$host$request_uri;
       }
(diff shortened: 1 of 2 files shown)


core/nginx/conf/nginx.conf
```diff
@@ -127,7 +127,7 @@ http {
       {% endif %}
 
       # If TLS is failing, prevent access to anything except certbot
-      {% if KUBERNETES_INGRESS != 'true' and TLS_ERROR and not (TLS_FLAVOR in [ 'mail-letsencrypt', 'mail' ]) %}
+      {% if KUBERNETES_INGRESS != 'true' and TLS_ERROR and not (TLS_FLAVOR in [ 'mail-letsencrypt', 'mail', 'notls' ]) %}
```
nextgens, line 130: I don't understand what this does. It either breaks notls if TLS_ERROR is set or does nothing (as it should never be set when notls is true)

audioscavenger: scratch that then. i still don't get what notls is for. sorry.

audioscavenger:
was fooled by what notls actually does. no idea what it's for.

PR #2207, opened 2022-01-31
Update webmail container configuration to support MESSAGE_SIZE_LIMIT

### LINKED ISSUE #2186: max attachement size with rainloop is 1mb
### Environment
 - [x] docker-compose
 - [ ] kubernetes
 - [ ] docker swarm

### Versions
1.9

## Description
cannot attach anything larger then 1,048,300 bytes to any email from rainloop

## Replication Steps
attach anything larger then 1,048,300 bytes to any email from rainloop and it will fails with http 413

## Expected behaviour
i can attach any file as large as the MESSAGE_SIZE_LIMIT as defined in (mailu.env - 33%)

## Logs
browser log only: for the xhr POST
````markdown
```
413 Request Entity Too Large
```
````

### PR DESCRIPTION
**Pull request by marioja**
## What type of PR?

bug-fix

### Related issue(s)
- Auto close an issue like: closes #2186 

## Prerequisites

### CODE REVIEW

**Comment by marioja** [comment]
The PR competes with #2187.  The major difference is that:

1. The webmail container supports the same limit as the front container for client_max_body_size (PR #2187 sets it to 0)
2. It was tested and works.  I can attach up to 29MB with a 50,000,000 message size limit.

File: webmails/rainloop/config/nginx-rainloop.conf [hunk 1]
```diff
@@ -12,13 +12,16 @@ server {
 
     index index.php;
 
+    # set maximum body size to configured limit
+    client_max_body_size {{ MESSAGE_SIZE_LIMIT|int + 8388608 }};
```
audioscavenger, line 16: you just reintroduced the bug i was fixing in [#967ebad](https://github.com/Mailu/Mailu/pull/2187/commits/967ebad94d9103309a60f2d5c3de6cb213974625)

marioja: Actually, I am following @nextgens comment https://github.com/Mailu/Mailu/pull/2187#discussion_r795542809 not to support setting message size limit 0.  I do not believe that it should be allowed.

marioja: @audioscavenger I discovered something that is probably also applicable to #2187.  If I recreate the webmail container from my repo, tag it mail/webmail:1.9 in my local docker host and run docker-compose up -d then the change to webmail is not effective and I cannot upload a 9MB attachment.  If I docker-compose restart webmail, then the same attachment will work.  This imply that if I start all containers at the same time, things don't work.  If I just restart webmail, the change in /etc/nginx/nginx.conf (client max body size set to 0 disable) seems to take effect.

audioscavenger: no idea what you mean, sorry. are you okay with limit=0 or not is the question

marioja: I said ealier that I agreed that @nextgens that size 0 is not desirable. As for what I am talking about: when I test your changes or mine I update the webmail container and then run mailu with docker-compose. When I do this it fails until I use docker-compose restart for webmail. Is it not how you test and do you not get the same results?

marioja: Ok, it took a while but I finally figured out what was going on.  The issue is that the Dockerfile for webmail starts the fastcgi process manager and then updates the 3 *.ini files that it uses.  This cause those configuration changes to only take place when restarting the container.  I have modified the Dockerfile to not start the fastcgi process manager and update the start.py to start it after making the configuration changes using jinja.  Now the MESSAGE_SIZE_LIMIT configured in the mailu.env file are respected in both nginx and php-fpm7 on the webmail container.  I have of course tested those changes.  This can be reviewed and is ready to pull back into master.

marioja:
@nextgens @Diman0 Gents, is there anything else that I need to do to move this forward?

Diman0:
It looks good to me. The functionality works. It is a nice touch  that config.py reloads nginx, however in general you must always recreate all containers upon changing mailu.env. But it does not hurt.

I tested uploading a  27MB attachment without issues using rainloop. Reverting back to 1.9 showed an error for attachments larger than 1 MB (as expected).

Diman0 (approved):
lgtm

ghostwheel42 (approved):
LGTM
I'm not sure why there is a separate config.py - but that's how it is done in the front container.

nextgens:
@mergifyio backport 1.9

Dennis14e:
Can this be (manually) backported to 1.9?
Apparently mergify couldn't do it automatically.
@nextgens

nextgens:
It will be, @Diman0 said he would work on it... but feel free to propose a PR to speedup the process (with 1.9 as base branch)

PR #2208, opened 2022-02-01
Add Password Complexity feature

### LINKED ISSUE #287: Password Complexity Requirements

0pc0deFR:
## What type of PR?

Feature

## What does this PR do?

This PR add the password complexity function (requested in #287). This option can be enabled by domain and it's applied to all users created on this domain.

### Related issue(s)
- Mention an issue like: #287 
- Auto close an issue like: closes #287 

## Prerequisites

### DIFF
--- a/core/admin/mailu/models.py
+++ b/core/admin/mailu/models.py
@@ -189,6 +189,7 @@ class Domain(Base):
     max_aliases = db.Column(db.Integer, nullable=False, default=-1)
     max_quota_bytes = db.Column(db.BigInteger, nullable=False, default=0)
     signup_enabled = db.Column(db.Boolean, nullable=False, default=False)
+    password_complexity_enabled = db.Column(db.Boolean, nullable=False, default=False)
 
     _dkim_key = None
     _dkim_key_on_disk = None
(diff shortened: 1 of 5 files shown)


0pc0deFR:
@nextgens I think it might be interesting to force 8 characters minimum but providing an additional option would allow administrators to choose the level of password complexity. In my proposal this option is disabled by default and can be enabled per domain.

audioscavenger:
can you please add an entry in mailu.env to enable it?

0pc0deFR:
@audioscavenger you want the activation to be done via an environment variable rather than on a particular domain?

nextgens:
I personally think that this is the wrong way to go; I suggest that we enforce 8 characters minimum (but no other restriction) and introduce an optional (default off, with a global config flag) way of checking password changes using https://haveibeenpwned.com & javascript

Diman0:
It would be nice to check the password with https://haveIbeenpowned.com. Unfortunately the API nowadays is a paid service. https://www.troyhunt.com/authentication-and-the-have-i-been-pwned-api/

Diman0:
That is nice. Then I also support your suggestion. Set a hard limit to 8 characters minimum from the password.
Optionally globally enabled via mailu.env, check ihavebeenpowned (https://haveibeenpwned.com/API/v3#PwnedPasswords).

ecdlguy:
Hi,

The password policy of our university states that the password...
- is at least 12 characters (20 for administrative accounts) long.
- contains upper and lower case letters, numbers and special characters (?!%+…).
- does not contain personal data like the name of a family member, your own pet, etc.
- is not found in the dictionary.
- is not just a simple password with a single number or one of the usual special characters ($, !, ?, #) at the beginning or end of the word.

Since im currently looking for an alternative for our current mailserver it would be great if the mailu password policy could be defined via the env-file, at least some basic aspects like the length. I don't see how I can use mailu for our department at the moment due to the hard coded minimum password length of 8.

cheers, Thorsten

nextgens:
You can tell them that Mailu's default password policy is compliant with [NIST 800-63-3](https://pages.nist.gov/800-63-3/sp800-63b.html#5111-memorized-secret-authenticators) but that theirs is not.

PR #2216, opened 2022-02-03
Add iptables+ipset as additional FAQ entry for fail2ban

### LINKED ISSUE #2214: fail2ban proposal - use ipset rather than plain iptables

## Before you open your issue

### Environment
 - [x] docker-compose
 - [ ] kubernetes
 - [ ] docker swarm

## Description
ipset is a nice addon for iptables, if you filter a lot of IPs with the same rule - as it is for blocking with fail2ban.

So the change (using the example from the faqs) would be something like:
```
[Definition]

actionstart = actionstart = ipset --create f2b-bad-auth iphash
              iptables -I DOCKER-USER -p tcp -m multiport --dports 1:1024 -m set --match-set f2b-bad-auth src -j DROP

actionstop = iptables -D DOCKER-USER -p tcp -m multiport --dports 1:1024 -m set --match-set f2b-bad-auth src -j DROP
             ipset --destroy f2b-bad-auth


actionban = ipset --test f2b-bad-auth <ip> ||  ipset --add f2b-bad-auth <ip>

actionunban = ipset --test f2b-bad-auth <ip> && ipset --del f2b-bad-auth <ip>
```

In the end you just have one iptable rule. Specially when dealing with bigger attacks, this come in handy, since having a lot of iptable rules might have a big impact on the network performance. I do have like 1000+ entries in the ipset, with no big load to be handled. This was very different when I had them as individual iptable rules in an own chain.

Do you think this is worth to have a PR to add this to the FAQs?

tkaefer:
## What type of PR?

### Related issue(s)
 closes #2214 

## Prerequisites

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -664,6 +664,8 @@ The above will block flagged IPs for a week, you can of course change it to you 
 The above will block flagged IPs for a week, you can of course change it to you needs.
 
 7. Add the /etc/fail2ban/action.d/docker-action.conf
+  
+Option 1: Use plain iptables
 
 .. code-block:: bash
 
@@ -682,6 +684,33 @@ The above will block flagged IPs for a week, you can of course change it to you 
   actionban = iptables -I f2b-bad-auth 1 -s <ip> -j DROP
   
   actionunban = iptables -D f2b-bad-auth -s <ip> -j DROP
+
+Using DOCKER-USER chain ensures that the blocked IPs are processed in the correct order with Docker. See more in: https://docs.docker.com/network/iptables/
+
+Option 2:  Use ipset together with iptables
+IMPORTANT: You have to install ipset on the host system, eg. `apt-get install ipset` on a Debian/Ubuntu system.
+
+See ipset homepage for details on ipset, https://ipset.netfilter.org/.
+
+ipset and iptables provide one big advantage over just using iptables: This setup reduces the overall iptable rules.
+There is just one rule for the bad authentications and the IPs are within the ipset. 
+Specially in larger setups with a high amount of brute force attacks this comes in handy.
+Using iptables with ipset might reduce the system load in such attacks significantly.
+
+.. code-block:: bash
+
+  [Definition]
+
+  actionstart = actionstart = ipset --create f2b-bad-auth iphash
+                iptables -I DOCKER-USER -p tcp -m multiport --dports 1:1024 -m set --match-set f2b-bad-auth src -j DROP
+
+  actionstop = iptables -D DOCKER-USER -p tcp -m multiport --dports 1:1024 -m set --match-set f2b-bad-auth src -j DROP
+               ipset --destroy f2b-bad-auth
+
+
+  actionban = ipset --test f2b-bad-auth <ip> ||  ipset --add f2b-bad-auth <ip>
+
+  actionunban = ipset --test f2b-bad-auth <ip> && ipset --del f2b-bad-auth <ip>
 
 Using DOCKER-USER chain ensures that the blocked IPs are processed in the correct order with Docker. See more in: https://docs.docker.com/network/iptables/
 


docs/faq.rst
```diff
@@ -685,6 +687,33 @@ The above will block flagged IPs for a week, you can of course change it to you
...
+  actionstart = actionstart = ipset --create f2b-bad-auth iphash
+                iptables -I DOCKER-USER -p tcp -m multiport --dports 1:1024 -m set --match-set f2b-bad-auth src -j DROP
+
+  actionstop = iptables -D DOCKER-USER -p tcp -m multiport --dports 1:1024 -m set --match-set f2b-bad-auth src -j DROP
+               ipset --destroy f2b-bad-auth
+
+
+  actionban = ipset --test f2b-bad-auth <ip> ||  ipset --add f2b-bad-auth <ip>
+
+  actionunban = ipset --test f2b-bad-auth <ip> && ipset --del f2b-bad-auth <ip>
```
nextgens, line 713: Why do you run a test first? The man page says that the command will be ignored if the entry is already there (or missing)

docs/faq.rst
```diff
@@ -685,6 +687,33 @@ The above will block flagged IPs for a week, you can of course change it to you
...
+Using iptables with ipset might reduce the system load in such attacks significantly.
+
+.. code-block:: bash
+
+  [Definition]
+
+  actionstart = actionstart = ipset --create f2b-bad-auth iphash
+                iptables -I DOCKER-USER -p tcp -m multiport --dports 1:1024 -m set --match-set f2b-bad-auth src -j DROP
+
+  actionstop = iptables -D DOCKER-USER -p tcp -m multiport --dports 1:1024 -m set --match-set f2b-bad-auth src -j DROP
```
nextgens, line 707: Why do you restrict it to TCP and lower port numbers?

I don't understand why port zero is excluded either :)

I would keep it simple and make it something like: ``iptables -I DOCKER-USER -m set --match-set f2b-bad-auth src -j DROP``

nextgens (changes requested):

docs/faq.rst
tkaefer: You'll get an exit code `1` and some "error" message, eg.:
```
# ipset --add f2b-bad-auth 37.x.x.x
ipset v7.10: Element cannot be added to the set: it's already added
# echo $?
1
``` 
ipset is ignoring it, but still being particular about it. 
That's why I thought it would be less "noisy" be doing checks upfront.

docs/faq.rst
tkaefer: I did match it to the existing documentation:
> actionstart = iptables -N f2b-bad-auth
>              iptables -A f2b-bad-auth -j RETURN
>              iptables -I DOCKER-USER -p tcp -m multiport --dports 1:1024 -j f2b-bad-auth

I did it different for my own setup, with a more configurable approach - but this would be different to anything that was documented before. It's more complex (with additional bash scripts called) and might be a bit confusing if you'll never dealt with fail2ban. 

It's documented here https://tkaefer.de/blog/2021/12/07/follow-up-docker-and-fail2ban-how-i-solved-it-for-me/

As said, it's more complex and might be too much for an FAQ.

docs/faq.rst
nextgens: The right syntax is ``ipset add f2b-bad-auth <ip> -exist`` ... like that there is no error thrown if it's removed/deleted twice.

```
Synopsis
ipset [ OPTIONS ] COMMAND [ COMMAND-OPTIONS ]

COMMANDS := { create | add | del | test | destroy | list | save | restore | flush | rename | swap | help | version | - }

OPTIONS := { -exist | -output { plain | save | xml } | -quiet | -resolve | -sorted | -name | -terse } 
```

docs/faq.rst
nextgens: That's a good reason :)

IMHO it's very confusing and error prone to selectively block a subset of ports... Please update the existing documentation too and let's make the block "global" (all protocols, all ports)

nextgens (approved):
LGTM

Diman0 (approved):
lgtm

PR #2221, opened 2022-02-09
Add support for custom NGINX config

easybe:
## What type of PR?

enhancement

## What does this PR do?

Add support for custom NGINX config. Including *.conf files in /etc/nginx/conf.d same as the default NGINX configuration gives the user more flexibility.

## Prerequisites

### DIFF
--- a/core/nginx/conf/nginx.conf
+++ b/core/nginx/conf/nginx.conf
@@ -248,6 +248,8 @@ http {
         proxy_pass http://$admin/internal$request_uri;
       }
     }
+
+    include /etc/nginx/conf.d/*.conf;
 }
 
 mail {
--- a/towncrier/newsfragments/2221.feature
+++ b/towncrier/newsfragments/2221.feature
@@ -1,0 +1,1 @@
+Add support for custom NGINX config in /etc/nginx/conf.d.


Diman0:
What are you trying to solve?
We already have https://github.com/Mailu/Mailu/blob/e48e2cdc1ae827f6e048f972f823d863d7a57ef9/core/nginx/conf/nginx.conf#L135 as documented here https://mailu.io/1.9/faq.html#how-can-i-override-settings

easybe:
> What are you trying to solve? We already have
> 
> https://github.com/Mailu/Mailu/blob/e48e2cdc1ae827f6e048f972f823d863d7a57ef9/core/nginx/conf/nginx.conf#L135
> 
> as documented here https://mailu.io/1.9/faq.html#how-can-i-override-settings

The problem with
https://github.com/Mailu/Mailu/blob/e48e2cdc1ae827f6e048f972f823d863d7a57ef9/core/nginx/conf/nginx.conf#L135
it that it is within a `server` block. I am seeking the possibility to provide my own `server` blocks.

ghostwheel42:
Sorry for the delay. 
Could you give an example why you would like to add your own server blocks?
Do you want to use the front container as reverse proxy for other services?

easybe:
> Do you want to use the front container as reverse proxy for other services?

Yes, exactly.

heshiming:
I second this. Basically what I want is another override inclusion under the "http{}" block so that I can run other services on the same server.

Diman0 (approved):
lgtm

nextgens (approved):
LGTM

Issue #2239, opened 2022-02-14
helm/k8s deployments seem to have problems with DNS resolution;  Mailu expects validated dnssec records


devfaz:
## Description
As the default k8s dns setup is not working correctly with dnssec, we will (like the docker setup) use unbound as resolver, but to allow us to continue using the internal k8s-dns system, we need a dns forwarder setting.

## Replication Steps
Setup mailu with mailu helm chart in k8s

I will provide a suitable fix myself.

nextgens:
This is related to #2163

We have talked about it on #mailu-dev recently. My preferred option would be to do the opposite: configure k8s' resolver to use unbound (that would run as a service in a pod) as upstream DNS resolver. The pro of doing it that way is to ensure cluster stability if the unbound pod is having problems...

https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/#coredns-configuration-equivalent-to-kube-dns explains how to configure/change the upstream for the two common cases (coredns or kube-dns). Maybe all we need is some documentation in the FAQ?

nextgens:
@devfaz which DNS resolver is your k8s deployment configured to use? what is its upstream?

Can you run ``dig @<ip> +adflag example.org`` on the various resolvers and track down where you are loosing the AD flag please?

devfaz:
@nextgens my adflag gets in the coredns of k8s. I reconfigured my cluster to use google dns (8.8.8.8). Asking the google-dns directly from my pod returns the ad-flag, asking my coredns (forwarding the request to google) returns an answer without ad-flag, so seems CoreDNS is (by default/generally?) not able to do dnssec correctly in k8s.

nextgens:
Yes, I do believe that it's a documentation problem more than anything else.

For dnsmasq you can make it perform the validation https://wiki.gentoo.org/wiki/Dnsmasq/en#DNSSEC or configure it to pass the validation of upstream servers through

devfaz:
Thanks for the link, but in openwrt its just a tick in a checkbox ;)

I will try to provide the requests PR.

devfaz:
@nextgens so after the tests yesterday i undo my changes (no unbound any longer), but now admin-pod complains again with
`CRITICAL:root:Your DNS resolver at 192.168.102.10 isn't doing DNSSEC validation; Please use another resolver or enable unbound via https://setup.mailu.io.
`
but if I start another pod in the same namespace and execute the dig-test DNSSEC seems to work

[...]

PR #2243, opened 2022-02-15
trying to fix dnssec check in admin

devfaz:
## What type of PR?

bug-fix

## What does this PR do?
fixes dnssec check for CoreDNS based dns-servers

### Related issue(s)
- #2239

### DIFF
--- a/core/admin/start.py
+++ b/core/admin/start.py
@@ -28,7 +28,7 @@ def test_DNS():
     import time
     # DNS stub configured to do DNSSEC enabled queries
     resolver = dns.resolver.Resolver()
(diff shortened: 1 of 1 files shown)


nextgens:
We'd need the same change in core/admin/mailu/utils.py too if we go for it

devfaz:
I currently don't think we should change this if coreDNS is changing it's behavior to return AD flag regardless of DNSSEC requested. This would allow us to detect outdated versions in customer setups and give some advise in our docs.

On the other side, we have to adapt our check if coreDNS is not fixing the issue, so let us just wait till we know if coreDNS is fixing the issue.

Issue #2164, opened 2022-01-14
Mailu 1.9 update requires recreation of docker-compose.yml file


Diman0:
There was an error in the docker-compose.yml created by the setup site. Not all services were configured to use the resolver (unbound). This has been addressed, but requires the docker-compose.yml to be updated. An update was also pushed to ensure a proper resolver is used which supports DNSSEC. With the old docker-compose.yml file the admin service will not start with the error that the resolver does not support DNSSEC.

To resolve your issue you can either:
- Regenerate your docker-compose.yml file on setup.mailu.io
  - Stop Mailu: docker-compose down
  - Replace docker-compose.yml with the file downloaded from the website. If you used different values in setup, then you must update mailu.env as well.
  - docker-compose up -d
- Manually update the docker-compose.yml file
1. Stop all services: docker-compose down
2. Open the docker-compose.yml file
3. Locate the resolver service and note down the IP address. For a default setup it is 192.168.203.254.
```
resolver:
   image: ${DOCKER_ORG:-mailu}/${DOCKER_PREFIX:-}unbound:${MAILU_VERSION:-1.9}
   env_file: mailu.env
   restart: always
   networks:
     default:
       ipv4_address: 192.168.203.254
```
  4.  Make sure all service (but resolver) have the following config
```
    depends_on:
      - resolver
    dns:
      - 192.168.203.254
```
5. Start all services: docker-compose up -d

Issue #2163, opened 2022-01-13
mailu_admin_1 is unhealthy (please see the logs).


medram:
### Environment
 - [x] docker-compose
 - [ ] kubernetes
 - [ ] docker swarm

### Versions
1.9

## Description
Every container is healthy except for mailu_admin_1, which is unhealthy I don't know why.
Please help :D

## Expected behaviour
mailu_admin_1 should be healthy.

## Logs
````markdown
```
CRITICAL:root:Your DNS resolver at 127.0.0.11 isn't doing DNSSEC validation; Please use another resolver or enable unbound via https://setup.mailu.io.
CRITICAL:root:Your DNS resolver at 127.0.0.11 isn't doing DNSSEC validation; Please use another resolver or enable unbound via https://setup.mailu.io.
CRITICAL:root:Your DNS resolver at 127.0.0.11 isn't doing DNSSEC validation; Please use another resolver or enable unbound via https://setup.mailu.io.
CRITICAL:root:Your DNS resolver at 127.0.0.11 isn't doing DNSSEC validation; Please use another resolver or enable unbound via https://setup.mailu.io.
CRITICAL:root:Your DNS resolver at 127.0.0.11 isn't doing DNSSEC validation; Please use another resolver or enable unbound via https://setup.mailu.io.
CRITICAL:root:Your DNS resolver at 127.0.0.11 isn't doing DNSSEC validation; Please use another resolver or enable unbound via https://setup.mailu.io.
CRITICAL:root:Your DNS resolver at 127.0.0.11 isn't doing DNSSEC validation; Please use another resolver or enable unbound via https://setup.mailu.io.
```
````

archerne:
I had this yesterday. If you recreate the docker-compose (on setup.mailu.io) and compare to yours there is most likely some unbound configuration that you need to update on a couple of the containers. You probably updated to the latest version but didnt change all the settings.

medram:
Thanks for your reply Archerne,
You're right, I did update to the latest version using 'docker-compose pull'.
can I use a new generated configs from setup.mailu.io and use it with my previous mailu data (that is located at '/mailu' folder), without clearing any cache or something like that?

Diman0:
There was an error in the docker-compose.yml created by the setup site. Not all services were configured with the resolver (unbound). This has been resolved.

To resolver your issue you can either.

- Regenerate your docker-compose.yml file on setup.mailu.io
- Manually update the docker-compose.yml file
[...]

Issue #2162, opened 2022-01-13
Strange behaviour after restart.


richardbeumer:
### Environment
 - [ ] docker-compose
 - [x] kubernetes
 - [ ] docker swarm

### Versions
1.8

## Description
After restarting my pods I suddenly get below error when trying to login. I suspect it's an issue in my database but I'm not sure. Please advise where to look further

## Logs
```
[2022-01-13 19:19:54,943] ERROR in app: Exception on /internal/auth/email [GET]
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/flask/app.py", line 2292, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/lib/python3.9/site-packages/flask/app.py", line 1815, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/lib/python3.9/site-packages/flask/app.py", line 1718, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/lib/python3.9/site-packages/flask/_compat.py", line 35, in reraise
    raise value
  File "/usr/lib/python3.9/site-packages/flask/app.py", line 1813, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/lib/python3.9/site-packages/flask/app.py", line 1799, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/app/mailu/internal/views/auth.py", line 24, in nginx_authentication
    headers = nginx.handle_authentication(flask.request.headers)
  File "/app/mailu/internal/nginx.py", line 59, in handle_authentication
    if (token.check_password(password) and
  File "/app/mailu/models.py", line 499, in check_password
    return hash.sha256_crypt.verify(password, self.password)
  File "/usr/lib/python3.9/site-packages/passlib/utils/handlers.py", line 757, in verify
    self = cls.from_string(hash, **context)
  File "/usr/lib/python3.9/site-packages/passlib/handlers/sha2_crypt.py", line 307, in from_string
    raise uh.exc.InvalidHashError(cls)
ValueError: not a valid sha256_crypt hash
```

richardbeumer:
Apperently my version was upgraded to `1.9`. I was using the `master` tag so all my containers where updated when restarted.
[...]

PR #2253, opened 2022-02-19
Workaround the infamous coredns feature

### LINKED ISSUE #2239: helm/k8s deployments seem to have problems with DNS resolution;  Mailu expects validated dnssec records

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Ensure that we set the ``DO`` flag on our queries to work around https://github.com/coredns/coredns/issues/5189

Add a FAQ entry to point users in the right direction in other cases (dnsmasq), discourage users from running Mailu without unbound

### Related issue(s)
- closes #2243
- closes #2239
- #2164
- #2163
- #2162
- #2135
- #1988

## Prerequisites

### DIFF
--- a/core/admin/start.py
+++ b/core/admin/start.py
@@ -37,11 +37,11 @@ def test_DNS():
             try:
                 result = resolver.resolve('example.org', dns.rdatatype.A, dns.rdataclass.IN, lifetime=10)
             except Exception as e:
-                log.critical("Your DNS resolver at %s is not working (%s). Please see https://mailu.io/1.9/faq.html#the-admin-container-won-t-start-and-its-log-says-critical-your-dns-resolver-isn-t-doing-dnssec-validation", ns, e);
+                log.critical("Your DNS resolver at %s is not working (%s). Please see https://mailu.io/master/faq.html#the-admin-container-won-t-start-and-its-log-says-critical-your-dns-resolver-isn-t-doing-dnssec-validation", ns, e);
             else:
                 if result.response.flags & dns.flags.AD:
                     break
-                log.critical("Your DNS resolver at %s isn't doing DNSSEC validation; Please see https://mailu.io/1.9/faq.html#the-admin-container-won-t-start-and-its-log-says-critical-your-dns-resolver-isn-t-doing-dnssec-validation.", ns)
+                log.critical("Your DNS resolver at %s isn't doing DNSSEC validation; Please see https://mailu.io/master/faq.html#the-admin-container-won-t-start-and-its-log-says-critical-your-dns-resolver-isn-t-doing-dnssec-validation.", ns)
             time.sleep(5)
 
 test_DNS()
--- a/core/admin/start.py
+++ b/core/admin/start.py
@@ -28,7 +28,7 @@ def test_DNS():
     import time
     # DNS stub configured to do DNSSEC enabled queries
     resolver = dns.resolver.Resolver()
-    resolver.use_edns(0, 0, 1232)
+    resolver.use_edns(0, dns.flags.DO, 1232)
     resolver.flags = dns.flags.AD | dns.flags.RD
     nameservers = resolver.nameservers
     for ns in nameservers:
--- a/core/admin/start.py
+++ b/core/admin/start.py
@@ -37,11 +37,11 @@ def test_DNS():
             try:
                 result = resolver.resolve('example.org', dns.rdatatype.A, dns.rdataclass.IN, lifetime=10)
             except Exception as e:
-                log.critical("Your DNS resolver at %s is not working (%s). Please use another resolver or enable unbound via https://setup.mailu.io.", ns, e);
+                log.critical("Your DNS resolver at %s is not working (%s). Please see https://mailu.io/1.9/faq.html#the-admin-container-won-t-start-and-its-log-says-critical-your-dns-resolver-isn-t-doing-dnssec-validation", ns, e);
             else:
                 if result.response.flags & dns.flags.AD:
                     break
-                log.critical("Your DNS resolver at %s isn't doing DNSSEC validation; Please use another resolver or enable unbound via https://setup.mailu.io.", ns)
+                log.critical("Your DNS resolver at %s isn't doing DNSSEC validation; Please see https://mailu.io/1.9/faq.html#the-admin-container-won-t-start-and-its-log-says-critical-your-dns-resolver-isn-t-doing-dnssec-validation.", ns)
             time.sleep(5)
 
 test_DNS()
--- a/setup/static/render.js
+++ b/setup/static/render.js
@@ -83,4 +83,11 @@ $(document).ready(function() {
 			$("#ipv6").hide();
 		}
 	});
+	$("#unbound").change(function() {
+		if ($(this).is(":checked")) {
+			$("#unbound").hide();
+		} else {
+			$("#unbound").show();
+		}
+	});
 });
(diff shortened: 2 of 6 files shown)


core/admin/start.py
```diff
@@ -37,11 +37,11 @@ def test_DNS():
             try:
                 result = resolver.resolve('example.org', dns.rdatatype.A, dns.rdataclass.IN, lifetime=10)
             except Exception as e:
-                log.critical("Your DNS resolver at %s is not working (%s). Please use another resolver or enable unbound via https://setup.mailu.io.", ns, e);
+                log.critical("Your DNS resolver at %s is not working (%s). Please see https://mailu.io/1.9/faq.html#the-admin-container-won-t-start-and-its-log-says-critical-your-dns-resolver-isn-t-doing-dnssec-validation", ns, e);
```
Diman0, line 40: This should link to master. Master is always available. The 1.9 documentation site will eventually not be available anymore.

core/admin/start.py
```diff
@@ -37,11 +37,11 @@ def test_DNS():
...
             else:
                 if result.response.flags & dns.flags.AD:
                     break
-                log.critical("Your DNS resolver at %s isn't doing DNSSEC validation; Please use another resolver or enable unbound via https://setup.mailu.io.", ns)
+                log.critical("Your DNS resolver at %s isn't doing DNSSEC validation; Please see https://mailu.io/1.9/faq.html#the-admin-container-won-t-start-and-its-log-says-critical-your-dns-resolver-isn-t-doing-dnssec-validation.", ns)
```
Diman0, line 44: This should link to master. Master is always available. The 1.9 documentation site will eventually not be available anymore.

setup/static/render.js
```diff
@@ -83,4 +83,11 @@ $(document).ready(function() {
 			$("#ipv6").hide();
 		}
 	});
+	$("#unbound").change(function() {
```
Diman0, line 86: It should check the property of the input resolver_enabled

setup/templates/steps/compose/03_expose.html
```diff
@@ -40,13 +40,16 @@
   <input class="form-control" type="text" name="subnet6" required value="{{ subnet6 }}:beef::/64">
 </div>
 
-<p>The unbound resolver enables Mailu to do DNSSEC verification, DNS root lookups and caching. This also helps the antispam service not to get blocked by the public or ISP DNS servers.</p>
 <div class="form-check form-check-inline">
+  <p>The unbound resolver enables Mailu to do DNSSEC verification, DNS root lookups and caching. This also helps the antispam service not to get blocked by the public or ISP DNS servers.</p>
   <label class="form-check-label">
     <input class="form-check-input" type="checkbox" name="resolver_enabled" value="true" checked>
```
Diman0, line 46: Should get a property id="resolver_enabled" which can be used in render.js for showing the warning message element.

Diman0 (dismissed):
See in-line comments. The main functionality works Ok. 
Thank you for adding the additional warning when unbound is unchecked. I hope this will prevent users from unchecking this without realizing the consequences.

I will submit a commit for the requested changes.

Diman0:
@nextgens  For me the PR is fine now. Since I pushed changes, please @mention me if you are happy now with the PR and I will approve it.

core/admin/start.py
```diff
@@ -37,11 +37,11 @@ def test_DNS():
             try:
                 result = resolver.resolve('example.org', dns.rdatatype.A, dns.rdataclass.IN, lifetime=10)
             except Exception as e:
-                log.critical("Your DNS resolver at %s is not working (%s). Please use another resolver or enable unbound via https://setup.mailu.io.", ns, e);
+                log.critical("Your DNS resolver at %s is not working (%s). Please see https://mailu.io/master/faq.html#the-admin-container-won-t-start-and-its-log-says-critical-your-dns-resolver-isn-t-doing-dnssec-validation", ns, e);
```
ghostwheel42, line 40: I think we could (and should) use /version here.
Just change the message
```
log.critical("Your DNS resolver at %s is not working (%s). Please see https://mailu.io/%s/faq.html#the-admin-container-won-t-start-and-its-log-says-critical-your-dns-resolver-isn-t-doing-dnssec-validation", ns, e, version);
```

And load the file on line ~8:
```
try:
    version = '.'.join(open('/version').read().split('.')[:2])
except:
    version = 'master'
```

ghostwheel42 (changes requested):
what about using /version ?

core/admin/start.py
Diman0: As long the user is not on an old version of Mailu (for which there is not a documentation site), this should work.

nextgens: IMHO that's all stuff that socrates (or its replacement) should do/deal with. Odds are this isn't the only place that would benefit from knowing the version number

nextgens:
@Diman0 I'm happy with it as is, I'd defer the version improvements to when we have socrates' replacement

Diman0 (approved):
LGTM

entrymon:
Any workaround on how to setup own resolver? 

I reported issue on Core DNS with details but would be good to know if any easy workaround meanwhile.

https://github.com/coredns/coredns/issues/5826

nextgens:
@entrymon you are creating issues everywhere but at the right place. Stop it.

Since you are using helm/k8s, it's on https://github.com/Mailu/helm-charts/ that you should report your problems.

What I gather from the situation is:
 - you are running an old version of coredns that doesn't have the fix (you need at least 1.9.4 to have the fix, you are running 1.9.3)
 - you are presumably running an old version of Mailu without the workaround too. None of the various comments you have left indicate which version of Mailu you are attempting to run

entrymon:
@nextgens  thanks for pointing me to helm but I don't see why it is helm chart issue. 

Also I upgraded my mailu and coreDNS both and was still getting the error so I posted it here.

Here is my MailU container image 
```
Containers:
  admin:
    Container ID:   docker://33984491a2832b6449b3f22bef283dffa15b21d8d15d9a54aeb1ed960a855429
    Image:          mailu/admin:1.9.45
    Image ID:       docker-pullable://mailu/admin@sha256:9b182ebcdde3723c15044c73b28f0d6189913645432ca6365d5b15d8ee398109
```

Here is my coreDNS container image 
```
Containers:
  coredns:
    Container ID:  docker://bfd0f1694c92c90c88fbbdd2cfd7ed0533e4f72aad9bfb05880d0e855cee9f84
    Image:         registry.k8s.io/coredns/coredns:v1.9.4
    Image ID:      docker-pullable://registry.k8s.io/coredns/coredns@sha256:b82e294de6be763f73ae71266c8f5466e7e03c69f3a1de96efd570284d35bb18
```

Appreciate your help.

nextgens:
This is a recurring issue that won't be fixed until the helm charts point to fixed versions and do what we do in Mailu: ship and enable unbound by default.

In your particular case, you seem to have several problems:
- running a buggy version of coredns (and a version of Mailu without workaround): I understand that is has now been fixed
- using an upstream DNS resolver that's either equally broken or just doesn't do DNSSEC validation

You can use the script on #2619 to investigate your upstream DNS resolution and identify where you loose the AD flag.

PR #2254, opened 2022-02-19
Send ISRG_X1 on port 25, make DANE pin that

### LINKED ISSUE #2138: Root CA is not included in certificate chain for DANE validation

## Before you open your issue

### Environment
 - [X] docker-compose
 - [ ] kubernetes
 - [ ] docker swarm

### Versions
1.9

## Description
 I just upgraded to 1.9 and am trying to integrate DANE validation.  I got the following DANE Survey Notice:

"Your TLSA record designates a root CA hash, but, as is common, the root CA is not included in your certificate chain.  It would need to be included to work with DANE-TA(2), but simpler to use an intermediate CA hash instead."

## Replication Steps
I set a DNS TLSA record as recommended, with 2 1 1 designation.  I have DNSSEC activated for the relevant domain.  Certificates are issued through Let's Encrypt.

## Expected behaviour
Full validation of the certificate chain.

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Ensure we send ISRG_X1 in the handshake on port 25 (non-interactive, size doesn't really matter).

Update the DANE pin to reflect the change.

I am not sure whether we will need to add --preferred-chain= in the future; This may be the case when letsencrypt decides to use X2/the ECDSA chain

This needs to be tested on a letsencrypt account that isn't mine (I'm opted in for the alternate cert chains)

### Related issue(s)
- closes #2138

## Prerequisites

There's already a towncrier news for it


nextgens:
bors try

core/nginx/letsencrypt.py
```diff
@@ -47,13 +47,24 @@ def format_for_nginx(fullchain, output):
         for cert in certs[:-1] if len(certs)>2 and os.getenv('LETSENCRYPT_SHORTCHAIN', default="False") else certs:
             pem.write(cert)
 
+def add_DANE_pin(chain, output):
+    with open(output, 'w') as pem:
+        with open(chain, 'r') as chain:
+            for line in chain:
+                pem.write(line)
```
ghostwheel42, line 54: No need to do this line by line. 
pem.write(chain.read())
would do the same. 

Is it certain there is a newline at eof? Otherwise appending the second Certificate below could fail.

core/nginx/letsencrypt.py
```diff
@@ -47,13 +47,24 @@ def format_for_nginx(fullchain, output):
...
 
 # Run certbot every day
 while True:
     subprocess.call(command)
     format_for_nginx('/certs/letsencrypt/live/mailu/fullchain.pem', '/certs/letsencrypt/live/mailu/nginx-chain.pem')
+    add_DANE_pin('/certs/letsencrypt/live/mailu/chain.pem', '/certs/letsencrypt/live/mailu/nginx-chain-DANE.pem')
     subprocess.call(command2)
     format_for_nginx('/certs/letsencrypt/live/mailu-ecdsa/fullchain.pem', '/certs/letsencrypt/live/mailu-ecdsa/nginx-chain.pem')
+    add_DANE_pin('/certs/letsencrypt/live/mailu-ecdsa/chain.pem', '/certs/letsencrypt/live/mailu-ecdsa/nginx-chain-DANE.pem')
     time.sleep(86400)
```
ghostwheel42, line 59: How/when is nginx restarted after changing this?

Maybe doing everything nginx related (format cert & concat chain & restart) in certbot's renewed hook would be better.

core/nginx/letsencrypt.py
nextgens: In fact since we're doing SPKI pinning we can use the cross-signed version of the root... so we don't need this at all.

core/nginx/letsencrypt.py
nextgens: It used to be once every hour... now the loop is once a day... so yeah good catch. I've moved the relevant code to the post-hook as suggested.

nextgens:
bors try

core/nginx/config.py
```diff
@@ -34,6 +34,26 @@
...
+    with open(fullchain, 'r') as pem:
+        cert = ''
+        for line in pem:
+            cert += line
+            if '-----END CERTIFICATE-----' in line:
+                certs += [cert]
+                cert = ''
+    with open(output, 'w') as pem:
+        for cert in certs[:-1] if len(certs)>2 and os.getenv('LETSENCRYPT_SHORTCHAIN', default="False") else certs:
+            pem.write(cert)
```
ghostwheel42, line 51: This can be done shorter and the `default="False"` is wrong (the string "False" is actually True here). What about this?

```
def format_for_nginx(fullchain, output):
    """ We may want to strip ISRG Root X1 out """
    if not os.path.exists(fullchain):
        return
    split = '-----END CERTIFICATE-----\n'
    with open(fullchain, 'r') as pem:
        certs = [f'{cert}{split}' for cert in pem.read().split(split) if cert]
    if len(certs)>2 and os.getenv('LETSENCRYPT_SHORTCHAIN'):
        del certs[-1]
    with open(output, 'w') as pem:
        pem.write(''.join(certs))
```

ghostwheel42 (changes requested):

core/nginx/config.py
nextgens: That's in f9869b1d

ghostwheel42 (approved):
lgtm

PR #2262, opened 2022-03-02
Add support for native mail replication

peterpramb:
## What type of PR?

`Feature`

## What does this PR do?

Dovecot includes native support for active/active [mail replication](https://doc.dovecot.org/configuration_manual/replication/) using dsync, an supported alternative as opposed to unreliable network file system usage (NFS, GlusterFS). This change provides basic configuration support for mail replication.

Limitations:
- Excluding specific users from replication is currently not possible. While Dovecot provides the `noreplicate` userdb field for this, it would require both modification of the Mailu database schema and configuration support in the `Admin` UI.
- Only `TCP` replication is supported yet, replication via `SSH` is currently not in scope.
- Replication via `TCPS` (`TLS`) should be possible via override config (not tested), but only on a global level and would therefore affect all enabled Dovecot services (see linked document for further details). For now it is recommended to use `IPsec` & co on the network layer.
- No configuration support for `Setup` has been added so far.
- No tests have been added so far (do we need any?).

Dependencies:
- Mailu/Podop#9

Note that while the provided code works, there are probably better ways to express it.

## Prerequisites

### DIFF
--- a/core/admin/mailu/configuration.py
+++ b/core/admin/mailu/configuration.py
@@ -56,6 +56,8 @@ DEFAULT_CONFIG = {
     'MESSAGE_RATELIMIT': '200/day',
     'MESSAGE_RATELIMIT_EXEMPTION': '',
     'RECIPIENT_DELIMITER': '',
+    'REPLICATION': False,
+    'REPLICATION_TARGET': '',
     # Web settings
     'SITENAME': 'Mailu',
     'WEBSITE': 'https://mailu.io',
@@ -86,6 +88,7 @@ DEFAULT_CONFIG = {
     'HOST_WEBDAV': 'webdav:5232',
     'HOST_REDIS': 'redis',
     'HOST_FRONT': 'front',
+    'HOST_DSYNC': None,
     'SUBNET': '192.168.203.0/24',
     'SUBNET6': None,
     'POD_ADDRESS_RANGE': None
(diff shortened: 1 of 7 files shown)


peterpramb:
Not sure where to properly place this hint:
- At least with `Podman` PMTU discovery does not work from within containers, therefore make sure that the `CNI` MTU fits the underlying transport network between both Mailu nodes, when mail replication is used. Otherwise replication will get stuck and never advance beyond auhentication.

Diman0:
If this would be accepted, this PR needs as well:
- automatic tests for the CI workflow
- More detailed documentation on what this functionality does.  What is it used for?
- Documentation on how to setup the other Mailu instance where the data is replicated to.
- Replication for the config. The config-export and config-import (https://mailu.io/1.9/cli.html#config-export) can be used for  this. Otherwise people will ask why only the mail is replicated and not the config.

ghostwheel42:
Maybe using something like https://github.com/lsyncd/lsyncd would be another way to sync everything (including configuration and fetchmail-state).
Though I have to admit that dsync would be the proper tool for syncing the mail storage...

peterpramb:
This PR is about exposing existing functionality to users, not implementing any configuration synchronization mechanism.

Generally it is assumed that both Mailu instances use the same shared database, but I agree that this should be made clear in the documentation.

peterpramb:
To be honest I've no idea how to properly implement that complex type of automated testing, as it would require:

* Spawning an external database
* Spawning two identially configured Mailu instances (beside hostnames of course)
* Enabling network connectivity between both Dovecot instances
* Inserting one testing mail into one Dovecot instance
* Waiting for replication
* Trying to read the testing mail from the other Dovecot instance

peterpramb:
Anyway, have to close this due to lack of time. Feel free to use it or parts of it for your own needs.

Issue #1750, opened 2021-02-03
podop raise exception on `username@[ip]` address 


jeanlucmongrain:
Using 1.8 master docker images in kubernetes.

Testing if it's an openrelay with http://www.aupads.org/test-relay.html and whenever the address is in format `username@[ip]` it raise exception:

```
Feb 03 10:25:36 mail postfix/smtpd[281512]: warning: socketmap:unix:/tmp/podop.socket:senderaccess socketmap server temporary error: unknown error
Feb 03 10:25:36 mail postfix/smtpd[281512]: warning: socketmap:unix:/tmp/podop.socket:senderaccess lookup error for "spamtest@[x.y.z.w]"
Feb 03 10:25:36 mail postfix/smtpd[281512]: NOQUEUE: reject: RCPT from santuario.pads.ufrj.br[146.164.48.5]: 451 4.3.5 <spamtest@[206.189.154.202]>: Sender address rejected: Server configuration error; from=<spamtest@[x.y.z.w]> to=<relaytest%antispam-ufrj.pads.ufrj.br@> proto=SMTP helo=<antispam-ufrj.pads.ufrj.br>
ERROR:root:Error when processing request
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/podop/postfix.py", line 104, in process_request
    result = await table.get(key)
  File "/usr/lib/python3.8/site-packages/podop/table.py", line 35, in get
    raise Exception(request.status)
Exception: 500
```

KeepMasterBranch:
created in podop in https://github.com/Mailu/Podop/issues/7

ghostwheel42:
Hi,
the problem is the call to `Email.resolve_domain(sender)` in the postfix_sender_access function (line 93).
The domain part of sender contains invalid IDNA characters ("[" and "]") and can't be decoded by idna.encode in IdnaDomain type decorator.
I think the internal backend functions for podop should check all input data _before_ running any operations - the input data is controlled from the ouside (via dovecot/postfix) and could trigger any kind of bad things (what about multiple "@" in the address for example).
Perhaps a simple try/except in postfix_sender_access would be a way to catch this, but I think validation is the better way to go.
Alex

PR #2265, opened 2022-03-05
Prevent an exception from being thrown if the username isn't an email address

### LINKED ISSUE #2261: Question:  Is login possible via port 25
If only port 25 is open to the world, is it possible for a brute force password attack to occur?

I'm seeing messages but someone told me the errors can't be trusted, that there is no actual way to log in via port 25.  But if this were true why would the error message indicate different?

```
2022/03/02 06:04:22 [info] 13#13: *89499 client 31.210.20.189:60509 connected to 0.0.0.0:25
2022/03/02 06:04:22 [error] 13#13: *89499 31.210.20.189 could not be resolved (3: Host not found) while in resolving client address, client: 31.210.20.189, server: 0.0.0.0:25
127.0.0.1 - - [02/Mar/2022:06:04:23 +0000] "GET /auth/email HTTP/1.0" 200 0 "-" "-"
2022/03/02 06:04:23 [info] 13#13: *89499 client login failed: "Authentication credentials invalid" while in http auth state, client: 31.210.20.189, server: 0.0.0.0:25, login: "database"
```

nextgens:
## What type of PR?

enhancement

## What does this PR do?

Mailu expects users to identify using an email address; some brute-force script don't send just an username and this leads to an exception... that makes the error message and return code vary.

This PR prevents the exceptions from being thrown in the first place

```
WARNING in nginx: Invalid user 'xxxx': (builtins.ValueError) invalid email address (no "@")"
```

### Related issue(s)
- closes #2261
- #1750

### DIFF
--- a/core/admin/mailu/internal/nginx.py
+++ b/core/admin/mailu/internal/nginx.py
@@ -94,7 +94,7 @@ def handle_authentication(headers):
         else:
             try:
                 user = models.User.query.get(user_email) if '@' in user_email else None
-                is_valid_user = bool(user)
+                is_valid_user = user is not None
             except sqlalchemy.exc.StatementError as exc:
                 exc = str(exc).split('\n', 1)[0]
                 app.logger.warn(f'Invalid user {user_email!r}: {exc}')
--- a/core/admin/mailu/internal/nginx.py
+++ b/core/admin/mailu/internal/nginx.py
@@ -93,9 +93,8 @@ def handle_authentication(headers):
             app.logger.warn(f'Received undecodable user/password from nginx: {raw_user_email!r}/{raw_password!r}')
         else:
             try:
-                if '@' in user_email:
-                    user = models.User.query.get(user_email)
-                    is_valid_user = True
+                user = models.User.query.get(user_email) if '@' in user_email else None
+                is_valid_user = bool(user)
             except sqlalchemy.exc.StatementError as exc:
                 exc = str(exc).split('\n', 1)[0]
                 app.logger.warn(f'Invalid user {user_email!r}: {exc}')
--- a/core/admin/mailu/internal/nginx.py
+++ b/core/admin/mailu/internal/nginx.py
@@ -93,10 +93,9 @@ def handle_authentication(headers):
             app.logger.warn(f'Received undecodable user/password from nginx: {raw_user_email!r}/{raw_password!r}')
         else:
             try:
-                user = models.User.query.get(user_email)
-                is_valid_user = True
-            except ValueError:
-                pass
+                if '@' in user_email:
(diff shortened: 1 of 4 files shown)


nextgens:
bors try

core/admin/mailu/internal/nginx.py
```diff
@@ -93,8 +93,8 @@ def handle_authentication(headers):
             app.logger.warn(f'Received undecodable user/password from nginx: {raw_user_email!r}/{raw_password!r}')
         else:
             try:
-                user = models.User.query.get(user_email)
-                is_valid_user = True
+                user = models.User.query.get(user_email) if '@' in user_email else None
+                is_valid_user = bool(user)
```
ghostwheel42, line 97: I'd do it this way: `is_valid_user = user is not None`
But I think this is only a matter of taste.

core/admin/mailu/internal/views/postfix.py
```diff
@@ -158,11 +159,13 @@ def postfix_sender_rate(sender):
...
-        return flask.jsonify("REJECT") if models.Domain.query.get(domain_name) else flask.abort(404)
-    else:
-        return flask.abort(404)
+    try:
+        if not is_void_address(sender):
+            localpart, domain_name = models.Email.resolve_domain(sender)
+            return flask.jsonify("REJECT") if models.Domain.query.get(domain_name) else flask.abort(404)
+    except sqlalchemy.exc.StatementError:
+        pass
+    return flask.abort(404)
```
ghostwheel42, line 168: Hmm. I don't think this was correct in the first place.
The function is_void address is really just checking if there's an '@' in sender.
But it strips '<' and '>' - do we need to do this here, too?

Maybe 'is_void address' can be removed and 'postfix_sender_access' should be this

```
if '@' in email:
    if email.startswith('<') and email.endswith('>'):
        email = email[1:-1]
    try:
        localpart, domain_name = models.Email.resolve_domain(sender)
        if models.Domain.query.get(domain_name):
            return flask.jsonify("REJECT")
    except sqlalchemy.exc.StatementError:
        pass
return flask.abort(404)
```

ghostwheel42 (changes requested):

core/admin/mailu/internal/nginx.py
nextgens: d723326b

core/admin/mailu/internal/views/postfix.py
nextgens: cd3eee4c

ghostwheel42 (approved):
lgtm

ghostwheel42 (approved):
now it lgtm ;)

PR #2277, opened 2022-03-11
Update RainLoop application.ini

the-djmaze:
`contacts_autosave` is part of `[defaults]`, not `[plugins]`

### DIFF
--- a/webmails/rainloop/defaults/application.ini
+++ b/webmails/rainloop/defaults/application.ini
@@ -15,5 +15,5 @@ custom_logout_link='/sso/logout'
 enable = On
 allow_sync = On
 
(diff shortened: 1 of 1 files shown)


nextgens (approved):
LGTM.

https://github.com/RainLoop/rainloop-webmail/blob/13d05dc2ff06ce06a689a7e33a5163ecf3b6c27e/rainloop/v/0.0.0/app/libraries/RainLoop/Config/Application.php#L300

the-djmaze:
> LGTM; The default is already "on" though

True, so it could also be removed.

ghostwheel42 (approved):
lgtm

Issue #2215, opened 2022-02-03
Brainstorming for 1.10 / 2.0 roadmap


Diman0:
# Description
Now 1.9 has been released for some time, it is time to discuss what we want to work on for the next release.

- Release date?
- Scope 1.10 (normal release) or 2.0 (release with breaking changes)?  
- What do we want to work on?

## Ideas

Ideas from the last meeting:
- introduce snappymail.
- rootless containers
  - probably requires init container
- replacement of Socrates. ghostwheel42 will work on this (if he finds time).
- A security/keys page for Mailu admin interface. 
  - On this page you can configure/change/generate keys for all your Mail domains.
  - DNS overview 
    - Have some UI that tells you if DNS is correctly configured.
- Have a look at BuildX
  - means we can build for multiple platforms including ARM
  - means we remove the build arm script
  - means CI will be changed massively.
  - For ARM, we could maybe build once per week to make sure the build time for normal builts is not too long.
- autoconfiguration for email clients?
  - automx: https://rseichter.github.io/automx2/?
  - integrate whatever we choose with mailu (part of mailu)?
  - also good time to drop starttls for imap/pop/smtp(sending mail).
    - So only support direct SSL/TLS
    - Could be done via environment_variable. When not configured, then starttls supported. If configured (default for new deployments) then it is disabled.
    - Another idea is to disable starttls, and report a custom error (when you use explicit starttls) that you must switch to implicit SSL/TLS port 465.
- Full 2 factor authentication with xoath2 
  -  Too large in scope for this release. But preparations could be made.
  - Means we need autoconfiguration. Otherwise the email client will not use xoath2.
  - means using single sign on via identity provider (which mailu could be as well). This opens the door to use other identity providers in the future.

Feel free to suggest your own ideas

## Misc
For small non-critical issues/features I suggest we do not put it on the roadmap, but simply offer anyone the chance to pick these up and submit a PR if they want it included.

What are your thoughts? Please share your feedback.
Regardless the above wall of text, feel free to mention any feature/issue you would like included in the next release.

## My 2 cents
For release date we could maybe aim around August/September?
We are all very busy with our personal life now. This should give enough time to implement new features. Of course a decided date is not set in stone. It could still be moved forward/backwards. 

https://github.com/Mailu/Mailu/blob/master/design/mailu-directory-structure.md
This is what I was thinking about for the scope. Changing the directory structure must be part of a major release with breaking changes. Do we want to make this change for this release or postpone this?

[...]

PR #2295, opened 2022-03-22
Switch from Rainloop to SnappyMail

Diman0:
## What type of PR?

Feature

## What does this PR do?
As discussed in the project meeting (#1582), we decided we want to switch from Rainloop to an alternative. Rainloop has multiple open security issues which were not patched for a long time. 

We decided to switch to SnappyMail because it is more secure and based on RainLoop. This means that users using RainLoop will still have a webmail that looks familiar for them.

This PR replaces RainLoop with SnappyMail.

### Related issue(s)
- #2215 
- #1582

## Prerequisites


Diman0:
bors try

Diman0:
bors try

ShadwDrgn:
Would love to see this merged because rainloop and roundcube are awful but snappymail is wonderful and actually has hotkeys :D sorry if this didn't contribute much.

tete2soja:
In addition, a critic issue is publicly available and the Rainloop team didn't respond to it:

* https://github.com/RainLoop/rainloop-webmail/issues/2142
* https://blog.sonarsource.com/rainloop-emails-at-risk-due-to-code-flaw

Can be a new reason to switch to SnappyMail.

webmails/snappymail/config/nginx-snappymail.conf
```diff
@@ -2,33 +2,42 @@ server {
...
     client_max_body_size {{ MESSAGE_SIZE_LIMIT|int + 8388608 }};
 
     location / {
-        try_files $uri /index.php?$query_string;
+        try_files $uri $uri/ /index.php$args;
     }
 
     location ~ \.php$ {
+        include /etc/nginx/fastcgi_params;
+
```
nextgens, line 26: I think you need a try_files here... or a way to check that the file actually exists.

See https://www.nginx.com/resources/wiki/start/topics/examples/phpfcgi/#connecting-nginx-to-php-fpm or https://nginx.org/en/docs/http/ngx_http_core_module.html#try_files (with two different locations depending on whether it does or doesn't exist)

core/nginx/conf/nginx.conf
```diff
@@ -190,6 +190,9 @@ http {
       {% endif %}
       {% if ADMIN == 'true' %}
        location {{ WEB_ADMIN }} {
+         {% if WEB_ADMIN != '/' %}
+         rewrite ^({{ WEB_ADMIN }})$ $1/ permanent;
```
nextgens, line 194: This seems unrelated... why is it required?

webmails/snappymail/Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+ENV SNAPPYMAIL_URL https://github.com/the-djmaze/snappymail/releases/download/v2.13.4/snappymail-2.13.4.zip
+
+RUN apk add --no-cache \
+      curl unzip \
+ && cd /var/www/webmail \
+ && curl -L -O ${SNAPPYMAIL_URL} \
+ && unzip -q *.zip \
+ && rm -f *.zip \
+ && find . -type d -exec chmod 755 {} \; \
+ && find . -type f -exec chmod 644 {} \; \
```
nextgens, line 75: The two lines could be simplified with: ``chmod -R u+w,a+rX /var/www/webmail``

webmails/snappymail/Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+
+# Install Snappymail from source
+ENV SNAPPYMAIL_URL https://github.com/the-djmaze/snappymail/releases/download/v2.13.4/snappymail-2.13.4.zip
+
+RUN apk add --no-cache \
+      curl unzip \
+ && cd /var/www/webmail \
+ && curl -L -O ${SNAPPYMAIL_URL} \
+ && unzip -q *.zip \
+ && rm -f *.zip \
```
nextgens, line 73: I'd pipe the curl output into unzip here

nextgens (changes requested):

webmails/snappymail/config/nginx-snappymail.conf
```diff
@@ -38,4 +47,9 @@ server {
     location ^~ /data {
       deny all;
     }
+
+    location /health {
+        add_header Content-Type text/plain;
+        return 200;
```
nextgens, line 53: I wonder if it's worth making it a PHP script to ensure that FPM is fine too.

nextgens (changes requested):

webmails/snappymail/config/nginx-snappymail.conf
ghostwheel42: This would be good. You don't need a script though, you can use `ping.path` and `ping.response` of the fpm daemon.

webmails/snappymail/config/nginx-snappymail.conf
Diman0: Good catch. I will change this.

core/nginx/conf/nginx.conf
Diman0: I don't remember. I think I had some issues with my reverse proxy and added this for testing purposes. I will remove it. It should not have been added.

webmails/snappymail/config/nginx-snappymail.conf
Diman0: Yes. With this we test both nginx and fpm-daemon. I will enable this and modify the health check in the Dockerfile.

Diman0:
> What are the plans for having this? Any ETA in plan?

This week I will try finishing this PR. Assuming the reviewers have no new findings, it should be part of master within a week.

Diman0:
bors try

webmails/snappymail/config/nginx-snappymail.conf
```diff
@@ -0,0 +1,62 @@
...
+        fastcgi_pass unix:/var/run/php7-fpm.sock;
+        fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
+        {% if WEB_WEBMAIL == '/' %}
+        fastcgi_param SCRIPT_NAME $fastcgi_script_name;
+        {% else %}
+        fastcgi_param SCRIPT_NAME {{WEB_WEBMAIL}}/$fastcgi_script_name;
+        {% endif %}
+    }
+
+    location ~ /\.ht {
```
nextgens, line 45: I'd rather we denied access to any hidden file here

nextgens (approved):
LGTM.

webmails/snappymail/config/nginx-snappymail.conf
nextgens: My understanding is that it shouldn't: in your example the dot is a parameter (after the '?').
What would be blocked is http://example.com/webmail/.example

Diman0: I have made the propose change

nextgens (approved):

ghostwheel42 (approved):
LGTM.

I was urprised the WEBMAIL variable in mailu.env is in fact only a bool: the only tests against it are == or != "none". So WEBMAIL=yesplease will work, too.

PR #2310, opened 2022-04-04
Update deprecated rspamd config option

henniaufmrenni:
## What type of PR?

Configuration update

## What does this PR do?

This is just a small config update to get rid of the following warning message:
`lua; antivirus.lua:109: CLAM_VIRUS [clamav]: Using attachments_only is deprecated. Please use scan_mime_parts = true instead`

As per the rspamd documentation https://rspamd.com/doc/modules/antivirus.html
> attachments_only = true; # Before 1.8.1
> scan_mime_parts = true; # After 1.8.1

The currently used version of rspamd is 3.1.

## Prerequisites

### DIFF
--- a/core/rspamd/conf/antivirus.conf
+++ b/core/rspamd/conf/antivirus.conf
@@ -1,6 +1,6 @@
 {% if ANTIVIRUS == 'clamav' %}
 clamav {
-  attachments_only = true;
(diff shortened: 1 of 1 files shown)


nextgens (approved):

Diman0 (approved):
lgtm

PR #2313, opened 2022-04-07
Update alpine-linux to 3.14.6 - Busybox security fix CVE-2022-28391

willofr:
## What type of PR?

Security fix

## What does this PR do?
Update Dockerfiles to use alpine-linux 3.14.6 which contains a security fix for Busybox
https://security.alpinelinux.org/vuln/CVE-2022-28391

## Prerequisites


nextgens:
Where do you think CVE-2022-28391 applies?

As far as I know we don't provide shell access in any of the containers... and I am fairly confident that we do not use netstat anywhere.

willofr:
I agree that netstat doesn't seem to be used in mailu, but, keeping a known vulnerability on a system is __never__ a good idea, especially RCE and EoP type vulns.
Can we say that it will never be possible to get a shell in a container by exploiting another vulnerability in Nginx, Postfix, Dovecot and so on? probably not.

nextgens:
This is a RCE in a component we don't use.

If another vulnerability leads to a shell being spawned, it will be with the same privileges (no EoP). Currently most daemons run as root in Mailu; if you want to improve upon the status-quo, start there ;)

willofr:
>Currently most daemons run as root in Mailu; if you want to improve upon the status-quo, start there ;)

+1, i missed this point

PR #2322, opened 2022-04-16
Remove Deprecated TLS Versions

GregTCLTK:
As outlined in [RFC-7525](http://www.rfc-editor.org/info/rfc7525), TLS 1.0 and TLS 1.1 are deprecated and should no longer be used. Users who still need TLS 1.0 and TLS 1.1 support for legacy devices can change the nginx config file and add the needed version(s). However, neither TLS 1.0 nor TLS 1.1 should be enabled by default.


nextgens:
This part of the configuration is specific to port 25: there any cryptography is better than no cryptography. Modern clients will support TLS_FALLBACK_SCSV (https://www.rfc-editor.org/rfc/rfc7507.html) that will prevent downgrade attacks.

PR #2323, opened 2022-04-16
Fix Postfix FileExistsError on startup

Pumba98:
## What type of PR?

bug-fix

## What does this PR do?

I'm running mailu with the mailu helm-chart on kubernetes. Sometimes when a Pod restarts I get the following error during startup:
```
Process Process-1:
File "/usr/lib/python3.9/multiprocessing/process.py", line 315, in _bootstrap
self.run()
File "/usr/lib/python3.9/multiprocessing/process.py", line 108, in run
self._target(*self._args, **self._kwargs)
File "/start.py", line 18, in start_podop
os.mkdir('/dev/shm/postfix',mode=0o700)
FileExistsError: [Errno 17] File exists: '/dev/shm/postfix'
INFO:MAIN:MTA-STS daemon starting...
```

But that does not prevent the container startup. When mails arrive it will fail with something like:

```
postfix/trivial-rewrite[94979]: warning: connect to /tmp/podop.socket: No such file or directory
postfix/trivial-rewrite[94979]: warning: table socketmap:unix:/tmp/podop.socket:transport lookup error: No such file or directory
postfix/trivial-rewrite[94979]: warning: socketmap:unix:/tmp/podop.socket:transport lookup error for "*"
```

I'm running this quick fix now since almost two months without problems. Maybe you got a better approach how to solve this, but this works fine for me.

### Related issue(s)
- none

## Prerequisites

### DIFF
--- a/core/postfix/start.py
+++ b/core/postfix/start.py
@@ -15,7 +15,8 @@ log.basicConfig(stream=sys.stderr, level=os.environ.get("LOG_LEVEL", "WARNING"))
 
 def start_podop():
     os.setuid(getpwnam('postfix').pw_uid)
-    os.mkdir('/dev/shm/postfix',mode=0o700)
+    if not os.path.exists('/dev/shm/postfix'):
+        os.mkdir('/dev/shm/postfix',mode=0o700)
     url = "http://" + os.environ["ADMIN_ADDRESS"] + "/internal/postfix/"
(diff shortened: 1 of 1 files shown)


nextgens (approved):
LGTM. Good catch, this is indeed a problem.

nextgens (approved):

ghostwheel42 (approved):

Issue #2293, opened 2022-03-22
[RFC] IPv6 support


nextgens:
The current ipv6 support in Mailu is broken (and has been for a very long time: see #2260, #2272 and #1789). This ticket aims to discuss what the final end-state should be.

Architecturally the choices are:
1. do ipv4 exclusively internally (get rid of SUBNET6)
  - pro: simple
  - con: won't make outbound ipv6 connections (partial ipv6 support)
2. dual-stack (what we are attempting to do but what has been broken for a very long time)
  - con: complicated, breaks silently and requires NAT
3. do ipv6 when requested internally (SUBNET or SUBNET6, not both)
  - con: requires NAT

When I say requires NAT, I mean we currently assign ULAs to SUBNET6 to prevent open-relays (in setup, not sure what helm does)... so any outbound connection will be NATed and may not have the right PTR record. Keep in mind that it's the smtp/postfix container that makes outbound connections "outside" from its own address (unless it's NATed).

nextgens:
@Niduroki @peterpramb @ghostwheel42 @Diman0 let's discuss it here.

I've discussed it with @Diman0 on #mailu-dev and the current view is that we are fine with 1.9 having broken v6 support and if we improve it it will be on master, once a consensus has emerged.

Niduroki:
I think a perfect solution would be able to work without NAT at all – NAT4, NAT64, or NAT66 – not all cloud runtimes support NAT66 from what I've gathered (notably podman, I guess? – I don't think Mailu supports other cloud runtimes?).

Meaning, that SUBNET could be a publicly routable IP-range (IPv4 *or* IPv6), and that internal access (authentication-less, that is) is only possible, if the sender-address is in `SUBNET`/`SUBNET6`.
(Since TCP handshakes make sure that the sender address isn't spoofed, this should be sound, too - at least when thinking about it for a minute)
Makes distributed, horizontally scaled, setups way more easy, too.

That's in a perfect world, at least - we'd definitely need to make sure, that every connection is TLS-encrypted in this scenario, too, and I'm not too certain, if that is a possibility in some of the backend connections, plus any misconfiguration would cause open-relays even in the IPv4 scenario.

SMTP rDNS/PTR being problematic could be fixed, by making sure, that the SMTP container gets assigned a fixed IP-address, when working without NAT. Same thing really with front, too, and its related `MX/A/AAAA` records

-----

Back to reality, though.

[...]

PR #2272, opened 2022-03-09
[WIP] Improved IPv6 support

Niduroki:
## What type of PR?

Enhancement, bug-fix, documentation

## What does this PR do?

Fix #2260

## Prerequisites
Before we can consider review and merge, please make sure the following list is done and checked.
If an entry in not applicable, you can check it or remove it from the list.

- [X] In case of feature or enhancement: documentation updated accordingly
- [ ] Unless it's docs or a minor change: add [changelog](https://mailu.io/master/contributors/workflow.html#changelog) entry file.

--------

Not yet completely tested, as I was a bit careless, and have run into letsencrypt certificate limits.

I have also ran into problems with the current master containers not accepting e-mails:  
`client login failed: "AUTH not supported" while in http auth state`  
Can somebody reproduce this perhaps?

Considering the problem ↑ seems to happen in the current master, too, rather than just my images, `AUTH not supported` doesn't sound really DNS/ipv4/ipv6 related, and otherwise everything seems to work when using IPv4 (which was one of the concerns, causing `ipv6=off` to be used, instead of making gunicorn listen on `[::]`) I decided to propose this PR as a WIP already.

If someone else could test the missing items for my, I'd be grateful, otherwise, expect the rest of the checks to be done sometime next week, when letsencrypt has lifted my limits.

----

* [ ] IPv4-only default setup:
  * [ ] Receive E-Mail via webmail
  * [X] Send E-Mail via webmail
  * [X] Some services:
     * [X] admin
     * [ ] rspamd in admin (accessible, but not thoroughly checked yet, thus no ☑️, yet)
     * [X] rainloop webmail (checking for SSO)
  * [ ] IPv6 delivered E-Mails are normally received

<details>
<summary>docker-compose.yml</summary>

```yml
version: '2.2'

services:

  # External dependencies
  redis:
    image: redis:alpine
    restart: always
    volumes:
      - "/home/mailu/redis:/data"
    depends_on:
      - resolver
    dns:
      - 192.168.203.254

  # Core services
  front:
    image: niduroki/mailu-nginx:master
    restart: always
    env_file: mailu.env
    logging:
      driver: json-file
    ports:
      - "[removed-ipv4]:80:80"
      - "[removed-ipv4]:443:443"
      - "[removed-ipv4]:25:25"
      - "[removed-ipv4]:465:465"
      - "[removed-ipv4]:587:587"
      - "[removed-ipv4]:110:110"
      - "[removed-ipv4]:995:995"
[...]

PR #2325, opened 2022-04-18
postfix: wrap IPv6 CIDRs in square brackets for RELAYNETS

pommi:
## What type of PR?

bug-fix

## What does this PR do?

This PR wraps IPv6 CIDRs in the `RELAYNETS` environment variable in square brackets for the postfix configuration.

The `RELAYNETS` environment variable is used for configuring both postfix `mynetworks` and rspamd `local_networks`. Postfix requires IPv6 addresses to be wrapped in square brackets (eg. `[2001:db8::]/64`).

When an IPv6 address is not wrapped in square brackets in the postfix configuration for `mynetworks` it results in this error while processing an incoming email from an IPv6 sender:
```
postfix/smtpd[340]: warning: 2001:db8::/64 is unavailable. unsupported dictionary type: 2001
postfix/smtpd[340]: warning: smtpd_client_event_limit_exceptions: 2001:db8::/64: table lookup problem
```

The sender sees an error and the incoming email is refused:
```
451 4.3.0 <unknown[2001:xxx:xxx:xxx:xxx:xxx:xxx:xxx]>: Temporary lookup failure
```

I tried to work around this issue by wrapping the IPv6 CIDR in square brackets in the `RELAYNETS` environment variable, but it segfaults rspamd, because it can't deal with this non-standard IPv6 notation used by postfix:
```
kernel: [4305632.603704] rspamd[1954299]: segfault at 0 ip 00007fb848983871 sp 00007ffe02cc6d1
8 error 4 in ld-musl-x86_64.so.1[7fb848948000+48000]
```

### Related issue(s)
- #2293
- #2272

## Prerequisites

### DIFF
--- a/core/postfix/start.py
+++ b/core/postfix/start.py
@@ -6,6 +6,7 @@ import shutil
 import multiprocessing
 import logging as log
 import sys
+import re
 
 from podop import run_server
 from pwd import getpwnam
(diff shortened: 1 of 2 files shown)


nextgens:
Good catch.

I'm usually reluctant to use the regexp engine for things like this... but here it does make things a lot easier.

Please add a newsfragment in towncrier

pommi:
Thanks @nextgens. Newsfragment added.

nextgens (approved):
LGTM

ghostwheel42 (approved):
lgtm

PR #2328, opened 2022-04-26
Feature: Configurable default spam threshold used for new users

enginefeeder101:
## What type of PR?

Feature

## What does this PR do?

This PR adds functionality to set a custom default spam threshold
for new users. The environment variable ``DEFAULT_SPAM_THRESHOLD`` is
used for this purpose. When not set, it defaults back to 80%, as the
default value was before.

If ``DEFAULT_SPAM_THRESHOLD`` is set to a value that Python cannot
parse as an integer, a ValueError is thrown. There is no error handling
for that case built-in. Should that be done?

## Prerequisites

### DIFF
--- a/docs/antispam.rst
+++ b/docs/antispam.rst
@@ -43,6 +43,8 @@ Rspamd rejects non-compliant email messages and email messages that contain viru
       
    66% (10/15) is less than 80%, so the email is classified as ham. This email message will go to the inbox folder. If the user wants email messages with a score of 10 (66%) to be classified as spam, then the user defined spam filter tolerance can be lowered to 65% in the administration web interface.
    
+   The default spam filter tolerance used for new users can be configured using the environment variable ``DEFAULT_SPAM_THRESHOLD``. See also: :ref:`_advanced_cfg`.
+
 .. image:: assets/screenshots/SpamFiltering.png
 
 The location in the administration web interface where the spam filter and spam filter tolerance can be configured.
--- a/docs/antispam.rst
+++ b/docs/antispam.rst
@@ -21,29 +21,29 @@ Rspamd rejects non-compliant email messages and email messages that contain viru
 1. When an email message is received or send by Postfix, it is scanned by Rspamd. If the email message receives a spam score between 6 and 15, a header is added to the email message with the spam score. This is an example of an email header for a spam score of 14::
 
     X-Spamd-Bar: ++++++++++++++
-    X-Spam-Level: ************** 
+    X-Spam-Level: **************
     Authentication-Results: test.mailu.io;
        dkim=pass header.d=example.com header.s=mailing header.b=ABCDE;
        dkim=pass header.d=example.com header.s=mailing header.b=ABCDE;
        dmarc=pass (policy=none) header.from=eventim.de;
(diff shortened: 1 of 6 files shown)


Diman0 (changes requested):
Thank you for submitting this PR. This PR is worthwhile. I can imagine that administrators want the possibility for setting a default threshold. But the PR requires some changes. We normally assign all environment variables to a global dictionary (app.config). This dictionary is used in other parts of Mailu.

So don't use the env var directly in models.py. Add the value to the global config.
In [configuration.py](https://github.com/Mailu/Mailu/blob/master/core/admin/mailu/configuration.py) add DEFAULT_SPAM_THRESHOLD to the default config (line 7+):
`'DEFAULT_SPAM_THRESHOLD': 80,`

In Class ConfigManager, in function init_app (see line 136+), you can add the code for assigning the value to the config:
`self.config['DEFAULT_SPAM_THRESHOLD'] = int(self.config['DEFAULT_SPAM_THRESHOLD'])`
This would be the ideal location to add error handling. It is not required. We don't have error handling for any of the other mailu.env environment variables.

Finally in models.py on line 511 you can use now:
`spam_threshold = db.Column(db.Integer, nullable=False,
        default=int(app.config['DEFAULT_SPAM_THRESHOLD]))`

The documentation changes looks fine.

Diman0:
At a glance it looks good to me. I will test and review the PR this week.

bors try

Diman0:
bors retry

docs/antispam.rst
```diff
@@ -43,6 +43,8 @@ Rspamd rejects non-compliant email messages and email messages that contain viru
       
    66% (10/15) is less than 80%, so the email is classified as ham. This email message will go to the inbox folder. If the user wants email messages with a score of 10 (66%) to be classified as spam, then the user defined spam filter tolerance can be lowered to 65% in the administration web interface.
    
+   The default spam filter tolerance used for new users can be configured using the environment variable ``DEFAULT_SPAM_THRESHOLD``. See also: :ref:`_advanced_cfg`.
```
Diman0, line 46: The build fails with the following error for the documentation:
`/docs/antispam.rst:46:undefined label: _advanced_cfg`

When referring to a bookmark you should not include the leading _ of the bookmark. 
Try changing ```:ref:`_advanced_cfg`.``` to ```:ref:`advanced_cfg` ```.

In the future I also recommend to try building and testing the images manually before submitting a PR or pushing commits to a PR. How to is described here:
https://mailu.io/1.9/contributors/environment.html#building-images
This way you can see beforehand that the build fails..

Diman0 (changes requested):
See in-line comment

Diman0 (dismissed):
See in-line comment

Diman0:
Everything should be fixed. Let's run the test suite

bors try

Diman0:
bors try

Diman0:
bors retry

Diman0:
bors retry

Diman0 (dismissed):
lgtm

Diman0:
bors try

Diman0 (approved):
lgtm. I tested the functionality works as expected

ghostwheel42 (approved):
Lgtm

PR #2331, opened 2022-04-29
Add data/fetchmail to mkdir command

### LINKED ISSUE #2314: small issue in https://setup.mailu.io/1.9/
Hello.
When wizard propose create folders it miss add folder for fetchmail, please add  mkdir -p  .... {.....,data/fetchmail}

ghostwheel42:
## What type of PR?

documentation

### Related issue(s)

closes #2314


ghostwheel42:
> Is it worth backporting it?

I don't think so. It's only in the docker stack documentation. Which IMO should be deprecated in the near future.
The docker-compose setup did create the data path without manual intervention AFAIK.

PR #2338, opened 2022-05-08
Update X-XSS-Protection to current recommendation

AvverbioPronome:
See:

- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection and
- https://cheatsheetseries.owasp.org/cheatsheets/HTTP_Headers_Cheat_Sheet.html#x-xss-protection

## What type of PR?

Slight enhancement

## What does this PR do?

This PR turns off the XSS auditor in the few browsers that still have one.

## Prerequisites

- [ ?] In case of feature or enhancement: documentation updated accordingly
- [x ] Unless it's docs or a minor change: add [changelog](https://mailu.io/master/contributors/workflow.html#changelog) entry file.

### DIFF
--- a/core/nginx/conf/nginx.conf
+++ b/core/nginx/conf/nginx.conf
@@ -117,7 +117,7 @@ http {
       add_header X-Frame-Options 'SAMEORIGIN';
       add_header X-Content-Type-Options 'nosniff';
       add_header X-Permitted-Cross-Domain-Policies 'none';
-      add_header X-XSS-Protection '1; mode=block';
+      add_header X-XSS-Protection '0';
       add_header Referrer-Policy 'same-origin';
 
       # mozilla autoconfiguration
--- a/core/nginx/conf/nginx.conf
+++ b/core/nginx/conf/nginx.conf
@@ -117,7 +117,6 @@ http {
       add_header X-Frame-Options 'SAMEORIGIN';
       add_header X-Content-Type-Options 'nosniff';
       add_header X-Permitted-Cross-Domain-Policies 'none';
-      add_header X-XSS-Protection '0';
       add_header Referrer-Policy 'same-origin';
 
       # mozilla autoconfiguration
--- a/towncrier/newsfragments/2338.misc
+++ b/towncrier/newsfragments/2338.misc
@@ -1,0 +1,1 @@
+Don't send the `X-XSS-Protection` http header anymore. 


ghostwheel42:
Hi! https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection states
`This means that if you do not need to support legacy browsers, it is recommended that you use [Content-Security-Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy) without allowing unsafe-inline scripts instead.`

I don't see Mailu implementing a `Content-Security-Policy` - I think disabling `X-XSS-Protection` should be accompanied by enabling a `Content-Security-Policy`, shouldn't it?

AvverbioPronome:
`Content-Security-Policy` should be enabled regardless. But CSP is in the proxied app domain, not for the reverse proxy to decide.

nextgens:
Let's remove the header (aka: don't set it explicitly) instead

AvverbioPronome:
Modern browsers ignore the header, whatever we set it to.

Old browsers will do whatever they want and might have different defaults. But they should mostly try and sanitize the page by themselves (that's what `1` without `mode=block` means), and that is exploitable by an attacker.

nextgens:
There's little we can do about old browsers being exploitable; I'm sure that you are aware of things such as https://github.com/Metnew/uxss-db

The question is whether there is still value in setting X-XSS-Protection explicitly.

What kind of vulnerability are we exposed to if we leave things as is? Are we talking about mutated XSS? memory corruption bugs? something else?

If the rationale is mutated XSS, I do believe that we should drop the header (on the basis that this is the long term solution in any case) and accept that old browsers are broken (on the basis that universal XSSes exist, are a lot easier to exploit -as they are generic- and won't be mitigated by implementing the proposed fix either).

AvverbioPronome:
We have two realistic alternatives:

- dropping the header: this means we don't care about old browsers -- and it could be fine, if it's an explicit decision.
- setting it to `0`: this means slightly enhancing the security of older browsers.

Both are fine. But `1`, with or without block, is a problem, and checkers like hardenize.com complain about it: they are happy if you set `0` and I haven't checked what happens if you omit the header altogether.

nextgens:
I vote for removing the header altogether. If you update your branch and create a towncrier entry I'll approve the PR

AvverbioPronome:
Ok, I shall do it tomorrow.

nextgens (approved):
LGTM

Diman0 (approved):
lgtm

PR #2357, opened 2022-05-24
Switch to ffdhe3072 to enable RFC 7919

nextgens:
## What type of PR?

enhancement

## What does this PR do?

The idea being:
- it's a "nothing up my sleeves" group
- it may help shave off some bytes of the SSL handshake; That being said, I doubt that clients that are modern enough to support this RFC won't offer an EC kex

https://raw.githubusercontent.com/internetstandards/dhe_groups/master/ffdhe3072.pem

## Prerequisites


nextgens:
The tradeoff (and why it wasn't done this way in the first place) is that using an RFC standardized group makes it a prime target for global adversaries (think intel agencies) as pre-computation is possible. There is a good argument for diversifying the groups in use.

Some related reading:
https://mta.openssl.org/pipermail/openssl-users/2019-March/010126.html
https://hg.nginx.org/nginx/rev/1aa9650a8154

I'm not sure we should merge this.

Diman0 (approved):
approved. We will create new PR documentation how to replace this dhparam with your own generated parameters.

PR #2360, opened 2022-05-28
roundcube: disable apache2 access log

pommi:
## What type of PR?

bug-fix

## What does this PR do?

It disables the access log of apache2 in the roundcube webmail container. Requests are already logged by the front container. The requests logged in the roundcube container contained contained the wrong client IP: the IP address of the front container.

----

Original PR:

~~Roundcube webmail is accessed through the nginx reverse proxy in the front container. Each access logline logged by apache2 in the roundcube container did not contain the actual client IP address, but the IP address of the front container, for example:~~

```
192.168.203.3 - - [28/May/2022:12:33:52 +0000] "POST /?_task=mail&_action=refresh HTTP/1.1" 200 677 "https://[REDACTED]/roundcube/?_task=mail&_mbox=INBOX" "Mozilla/5.0 (X11; Linux x86_64; rv:100.0) Gecko/20100101 Firefox/100.0"
^
IP address of the front container
```

~~By enabling the apache2 remoteip module and configuring it to get the actual client IP address from the X-Forwarded-For header, it logs the correct client IP address to the access log.~~

### Related issue(s)
- None

## Prerequisites

### DIFF
--- a/towncrier/newsfragments/2360.bugfix
+++ b/towncrier/newsfragments/2360.bugfix
@@ -1,0 +1,1 @@
+roundcube: log actual client ip by using apache2 remoteip
--- a/towncrier/newsfragments/2360.bugfix
+++ b/towncrier/newsfragments/2360.bugfix
@@ -1,1 +1,0 @@
-roundcube: log actual client ip by using apache2 remoteip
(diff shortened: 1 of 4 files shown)


ghostwheel42:
Hi. Good idea to use mod_remoteip here.
Wouldn't it be possible to disable the access log completely?
The front container should already log all requests, shouldn't it?
roundcube AFAIK handles the X-Forwarded-headers.

pommi:
@ghostwheel42 Here you go :slightly_smiling_face: If you're ok with it, I'm keeping the original commit in the PR for reference.

Diman0:
bors try

Diman0:
looks like a network error caused the build to fail

bors try

Diman0 (approved):
It looks good to me. I can see that  in the roundcube container the access log is disabled.

ghostwheel42 (approved):
LGTM

PR #2364, opened 2022-05-30
Update Dockerfile

twekkel:
apt is intended for interactive usage, for scripts use apt-get (https://manpages.debian.org/bullseye/apt/apt.8.en.html) to avoid warnings.

## What type of PR?

minor enhancement

## What does this PR do?

replace apt with apt-get to avoid below warning

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

## Prerequisites

-

PR #2367, opened 2022-06-05
Roundcube: add button to taskmenu for mailu

ItzYanick:
## What type of PR?

enhancement

## What does this PR do?

Adds a button to the roundcube interface. This button gets you back to the admin interface. Now you can switch between both without any issues

### Related issue(s)
- Did not found any but was bothering me

## Prerequisites

### DIFF
--- a/webmails/roundcube/mailu.js
+++ b/webmails/roundcube/mailu.js
@@ -1,0 +1,9 @@
+rcmail.addEventListener('init', function(evt) {
+      var btn = document.createElement('a');
+      btn.setAttribute("href", "/admin");
+      btn.setAttribute("role", "button");
+      btn.setAttribute("class", "showurl");
+      btn.innerHTML = '<span class="inner">To Mailu</span>';
+
+      document.getElementById("taskmenu").appendChild(btn);
+});
--- a/webmails/roundcube/mailu.js
+++ b/webmails/roundcube/mailu.js
@@ -1,6 +1,6 @@
 rcmail.addEventListener('init', function(evt) {
       var btn = document.createElement('a');
-      btn.setAttribute("href", "/admin");
+      btn.setAttribute("href", "{{ WEB_ADMIN }}");
       btn.setAttribute("role", "button");
       btn.setAttribute("class", "showurl");
       btn.innerHTML = '<span class="inner">To Mailu</span>';
--- a/webmails/roundcube/mailu.js
+++ b/webmails/roundcube/mailu.js
@@ -1,9 +1,11 @@
-rcmail.addEventListener('init', function(evt) {
-      var btn = document.createElement('a');
-      btn.setAttribute("href", "{{ WEB_ADMIN }}");
-      btn.setAttribute("role", "button");
-      btn.setAttribute("class", "showurl");
-      btn.innerHTML = '<span class="inner">To Mailu</span>';
+if ({{ ADMIN }}) {
+      rcmail.addEventListener('init', function(evt) {
+            var btn = document.createElement('a');
+            btn.setAttribute("href", "{{ WEB_ADMIN }}");
+            btn.setAttribute("role", "button");
+            btn.setAttribute("class", "showurl");
+            btn.innerHTML = '<span class="inner">To {{ SITENAME }}</span>';
 
-      document.getElementById("taskmenu").appendChild(btn);
-});
+            document.getElementById("taskmenu").appendChild(btn);
+      });
+}
--- a/towncrier/newsfragments/2367.feature
+++ b/towncrier/newsfragments/2367.feature
@@ -1,0 +1,1 @@
+Adds a button to the roundcube interface that gets you back to the admin interface
--- a/webmails/roundcube/Dockerfile
+++ b/webmails/roundcube/Dockerfile
@@ -63,6 +63,7 @@ RUN set -eu \
 # enable database_attachments (and memcache?)
 
 COPY mailu.php /var/www/html/plugins/mailu/mailu.php
+COPY mailu.js /var/www/html/plugins/mailu/mailu.js
 COPY php.ini /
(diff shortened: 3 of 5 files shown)


ItzYanick:
I do not know if any of the Prerequisites is needed for such a small change so I will just leave them unticked for now. Can change later if needed tho.

ItzYanick:
Build failed because of docs? My changes could not have done this. #2364 seems to have the same problem

webmails/roundcube/mailu.js
```diff
@@ -0,0 +1,9 @@
+rcmail.addEventListener('init', function(evt) {
+      var btn = document.createElement('a');
+      btn.setAttribute("href", "/admin");
```
Diman0, line 3: You make the assumption that admin is always available on endpoint /admin. This is incorrect. It is defined in environment variable WEB_ADMIN. The path should be derived from this env var.

Diman0 (changes requested):
See in-line comment. 
Even though it is a small change, please add a (changelog)[https://mailu.io/master/contributors/workflow.html#changelog] entry.

Diman0:
bors try

ItzYanick:
added changelog

ItzYanick:
bors try

Diman0:
bors try

webmails/roundcube/mailu.js
```diff
@@ -0,0 +1,9 @@
+rcmail.addEventListener('init', function(evt) {
+      var btn = document.createElement('a');
+      btn.setAttribute("href", "{{ WEB_ADMIN }}");
+      btn.setAttribute("role", "button");
+      btn.setAttribute("class", "showurl");
+      btn.innerHTML = '<span class="inner">To Mailu</span>';
```
Diman0, line 6: Please change this to
btn.innerHTML = '<span class="inner">To {{ SITENAME }}</span>';

Not everyone uses Mailu as the main site name. In SITENAME in mailu.env you can configure the name for the site. By default this is Mailu, but that is not the case for all users. Some users do not want Mailu branding at all.

Diman0 (changes requested):
The PR itself works fine. 
See in-line comment. I did not think before about this. Not all users use Mailu as sitename. The value of SITENAME should be used in case someone does not use the Mailu branding.

There is also another topic. Multiple (or maybe even many) users have admin NOT exposed to the internet. For all these users we are introducing a broken 'back to the admin page' button.

I'm not certain if we should make this button optional. It would mean adding YET another setting to mailu.env. We have already to many settings here. So should all these users deal with it? They have a broken back to mailu button. Or do we add yet another config setting to mailu.env? I will ask the opinion of other people in the mailu dev channel and will come back to you.

Diman0:
`There is also another topic. Multiple (or maybe even many) users have admin NOT exposed to the internet. For all these users we are introducing a broken 'back to the admin page' button..`

Ignore this. If the administrator blocked access, then it is the administrators problem. 
The default in Mailu is that admin is reachable. If admin is enabled, then the button should be enabled.  if admin is disabled, then the button should not be shown. The mailu.env setting is `ADMIN` for this. It is TRUE when admin is enabled, and FALSE when admin is disabled.

That means there should still be an IF statement around the button which checks for {{ ADMIN  }}.

ghostwheel42:
> Ignore this. If the administrator blocked access, then it is the administrators problem.

Adding `if ({{ ADMIN }})` in front of the `rcmail.addEventListener()` in mailu.js would hide the button, wouldn't it?

Diman0:
> > Ignore this. If the administrator blocked access, then it is the administrators problem.
> 
> Adding `if ({{ ADMIN }})` in front of the `rcmail.addEventListener()` in mailu.js would hide the button, wouldn't it?

It would hide the button when admin is disabled. So in the setup.mailu.io you disabled admin access. In this case it write ADMIN=FALSE to the mailu.env file.
ADMIN is true in mailu.env when you enabled Admin in setup.mailu.io. 

The situation I was talking about (the ignore this part), is when users block access to admin. E.g. make it only reachable via the local network and not via the external (WAN) network.

ghostwheel42:
> The situation I was talking about (the ignore this part), is when users block access to admin. E.g. make it only reachable via the local network and not via the external (WAN) network.

Ah, okay. The button would only work on the "local network" then.
Admins can deceide themselves what to do when accessing the admin url from outside (display a message, redirect, etc.).

Diman0:
It looks good to me. But I will first run some tests with admin disabled/enabled and a non-standard site name.
There are always many different scenario's to consider. Even with a small PR like this :grin: 

bors try

Diman0 (approved):
Tests went fine. LGTM. Thank you for your contribution.

ghostwheel42:
> Tests went fine. LGTM. Thank you for your contribution.

It does what it should, but I think "To SITENAME" is not really the correct description for the button and below "settings" it's not the correct place. Also the button is only showing up in the Elastic skin.

Adding a section "Server" to the list of settings seems to be a better place.

I'm not against adding this, but I think using the section and opening the admin interface in a new tab (or don't open webmail in a new tab when coming from admin) would be a better way.

ItzYanick:
Maybe the "To Admin" would be alright then because it also shows "Sign in Admin" at the login screen as well.

I don't want to have it in the Settings because I want a quick and easy way to switch. In the Admin Panel you can also click on "Webmail" to quickly get there so why not in the webmail screen as well. I have to re-investigate that if there is a plugin API of Roundcube that allows me adding buttons without messing with the UI. If there is not, I could offer to add the button on the left side *only* in the Elastic skin and in the other skins maybe in the settings.

I also think that the Admin Panel should open the webmail in the same tab instead of a own. This way we have a way more clean approach between both panels. But that would be a different Pull Request.

ghostwheel42:
I've replaced this by #2591

PR #2382, opened 2022-06-27
Handle IPv6 resolvers correctly

TheEnbyperor:
## What type of PR?
Bug fix

## What does this PR do?
NGINX front container currently does not handle IPv6 resolver addresses correctly, this PR fixes this.

## Prerequisites

### DIFF
--- a/core/nginx/config.py
+++ b/core/nginx/config.py
@@ -2,6 +2,7 @@
 
 import os
 import logging as log
+import ipaddress
 import sys
 from socrate import system, conf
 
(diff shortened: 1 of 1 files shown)


nextgens:
Can you provide an example of resolv.conf that the current code fails to parse please? I don't understand what the proposed change does.

TheEnbyperor:
With a `resolv.conf` such as
```
nameserver 2a0d:1a40:7900::1
```
which results in the following in the `nginx.conf` file
```
resolver 2a0d:1a40:7900::1 valid=30s;
```

NGINX tries to parse this as a hostname an port, the correct syntax is as follows
```
resolver [2a0d:1a40:7900::1] valid=30s;
```

nextgens:
Okay now I understand what this does. Let's try to do with without any additional module; what about something like this instead?

```
resolver = content[content.index("nameserver") + 1]
args['RESOLVER'] = '[{}]'.format(resolver) if ':' in resolver else resolver
```

Can you please create a towncrier/newsfragment entry to describe the change too?

ghostwheel42:
Will be fixed via  #2421

PR #2405, opened 2022-08-04
Remove support for TLSv1 and TLSv1.1 in STARTTLS on port 25

johanatandromeda:
## What type of PR?

enhancement

## What does this PR do?

Removes TLSv1 and TLSv1.1 support for STARTTLS on port 25

### Related issue(s)
Did not create an issue

## Prerequisites


nextgens:
Sorry but no. Any encryption beats no encryption, this is why we have a special configuration for port 25.

Modern clients have downgrade protection anyway.

PR #2409, opened 2022-08-05
Fix error in reverse proxy example for how to include nginx override.

### LINKED ISSUE #2232: nginx.conf manual override not working as in docs
### Environment
 - [x ] docker-compose
 - [ ] kubernetes
 - [ ] docker swarm

Mailu 1.9

## Description

In Override Mailu configuration https://mailu.io/master/reverse.html#override-mailu-configuration it is mentioned that in case someone wants to change the nginx config, they should just do a volume override like this.
`  volumes:
    - "$ROOT/certs:/certs"
    - "$ROOT/nginx.conf:/etc/nginx/nginx.conf"
`

I tried this, but seems like the actual 
core/nginx/config.py

is always run at container startup and will override it again.

## Replication Steps
Mount nginx.conf in docker compose
 - "$ROOT/nginx.conf:/etc/nginx/nginx.conf"
and then edit the file. Restart the container.

## Expected behaviour
Expected from logs would be the possibility to keep the local changes.

## Workaround
I am now building my own docker container and modify the file before, but that is much more cumbersome than the volume override.

Thanks for the great work.

Diman0:
## What type of PR?

documentation

## What does this PR do?
The reverse proxy documentation contained an error in an example section. The nginx override file is incorrectly overridden in the sample. Also clarified how the file is included and that you should define location blocks (because it is included in the main mailu server block).

### Related issue(s)
- closes #2232 

## Prerequisites

- [n/a] Unless it's docs or a minor change: add [changelog](https://mailu.io/master/contributors/workflow.html#changelog) entry file.

### DIFF
--- a/docs/reverse.rst
+++ b/docs/reverse.rst
@@ -243,25 +243,10 @@ Refer to the Traefik documentation for more details.
 Override Mailu configuration
 ----------------------------
 
-If you do not have the resources for running a separate reverse proxy, you could override Mailu reverse proxy configuration by using a Docker volume.
-Simply store your configuration file (Nginx format), in ``/mailu/overrides/nginx.conf`` for instance.
-All *.conf files will be included in the main server block of Mailu in nginx which listens on port 80/443.
+If you do not have the resources for running a separate reverse proxy, you could override Mailu reverse proxy configuration by using :ref:`an override<override-label>`.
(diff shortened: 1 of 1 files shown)


ghostwheel42 (approved):
Lgtm

Diman0:
this time it will build

bors try

nextgens (approved):

PR #2410, opened 2022-08-06
The MESSAGE_SIZE_LIMIT is now used for the clamav configuration

### LINKED ISSUE #1646: make ClamAV respect MESSAGE_SIZE_LIMIT
Currently ClamAV has hardcoded file size limits:
https://github.com/Mailu/Mailu/blob/master/optional/clamav/conf/clamd.conf#L44-L45

I think we should adjust these limits to also use MESSAGE_SIZE_LIMIT, as otherwise you have the confusing situation that big emails are allowed, but big attachments are getting blocked.

Diman0:
## What type of PR?

bug-fix

## What does this PR do?
The MESSAGE_SIZE_LIMIT is now used for the clamav configuration (MaxScanSize, MaxFileSize).

This resolves the situation where MESSAGE_SIZE_LIMIT was set higher than the hard-coded values in the clamav configuration.

For MaxScanSize we used a hardcoded value that was 3 times higher than the default value. I kept this in place.
I have set MaxFileSize a little bit higher than MESSAGE_SIZE_LIMIT. This makes sure emails can always be scanned (assuming they don't have too large archives as attachments).

### Related issue(s)
- closes #1646

## Prerequisites

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -37,7 +37,7 @@ The ``POSTMASTER`` is the local part of the postmaster email address. It is
 recommended to setup a generic value and later configure a mail alias for that
 address.
 
-The ``WILDCARD_SENDERS`` setting is a comma delimited list of user email addresses 
+The ``WILDCARD_SENDERS`` setting is a comma delimited list of user email addresses
 that are allowed to send emails from any existing address (spoofing the sender).
 
 The ``AUTH_RATELIMIT_IP`` (default: 60/hour) holds a security setting for fighting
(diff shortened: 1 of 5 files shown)


Diman0:
Discussed during the last PR triage meeting.
Rather than MESSAGE_SIZE_LIMIT determining the clamav config, it should be a separate setting. For security purposes this setting should not be to high.

If possible introduce an override for this. If that is not possible we will introduce two new mailu.env variables.

Diman0:
There is no include functionality for the clamd.conf file. That means it does not fit with the mailu/overrides folder. 
It should not be too easy to change these settings (MaxScanSize, MaxFileSize). Setting them to high makes your vulnerable for email bombs (attachments with huge files in a small compressed archive).

People who know what they do, can simply create an host mount for directly changing the clamd config file

PR #2412, opened 2022-08-07
Feature: sender relay

joniw:
## What type of PR?

Feature

## What does this PR do?

Supports setting up [sender dependent relayhost maps](https://www.postfix.org/postconf.5.html#sender_dependent_relayhost_maps) for users.

The code adds a menu item for users which allows to add relay host entries for this user. Every added account allows the user to send mails from the relay mail address. The mail is relayed over the relay server with the relay mail login.

When an account is saved, an actual login is tried against the server to prevent the use of invalid mail accounts.

### Related issue(s)

None

## Prerequisites

### DIFF
--- a/core/admin/mailu/internal/views/postfix.py
+++ b/core/admin/mailu/internal/views/postfix.py
@@ -180,3 +180,46 @@ def idna_encode(addresses):
         for (localpart, domain) in
         (address.rsplit("@", 1) for address in addresses)
     ]
+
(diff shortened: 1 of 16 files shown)


Diman0:
This PR does not align with the main goals of the project. For this reason we will not merge it. In the future please first create an issue to get feedback. Another option is to get in touch in the matrix mailu-dev channel.

PR #2413, opened 2022-08-14
bump to alpinelinux 3.14.7 - OpenSSL security fix

willofr:
## What type of PR?

Security fix

## What does this PR do?
update alpinelinux to 3.14.7 for an OpenSSL security fix: https://alpinelinux.org/posts/Alpine-3.13.11-3.14.7-3.15.5-released.html

## Prerequisites


nextgens:
We're not affected by either of the CVEs that are fixed in this release

PR #2425, opened 2022-08-20
Hotfix for workflow. For build step do not build from cache.

Diman0:
Hotfix for workflow. For build step do not build from cache.
Make sure cache layers are unique by using ${{ github.ref }}

The build job uses the cache-from. This is not required. Due to non-unique cache keys it also resulted the build step building from cache. 
As a precaution the ${{ github.ref }} and ${{ github.run_id }} are appended to the cache key for the build cache layers. This should make sure the cache keys are unique among workflow runs

## What type of PR?

bug-fix

### Related issue(s)
n/a

## Prerequisites

- [n/a] In case of feature or enhancement: documentation updated accordingly
- [n/a ] Unless it's docs or a minor change: add [changelog](https://mailu.io/master/contributors/workflow.html#changelog) entry file.

### DIFF
--- a/.github/workflows/build_test_deploy.yml
+++ b/.github/workflows/build_test_deploy.yml
@@ -134,8 +134,7 @@ jobs:
           load: false
           push: false
           set: |
-            *.cache-from=type=gha,scope=build-${{ matrix.target }}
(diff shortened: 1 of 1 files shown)


nextgens (approved):
LGTM

PR #2426, opened 2022-08-22
Update docs to mention .inc.php for roundcube

DannyDaemonic:
My recent patch updated the roundcube overrides to use .inc.php vs .inc (as is done in roundcube and as suggested by roundcube plugin docs). I updated the overrides section in the docs but a page that links to the overrides section still had the old ".inc" file name. This fixes that oversight. Sorry about that.

## What type of PR?

documentation

## What does this PR do?

Updates configuration.rst to also reflect the new extension.

## Prerequisites
This patch neither adds features nor requires a towncrier.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -332,7 +332,7 @@ If ``ROUNDCUBE_PLUGINS`` is not set the following plugins are enabled by default
 
 To disable all plugins just set ``ROUNDCUBE_PLUGINS`` to ``mailu``.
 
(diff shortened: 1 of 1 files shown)


DannyDaemonic:
At some point I may be up for writing an separate document just covering overrides. It would be good to have all override information in one place along with some examples, rather than just a quick mention in the faq that gets linked to from elsewhere.

nextgens (approved):

Diman0 (approved):

Diman0:
bors retry

Diman0:
bors retry

PR #2430, opened 2022-08-27
Prevent signups for accounts where an alias exists

nextgens:
## What type of PR?

Feature

## What does this PR do?

This prevent signups for accounts where an SQL like alias exists; we already do it for non-SQL like aliases

## Prerequisites

- [] In case of feature or enhancement: documentation updated accordingly


Diman0:
1. Created domain google.com and enabled self-registration.
2. Added alias with sql-like enabled for alias 'ex%' and another with sa_ple.
3. Add alias with specific alias 'word'
4. Tried to self-register the accounts example@google.com, sample@google.com and word@google.com. All attempts received the error that the email was already in use. It is not possible to use an email address already in use by an alias.

PR #2440, opened 2022-08-31
The ARM wheels don't work

### LINKED ISSUE #2439: using piwheels.org python wheels in arm images fails at runtime
Specifically for @Diman0 

As described in https://github.com/Mailu/Mailu/issues/2383#issuecomment-1230552614 and in https://github.com/Mailu/Mailu/issues/1973#issuecomment-1232678585 the master-arm versions of postfix and admin (and maybe others?) images are using the wheels from piwheels.org in an attempt to drastically cut the build time for arm.
Unfortunately, this does not work at runtime as said in https://github.com/piwheels/packages/issues/312 (piwheels does support only Rasbian, not alpine)
The previous versions (namely diman/xxx:1.11-arm) are working well.

see eg in postfix Dockerfile:
`/bin/sh -c pip install --no-cache-dir --extra-index-url=https://www.piwheels.org/simple --only-binary=:all: postfix-mta-sts-resolver==1.0.1 || (apk add --no-cache --virtual .build-deps gcc musl-dev python3-dev py3-wheel libffi-dev   && pip3 install postfix-mta-sts-resolver==1.0.1   && apk del .build-deps )`

or in admin Dockerfile:
`/bin/sh -c set -eu  && apk add --no-cache libressl curl postgresql-libs mariadb-connector-c  && pip install --no-cache-dir --extra-index-url=https://www.piwheels.org/simple -r requirements.txt --only-binary=:all: --no-binary=Flask-bootstrap,PyYAML,SQLAlchemy  || ( apk add --no-cache --virtual build-dep libressl-dev libffi-dev python3-dev build-base postgresql-dev mariadb-connector-c-dev cargo  && pip install --upgrade pip  && pip install -r requirements.txt  && apk del --no-cache build-dep )` 

The result is that admin cannot start at all (see https://github.com/Mailu/Mailu/issues/2383#issuecomment-1230495806), and postfix starts but not the mta-sts daemon (see https://github.com/Mailu/Mailu/issues/1973#issuecomment-1232670764)

This also means that we definitely need to have:
1. some automated tests before publishing arm images. I understand github does not provide arm environments for testing but if needed, I think we can use local raspberry pies (eg with https://github.com/nektos/act or https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners) 
2. "fixed" tags allowing to come back to previous versions (only master-arm is currently published and replaced by new versions, so no fallback possible to a previous version)

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Remove piwheels to ensure we always rebuild on ARM

### Related issue(s)
- closes #2439
- #1200

### DIFF
--- a/core/admin/Dockerfile
+++ b/core/admin/Dockerfile
@@ -37,7 +37,7 @@ WORKDIR /app
 COPY requirements-prod.txt requirements.txt
 RUN set -eu \
  && apk add --no-cache libressl curl postgresql-libs mariadb-connector-c \
- && pip install --no-cache-dir --extra-index-url=https://www.piwheels.org/simple -r requirements.txt --only-binary=:all: --no-binary=Flask-bootstrap,PyYAML,SQLAlchemy \
+ && pip install --no-cache-dir -r requirements.txt --only-binary=:all: --no-binary=Flask-bootstrap,PyYAML,SQLAlchemy \
  || ( apk add --no-cache --virtual build-dep libressl-dev libffi-dev python3-dev build-base postgresql-dev mariadb-connector-c-dev cargo \
  && pip install --upgrade pip \
  && pip install -r requirements.txt \
--- a/core/postfix/Dockerfile
+++ b/core/postfix/Dockerfile
@@ -22,7 +22,7 @@ RUN apk add --no-cache --virtual .build-deps gcc musl-dev python3-dev \
 
 # Image specific layers under this line
 # Building pycares from source requires py3-wheel and libffi-dev packages
-RUN pip install --no-cache-dir --extra-index-url=https://www.piwheels.org/simple --only-binary=:all: postfix-mta-sts-resolver==1.0.1 || (apk add --no-cache --virtual .build-deps gcc musl-dev python3-dev py3-wheel libffi-dev \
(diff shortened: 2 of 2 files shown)


Diman0 (changes requested):
Is this valid? If I check the docs (https://pip.pypa.io/en/stable/cli/pip_install/):
```
--only-binary <format_control>

    Do not use source packages. Can be supplied multiple times, and each time adds to the existing value. 
    Accepts either “:all:” to disable all source packages, “:none:” to empty the set, or one or more package
    names with commas between them. 
    Packages without binary distributions will fail to install when this option is used on them.
```

You are currently using ```--only-binary=:all: --no-binary=Flask-bootstrap,PyYAML,SQLAlchemy```.
The first parameter specifies that ALL packages will not be built from source. If no binary x64/arm binary package can be found, then pip will fail. Then you provide a list of 3 packages that must always be built from source.

I'd say we remove `--only-binary:all`` and replace this with ``--prefer-binary``. The latter means that binary wheels will be used IF they can be found. If they are not found, then we will still build from source.

Diman0:
Nvm. I missed the OR ( || ) operator. If it fails, then the right part of the OR ( || ) operator will build the packages.

Diman0 (approved):
lgtm

Issue #2442, opened 2022-09-01
rspamc doesn't seem to correctly teach HAM / spam filtering almost doesn't work


woj-tek:
(this is somewhat related to #1438 but it's closed)

## Before you open your issue
- [x] Check if no issue or pull-request for this already exists.
- [x] Check [documentation](https://mailu.io/master/) and [FAQ](https://mailu.io/master/faq.html). (Tip, use the search function on the documentation page)
- [x] You understand `Mailu` is made by volunteers in their **free time** — be conscise, civil and accept that delays can occur.
- [x] The title of the issue should be short and simple. It should contain specific terms related to the actual issue. Be specific while writing the title.

### Environment
 - [ ] docker-compose
 - [x] kubernetes
 - [ ] docker swarm

### Versions
(managed by fluxcd)
```
mailu-prod-mailu                      	0.3.3   	False    	True 	pulled 'mailu' chart with version '0.3.3'
```

Images (output of, `kubectl -n mailu-prod get pods -o jsonpath="{.items[*].spec.containers[*].image}"`, based on https://kubernetes.io/docs/tasks/access-application-cluster/list-all-running-container-images/... seems wrong though):
```
mailu/admin:1.9.26
mailu/clamav:1.9.26
mailu/dovecot:1.9.26
mailu/nginx:1.9.26
mailu/nginx:1.9.26
mailu/nginx:1.9.26
mailu/postfix:1.9.26
redis:5-alpine
mailu/roundcube:1.9.26
mailu/rspamd:1.9.26
```
As hashes (`kubectl get pods -o jsonpath="{.items[*].status.containerStatuses[*].imageID}" | tr -s '[[:space:]]' '\n' | sort | uniq -c`0:
```
      1 docker.io/library/redis@sha256:1c465cac273d3e0a63ca9896cb78eb75bad9c80e9a0b3ea0b5a0c6d1f12c3232
      1 docker.io/mailu/admin@sha256:321e84b2361a65d02aa0c47ddf63b7f048ee701f0b75a0069cda7e1757c84f96
      1 docker.io/mailu/clamav@sha256:d7a5a568fa858cd57a2c5305e172d40a8a32567216042d1a2be95332133490f2
      1 docker.io/mailu/dovecot@sha256:afe7dced133bd8152862bde2db29ff4eb60c67f25b3c9261df26e6af856acdd3
[...]

PR #2446, opened 2022-09-05
Update antispam.rst

woj-tek:
## What type of PR?

documentation

## What does this PR do?

Change hostname of rspamd (`antispam` wasn't the one available in our k8s deployment) and add rspamc call to teach bayes filter with ham/spam.

### Related issue(s)
- https://github.com/Mailu/Mailu/issues/2442

## Prerequisites

- [-] In case of feature or enhancement: documentation updated accordingly
- [-] Unless it's docs or a minor change: add [changelog](https://mailu.io/master/contributors/workflow.html#changelog) entry file.

### DIFF
--- a/docs/antispam.rst
+++ b/docs/antispam.rst
@@ -58,15 +58,17 @@ If you already have an existing mailbox and want Mailu to learn them all as ham 
 
 .. code-block:: bash
 
-  rspamc -h antispam:11334 -P mailu -f 13 fuzzy_add /mail/user\@example.com/.Ham_Learn/cur/
-
-This should learn every file located in the ``Ham_Learn`` folder from user@example.com 
+  rspamc -h mailu-rspamd:11334 -P mailu -f 13 fuzzy_add /mail/user\@example.com/.Ham_Learn/cur/
+  rspamc -h mailu-rspamd:11334 -P mailu learn_ham /mail/user\@example.com/.Ham_Learn/cur/
+  
+This should learn every file located in the ``Ham_Learn`` folder from user@example.com (first line adds mail to fuzzy module, second one to bayes one)
 
 Likewise, to learn all messages within the folder ``Spam_Learn`` as spam messages :
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


nextgens:
I am sorry but this isn't mergeable.

Please follow instructions from https://mailu.io/master/contributors/workflow.html#pr-workflow

1) It needs to be targeting master (and not 1.9)
2) the hostname is the right one, if it doesn't match on k8s, it's a problem that should be reported on https://github.com/Mailu/helm-charts/

woj-tek:
> It needs to be targeting master (and not 1.9)

My bad - I clicked on "edit on GitHub" just after navigating to https://mailu.io/, which defaults to stable 1.9. I'll create propper one

> the hostname is the right one, if it doesn't match on k8s, it's a problem that should be reported on https://github.com/Mailu/helm-charts/

I digged a little bit more and it seems it's "by design" - Deployment and Service name is configured to be prefixed with configured mailu name (vide https://github.com/Mailu/helm-charts/blob/eccf0342c81609c9b770205eb12c02683396f93a/mailu/templates/rspamd.yaml#L12-L12 and https://github.com/Mailu/helm-charts/blob/eccf0342c81609c9b770205eb12c02683396f93a/mailu/templates/rspamd.yaml#L164-L164)

Keeping that in mind, would it be acceptable to mention in the documentation as a sidenote/comment, that in case of k8s deployment hostname may be different and it's recommendable to check it using `kubectl` command, and that by default it should be `mailu-rspamd`?

nextgens:
I really think that it should be changed on the helm side of things...

Isn't it the hostname parameter that needs changing?
https://github.com/Mailu/helm-charts/blob/eccf0342c81609c9b770205eb12c02683396f93a/mailu/templates/rspamd.yaml#L37

IMHO it makes no sense for it to be different from what we use in docker-compose... if it is, it means that migrations from one to the other will result in lost history of the rspamd scans (as per https://github.com/Mailu/helm-charts/issues/95)

woj-tek:
Done: https://github.com/Mailu/helm-charts/issues/189

PR #2458, opened 2022-09-26
Fix: Don't update updated_at on quota_bytes_used change

### LINKED ISSUE #1363: Strange behavior of user updated_at field
I have mailboxes with updated_at date of today. After some testing i found that this field is updated once a mailbox is receiving a new mail. Can somebody confirm this?

I use the default sqlite backend.
EDIT:

seems like /internal/dovecot/quota/* endpoint is updating the user entry and refreshing the changed_at date for that row.

DjVinnii:
## What type of PR?

bug-fix

## What does this PR do?

This PR makes sure that the `updated_at` field is not updated when `quota_bytes_used` is updated. All other updates to the `User` model still updates the `updated_at` field. 

This is done by explicitly using an method in the `Base` class triggering [`flag_modified`][url-flag-modified].

### Related issue(s)
- closes #1363

## Prerequisites

[url-flag-modified]: https://docs.sqlalchemy.org/en/14/orm/session_api.html#sqlalchemy.orm.attributes.flag_modified

### DIFF
--- a/core/admin/mailu/internal/views/dovecot.py
+++ b/core/admin/mailu/internal/views/dovecot.py
@@ -33,7 +33,7 @@ def dovecot_quota(ns, user_email):
     user = models.User.query.get(user_email) or flask.abort(404)
     if ns == "storage":
         user.quota_bytes_used = flask.request.get_json()
-        user.flag_updated_at_as_modified()
+        user.dont_change_updated_at()
         models.db.session.commit()
     return flask.jsonify(None)
 
--- a/core/admin/mailu/internal/views/dovecot.py
+++ b/core/admin/mailu/internal/views/dovecot.py
@@ -33,6 +33,7 @@ def dovecot_quota(ns, user_email):
     user = models.User.query.get(user_email) or flask.abort(404)
     if ns == "storage":
         user.quota_bytes_used = flask.request.get_json()
+        user.flag_updated_at_as_modified()
         models.db.session.commit()
     return flask.jsonify(None)
 
--- a/core/admin/mailu/internal/views/fetch.py
+++ b/core/admin/mailu/internal/views/fetch.py
@@ -27,6 +27,7 @@ def fetch_done(fetch_id):
     fetch = models.Fetch.query.get(fetch_id) or flask.abort(404)
     fetch.last_check = datetime.datetime.now()
     fetch.error_message = str(flask.request.get_json())
+    fetch.dont_change_updated_at()
     models.db.session.add(fetch)
     models.db.session.commit()
     return ""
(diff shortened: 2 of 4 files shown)


ghostwheel42:
Hi and thanks for your contribution.
I think we should explicitely flag "updated_at" as modified in dovecot_quota() (and also in fetch_done() for last_check and error_message). This can be done by adding a method to the Base class and calling it explicitely.
The way you did it will not update the "updated_at" timestamp even if other attributes are changed (for example when importing a configuration).

DjVinnii:
Thanks for the feedback @ghostwheel42. It's true that `updated_at` won't be updated if other fields are changed in combination with `quota_bytes_used`. I had a quick look if there are multiple places where `quota_bytes_used` is being updated (mainly in combination with other fields), but it looks like it is only updated in `dovecot_quota()`. If configuration is being imported, doesn't this configuration have the `updated_at` field populated an thus be updated nevertheless? 

However, I agree it might be better to add a method in the Base class and call it explicitly. This probably removes a bit of overhead, because not every model change is being checked.

DjVinnii:
@ghostwheel42 I created a method in the `Base` class that can be used to flag `updated_at` as modified. If needed, this can be changed to mark any field as modified by a given parameter. But I didn't see any use case for this at the moment.

You also talked about implementing this in `fetch_done()`. I'm not using fetching myself, so I wasn't quite sure how to test this. Especially if `flag_modified()` can be triggered before `models.db.session.add(fetch)` or after.

ghostwheel42:
I'm sorry for the delay. Somehow lost track of this.
This looks good, but could you please rename the method to its purpose (and not to what it does):
Instead of `flag_updated_at_as_modified` it could be `dont_touch_updated_at` or sth like this.
Yould you also call this in this function: https://github.com/Mailu/Mailu/blob/ff4d3f4d375b020867a737fe5e44d3ae14e41705/core/admin/mailu/internal/views/fetch.py#L26

DjVinnii:
> I'm sorry for the delay. Somehow lost track of this. 

No worries. I'm well aware that Mailu is maintained by volunteers.

> This looks good, but could you please rename the method to its purpose (and not to what it does): Instead of `flag_updated_at_as_modified` it could be `dont_touch_updated_at` or sth like this. Yould you also call this in this function:
> 
> https://github.com/Mailu/Mailu/blob/ff4d3f4d375b020867a737fe5e44d3ae14e41705/core/admin/mailu/internal/views/fetch.py#L26

Sure!  I renamed the method to `dont_change_updated_at`. Also calling the method in `fetch_done()` is added now.

ghostwheel42:
bors try

ghostwheel42 (commented):
LGTM

nextgens (approved):
LGTM

Diman0 (approved):
Lgtm

PR #2464, opened 2022-10-05
Introduce RESTful API

### LINKED ISSUE #445: RESTful API
A RESTful API would allow third party applications to interact with Mailu in a defined way.

* Creating/editing/listing domains, users, aliases etc.
* Authentication via API token
* Implementation as a separate, optional docker container

Colud be based on the OpenAPI-Specification
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md
with Swagger and Python Flask server
https://editor.swagger.io/

Diman0:
## What type of PR?

Feature

## What does this PR do?
Introduces a RESTful API for changing the complete Mailu config.
Anything that can be configured in the web administration interface, can also be configured via the Mailu RESTful API.

Via the swagger.json endpoint the complete OpenAPI specification can be retrieved.
Via the endpoint swaggerui, a web client is available which shows all the endpoints, data models and allows you to submit requests.

See docs/api.rst and docs/configuration.rst for details for enabling it.

### Related issue(s)
- closes #445 

## Prerequisites

### DIFF
--- a/core/admin/mailu/__init__.py
+++ b/core/admin/mailu/__init__.py
@@ -70,10 +70,12 @@ def create_app_from_config(config):
         return utils.flask_babel.format_datetime(value) if value else ''
 
     # Import views
-    from mailu import ui, internal, sso
+    from mailu import ui, internal, sso, api
     app.register_blueprint(ui.ui, url_prefix=app.config['WEB_ADMIN'])
     app.register_blueprint(internal.internal, url_prefix='/internal')
     app.register_blueprint(sso.sso, url_prefix='/sso')
+    if app.config.get('API_TOKEN'):
+        api.register(app, web_api_root=app.config.get('WEB_API'))
     return app
 
 
--- a/core/admin/mailu/api/__init__.py
+++ b/core/admin/mailu/api/__init__.py
@@ -1,0 +1,24 @@
+from flask import redirect, url_for, Blueprint
+from flask_restx import apidoc
+from . import v1 as APIv1
+
+def register(app, web_api_root):
+
+    APIv1.app = app
+    # register api bluprint(s)
+    apidoc.apidoc.url_prefix = f'{web_api_root}/v{int(APIv1.VERSION)}'
+    APIv1.api_token = app.config['API_TOKEN']
+    app.register_blueprint(APIv1.blueprint, url_prefix=f'{web_api_root}/v{int(APIv1.VERSION)}')
+
+    # add redirect to current api version
+    redirect_api = Blueprint('redirect_api', __name__)
+    @redirect_api.route('/')
+    def redir():
+        return redirect(url_for(f'{APIv1.blueprint.name}.root'))
+    app.register_blueprint(redirect_api, url_prefix=f'{web_api_root}')
+
+    # swagger ui config
+    app.config.SWAGGER_UI_DOC_EXPANSION = 'list'
+    app.config.SWAGGER_UI_OPERATION_ID = True
+    app.config.SWAGGER_UI_REQUEST_DURATION = True
+    app.config.RESTX_MASK_SWAGGER = False
--- a/core/admin/mailu/api/common.py
+++ b/core/admin/mailu/api/common.py
@@ -1,0 +1,42 @@
+from .. import models, utils
+from . import v1
+from flask import request
+import flask
+import hmac
+from functools import wraps
+from flask_restx import abort
+from sqlalchemy.sql.expression import label
+
+def fqdn_in_use(name):
+    d = models.db.session.query(label('name', models.Domain.name))
+    a = models.db.session.query(label('name', models.Alternative.name))
+    r = models.db.session.query(label('name', models.Relay.name))
+    u = d.union_all(a).union_all(r).filter_by(name=name)
+    if models.db.session.query(u.exists()).scalar():
+        return True
+    return False
+
+""" Decorator for validating api token for authentication """
+def api_token_authorization(func):
+    @wraps(func)
+    def decorated_function(*args, **kwds):
+        client_ip = flask.request.headers.get('X-Real-IP', flask.request.remote_addr)
+        if utils.limiter.should_rate_limit_ip(client_ip):
+            abort(429, 'Too many attempts from your IP (rate-limit)' )
+        if not request.headers.get('Authorization'):
+            abort(401, 'A valid Bearer token is expected which is provided as request header')
+        #Client provides 'Authentication: Bearer <token>'
+        if (' ' in request.headers.get('Authorization')
+                and not hmac.compare_digest(request.headers.get('Authorization'), 'Bearer ' + v1.api_token)):
+            utils.limiter.rate_limit_ip(client_ip)
+            flask.current_app.logger.warn(f'Invalid API token provided by {client_ip}.')
+            abort(403, 'A valid Bearer token is expected which is provided as request header')
+        #Client provides 'Authentication: <token>'
+        elif (' ' not in request.headers.get('Authorization')
+                and not hmac.compare_digest(request.headers.get('Authorization'), v1.api_token)):
+            utils.limiter.rate_limit_ip(client_ip)
+            flask.current_app.logger.warn(f'Invalid API token provided by {client_ip}.')
+            abort(403, 'A valid Bearer token is expected which is provided as request header')
+        flask.current_app.logger.info(f'Valid API token provided by {client_ip}.')
+        return func(*args, **kwds)
+    return decorated_function
--- a/core/admin/mailu/api/v1/__init__.py
+++ b/core/admin/mailu/api/v1/__init__.py
@@ -1,0 +1,43 @@
+from flask import Blueprint
+from flask_restx import Api, fields
+
+
+VERSION = 1.0
+api_token = None
+
+blueprint = Blueprint(f'api_v{int(VERSION)}', __name__)
+
+authorization = {
+    'Bearer': {
+        'type': 'apiKey',
+        'in': 'header',
+        'name': 'Authorization'
+    }
+}
+
+api = Api(
+    blueprint, version=f'{VERSION:.1f}',
+    title='Mailu API', default_label='Mailu',
+    validate=True,
+    authorizations=authorization,
+    security='Bearer',
+    doc='/'
+)
+
+response_fields = api.model('Response', {
+    'code': fields.Integer,
+    'message': fields.String,
+})
+
+error_fields = api.model('Error', {
+    'errors': fields.Nested(api.model('Error_Key', {
+        'key': fields.String,
+        'message':fields.String
+    })),
+    'message': fields.String,
+})
+
+from . import domains
+from . import alias
+from . import relay
+from . import user
--- a/core/admin/mailu/api/v1/alias.py
+++ b/core/admin/mailu/api/v1/alias.py
@@ -1,0 +1,127 @@
+from flask_restx import Resource, fields, marshal
+from . import api, response_fields
+from .. import common
+from ... import models
+
+db = models.db
+
+alias = api.namespace('alias', description='Alias operations')
+
+alias_fields_update = alias.model('AliasUpdate', {
+    'comment': fields.String(description='a comment'),
+    'destination': fields.List(fields.String(description='alias email address', example='user@example.com')),
+    'wildcard': fields.Boolean(description='enable SQL Like wildcard syntax')
+})
+
+alias_fields = alias.inherit('Alias',alias_fields_update, {
+  'email': fields.String(description='the alias email address', example='user@example.com', required=True),
+  'destination': fields.List(fields.String(description='alias email address', example='user@example.com', required=True)),
+
+})
+
+
+@alias.route('')
+class Aliases(Resource):
+    @alias.doc('list_alias')
+    @alias.marshal_with(alias_fields, as_list=True, skip_none=True, mask=None)
+    @alias.doc(security='Bearer')
+    @common.api_token_authorization
+    def get(self):
+        """ List aliases """
+        return models.Alias.query.all()
+
+    @alias.doc('create_alias')
+    @alias.expect(alias_fields)
+    @alias.response(200, 'Success', response_fields)
+    @alias.response(400, 'Input validation exception', response_fields)
+    @alias.response(409, 'Duplicate alias', response_fields)
+    @alias.doc(security='Bearer')
+    @common.api_token_authorization
+    def post(self):
+        """ Create a new alias """
+        data = api.payload
+
+        alias_found = models.Alias.query.filter_by(email = data['email']).first()
+        if alias_found:
+          return { 'code': 409, 'message': f'Duplicate alias {data["email"]}'}, 409
+
+        alias_model = models.Alias(email=data["email"],destination=data['destination'])
+        if 'comment' in data:
+          alias_model.comment = data['comment']
+        if 'wildcard' in data:
+          alias_model.wildcard = data['wildcard']
+        db.session.add(alias_model)
+        db.session.commit()
+
+        return {'code': 200, 'message': f'Alias {data["email"]} to destination {data["destination"]} has been created'}, 200
+
+@alias.route('/<string:alias>')
+class Alias(Resource):
+    @alias.doc('find_alias')
+    @alias.response(200, 'Success', alias_fields)
+    @alias.response(404, 'Alias not found', response_fields)
+    @alias.doc(security='Bearer')
+    @common.api_token_authorization
+    def get(self, alias):
+        """ Find alias """
+        alias_found = models.Alias.query.filter_by(email = alias).first()
+        if alias_found is None:
+          return { 'code': 404, 'message': f'Alias {alias} cannot be found'}, 404
+        else:
+          return marshal(alias_found,alias_fields), 200
+
+    @alias.doc('update_alias')
+    @alias.expect(alias_fields_update)
+    @alias.response(200, 'Success', response_fields)
+    @alias.response(404, 'Alias not found', response_fields)
+    @alias.response(400, 'Input validation exception', response_fields)
+    @alias.doc(security='Bearer')
+    @common.api_token_authorization
+    def put(self, alias):
+      """ Update alias """
+      data = api.payload
+      alias_found = models.Alias.query.filter_by(email = alias).first()
+      if alias_found is None:
+        return { 'code': 404, 'message': f'Alias {alias} cannot be found'}, 404
+      if 'comment' in data:
+        alias_found.comment = data['comment']
+      if 'destination' in data:
+        destination_csl = ",".join(data['destination'])
+        alias_found.destination = destination_csl
+      if 'wildcard' in data:
+        alias_found.wildcard = data['wildcard']
+      db.session.add(alias_found)
+      db.session.commit()
+      return {'code': 200, 'message': f'Alias {alias} has been updated'}
+
+    @alias.doc('delete_alias')
+    @alias.response(200, 'Success', response_fields)
+    @alias.response(404, 'Alias not found', response_fields)
+    @alias.doc(security='Bearer')
+    @common.api_token_authorization
+    def delete(self, alias):
+      """ Delete alias """
+      alias_found = models.Alias.query.filter_by(email = alias).first()
+      if alias_found is None:
+        return { 'code': 404, 'message': f'Alias {alias} cannot be found'}, 404
+      db.session.delete(alias_found)
+      db.session.commit()
+      return {'code': 200, 'message': f'Alias {alias} has been deleted'}, 200
+
+@alias.route('/destination/<string:domain>')
+class AliasWithDest(Resource):
+    @alias.doc('find_alias_filter_domain')
+    @alias.response(200, 'Success', alias_fields)
+    @alias.response(404, 'Alias or domain not found', response_fields)
+    @alias.doc(security='Bearer')
+    @common.api_token_authorization
+    def get(self, domain):
+        """ Find aliases of domain """
+        domain_found = models.Domain.query.filter_by(name=domain).first()
+        if domain_found is None:
+          return { 'code': 404, 'message': f'Domain {domain} cannot be found'}, 404
+        aliases_found = domain_found.aliases
+        if aliases_found.count == 0:
+          return { 'code': 404, 'message': f'No alias can be found for domain {domain}'}, 404
+        else:
+          return marshal(aliases_found, alias_fields), 200
--- a/core/admin/mailu/api/v1/alias.py
+++ b/core/admin/mailu/api/v1/alias.py
@@ -77,7 +77,7 @@ class Alias(Resource):
     @alias.response(400, 'Input validation exception', response_fields)
     @alias.doc(security='Bearer')
     @common.api_token_authorization
-    def put(self, alias):
+    def patch(self, alias):
       """ Update alias """
       data = api.payload
       alias_found = models.Alias.query.filter_by(email = alias).first()
@@ -86,8 +86,7 @@ class Alias(Resource):
       if 'comment' in data:
         alias_found.comment = data['comment']
       if 'destination' in data:
-        destination_csl = ",".join(data['destination'])
-        alias_found.destination = destination_csl
+        alias_found.destination = data['destination']
       if 'wildcard' in data:
         alias_found.wildcard = data['wildcard']
       db.session.add(alias_found)
--- a/core/admin/mailu/api/v1/domains.py
+++ b/core/admin/mailu/api/v1/domains.py
@@ -1,0 +1,410 @@
+import validators
+from flask_restx import Resource, fields, marshal
+from . import api, response_fields, user
+from .. import common
+from ... import models
+
+db = models.db
+
+dom = api.namespace('domain', description='Domain operations')
+alt = api.namespace('alternative', description='Alternative operations')
+
+domain_fields = api.model('Domain', {
+    'name': fields.String(description='FQDN (e.g. example.com)', example='example.com', required=True),
+    'comment': fields.String(description='a comment'),
+    'max_users': fields.Integer(description='maximum number of users', min=-1, default=-1),
+    'max_aliases': fields.Integer(description='maximum number of aliases', min=-1, default=-1),
+    'max_quota_bytes': fields.Integer(description='maximum quota for mailbox', min=0),
+    'signup_enabled': fields.Boolean(description='allow signup'),
+    'alternatives': fields.List(fields.String(attribute='name', description='FQDN', example='example2.com')),
+})
+
+domain_fields_update = api.model('DomainUpdate', {
+    'comment': fields.String(description='a comment'),
+    'max_users': fields.Integer(description='maximum number of users', min=-1, default=-1),
+    'max_aliases': fields.Integer(description='maximum number of aliases', min=-1, default=-1),
+    'max_quota_bytes': fields.Integer(description='maximum quota for mailbox', min=0),
+    'signup_enabled': fields.Boolean(description='allow signup'),
+    'alternatives': fields.List(fields.String(attribute='name', description='FQDN', example='example2.com')),
+})
+
+domain_fields_get = api.model('DomainGet', {
+    'name': fields.String(description='FQDN (e.g. example.com)', example='example.com', required=True),
+    'comment': fields.String(description='a comment'),
+    'max_users': fields.Integer(description='maximum number of users', min=-1, default=-1),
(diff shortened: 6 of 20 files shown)


Diman0:
bors try

Diman0:
bors try

Diman0:
bors try

Diman0:
bors retry

core/admin/mailu/api/v1/user.py
```diff
@@ -0,0 +1,276 @@
...
+}.update(base_model))
+
+user_fields_put = api.model('UserUpdate', {
+    'raw_password': fields.String(description='The raw (plain text) password of the user. Mailu will hash the password using PBKDF2-HMAC-SHA256', example='secret'),
+}.update(base_model))
+"""
+
+user_fields_get = api.model('UserGet', {
+    'email': fields.String(description='The email address of the user', example='John.Doe@example.com', attribute='_email'),
+    'password': fields.String(description='PBKDF2-HMAC-SHA256 based password of the user. For more info see passlib.hash.pbkdf2_sha256', example='$pbkdf2-sha256$1$.6UI/S.nXIk8jcbdHx3Fhg$98jZicV16ODfEsEZeYPGHU3kbrUrvUEXOPimVSQDD44'),
```
nextgens, line 52: That's not accurate anymore; nowadays we use bcrypt!

core/admin/mailu/api/common.py
```diff
@@ -0,0 +1,33 @@
...
+import flask
+import hmac
+from functools import wraps
+from flask_restx import abort
+
+def fqdn_in_use(*names):
+    for name in names:
+        for model in models.Domain, models.Alternative, models.Relay:
+            if model.query.get(name):
+                return model
```
nextgens, line 13: This is the least efficient implementation I can think of... and I'm confused about the purpose: why don't we return a boolean?

I suggest we either build a proper query and join all the tables... or alternatively map() the relevant get()s and return on first hit.

core/admin/mailu/api/v1/user.py
```diff
@@ -0,0 +1,276 @@
...
+
+user_fields_post = api.model('UserCreate', {
+    'email': fields.String(description='The email address of the user', example='John.Doe@example.com', attribute='_email', required=True),
+    'raw_password': fields.String(description='The raw (plain text) password of the user. Mailu will hash the password using PBKDF2-HMAC-SHA256', example='secret', required=True),
+}.update(base_model))
+
+user_fields_put = api.model('UserUpdate', {
+    'raw_password': fields.String(description='The raw (plain text) password of the user. Mailu will hash the password using PBKDF2-HMAC-SHA256', example='secret'),
+}.update(base_model))
+"""
```
nextgens, line 48: Why do we keep the comment here?

nextgens (changes requested):

core/admin/mailu/api/v1/user.py
Diman0: Aren't we using it in models.py? 
I'm referring to:
https://github.com/Mailu/Mailu/blob/8a90f83bd039b2c24fc892ef03c93439caed1496/core/admin/mailu/models.py#L592

I remember we also have a credentials cache. Is this cache using pbkdf2-hmac and for cold stored passwords we use bcrypt?

What would be an accurate description?
Is this fine?
'Bcrypt SHA256 based password of the user. For more info see passlib.hash.bcrypt_sha256', example='$bcrypt-sha256$v=2,t=2b,r=12$fmsAdJbYAD1gGQIE5nfJq.$zLkQUEs2XZfTpAEpcix/1k5UTNPm0jO')

core/admin/mailu/api/common.py
Diman0: It is a helper function, that is used to check if a domain is already used.
When creating a new item with a new domain, we must check if there is not already a Domain, Alternative or Relayed domain with the same name. This helper function helps checking this. It is used in 5 locations. For example:
https://github.com/Diman0/Mailu/blob/0875975e32a661c246b215e6232e141eda3f0719/core/admin/mailu/api/v1/domains.py#L100

core/admin/mailu/api/v1/user.py
Diman0: I forgot removing it. I will remove it.

core/admin/mailu/api/v1/user.py
nextgens: Yes, this line is the credential cache.

"Hash of the user's password; Mailu supports most of what passlib.hash does. Example='$bcrypt-sha256$v=2,t=2b,r=12$fmsAdJbYAD1gGQIE5nfJq.$zLkQUEs2XZfTpAEpcix/1k5UTNPm0jO')"

Diman0:
bors try

Diman0:
bors retry

smolinari:
LGTM! :grinning: I'm very interested in this, so posting to watch progress. 

Scott

core/admin/mailu/api/common.py
```diff
@@ -0,0 +1,33 @@
...
+        client_ip = flask.request.headers.get('X-Real-IP', flask.request.remote_addr)
+        if utils.limiter.should_rate_limit_ip(client_ip):
+            abort(429, 'Too many attempts from your IP (rate-limit)' )
+        if (request.args.get('api_token') == '' or
+            request.args.get('api_token') == None):
+            abort(401, 'A valid API token is expected as query string parameter')
+        if not hmac.compare_digest(request.args.get('api_token'), v1.api_token):
+            utils.limiter.rate_limit_ip(client_ip)
+            flask.current_app.logger.warn(f'Invalid API token provided by {client_ip}.')
+            abort(403, 'A valid API token is expected as query string parameter')
```
ghostwheel42, line 29: Is using a token in the Query string the best way to do authentication?
Shouldn't this be a HTTP request header `Authorization: Bearer <TOKEN>`?

core/admin/mailu/api/common.py
```diff
@@ -0,0 +1,33 @@
...
+        else:
+            flask.current_app.logger.info(f'Valid API token provided by {client_ip}.')
```
ghostwheel42, line 31: I think you should remove this `else:`.
So it's more in-line withe the above `if ...: abort()`.

ghostwheel42 (dismissed):
Hi @Diman0, I've finally started a review.
Looks promising. I'll still have to look into the "core" parts (alias, domains, relay, user) and to take a look at the swagger ui.

ghostwheel42:
I have a questin about `from flask_restx import abort`.
Is it enough to just call `abort()` or do we need to `return abort()` like in the rest af the flask suite?

core/admin/mailu/api/common.py
nextgens: Agreed, it would make more sense with a header.

Requests parameters can be leaked in referrers

core/admin/mailu/api/common.py
nextgens: I have found it useful to print() the request you ask sqlalchemy to make. A Naked get() like this is a 'SELECT * FROM' whereas ideally you want a 'WHERE EXISTS' query with all the tables and all the names in a single query.

docs/configuration.rst
```diff
@@ -189,23 +189,32 @@ An example:
...
+.. _advanced_settings:
 
 Advanced settings
 -----------------
 
-The ``CREDENTIAL_ROUNDS`` (default: 12) setting is the number of rounds used by the 
-password hashing scheme. The number of rounds can be reduced in case faster 
-authentication is needed or increased when additional protection is desired. 
-Keep in mind that this is a mitigation against offline attacks on password hashes, 
+The ``API`` (default: False) setting controls if the API endpoint is publicly
```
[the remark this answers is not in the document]
Diman0: I was thinking about that if API is disabled, then you could still directly go to admin by changing the network mapping in your docker-compose.yml file.

But it is better not to advertise this. I will remove the word `publicly`.

core/admin/mailu/configuration.py
```diff
@@ -69,6 +69,9 @@
     'LOGO_URL': None,
     'LOGO_BACKGROUND': None,
     # Advanced settings
+    'API' : 'false',
```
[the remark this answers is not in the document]
Diman0: Good catch. This needs to be changed.

core/admin/mailu/__init__.py
```diff
@@ -70,10 +70,12 @@ def format_datetime(value):
         return utils.flask_babel.format_datetime(value) if value else ''
 
     # Import views
-    from mailu import ui, internal, sso
+    from mailu import ui, internal, sso, api
     app.register_blueprint(ui.ui, url_prefix=app.config['WEB_ADMIN'])
     app.register_blueprint(internal.internal, url_prefix='/internal')
     app.register_blueprint(sso.sso, url_prefix='/sso')
+    if app.config.get('API_TOKEN'):
```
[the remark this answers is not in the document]
Diman0: I didn't even know that was possible. Maybe you can show me how to do this sometime? I will then do this for future situations.

core/admin/mailu/api/__init__.py
```diff
@@ -0,0 +1,27 @@
+from flask import redirect, url_for
+from flask_restx import apidoc
+from . import v1
```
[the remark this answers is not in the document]
Diman0: I can do this. However I'd liked that  the line `API=v1` makes it really clear what version is used by specifying it a second time via `API=v1`. But I will change this.

core/admin/mailu/api/common.py
Diman0: I wasn't aware of that it could be leaked. I thought it was safe due to the usage of HTTPS.  I will have a look at changing it to using Bearer token. That is not too difficult.

core/admin/mailu/api/common.py
Diman0: Agreed

core/admin/mailu/api/common.py
ghostwheel42: Okay, I've googled a bit and this is how it's done:
```
from sqlalchemy.sql.expression import label

def fqdn_in_use(*names):
    d = models.db.session.query(label('name', models.Domain.name))
    a = models.db.session.query(label('name', models.Alternative.name))
    r = models.db.session.query(label('name', models.Relay.name))
    u = d.union_all(a).union_all(r)
    for name in names:
        if u.filter_by(name=name).count():
            return True
    return False
```

@Diman0 could you include this?

We can probably use this at many other locations in mailu.

core/admin/mailu/__init__.py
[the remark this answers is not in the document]
ghostwheel42: There's nothing special about this - just checkout and continue developing.
It's like using master as base.

core/admin/mailu/api/common.py
nextgens: .exists() instead of count(), no?

nextgens: You may even be able to get rid of the for loop and make a single query: with .filter(or_(*search_args)) ... and you'd build search_args potentially without the union

ghostwheel42: Yes, wanted to write this. 
I'll improve it. Just one query is way better.

Diman0: In the api blueprint, only a single name must be checked. So instead of *names, the function could require name as parameter and we could rewrite is as:
```
from sqlalchemy.sql.expression import label

def fqdn_in_use(name):
    d = models.db.session.query(label('name', models.Domain.name))
    a = models.db.session.query(label('name', models.Alternative.name))
    r = models.db.session.query(label('name', models.Relay.name))
    if d.union_all(a).union_all(r).filter_by(name=name).count():
        return True
    return False
```

nextgens: Well, if we don't need to support checking for multiple names we can do simpler and three queries isn't that bad :)

Please don't use .count() as that makes it iterate over all the data/indexes; use .exists() which will bail out on first match found.

Diman0:
bors try

core/admin/mailu/api/common.py
ghostwheel42: this uses exists:
```
from sqlalchemy.sql.expression import label

def fqdn_in_use(name):
    d = models.db.session.query(label('name', models.Domain.name))
    a = models.db.session.query(label('name', models.Alternative.name))
    r = models.db.session.query(label('name', models.Relay.name))
    u = d.union_all(a).union_all(r).filter_by(name=name)
    if models.db.session.query(u.exists()).scalar():
        return True
    return False
```

core/admin/mailu/api/common.py
```diff
@@ -20,14 +22,12 @@ def decorated_function(*args, **kwds):
         client_ip = flask.request.headers.get('X-Real-IP', flask.request.remote_addr)
         if utils.limiter.should_rate_limit_ip(client_ip):
             abort(429, 'Too many attempts from your IP (rate-limit)' )
-        if (request.args.get('api_token') == '' or
-            request.args.get('api_token') == None):
-            abort(401, 'A valid API token is expected as query string parameter')
-        if not hmac.compare_digest(request.args.get('api_token'), v1.api_token):
+        if not request.headers.get('Authorization'):
+            abort(401, 'A valid API token is expected which is provided as request header')
+        if not hmac.compare_digest(request.headers.get('Authorization'), v1.api_token):
```
nextgens, line 27: Hmm, you do need to have the scheme here. Please use "Bearer" as in RFC 6750 and suggested by @ghostwheel42

core/admin/mailu/api/common.py
```diff
@@ -5,13 +5,15 @@
...
-    for name in names:
-        for model in models.Domain, models.Alternative, models.Relay:
-            if model.query.get(name):
-                return model
-    return None
+def fqdn_in_use(name):
+    d = models.db.session.query(label('name', models.Domain.name))
+    a = models.db.session.query(label('name', models.Alternative.name))
+    r = models.db.session.query(label('name', models.Relay.name))
+    if d.union_all(a).union_all(r).filter_by(name=name).count() > 0:
```
nextgens, line 14: Please use the .exists() variant as suggested in https://github.com/Mailu/Mailu/pull/2464#discussion_r1032557167

nextgens (dismissed):

core/admin/mailu/api/common.py
Diman0: The version of SwaggerUI included with flask restx does not support this out of box.
It does correctly use the request header Authorization, but it does not prefix the entered token value with `Bearer`. I wouldn't be surprised if this applies for many other clients as well.
Or you must manually enter as value `Bearer <token>` instead of `<token>` in SwaggerUI.

I could change the check to
```
hmac.compare_digest(request.headers.get('Authorization'), 'Bearer ' + v1.api_token):
```
But that does mean that the client must know that the token must be prepended with Bearer. 

We could check for both variants? check for
- `<token>`
- `Bearer <token>`

nextgens: In that case, please check whether there is a space and if there is ensure that the first part is "Bearer"

core/admin/mailu/api/common.py
Diman0: Thank you. I will use this.

core/admin/mailu/api/v1/user.py
```diff
@@ -0,0 +1,237 @@
...
+    'reply_enddate': fields.Date(description='End date for automatic replies in YYYY-MM-DD format.', example='2022-02-22'),
+    'displayed_name': fields.String(description='The display name of the user within the Admin GUI', example='John Doe'),
+    'spam_enabled': fields.Boolean(description='Enable the spam filter'),
+    'spam_mark_as_read': fields.Boolean(description='Enable marking spam mails as read'),
+    'spam_threshold': fields.Integer(description='The user defined spam filter tolerance', example='80'),
+})
+
+
+@user.route('')
+class Users(Resource):
```
nextgens, line 64: User.allow_spoofing is missing

core/admin/mailu/api/v1/alias.py
```diff
@@ -0,0 +1,127 @@
...
+    @alias.response(200, 'Success', response_fields)
+    @alias.response(400, 'Input validation exception', response_fields)
+    @alias.response(409, 'Duplicate alias', response_fields)
+    @alias.doc(security='Bearer')
+    @common.api_token_authorization
+    def post(self):
+        """ Create a new alias """
+        data = api.payload
+
+        alias_found = models.Alias.query.filter_by(email = data['email']).all()
```
nextgens, line 44: Please use the construct with .exists() or at least switch to .first() instead of .all()

core/admin/mailu/api/v1/alias.py
```diff
@@ -0,0 +1,127 @@
...
+@alias.route('/<string:alias>')
+class Alias(Resource):
+    @alias.doc('find_alias')
+    @alias.response(200, 'Success', alias_fields)
+    @alias.response(404, 'Alias not found', response_fields)
+    @alias.doc(security='Bearer')
+    @common.api_token_authorization
+    def get(self, alias):
+        """ Find alias """
+        alias_found = models.Alias.query.filter_by(email = alias).all()
```
nextgens, line 67: .first() instead of .all()

core/admin/mailu/api/v1/relay.py
```diff
@@ -0,0 +1,119 @@
...
+    @relay.doc(security='Bearer')
+    @common.api_token_authorization
+    def post(self):
+        """ Create relay """
+        data = api.payload
+
+        if not validators.domain(name):
+            return { 'code': 400, 'message': f'Relayed domain {name} is not a valid domain'}, 400
+
+        relay_found = models.Relay.query.filter_by(name=data['name']).all()
```
nextgens, line 47: Is this used? I can't see where

nextgens (changes requested):

Diman0:
bors try

Diman0:
bors try

core/nginx/conf/nginx.conf
```diff
@@ -244,6 +244,13 @@ http {
       {% endif %}
       {% endif %}
 
+      {% if API == 'true' %}
```
nextgens, line 247: I wonder if we still need to test against strings here now that we normalize in master

core/admin/mailu/api/v1/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+}
+
+api = Api(
+    blueprint, version=f'{VERSION:.1f}',
+    title='Mailu API', default_label='Mailu',
+    validate=True,
+    authorizations=authorization,
+    security='Bearer',
+    doc='/swaggerui/'
+)
```
nextgens, line 25: Is there a way to redirect /api/v1/ to /api/v1/swaggerui/ if there is no parameter ?

nextgens (dismissed):
I get a bunch of warnings in the logs of admin:
```
storage-admin-1  | /app/venv/lib/python3.10/site-packages/flask_restx/api.py:323: UserWarning: The setup method 'add_url_rule' can no longer be called on the blueprint 'api_v1'. It has already been registered at least once, any changes will not be applied consistently.
storage-admin-1  | Make sure all imports, decorators, functions, etc. needed to set up the blueprint are done before registering it.
storage-admin-1  | This warning will become an exception in Flask 2.3.
storage-admin-1  |   app_or_blueprint.add_url_rule(self.prefix or "/", "root", self.render_root)
```

core/admin/mailu/api/v1/domains.py
```diff
@@ -0,0 +1,410 @@
...
+        """ List users from domain """
+        if not validators.domain(domain):
+            return { 'code': 400, 'message': f'Domain {domain} is not a valid domain'}, 400
+        domain_found = models.Domain.query.get(domain)
+        if not domain_found:
+            return { 'code': 404, 'message': f'Domain {domain} does not exist'}, 404
+        return  models.User.query.filter_by(domain=domain_found).all()
+
+@alt.route('')
+class Alternatives(Resource):
```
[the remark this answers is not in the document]
Diman0: That is correct. In the main administration interface it is also not possible to update an alternative. You only have
- Create
- Delete

in the admin interface

core/admin/mailu/api/v1/__init__.py
nextgens: I am just worried that if you browse to /api/ you get redirected to /api/v1/ that displays nothing useful. IMHO the documentation or a placeholder listing versions would be more useful.

Diman0: /api goes to the login page. 
/api/ will redirect to /api/v1./ /api/v1 will show the documentation.

Diman0:
I cannot do anything about the warnings. The warnings are due to the flask-restx module. There is no new version yet. In the future once the warnings are addressed in flask-restx, we could update the package.

bors try

Diman0:
bors retry

nextgens:
bors try

nextgens (dismissed):
LGTM.

The label Needs2 was set by @ghostwheel42 ...

docs/configuration.rst
```diff
@@ -198,23 +198,31 @@ An example:
...
-password hashing scheme. The number of rounds can be reduced in case faster 
-authentication is needed or increased when additional protection is desired. 
-Keep in mind that this is a mitigation against offline attacks on password hashes, 
+The ``API`` (default: False) setting controls if the API endpoint is reachable.
+
+The ``WEB_API`` (default: /api) setting configures the endpoint that the API
+listens on publicly&interally. The path must always start with a leading slash.
+
+The ``API_TOKEN`` (default: None) enables the API endpoint. This token must be
+passed as request header to the API as authentication token.
```
ghostwheel42, line 212: I think it is a little bit unclear (and maybe redundant) waht the API, WEB_API and API_TOKEN variables do.

Do I understand correctly, that
1. by setting API_TOKEN only: I enable the API internally on /api
2. by setting WEB_API only: nothing happens
3. by setting API only: nothing happens
4. by setting API and API_TOKEN: the api is exposed to the world (path can be changed by WEB_API)

I think we should describe this clearer.

core/admin/mailu/api/v1/alias.py
```diff
@@ -0,0 +1,127 @@
...
+          return marshal(alias_found,alias_fields), 200
+
+    @alias.doc('update_alias')
+    @alias.expect(alias_fields_update)
+    @alias.response(200, 'Success', response_fields)
+    @alias.response(404, 'Alias not found', response_fields)
+    @alias.response(400, 'Input validation exception', response_fields)
+    @alias.doc(security='Bearer')
+    @common.api_token_authorization
+    def put(self, alias):
```
ghostwheel42, line 80: I think all the "put" methods should really be "patch".
We are changing the existing object "in place", so this is a patch not a put.

ghostwheel42 (dismissed):
Thanks for all the work put into this.
I think it's a good start, but I have some requests for changes.

We can start like this, but I think that by doing this generically we'd make it more robust and also better managable.
We have a model.py and schema.py knowing everything about the config - the config-import and config-export are already as generic as possible, so we should use this for the rest api, too.

I didn't test API interaction as thorough as I'd like, so I don't know if we would need to switch to v2 really quick when making those changes.

core/nginx/conf/nginx.conf
```diff
@@ -244,6 +244,13 @@ http {
       {% endif %}
       {% endif %}
 
+      {% if API %}
+      location ~ {{ WEB_API }} {
```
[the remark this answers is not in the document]
Diman0: This is handled via the global config object in `core/admin/mailu/configuration.py`. Here the default is set to `/api`.

docs/configuration.rst
Diman0: What is not clear about "The API_TOKEN (default: None) `enables the API endpoint.`?"

All three settings are required. Some people will want the option to make it publicly exposed or not. some people will want to change the end point (/api), just like the other end points like WEB_ADMIN. For this first version the API token is static. In a future PR we could explore configuring the API token per user via the web admin.

by setting API_TOKEN only: I enable the API internally on /api
=>
That is correct.

by setting WEB_API only: nothing happens
=>
That is correct.

by setting API only: nothing happens
=> 
That is correct.

by setting API and API_TOKEN: the api is exposed to the world (path can be changed by WEB_API)
=>
That is correct.

I could add a paragraph above these 3 settings that states that `API` and `API_TOKEN`  should be both configured for using the API. I could also add this to the `docs/api.rst`

core/admin/mailu/api/v1/alias.py
Diman0: I was not even aware that there was a HTTP PATCH method. 
It is not mentioned in the [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-put) that describes all HTTP methods.

But reading the decription in [RFC5789](https://www.rfc-editor.org/rfc/rfc5789#section-2) I agree that the HTTP method PATCH is a better fit.

I will check if this can be changed, and will change it if flask-restx supports this.

docs/configuration.rst
smolinari: If I may, I think the naming could be better:

API -> API_ENABLED
WEB_API -> API_PATH
API_TOKEN -> is ok as is. 

??? 

Scott

Diman0:
This PR is originally based on two small commits from ghostweels apiv1 branch. I've force pushed changes to reflect this.

docs/configuration.rst
Diman0: No. I used the same naming conventions as the other components. I don't want to mix naming schemes. 
ADMIN also uses `ADMIN` for enabling (exposing publicly) and `WEB_ADMIN` for the endpoint.
WEBMAIL also uses `WEBMAIL` for enabling (exposing publicly) and `WEB_WEBMAIL` for the endpoint.



Maybe configuring the API to be publicly reachable (`API`), should be moved from the configuration reference to the setup utility. We remove `API` from the documentation. This is how we currently do it for ADMIN & WEBMAIL. This makes it more consistent.

core/nginx/conf/nginx.conf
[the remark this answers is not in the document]
ghostwheel42: Unfortunately it is not. This is the front or nginx container and no code from admin runs in there.
You have to sanitize the environment everywhere you use it until we have moved all environment sanitization into the base container.

docs/configuration.rst
ghostwheel42: I fear we will have complaints about setting WEB_API and the api remaining on "/api" internally.
And complaints about setting API to true and "nothing happens".
I know Mailu uses this kind of settings in other places, but we could start making it more accessible.

I think we should not put API into the setup utility, as it is not needed there.
Only variables changing the compose-file need to be in the setup utility.
Everything else can be changed "on the fly" - and maybe at some time in the admin interface, too.

Diman0: That is not how the setup utility is currently designed. Many settings are for the mailu.env file only. The goal of the setup utility is to enable less experienced users to configure a new mailu installation. Ideally users new have to touch mailu.env.

>I fear we will have complaints about setting WEB_API and the api remaining on "/api" internally.
>And complaints about setting API to true and "nothing happens".
>I know Mailu uses this kind of settings in other places, but we could start making it more accessible.

Using the setup utility will prevent this. In the setup utility we will enforce that you cannot provide a non-empty value for API_TOKEN and WEB_API, if the API is enabled. Just like when you enable ADMIN or WEBMAIL in the setup utility.

Modifying the mailu.env file is scary and error prone for many of our users. 

>I fear we will have complaints about setting WEB_API and the api remaining on "/api" internally.

I do not agree. This is in line with all our internal components. They all have static endpoints defined. It is by design that you cannot change the internal endpoints (/ui, /sso, /internal, /api, etc).

core/nginx/conf/nginx.conf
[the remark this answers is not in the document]
Diman0: Ah yes. You are correct. The default has been added to the nginx.conf via your commit https://github.com/Mailu/Mailu/pull/2464/commits/be407813946665ce571a5124a82f08920a32cd44.

docs/configuration.rst
ghostwheel42: > I do not agree. This is in line with all our internal components.

Hi @Diman0 - I never said this is not "in line" - other setup parameters can also be hard to understand.

But it's okay for me to configure it like this. Maybe we will have a big config-change later where we eill separate the things which can be configured at runtime from those which can not. Things like the api for example could be (de)activated at runtime and also the token could be changed live in the admin interface.

nextgens (approved):

ghostwheel42:
bors try

ghostwheel42:
bors try

ghostwheel42:
bors try

ghostwheel42 (approved):

why2lyj:
@Diman0 tkx for this PR

smolinari:
:clap: - Yes!! Thank you all for the effort! :+1: 

Scott

PR #2483, opened 2022-10-20
Introduce FETCHMAIL_ENABLED

### LINKED ISSUE #2127: Suggestion: Only show "Fetched accounts" menu entry if fetchmail is enabled
I think it would be neat if the sidebar entry `Fetched accounts` in the admin portal is only displayed when fetchmail is really enabled. When fetchmail is disabled, the tab may cause confusion for users, as it has no functionality in this case.

DjVinnii:
## What type of PR?

Enhancement

## What does this PR do?
Add `FETCHMAIL_ENABLED` to enable/disable the Fetchmail functionality in the Admin UI.

### Related issue(s)
- closes #2127

## Prerequisites

### DIFF
--- a/core/admin/mailu/configuration.py
+++ b/core/admin/mailu/configuration.py
@@ -16,7 +16,7 @@ DEFAULT_CONFIG = {
     'DOMAIN_REGISTRATION': False,
     'TEMPLATES_AUTO_RELOAD': True,
     'MEMORY_SESSIONS': False,
-    'FETCHMAIL_ENABLED': False,
+    'FETCHMAIL_ENABLED': True,
     # Database settings
     'DB_FLAVOR': None,
     'DB_USER': 'mailu',
--- a/core/admin/mailu/configuration.py
+++ b/core/admin/mailu/configuration.py
@@ -16,13 +16,14 @@ DEFAULT_CONFIG = {
     'DOMAIN_REGISTRATION': False,
     'TEMPLATES_AUTO_RELOAD': True,
     'MEMORY_SESSIONS': False,
+    'FETCHMAIL_ENABLED': False,
     # Database settings
     'DB_FLAVOR': None,
     'DB_USER': 'mailu',
     'DB_PW': None,
     'DB_HOST': 'database',
     'DB_NAME': 'mailu',
-    'SQLITE_DATABASE_FILE':'data/main.db',
+    'SQLITE_DATABASE_FILE': 'data/main.db',
     'SQLALCHEMY_DATABASE_URI': 'sqlite:////data/main.db',
     'SQLALCHEMY_TRACK_MODIFICATIONS': False,
     # Statistics management
@@ -59,7 +60,7 @@ DEFAULT_CONFIG = {
     # Web settings
     'SITENAME': 'Mailu',
     'WEBSITE': 'https://mailu.io',
-    'ADMIN' : 'none',
+    'ADMIN': 'none',
     'WEB_ADMIN': '/admin',
     'WEB_WEBMAIL': '/webmail',
     'WEBMAIL': 'none',
--- a/core/admin/mailu/configuration.py
+++ b/core/admin/mailu/configuration.py
@@ -16,7 +16,7 @@ DEFAULT_CONFIG = {
     'DOMAIN_REGISTRATION': False,
     'TEMPLATES_AUTO_RELOAD': True,
     'MEMORY_SESSIONS': False,
-    'FETCHMAIL_ENABLED': True,
+    'FETCHMAIL_ENABLED': False,
     # Database settings
     'DB_FLAVOR': None,
     'DB_USER': 'mailu',
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -102,8 +102,8 @@ support or e.g. mismatching TLS versions to deliver emails to Mailu.
 
 .. _fetchmail:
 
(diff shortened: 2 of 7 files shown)


core/admin/mailu/configuration.py
```diff
@@ -16,13 +16,14 @@
     'DOMAIN_REGISTRATION': False,
     'TEMPLATES_AUTO_RELOAD': True,
     'MEMORY_SESSIONS': False,
+    'FETCHMAIL_ENABLED': False,
```
ghostwheel42, line 19: I think the default must be True to not break existing deployments not having the environment variable set.

ghostwheel42 (changes requested):
Looks good. But I'd change the default.

docs/configuration.rst
```diff
@@ -102,6 +102,9 @@ support or e.g. mismatching TLS versions to deliver emails to Mailu.
 
 .. _fetchmail:
 
+When ``FETCHMAIL_ENABLED`` is set to ``True``, the fetchmail functionality is shown in the admin interface.
```
ghostwheel42, line 105: Maybe this should be changed from "is shown" to "is enabled", as you'll get a 404 for all calls.

ghostwheel42:
Hmm. Do we need to also test the FETCHMAIL_ENABLED flag in the fetchmail container itself?
People would expect fetchmail not to run when this is set to false.

core/admin/mailu/configuration.py
DjVinnii: Yeah, that seems reasonable. I'll change te default to `True`.

DjVinnii:
> Hmm. Do we need to also test the FETCHMAIL_ENABLED flag in the fetchmail container itself? People would expect fetchmail not to run when this is set to false.

That shouldn't be too hard to implement. I think it makes fetchmail more predicatable. If `FETCHMAIL_ENABLED` is set to `False`, I (as user) won't expect fetchmail to do anything (even if configured before).

DjVinnii:
I implemented a simple check in `fetchmail.py` which skips the `run()` function and gets back to the sleep. I came to my attention that there is no logging in the while loop. In my opinion, it could be useful to have some simple form of logging, showing the sleep period (`FETCHMAIL_DELAY`) and if the fetching is being executed (`FETCHMAIL_ENABLED`). Should I implement this as well in this PR?

ghostwheel42:
> In my opinion, it could be useful to have some simple form of logging, showing the sleep period (`FETCHMAIL_DELAY`) and if the fetching is being executed (`FETCHMAIL_ENABLED`). Should I implement this as well in this PR?

Yes, sure. I think it will be useful if people are seeing what's happening inside the container.

DjVinnii:
I added some simple logging for `FETCHMAIL_ENABLED` and `FETCHMAIL_DELAY`. Now, it is better visible how long fetchmail is waiting for the next run, and if it is skipping due `FETCHMAIL_DISABLED`. 

It might be nice to implement better logging in the future. This could include timestamps and log levels for example.

core/admin/mailu/configuration.py
```diff
@@ -16,13 +16,14 @@
     'DOMAIN_REGISTRATION': False,
     'TEMPLATES_AUTO_RELOAD': True,
     'MEMORY_SESSIONS': False,
+    'FETCHMAIL_ENABLED': True,
```
Diman0, line 19: Do we really want to se this to true?
The expected default behaviour is that if fetchmail is not configured it is disabled. 

It does not have to be backwards compatible. When a new Mailu version is released, we always state that you must regenerate your docker-compose.yml & mailu.env file (which will contain the new env var).

Of course this will also be mentioned in the release notes.

More importantly, the fetchmail container itself will only run if the env var is explicitly set to true. So for people not updating their mailu.env file, they will still see fetchmail in the admin GUI, but will have a non working fetchmail container.

DjVinnii: Hmm. Initially I did set the default to False. Because it is an optional component, it seems reasonable to me that it should be enabled explicitly.  However @ghostwheel42 mentioned that it might be better if we don't break existing installations by defaulting to False.

> More importantly, the fetchmail container itself will only run if the env var is explicitly set to true. So for people not updating their mailu.env file, they will still see fetchmail in the admin GUI, but will have a non working fetchmail container.

In the current state, fetchmail is ran by default. It only skips the runs if `FETCHMAIL_ENABLED` is set to False. This is the same for enabling the UI for fetchmail.

I guess there is something to say for both defaults, but in my opinion defaulting to false and mentioning it in release notes seems the most logical.

Diman0: @ghostwheel42 
Do you agree with us? We always explicitly mention in the release notes that the docker-compose.yml file must be regenerated. When you do this, the mailu.env file will be correctly configured.

Of course the release notes will also mention this new setting for the people who `feel lucky` and rather manually update the docker-compose.yml file & mailu.env file upon upgrade.

Existing deployments are not affected. Only new deployments on the new mailu release. If people don't use setup.mailu.io and also do not read the release notes, then it is not our problem.

ghostwheel42: Yes, sure. If it’s documented it’s okay.

nextgens (approved):
LGTM

Diman0 (dismissed):
lgtm

Diman0 (approved):

PR #2488, opened 2022-10-23
Properly cast DEFER_ON_TLS_ERROR as bool in Jinja2

lsde:
## What type of PR?

bug-fix

## What does this PR do?

It fixes DEFER_ON_TLS_ERROR casting when setting it to False trough environment.

Without the fix, the value of DEFER_ON_TLS_ERROR is a string, thus is always evaulated as True


nextgens:
According to the tests this leads to:

```
jinja2.exceptions.TemplateRuntimeError: No filter named 'bool' found.
  File "/start.py", line 59, in <module>
    conf.jinja(postfix_file, os.environ, os.path.join("/etc/postfix", os.path.basename(postfix_file)))
  File "/usr/lib/python3.9/site-packages/socrate/conf.py", line 15, in jinja
    result = jinja2.Template(source.read()).render(environ)
  File "/usr/lib/python3.9/site-packages/jinja2/environment.py", line 1301, in render
    self.environment.handle_exception()
  File "/usr/lib/python3.9/site-packages/jinja2/environment.py", line 936, in handle_exception
    raise rewrite_traceback_stack(source=source)
  File "<template>", line 59, in top-level template code
```

ghostwheel42:
Please just modify start.py and set
os.environ['DEFER_ON_TLS_ERROR'] = os.environ.get('DEFER_ON_TLS_ERROR', 'true').lower() in ('yes', 'true')

We'll soon normalize all environment variables in a single place but this will help in the meantime.

lsde:
oh, sorry, my Jinja and Ansible skills got so interwowen that i totally forgot "bool" filter is not part of Jina itself. This is one of those 'i just know it, why trying it live' moments 🤦.

ok, will do. thanks and keep up the good work!

PR #2494, opened 2022-10-26
Improve ISSUE_TEMPLATE.md

ghostwheel42:
## What type of PR?

documentation

## What does this PR do?

I'm having trouble separating the user input from the instructions in the issue template.
This puts the instructions in comments so the rendered markdown is almost empty.


Diman0 (approved):
Looks good to me. Let's give it a try. We can always modify it if the comment is not shown well.

PR #2500, opened 2022-10-31
Password policy enforcement

### LINKED ISSUE #287: Password Complexity Requirements

nextgens:
## What type of PR?

Feature

## What does this PR do?

It enforces that all new passwords set by users are at least 8 characters in length and checks all users' passwords at login time against HIBP.

The HIBP part requires javascript and Mailu to be accessed over HTTPS to work but degrades gracefully (no message will be shown if the requirements are not met).

It was a conscious choice to implement it at this level: administrators can set weaker passwords using non-HTTP based interfaces.

### Related issue(s)
- close #2208
- close #287

## Prerequisites


core/admin/mailu/sso/views/base.py
```diff
@@ -40,7 +40,10 @@ def login():
             flask_login.login_user(user)
             response = flask.redirect(destination)
             response.set_cookie('rate_limit', utils.limiter.device_cookie(username), max_age=31536000, path=flask.url_for('sso.login'), secure=app.config['SESSION_COOKIE_SECURE'], httponly=True)
-            flask.current_app.logger.info(f'Login succeeded for {username} from {client_ip}.')
+            flask.current_app.logger.info(f'Login succeeded for {username} from {client_ip} pwned={form.pwned.data}.')
+            breaches = int(form.pwned.data)
```
ghostwheel42, line 44: Does this need a try/except?
I think we should add it to not spam the logs if the field is empty or filled with garbage.

```python
            try:
                breaches = int(form.pwned.data)
            except ValueError:
                breaches = -1
```

core/admin/mailu/ui/views/domains.py
```diff
@@ -93,6 +93,13 @@ def domain_signup(domain_name=None):
         del form.pw
         del form.pw2
     if form.validate_on_submit():
+        if not flask_login.current_user.is_authenticated and len(form.pw.data) < 8:
+            flask.flash("This password is too short.", "error")
+            return flask.render_template('domain/signup.html', form=form)
+        breaches = int(form.pwned.data)
```
ghostwheel42, line 99: Same try/except applies here.

core/admin/mailu/ui/views/users.py
```diff
@@ -28,6 +28,15 @@ def user_create(domain_name):
         form.quota_bytes.validators = [
             wtforms.validators.NumberRange(max=domain.max_quota_bytes)]
     if form.validate_on_submit():
+        if len(form.pw.data) < 8:
+            flask.flash("This password is too short.", "error")
+            return flask.render_template('user/create.html',
+                domain=domain, form=form)
+        breaches = int(form.pwned.data)
```
ghostwheel42, line 35: and here

core/admin/mailu/ui/views/users.py
```diff
@@ -60,6 +69,15 @@ def user_edit(user_email):
         form.quota_bytes.validators = [
             wtforms.validators.NumberRange(max=max_quota_bytes)]
     if form.validate_on_submit():
+        if len(form.pw.data) < 8:
+            flask.flash("This password is too short.", "error")
+            return flask.render_template('user/edit.html', form=form, user=user,
+                domain=user.domain, max_quota_bytes=max_quota_bytes)
+        breaches = int(form.pwned.data)
```
ghostwheel42, line 76: and here.

core/admin/mailu/ui/views/users.py
```diff
@@ -119,6 +137,13 @@ def user_password(user_email):
         if form.pw.data != form.pw2.data:
             flask.flash('Passwords do not match', 'error')
         else:
+            if len(form.pw.data) < 8:
+                flask.flash("This password is too short.", "error")
+                return flask.render_template('user/password.html', form=form, user=user)
+            breaches = int(form.pwned.data)
```
ghostwheel42, line 143: and here ;)

core/admin/mailu/ui/views/users.py
```diff
@@ -170,6 +195,13 @@ def user_signup(domain_name=None):
         if domain.has_email(form.localpart.data) or models.Alias.resolve(form.localpart.data, domain_name):
             flask.flash('Email is already used', 'error')
         else:
+            if len(form.pw.data) < 8:
+                flask.flash("This password is too short.", "error")
+                return flask.render_template('user/signup.html', domain=domain, form=form)
+            breaches = int(form.pwned.data)
```
ghostwheel42, line 201: and here.
I wonder if we could somehow get rid of this code-duplication.

```python
    if msg := utils.isBadOrPwned(form):
        flask.flash(msg, "error")
        return flask.render_template('user/signup.html', domain=domain, form=form)

# utils.py:
def isBadOrPwned(form):
        if len(form.pw.data) < 8:
            return "This password is too short."
        try:
            breaches = int(form.pwned.data)
        except ValueError:
            breaches = -1
        if breaches > 0:
            return f"This password appears in {breaches} data breaches! Please change it."
    return None
```

core/admin/assets/content/assets/app.js
```diff
@@ -75,5 +125,12 @@ $('document').ready(function() {
         $('form :input').prop('disabled', true);
     }
 
+    if (window.isSecureContext) {
+        $("#HIBPpw").change(function(){
```
ghostwheel42, line 129: I think the password fields are missing this HIBPpw id.
Shouldn't this just be "#pw"?

core/admin/mailu/sso/views/base.py
nextgens: Done in 66250e39

core/admin/assets/content/assets/app.js
nextgens: Indeed; that's 24b2c7c0

I was playing with ":password" as a selector and then decided to override it in the template (but haven't committed the files) ... your suggested approach is better.

nextgens:
bors try

nextgens:
bors try

ghostwheel42:
Hmm, this still does nothing when testing (no request to api.pwnedpasswords.com).

I think we should also try to stop early (ie not submit the form) when changing a password.
"Flash" should be callable via javascript and if we stop posting the form the browser won't suggest to save the (not) changed password.

nextgens:
> Hmm, this still does nothing when testing (no request to api.pwnedpasswords.com).
> 

Are you testing it from a secure context? If not it won't work (crypto.subtle won't be available). If we want to make it work there too we need to ship our own implementation of SHA1 in pure JS.

> I think we should also try to stop early (ie not submit the form) when changing a password. "Flash" should be callable via javascript and if we stop posting the form the browser won't suggest to save the (not) changed password.

I thought about it but the idea on the login form is to let the administrator know when users are using compromised passwords. Right now it's purely informative (we don't force the user to change) ... but we also ensure we let the administrator know (in the logs of admin).

ghostwheel42:
bors try

Diman0:
bors retry

nextgens:
Hmm, after the last few changes I end up with two separate input forms with the id='pwned'

ghostwheel42:
bors try

ghostwheel42:
bors try

ghostwheel42 (approved):
LGTM

PR #2503, opened 2022-11-01
Fix build action dependencies

fastlorenzo:
## What type of PR?

Bug fix

## What does this PR do?

Adding a missing check for x86_64 to install gcc, etc.  Otherwise the build pipeline will throw an error if not using the cache that it's missing gcc and the other libraries.

### Related issue(s)

N/A

## Prerequisites


nextgens:
So we got to the bottom of it. This has nothing to do with the cache... Currently the build fails if you add dependencies (which you did, attempting to build your openid branch) that have no wheels.

We'll work on a better solution for this; feel free to create an issue so that we don't forget.

PR #2504, opened 2022-11-01
Remove superfluous cache export entry for arm

Diman0:
## What type of PR?

fix

## What does this PR do?
Fix arm build. The arm job contained one cache export entry too many. Only one cache export entry is allowed.

## Prerequisites

- [n/a] Unless it's docs or a minor change: add [changelog](https://mailu.io/master/contributors/workflow.html#changelog) entry file.


Diman0:
bors try

ghostwheel42 (approved):

PR #2505, opened 2022-11-01
Fix building wheels when deps need to compile

ghostwheel42:
## What type of PR?

bug-fix

## What does this PR do?

This installs build requiremnets (compiler, etc.) when tehre are no pre-built wheels an the current architecture.


nextgens (approved):
LGTM

PR #2507, opened 2022-11-01
Fix the CI

nextgens:
We should install the dependencies everywhere where we may have to rebuild the wheels.

If other people use other arch and want their builds to go faster we can whitelist them too after they have confirmed it works.


ghostwheel42 (commented):
LGTM

PR #2513, opened 2022-11-02
fix(auto-reply): include start and end dates in the auto-reply period

### LINKED ISSUE #2512: Include start and end dates in the auto-reply period
### Environment

- [X] docker-compose
- [ ] kubernetes
- [ ] docker swarm

### Version

- Version: `1.9`

## Description
The administration interface provides an [auto-reply](https://mailu.io/master/webadministration.html#auto-reply) page where automatic replies can be configured with start and end dates. Unfortunately both the start date and the end date are not included in the auto-reply period (i.e. no auto replies are being sent on these two days). To work around this issue you have to insert the day before your vacation as start date and the day after your vacation as end date. This is not intuitive.

## Replication Steps
Activate the auto-reply feature, insert subject and body text and the current date as "start of vacation" ("end of vacation" has to be a date in the future). Then send an email from another email account (external) to your email address hosted on Mailu.

## Observed behaviour
No auto reply message received by the sender.

## Expected behaviour
Auto reply message received by the sender.

To verify this behaviour you can add yesterday's date as "start of vacation" date and send another email to your Mailu account ... the sender will receive an auto reply message ... 
The same applies to the "end of vacation" date.

## Logs
n/a

bb-wkr:
## What type of PR?
bug-fix

## What does this PR do?
Include start and end dates in the auto-reply period

### Related issue(s)
closes #2512

## Prerequisites
If an entry is not applicable, you can check it or remove it from the list.

### DIFF
--- a/core/admin/mailu/models.py
+++ b/core/admin/mailu/models.py
@@ -546,8 +546,8 @@ class User(Base, Email):
         now = date.today()
         return (
             self.reply_enabled and
-            self.reply_startdate < now and
-            self.reply_enddate > now
+            self.reply_startdate <= now and
+            self.reply_enddate >= now
         )
 
     @property
--- a/towncrier/newsfragments/2512.bugfix
+++ b/towncrier/newsfragments/2512.bugfix
@@ -1,0 +1,1 @@
+Fix: include start and end dates in the auto-reply period


ghostwheel42 (approved):
Ah, yes. That's the way it should be.

nextgens (approved):
Thank you for your contribution!

PR #2514, opened 2022-11-02
Update deps

ghostwheel42:
## What type of PR?

update python dependencies

## What does this PR do?

Update python deps in base image


ghostwheel42:
bors try

ghostwheel42:
I think this can be merged now. But please test it before, as our test-coverage is probably not great yet.
I've pinned Flask-sqlalchemy to be <3.0 for now, as mailu fails building with errors in models.py/schema.py
We'll need to fix this later.

ghostwheel42:
bors try

Diman0:
@ghostwheel42 I know it is not caused by changes of this PR, but the date picker for  the automatic reply is horrible. This applies to the previous date picker and the new date picker introduced by this PR. 

Is it difficult to change the date picker to:
- Use the current date as start & end date. Most likely you don't have to change the year anymore with this change.
- Limit the date range. Let's say from 2020 to 2040. This makes switching the year easier. Now it is difficult to change the year with the mouse due to the scrollbar going from year 1 up to year 9999+.

If it is not easy to change it, I'd say keep it as it is. People can also type in the date (although that is not obvious).

Diman0:
We also do not check for the automatic reply, the end date is really after the start date. I think that is something for a different PR. This PR is for updating the deps.

Diman0:
The admin interface works fine. I can also access rspamd. The automatic tests  confirm that:
- email receiving/sending works.
- aliases work
- clamav&rspamd works
- podop works
- internal endpoint works

I also receive the assert error when running `docker-compose exec admin flask mailu`. As you stated already, this assert error is not shown when you use an actual flask mailu command. It is only displayed when checking all options.

Diman0:
Maybe we can resolve the assert error by adding `@with_appcontext` on line 22 here?
https://github.com/Mailu/Mailu/blob/745c211c4abb17a26930902e8190f19d85b84352/core/admin/mailu/manage.py#L22

ghostwheel42:
> Maybe we can resolve the assert error by adding `@with_appcontext` on line 22 here?
> 
> https://github.com/Mailu/Mailu/blob/745c211c4abb17a26930902e8190f19d85b84352/core/admin/mailu/manage.py#L22

no, unfortunately we can't. I think this must be a bug in flask/click - the help function is a click thing IMHO.

Diman0:
It is only an annoying message when you run `flask mailu` without a command. Normally users don't do this. I don't see this as a show stopper.

The date picker changes should be done in a different PR. This PR is about updating the deps.

Diman0 (approved):
lgtm

PR #2524, opened 2022-11-10
Update the webmail images

### LINKED ISSUE #1521: Build webmails on php:fpm-alpine

Diman0:
Update the webmail images.
Roundcube
  - Switch to base image (alpine)
  - Switch to php-fpm

SnappyMail
  - Switch to base image
  - Upgrade php7 to php8.

## What type of PR?

Feature

## What does this PR do?
Update the webmail images.
Roundcube
  - Switch to base image (alpine)
  - Switch to php-fpm

SnappyMail
  - Switch to base image
  - Upgrade php7 to php8.

### Related issue(s)
- closes #1521

## Prerequisites


webmails/roundcube/start.py
```diff
@@ -60,52 +61,56 @@
...
 try:
-    result = subprocess.check_output(["/var/www/html/bin/initdb.sh", "--dir", "/var/www/html/SQL"],
+    result = subprocess.check_output(["/var/www/webmail/bin/initdb.sh", "--dir", "/var/www/webmail/SQL"],
                                      stderr=subprocess.STDOUT)
     print(result.decode())
 except subprocess.CalledProcessError as exc:
     err = exc.stdout.decode()
     if "already exists" in err:
-        print("Already initialized")
+        print("Already initialzed")
```
nextgens, line 84: ?

webmails/snappymail/Dockerfile
```diff
@@ -74,7 +66,7 @@ RUN cd /var/www/webmail \
 COPY login/include.php /var/www/webmail/include.php
 COPY login/sso.php /var/www/webmail/sso.php
 
-COPY start.py /start.py
+COPY start.py /
```
nextgens, line 69: Elsewhere we specify the full path for the destination

webmails/roundcube/config/php.ini
```diff
@@ -2,3 +2,7 @@ expose_php=Off
 date.timezone={{ TZ }}
 upload_max_filesize = {{ MAX_FILESIZE }}M
 post_max_size = {{ MAX_FILESIZE }}M
+suhosin.session.encrypt=Off
```
nextgens, line 5: As discussed on #mailu-dev I don't think that's a good idea.

I suggest we remove suhosin and add snuffleupagus instead

webmails/roundcube/start.py
```diff
@@ -60,52 +61,56 @@
 context["PLUGINS"] = ",".join(f"'{p}'" for p in plugins)
 
 # add overrides
-context["INCLUDES"] = sorted(inc for inc in os.listdir("/overrides") if inc.endswith((".inc", ".inc.php"))) if os.path.isdir("/overrides") else []
```
nextgens, line 63: We should document it somewhere if we change this

webmails/roundcube/config/nginx-roundcube.conf
```diff
@@ -0,0 +1,62 @@
...
+    location ~ /\. {
+        deny all;
+    }
+
+    location ^~ /data {
+       deny all;
+    }
+
+    location = /ping {
+        allow 127.0.0.1;
```
nextgens, line 54: The healthcheck goes to http://localhost ... that may resolve to an ipv6

webmails/roundcube/Dockerfile
```diff
@@ -1,72 +1,56 @@
...
-COPY config.inc.carddav.php /var/www/html/plugins/carddav/config.inc.php
 
 EXPOSE 80/tcp
 VOLUME /data
 VOLUME /overrides
 
 CMD /start.py
 
-HEALTHCHECK CMD curl -f -L -H 'User-Agent: health' http://localhost/ || exit 1
+HEALTHCHECK CMD curl -f -L http://localhost/ping || exit 1
```
nextgens, line 54: See comment about IPv6 below

webmails/roundcube/start.py
Diman0: I will fix the typo

webmails/snappymail/Dockerfile
Diman0: That was for a temporary test. I forgot to change it back.

webmails/roundcube/config/php.ini
Diman0: I will remove it and will add snuffleupagus. But this does mean we must test if it breaks anything.

webmails/roundcube/start.py
Diman0: I started the PR before the new includes logic. The includes logic with both extensions should be used. I will change this back

webmails/roundcube/config/nginx-roundcube.conf
Diman0: I will add allow ::1

webmails/roundcube/config/php.ini
Diman0: I have removed it, but snuffleupagus is not available for alpine 3.16. the edge repo only has a package for php7. We use php8.

nextgens:
LGTM; I'll let @ghostwheel42 do the ``set -euxo pipefail`` changes

webmails/snappymail/Dockerfile
ghostwheel42: I changed it everywhere to the folder only in the base PR. So IMHO `COPY start.py /` is fine for me.

webmails/roundcube/Dockerfile
ghostwheel42: You removed the User-Agent - do you still filter the checks from the access log?
This can be done in the nginx /ping location

webmails/snappymail/Dockerfile
Diman0: I will make sure it is consistent as nextgens mentioned in the chat

webmails/roundcube/Dockerfile
Diman0: the webmails have the access log disabled.

Diman0:
bors try

nextgens (approved):
LGTM

PR #2525, opened 2022-11-11
Switch to GrapheneOS's hardened_malloc

nextgens:
## What type of PR?

Feature

## What does this PR do?

Switch to GrapheneOS's hardened_malloc

This was suggested during the dev meeting of the 18/09/22.

It may break things and it may make things unbearably slow... but it should also make the exploitation of memory corruption bugs a lot harder.

## Prerequisites

### DIFF
--- a/core/base/Dockerfile
+++ b/core/base/Dockerfile
@@ -12,7 +12,8 @@ ARG MAILU_GID=1000
 RUN set -euxo pipefail \
   ; addgroup -Sg ${MAILU_GID} mailu \
   ; adduser -Sg ${MAILU_UID} -G mailu -h /app -g "mailu app" -s /bin/bash mailu \
-  ; apk add --no-cache bash ca-certificates curl python3 tzdata
+  ; apk add --no-cache bash ca-certificates curl python3 tzdata \
+  ; apk add --repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing hardened-malloc
 
 WORKDIR /app
 
--- a/core/base/Dockerfile
+++ b/core/base/Dockerfile
@@ -14,11 +14,17 @@ RUN set -euxo pipefail \
   ; adduser -Sg ${MAILU_UID} -G mailu -h /app -g "mailu app" -s /bin/bash mailu \
   ; apk add --no-cache bash ca-certificates curl python3 tzdata \
   ; machine="$(uname -m)" \
-  ; ( [[ "${machine}" == x86_64 ]] || \
-      [[ "${machine}" == armv8* ]] || \
-      [[ "${machine}" == aarch64 ]] ) && \
-    echo "Installing hardened-malloc" && \
-    apk add --repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing hardened-malloc
+  ; if [[ "${machine}" == x86_64 || "${machine}" == armv8* || "${machine}" == aarch64 ]] \
+  ; then \
+      echo "Installing hardened-malloc" \
+  ;   apk add --repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing hardened-malloc \
+  ; else \
+      echo "Faking the install of hardened-malloc" \
+  ;   cp /lib/ld-musl-${machine}.so.1 /usr/lib/libhardened_malloc.so \
+  ;   touch /usr/lib/libhardened_malloc.so.faked \
+  ; fi
+
+ENV LD_PRELOAD=/usr/lib/libhardened_malloc.so
 
 WORKDIR /app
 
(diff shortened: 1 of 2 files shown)


nextgens:
You can validate it's in use with the following command:

```
lsof /usr/lib/libhardened_malloc.so
```

HorayNarea:
Are you sure that just installing the lib also enables it as the default malloc? …this is very unlikely since the package only include 2 `.so` files :thinking:

nextgens:
> Are you sure that just installing the lib also enables it as the default malloc? …this is very unlikely since the package only include 2 `.so` files thinking

Have you tried it? I gave a command above to test whether it's in use.

HorayNarea:
Yes I checked:

```
bash-5.1# lsof /usr/lib/libhardened_malloc.so
bash-5.1#
```

If you want to override/preload a library system wide, the dynamic linker has to be told to use this library.
On glibc based distros this is done in `/etc/ld.so.conf` as is documented in the [hardened_malloc repo](https://github.com/GrapheneOS/hardened_malloc#traditional-linux-based-operating-systems)

I believe you got the "how to use hardened_malloc in alpine" and `lsof` command from [this site](https://dustri.org/b/using-hardened_malloc-in-alpine-linux.html), which also contains the part this PR is missing:

> Note that musl doesn't use /etc/ld.so.conf but [/etc/ld-musl-$(ARCH).path](https://musl.libc.org/doc/1.1.24/manual.html) instead.

Unfortunately I could not get the dynamic linker to load hardened_malloc that way.

Only by adding

```Dockerfile
ENV LD_PRELOAD=/usr/lib/libhardened_malloc.so
```

after installing hardened_malloc was I able to get it to actually use it:

```
bash-5.1# lsof /usr/lib/libhardened_malloc.so
COMMAND PID USER  FD   TYPE DEVICE SIZE/OFF NODE NAME
bash      1 root mem    REG   0,77    30640 1902 /usr/lib/libhardened_malloc.so
lsof     10 root mem    REG   0,77    30640 1902 /usr/lib/libhardened_malloc.so
lsof     11 root mem    REG   0,77    30640 1902 /usr/lib/libhardened_malloc.so
```

core/base/Dockerfile
```diff
@@ -12,7 +12,19 @@ ARG MAILU_GID=1000
...
+  ; apk add --no-cache bash ca-certificates curl python3 tzdata \
+  ; machine="$(uname -m)" \
+  ; if [[ "${machine}" == x86_64 || "${machine}" == armv8* || "${machine}" == aarch64 ]] \
+  ; then \
+      echo "Installing hardened-malloc" \
+  ;   apk add --repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing hardened-malloc \
+  ; else \
+      echo "Faking the install of hardened-malloc" \
+  ;   cp /lib/ld-musl-${machine}.so.1 /usr/lib/libhardened_malloc.so \
+  ;   touch /usr/lib/libhardened_malloc.so.faked \
```
ghostwheel42, line 24: I think you can set LD_PRELOAD to /usr/lib/libhardened_malloc.so without any side-effect if it's not present.
Faking is not necessary and will make people using ldd only think it would be enabled when its not.

core/base/Dockerfile
```diff
@@ -12,7 +12,19 @@ ARG MAILU_GID=1000
 RUN set -euxo pipefail \
   ; addgroup -Sg ${MAILU_GID} mailu \
   ; adduser -Sg ${MAILU_UID} -G mailu -h /app -g "mailu app" -s /bin/bash mailu \
-  ; apk add --no-cache bash ca-certificates curl python3 tzdata
+  ; apk add --no-cache bash ca-certificates curl python3 tzdata \
+  ; machine="$(uname -m)" \
+  ; if [[ "${machine}" == x86_64 || "${machine}" == armv8* || "${machine}" == aarch64 ]] \
+  ; then \
+      echo "Installing hardened-malloc" \
+  ;   apk add --repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing hardened-malloc \
```
ghostwheel42, line 20: Can you add a --no-cache here?

ghostwheel42 (changes requested):

core/base/Dockerfile
nextgens: Fair point. I know that other linkers complain when this is invalid... I haven't tested with musl.

Done in 81628149a234985e1fba07881ac3448078ce2bc3

core/base/Dockerfile
nextgens: Sure, that's 9b2f018be615400ed78befdcdd6649e8b815fa8d

ghostwheel42 (approved):
LGTM

PR #2535, opened 2022-11-15
fix the linux/arm/v7 build

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

The arm builder is running aarch64 ... and there is no package for arm/v7

### DIFF
--- a/core/base/Dockerfile
+++ b/core/base/Dockerfile
@@ -19,9 +19,9 @@ RUN set -euxo pipefail \
     && apk add --no-cache --repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing hardened-malloc
 
 ENV LD_PRELOAD=/usr/lib/libhardened_malloc.so
-ENV CXXFLAGS="-g -O2 -fdebug-prefix-map=/app=. -fstack-protector-strong -Wformat -Werror=format-security"
-ENV CFLAGS="-g -O2 -fdebug-prefix-map=/app=. -fstack-protector-strong -Wformat -Werror=format-security"
-ENV CPPFLAGS="-Wdate-time -D_FORTIFY_SOURCE=2"
+ENV CXXFLAGS="-g -O2 -fdebug-prefix-map=/app=. -fstack-protector-strong -Wformat -Werror=format-security -pie -fPIE"
+ENV CFLAGS="-g -O2 -fdebug-prefix-map=/app=. -fstack-protector-strong -Wformat -Werror=format-security -pie -fPIE"
+ENV CPPFLAGS="-Wdate-time -D_FORTIFY_SOURCE=2 -pie -fPIE"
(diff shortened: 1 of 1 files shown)


nextgens:
I have tested it locally on the builder and it seems to do the job

core/base/Dockerfile
```diff
@@ -8,11 +8,13 @@ ENV TZ=Etc/UTC LANG=C.UTF-8
...
+ARG TARGETPLATFORM
 
 RUN set -euxo pipefail \
   ; addgroup -Sg ${MAILU_GID} mailu \
   ; adduser -Sg ${MAILU_UID} -G mailu -h /app -g "mailu app" -s /bin/bash mailu \
   ; apk add --no-cache bash ca-certificates curl python3 tzdata \
+  ; if [[ "$TARGETPLATFORM" == 'linux/arm/v7' ]]; then exit;fi \
   ; machine="$(uname -m)" \
   ; if [[ "${machine}" == x86_64 || "${machine}" == armv8* || "${machine}" == aarch64 ]] \
   ; then \
```
ghostwheel42, line 20: I think this early exit will confuse people adding more commands at the end.
Why not just do it like this:
```
  ; [[ "$TARGETPLATFORM" != 'linux/arm/v7' ]] \
    && machine="$(uname -m)" \
    && [[ "${machine}" == x86_64 || "${machine}" == armv8* || "${machine}" == aarch64 ]] \
    && apk add --no-cache --repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing hardened-malloc
```

ghostwheel42 (changes requested):

core/base/Dockerfile
nextgens: Because the shell isn't bash, it's sh

nextgens: But even with the SH syntax, the ``set -euxo pipefail`` will produce an error as test returns false

ghostwheel42: I don't see bash specifics in my proposal. But yes, it needs to be negated:
```
  ; machine="$(uname -m)"
    ! [[ "${TARGETPLATFORM}" != linux/arm/v7 && ( "${machine}" == x86_64 || "${machine}" == armv8* || "${machine}" == aarch64 ) ]] \
    || apk add --no-cache --repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing hardened-malloc
```

If you don't like this please wrap the TARGETPLATFORM if/then/fi around the other lines, and don't exit early.

nextgens:
@ghostwheel42 ready for another review

ghostwheel42 (dismissed):
lgtm

Diman0:
bors try

Diman0 (approved):
lgtm

Issue #2536, opened 2022-11-17
Rspamd can not start


bsiara:
## Description

I have a problem with start mailu project with rspamd. During start I get error: 

`antispam_1   | 2022-11-17 08:27:25 #1(main) <1b64bb>; main; main: rspamd 3.2 is loading configuration, build id: release
mailu_antispam_1 exited with code 132`

I thinked that problem in my cpu, which don't support avx. I pulled mailu project on target machine and bulded rspamd docker image, but result with the same error.

### Environment

- [ amd64 ] docker-compose
```
cat /proc/cpuinfo 
processor	: 0
vendor_id	: GenuineIntel
cpu family	: 6
model		: 28
model name	: Intel(R) Atom(TM) CPU  230   @ 1.60GHz
stepping	: 2
microcode	: 0x212
cpu MHz		: 1596.847
cache size	: 512 KB
physical id	: 0
siblings	: 2
core id		: 0
cpu cores	: 1
apicid		: 0
initial apicid	: 0
fpu		: yes
fpu_exception	: yes
cpuid level	: 10
wp		: yes
flags		: fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx lm constant_tsc arch_perfmon pebs bts nopl cpuid aperfmperf pni dtes64 monitor ds_cpl tm2 ssse3 cx16 xtpr pdcm movbe lahf_lm dtherm
bugs		:
bogomips	: 3193.67
clflush size	: 64
cache_alignment	: 64
address sizes	: 32 bits physical, 48 bits virtual
power management:

processor	: 1
vendor_id	: GenuineIntel
cpu family	: 6
model		: 28
model name	: Intel(R) Atom(TM) CPU  230   @ 1.60GHz
stepping	: 2
microcode	: 0x212
cpu MHz		: 1596.836
cache size	: 512 KB
physical id	: 0
siblings	: 2
core id		: 0
cpu cores	: 1
apicid		: 1
initial apicid	: 1
fpu		: yes
fpu_exception	: yes
cpuid level	: 10
wp		: yes
[...]

PR #2539, opened 2022-11-17
Upgrade alpine, make setup use the base image

nextgens:
## What type of PR?

enhancement

## What does this PR do?

Upgrade alpine, make setup use the base image, introduce a health-check, drop privileges. Drop privileges on admin too.

It may or may not help #2536


nextgens:
bors try

nextgens:
See if it helps with the warnings.
bors try

nextgens:
Ask bors if that broke anything.

bors try

nextgens:
Try the new test too.

bors try

nextgens:
bors try

nextgens:
bors try

Diman0:
bors try

It looks good to me. I will do some smoke testing before approving it.

Diman0 (changes requested):
admin uses /mailu/data.
fetchmail uses /mailu/data/fetchmail
So the chown for admin, changes the file permissions for fetchmail. It appears not to result in problematic issues for fetchmail. The real solution is to use `/mailu/data/admin` for admin in my opinion. Of course this means I must mention in the release notes that users must manually mv these files when upgrading to the new mailu version. If you choose to do this, the newsfragment should also warn about this.

For rspamd we must also use chown to change permissions for the host mounted folders:
- /var/lib/rspamd (filter folder in mailu data folder)
  - os.system("chown rspamd:rspamd -R /var/lib/rspamd")
  - a problem is that antivirus uses the EXACT same folder /mailu/filter. Maybe this PR is the time to introduce the folder /mailu/filter/antivirus & /mailu/filter/fetchmail?
- /etc/rspamd/override.d (/overrides/rspamd in mailu data folder
  - os.system("chown rspamd:rspamd -R /etc/rspamd/override.d")
  - This is currently not possible, because we mount  the overrides folder as read-only. We could ignore this and then it is the user's problem to make sure it is readable? Or we could remove the :ro flag on the host mount.

nextgens:
I have done the minimum in 44c4758; I suggest we rethink the folder hierarchy and instruction to users in another PR

bors try

nextgens:
bors try

Diman0 (approved):
lgtm

PR #2556, opened 2022-11-24
deprecate swarm in 1.9

nextgens:
This makes it clear that Swam shouldn't be used for new deployments.

I will send another PR against master to get rid of it completely.


ghostwheel42 (approved):
LGTM

PR #2558, opened 2022-11-24
Don't do it as root

nextgens:
A naive attempt to ensure we don't run the PHP stuff as root; without it we mess the permissions up and fail to upgrade the database schema of roundcube


nextgens:
bors try

ghostwheel42 (approved):

PR #2561, opened 2022-11-25
set the umask when using config-export

nextgens:
## What type of PR?

enhancement

## What does this PR do?

Set a better umask when using config-export


ghostwheel42 (approved):
LGTM

PR #2562, opened 2022-11-26
Dynamic address resolution everywhere

### LINKED ISSUE #1013: Refactor the way our containers communicate

### LINKED ISSUE #1341: HOST_* and *_ADDRESS resolution errors for docker-compose deployments

### LINKED ISSUE #1430: mailu/nginx:1.7 does not start

nextgens:
## What type of PR?

enhancement

## What does this PR do?

Use dynamic address resolution everywhere.
Derive a new key for admin/SECRET_KEY
Cleanup the environment

This should allow restarting containers.

### Related issue(s)
- closes #1341
- closes #1013
- closes #1430

## Prerequisites

### DIFF
--- a/core/admin/mailu/configuration.py
+++ b/core/admin/mailu/configuration.py
@@ -1,7 +1,6 @@
 import os
 
 from datetime import timedelta
-from socrate import system
 import ipaddress
 
 DEFAULT_CONFIG = {
@@ -83,17 +82,6 @@ DEFAULT_CONFIG = {
     'PROXY_AUTH_WHITELIST': '',
     'PROXY_AUTH_HEADER': 'X-Auth-Email',
     'PROXY_AUTH_CREATE': False,
-    # Host settings
-    'HOST_IMAP': 'imap',
-    'HOST_LMTP': 'imap:2525',
-    'HOST_POP3': 'imap',
-    'HOST_SMTP': 'smtp',
-    'HOST_AUTHSMTP': 'smtp',
-    'HOST_ADMIN': 'admin',
-    'HOST_WEBMAIL': 'webmail',
-    'HOST_WEBDAV': 'webdav:5232',
-    'HOST_REDIS': 'redis',
-    'HOST_FRONT': 'front',
     'SUBNET': '192.168.203.0/24',
     'SUBNET6': None
 }
@@ -110,19 +98,6 @@ class ConfigManager:
 
     def __init__(self):
         self.config = dict()
-
-    def get_host_address(self, name):
-        # if MYSERVICE_ADDRESS is defined, use this
-        if f'{name}_ADDRESS' in os.environ:
-            return os.environ.get(f'{name}_ADDRESS')
-        # otherwise use the host name and resolve it
-        return system.resolve_address(self.config[f'HOST_{name}'])
-
-    def resolve_hosts(self):
-        for key in ['IMAP', 'POP3', 'AUTHSMTP', 'SMTP', 'REDIS']:
-            self.config[f'{key}_ADDRESS'] = self.get_host_address(key)
-        if self.config['WEBMAIL'] != 'none':
-            self.config['WEBMAIL_ADDRESS'] = self.get_host_address('WEBMAIL')
 
     def __get_env(self, key, value):
         key_file = key + "_FILE"
@@ -144,11 +119,14 @@ class ConfigManager:
         # get current app config
         self.config.update(app.config)
         # get environment variables
+        for key in os.environ:
+            if key.endswith('_ADDRESS'):
+                self.config[key] = os.environ[key]
+
         self.config.update({
             key: self.__coerce_value(self.__get_env(key, value))
             for key, value in DEFAULT_CONFIG.items()
         })
-        self.resolve_hosts()
 
         # automatically set the sqlalchemy string
         if self.config['DB_FLAVOR']:
--- a/core/base/libs/socrate/socrate/system.py
+++ b/core/base/libs/socrate/socrate/system.py
@@ -28,7 +28,7 @@ def set_env(required_secrets=[]):
     secret_key = os.environ.get('SECRET_KEY')
     if not secret_key:
         try:
-            secret_key = open(env.get("SECRET_KEY_FILE"), "r").read().strip()
+            secret_key = open(os.environ.get("SECRET_KEY_FILE"), "r").read().strip()
         except Exception as exc:
             log.error(f"Can't read SECRET_KEY from file: {exc}")
             raise exc
--- a/core/base/libs/socrate/socrate/system.py
+++ b/core/base/libs/socrate/socrate/system.py
@@ -15,6 +15,14 @@ def resolve_hostname(hostname):
     except Exception as e:
         log.warn("Unable to lookup '%s': %s",hostname,e)
         raise e
+
+def _coerce_value(value):
+    if isinstance(value, str) and value.lower() in ('true','yes'):
+        return True
+    elif isinstance(value, str) and value.lower() in ('false', 'no'):
+        return False
+    return value
+
 def set_env(required_secrets=[]):
     """ This will set all the environment variables and retains only the secrets we need """
     secret_key = os.environ.get('SECRET_KEY')
@@ -28,6 +36,11 @@ def set_env(required_secrets=[]):
     # derive the keys we need
     for secret in required_secrets:
         os.environ[f'{secret}_KEY'] = hmac.new(bytearray(secret_key, 'utf-8'), bytearray(secret, 'utf-8'), 'sha256').hexdigest()
+
+    return {
+            key: _coerce_value(os.environ.get(key, value))
+            for key, value in os.environ.items()
+           }
 
 def clean_env():
     """ remove all secret keys """
--- a/core/base/libs/socrate/socrate/system.py
+++ b/core/base/libs/socrate/socrate/system.py
@@ -1,7 +1,8 @@
+import hmac
+import logging as log
+import os
 import socket
 import tenacity
-from os import environ
-import logging as log
 
 @tenacity.retry(stop=tenacity.stop_after_attempt(100),
                 wait=tenacity.wait_random(min=2, max=5))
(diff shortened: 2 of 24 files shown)


nextgens:
bors try

Diman0:
bors retry

webmails/start.py
```diff
@@ -13,14 +13,13 @@
 env = os.environ
 
 logging.basicConfig(stream=sys.stderr, level=env.get("LOG_LEVEL", "WARNING"))
+system.set_env(['ROUNDCUBE','SNUFFLEUPAGUS'])
```
lub, line 16: Are multiple webmails even supported? Otherwise we could just use something like `WEBMAIL` instead.

webmails/start.py
```diff
@@ -43,17 +42,6 @@
...
     exit(1)
 
-# derive roundcube secret key
-secret_key = env.get("SECRET_KEY")
-if not secret_key:
-    try:
-        secret_key = open(env.get("SECRET_KEY_FILE"), "r").read().strip()
-    except Exception as exc:
-        print(f"Can't read SECRET_KEY from file: {exc}", file=sys.stderr)
-        exit(2)
```
lub, line 53: As discussed in [chat](https://matrix.to/#/!MINuyJjJSrfowljYCK:tedomum.net/$1669479122105539aLcXz:matrix.org?via=tedomum.net&via=matrix.org&via=tchncs.de) I think this would break setups that use SECRET_KEY_FILE, but except the newsfragment/changelog entry it's currently undocumented anyways.

Alternatively I guess it should be also possible to move this to `base/libs/socrate/system.py`

lub (commented):
I've added some comments with suggestions. Good idea to put some things in a central lib.

Haven't tested it on my own installation, yet.

optional/fetchmail/fetchmail.py
```diff
@@ -79,7 +61,7 @@ def run(debug):
                 protocol=fetch["protocol"],
                 host=escape_rc_string(fetch["host"]),
                 port=fetch["port"],
-                smtphost=smtphostport if fetch['scan'] else lmtphostport,
+                smtphost=f'{os.environ["FRONT_ADDRESS"]}:25' if fetch['scan'] else f'{os.environ["IMAP_ADDRESS"]}:2525',
```
lub, line 64: Before `smtphostport` pointed to `SMTP_HOST`

```suggestion
                smtphost=f'{os.environ["SMTP_ADDRESS"]}' if fetch['scan'] else f'{os.environ["IMAP_ADDRESS"]}:2525',
```

nextgens: Done in d5f1f80e ; the syntax with ":" wasn't correct anyways

webmails/start.py
nextgens: The idea is to have one key per usage. If we have different pieces requiring keys we generate a different key each time... the key name itself doesn't matter.

optional/fetchmail/fetchmail.py
```diff
@@ -118,8 +100,9 @@ def run(debug):
     os.chmod("/data/fetchids", 0o700)
     os.setgid(id_fetchmail.pw_gid)
     os.setuid(id_fetchmail.pw_uid)
+    system.set_env()
     while True:
-        delay = int(os.environ.get("FETCHMAIL_DELAY", 60))
+        delay = int(os.environ.get("FETCHMAIL_DELAY", 900))
```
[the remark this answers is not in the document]
nextgens: This is the default in case the variable can't be parsed. We define it to 600 in the default config and that's what should will be used (rather than the fallback value here)

docs/configuration.rst
```diff
@@ -249,32 +249,23 @@ virus mails during SMTP dialogue, so the sender will receive a reject message.
...
-resolve the hostnames.
+- ``ADMIN_ADDRESS``
+- ``ANTISPAM_ADDRESS``
+- ``ANTIVIRUS_ADDRESS``
+- ``FRONT_ADDRESS``
+- ``IMAP_ADDRESS``
+- ``REDIS_ADDRESS``
+- ``SMTP_ADDRESS``
+- ``WEBDAV_ADDRESS``
+- ``WEBMAIL_ADDRESS``
```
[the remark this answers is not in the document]
nextgens: IMHO that's a different PR. We should probably craft a SELinux policy to make it explicit... but I personally don't see the point in documenting configurations we explicitly say are not supported.

core/admin/mailu/configuration.py
```diff
@@ -111,18 +99,9 @@ class ConfigManager:
...
-        # otherwise use the host name and resolve it
-        return system.resolve_address(self.config[f'HOST_{name}'])
-
     def resolve_hosts(self):
-        for key in ['IMAP', 'POP3', 'AUTHSMTP', 'SMTP', 'REDIS']:
-            self.config[f'{key}_ADDRESS'] = self.get_host_address(key)
-        if self.config['WEBMAIL'] != 'none':
-            self.config['WEBMAIL_ADDRESS'] = self.get_host_address('WEBMAIL')
+        for key in ['ADMIN', 'FRONT', 'SMTP', 'IMAP', 'REDIS', 'ANTIVIRUS:', 'ANTISPAM', 'WEBMAIL', 'WEBDAV']:
+            self.config[f'{key}_ADDRESS'] = os.environ.get(f'{key}_ADDRESS')
```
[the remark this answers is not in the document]
nextgens: I've removed resolve_hosts in 9be5c69f. Yes we only use a subset... the idea of defining them all here was to make it easy for the k8s/helm people to know what is used.

webmails/start.py
nextgens: I've reintroduced it in 09f3cebe

nextgens:
Looks like I have made a mess of the commit history :/

bors try

nextgens:
bors try

nextgens:
bors try

Diman0 (dismissed):
In `core/base/libs/socrate/test.py` the unit tests for resolve_address and get_host_address_from_environment can be removed.

What does the patch for the managesieve roundcube plugin do? Why is it required? 

We are using static hostname assignments in the Dockerfile of the base image (e.g. ADMIN_ADDRESS=admin). This means this will break for anyone using non-standard service names in their docker-compose file or whatever orchestrator they use. I'm fine with this. They can use the *_ADDRESS env vars. But maybe we should mention this in the release notes.

Everything else looks fine. It is wonderful to restart the mailu stack &, see that all IPs have changed, and then still have a functioning mailu stack :+1:

nextgens:
> In `core/base/libs/socrate/test.py` the unit tests for resolve_address and get_host_address_from_environment can be removed.
> 

ACK

> What does the patch for the managesieve roundcube plugin do? Why is it required?
> 

It wasn't meant to be part of this PR... I've submitted a separate PR for it with the appropriate explanation. Do you want me to force-push the change to get rid of the commit?

> We are using static hostname assignments in the Dockerfile of the base image (e.g. ADMIN_ADDRESS=admin). This means this will break for anyone using non-standard service names in their docker-compose file or whatever orchestrator they use. I'm fine with this. They can use the *_ADDRESS env vars. But maybe we should mention this in the release notes.
> 

There will be quite a few things to mention: should we start a draft for it somewhere?

> Everything else looks fine. It is wonderful to restart the mailu stack &, see that all IPs have changed, and then still have a functioning mailu stack +1

Agreed :)

nextgens:
bors try

Diman0:
> > In `core/base/libs/socrate/test.py` the unit tests for resolve_address and get_host_address_from_environment can be removed.
> 
> ACK
> 
> > What does the patch for the managesieve roundcube plugin do? Why is it required?
> 
> It wasn't meant to be part of this PR... I've submitted a separate PR for it with the appropriate explanation. Do you want me to force-push the change to get rid of the commit?
> 
> > We are using static hostname assignments in the Dockerfile of the base image (e.g. ADMIN_ADDRESS=admin). This means this will break for anyone using non-standard service names in their docker-compose file or whatever orchestrator they use. I'm fine with this. They can use the *_ADDRESS env vars. But maybe we should mention this in the release notes.
> 
> There will be quite a few things to mention: should we start a draft for it somewhere?
> 
> > Everything else looks fine. It is wonderful to restart the mailu stack &, see that all IPs have changed, and then still have a functioning mailu stack +1
> 
> Agreed :)

You force pushed the changes away (the sieve patch). That is fine with me. It is better to use the other PR for this. I will approve the PR as soon as BORS try finishes.

lub (changes requested):
One suggestion, otherwise seems to look fine

core/base/libs/socrate/socrate/system.py
```diff
@@ -15,24 +16,32 @@ def resolve_hostname(hostname):
...
-    ip_address = resolve_hostname(hostname)
-    if ":" in ip_address:
-        ip_address = "[{}]".format(ip_address)
-    return ip_address + "".join(":" + port for port in rest)
+def set_env(required_secrets=[]):
+    """ This will set all the environment variables and retains only the secrets we need """
+    secret_key = os.environ.get('SECRET_KEY')
+    if not secret_key:
+        try:
+            secret_key = open(env.get("SECRET_KEY_FILE"), "r").read().strip()
```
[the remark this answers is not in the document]
nextgens: Doh; that's df924b08

Diman0 (approved):
LGTM

PR #2563, opened 2022-11-27
Close #1483: remove postfix's /queue/pid/master.pid

### LINKED ISSUE #1483: postfix throttling afer a hard shutdown
After a hard shutdown, I noticed that postfix would not restart because of master.pid is found in /queue/pid/master.pid
We should check and remove this file during container start up (start.py)

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Remove postfix's /queue/pid/master.pid on startup if there is no other instance running.

### Related issue(s)
- closes #1483

## Prerequisites


Diman0 (approved):
LGTM

PR #2565, opened 2022-11-27
Fix DB downgrade

nextgens:
Fix DB downgrade. This isn't used in normal operations but is wrong nevertheless.


ghostwheel42 (approved):
LGTM

PR #2566, opened 2022-11-27
Make it clear that we don't delete users

### LINKED ISSUE #1820: [Enhancement] List of previously used addresses/aliases to block accidental reuse and privacy implications
I did search for existing issues first. Not quite sure how this feature would be worded so I may have missed a discussion.

I propose a simple feature, definitely easily overrideable if need be and with a toggle to turn it off, that permanently keeps track of used addresses so you cannot accidentally reuse a previously used one and leak data.

Address reuse is a problem for obvious reasons:
- intermingling different identities
- therefore leaking private information 

Storing the addresses can be done privacy friendly: simply store exclusively the address itself, no metadata.

On a side note: Deleting accounts does not delete the actual folders. I am doing this manually so far, but it'd be nice if *delete* actually meant *delete*.

As usual I have to praise how well Mailu works in general. I realized I've been using it for years now, and apart from some inconveniences like this or #1700 #1328 there hasn't been an issue, and it's more lightweight than the competitors. I wish you all the best!

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Make it clear that we don't delete users. Users can and should be disabled when not in use anymore.

### Related issue(s)
- closes #1820

## Prerequisites


Diman0 (dismissed):
The newsfragment/changelog is missing.

With these changes it is only possible to remove users via the config import/export or the restful api. That is good. There should still be an option to delete an user via an advanced method.

Shouldn't we document somewhere that it's not possible to delete users?
Maybe as a question/answer in the FAQ section?
It makes sense to me that users will be confused that there is no delete button anymore. Having a delete button is pretty basic functionality after all.

I'm fine with creating the new FAQ entry that will explain how users can still be deleted and what manual post-deletion tasks are required (delete the imap folder of the user on file system).

nextgens:
> The newsfragment/changelog is missing.
> 

Fixed

> With these changes it is only possible to remove users via the config import/export or the restful api. That is good. There should still be an option to delete an user via an advanced method.
> 

I disagree; I don't think the REST API should allow user deletion.

> Shouldn't we document somewhere that it's not possible to delete users? Maybe as a question/answer in the FAQ section? It makes sense to me that users will be confused that there is no delete button anymore. Having a delete button is pretty basic functionality after all.
> 
> I'm fine with creating the new FAQ entry that will explain how users can still be deleted and what manual post-deletion tasks are required (delete the imap folder of the user on file system).

That's a good idea

ghostwheel42:
I think we need to create a process for user deletion.
Mailu's database schema is designed with "on delete cascade", so deleting a domain would also delete all users while keeping the mail folders in place.
This is one reason why it takes me so long to chnage this as we decided in out meeting.
I think - at the end - deletion should be possible via API. But this needs some code in the backend (imap container) to handle it.

ghostwheel42:
The original issue is also talking about aliases. 
I think handling the whole "block after deletion" process would need a bigger change in how mailu handled this (extra table with blocked addresses and Grace periods also showing the original owner to re-enable an accidentally deleted address).

@Diman0  Maybe we could document on how to set an alias to refuse incoming emails instead of deleting. This is possible when there's no target configured, isn't it?

Diman0:
I added a draft text for the FAQ entry in https://demo.hedgedoc.org/TG_6KhdWRuaUPY-tFBOFIQ?both#.
Please add feedback via that page. Ping me here or in the mailu dev chat when you have added your 2 cents.

Diman0:
@ghostwheel42  I could not use the script. It contained errors. `docker-compose` was incorrectly written as `docker compose`. This caused the script to fail. It worked for me after changing `docker compose` to `docker-compose`.

I reckon you are using docker compose v2? 
I have no experience with docker compose v2 yet.
At this moment, we still require docker-compose (https://mailu.io/master/compose/requirements.html). I see on the docker website that docker-compose (v1) is deprecated. I'm fine with switching to docker-compose v2. I guess we need a new PR to update all instructions to use docker compose v2 ( the requirements section, examples on the CLI page, the setup utility etc).

Diman0:
bors try

Diman0 (changes requested):
The script does not contain the `sudo` command for each `rm` command. This can fail if the user is not the owner of the maildirs. 
I will update the faq entry that you need to run the commands with root privileges.

@ghostwheel42 
The script can only find disabled users who have a mailbox. Someone must have accessed his email at least once and then the user must have been disabled. Only then the user is found and the script provides the delete commands.

I think this is acceptable? I will explicitly document this in the FAQ entry.

@nextgens @ghostwheel42 Please provide your opinion.

docs/faq.rst
```diff
@@ -393,6 +393,46 @@ Technical issues
...
+How to delete users?
+````````````````````
+
+From the web administration interface, when a user is deleted, the user is only disabled. When a user is not enabled, this user:
+
+* cannot send/receive email
+* cannot access Mailu (admin/webmail)
+* cannot access the email box via pop3/imap
+
+It is not possible to delete users via the Mailu web administration interface. The main reason is to prevent email address reusage. If a user was deleted, it can be recreated and used by someone else. It is not clear that the email address has been used by someone else previously. This new user might receive emails which were meant to be received by the previous user. Disabling the user, prevents the email address to be reused by mistake.
```
nextgens, line 407:
```suggestion
It is not possible to delete users via the Mailu web administration interface. The main reason is to prevent email address reuse. If a user was deleted, it can be recreated and used by someone else. It is not clear that the email address has been used by someone else previously. This new user might receive emails which were meant for the previous user. Disabling the user, prevents the email address to be reused by mistake.
```

nextgens:
I wouldn't change anything to involve sudo. Test whether the script runs as root, error out if it it doesn't... but don't force people to use sudo please.

@Diman0  I don't see what kind of data we would delete for users who do not have a mailbox. The script does provides an option to delete files unknown to Mailu.

Diman0:
> @Diman0 I don't see what kind of data we would delete for users who do not have a mailbox. The script does provides an option to delete files unknown to Mailu.

@nextgens Well we would delete the user from mailu. You cannot delete users via admin anymore. I will just add this as note to the FAQ entry. 

Regarding sudo, I'm not talking about the script, but the commands provided by the script.  Should I add a note that you must actually run the commands as root? 
E.g.
`rm -rf '/mailu/mail/test2@example.com'`
Normally the mail folder is owned by root and the above fails. Do we expect the user to know that he must add sudo or switch to the root account? Or will they create issues that the commands do not work?

nextgens:
> > @Diman0 I don't see what kind of data we would delete for users who do not have a mailbox. The script does provides an option to delete files unknown to Mailu.
> 
> @nextgens Well we would delete the user from mailu. You cannot delete users via admin anymore. I will just add this as note to the FAQ entry.
> 

Ah. I think that's an edge-case that we don't need to be concerned about.

> Regarding sudo, I'm not talking about the script, but the commands provided by the script. Should I add a note that you must actually run the commands as root? E.g. `rm -rf '/mailu/mail/test2@example.com'` Normally the mail folder is owned by root and the above fails. Do we expect the user to know that he must add sudo or switch to the root account? Or will they create issues that the commands do not work?

I personally think that it's fine like this... but if it bothers you, please do.

Diman0:
Ok, then we leave it as it is.  I fixed  the typo you found (thank you for spotting). As a sanity check lets run bors try. Afer that I will approve the PR.

bors try

Diman0 (approved):
LGTM

elandorr:
Excuse me for not reading the entire discussion, I only got an e-mail for #1820, but I'd like to say I definitely want to be able to delete users/domains/anything. The entire point of self-hosting is regaining control, after all.

It's very relevant to privacy: Sometimes you want to have something erased 100%, no traces.

It's also a matter of metadata, 'disabling' doesn't get rid of that.

My original post in 2021 seems simple enough: An extra table with a one-way list only, no additional metadata, as it's not required. I simply kept such a table myself until now.

Another thing that comes to mind is storage: When I wrote that post deletion did not delete anything, meaning you had all kinds of generated data in the folder (indices and whatnot, settings, filters etc). Presumably disabling doesn't clean any of that, either.

Diman0:
It does much more than that.  See [this link](https://mailu.io/master/faq.html#how-to-delete-users) for more information. All data of users can be deleted now. This includes the data from the roundcube database as well.

elandorr:
@Diman0 Thanks! I tried on 1.9.50. Made a new account, logged into rainloop to have the folder created, and 'deleted' it via webadmin. (if it's just disabling, mind making it clear by using the word 'disable'?)

The script now returns the commands as stated in the docs. But even before running anything, I can still re-create a new account with the same name even before.

Am I missing something, or is this a bug? I feel like I'm not on the current version somehow. In a previous changelog you added a version display, and that also does not show up, but all the images are 1.9.50.

elandorr:
Thanks! As long as we get to fully delete stuff when needed, we're good. Will have to play with it when available to understand how you intended it to work in practice. Sometimes you want to keep an old identity blacklisted, but still delete every other trace etc.

I don't quite understand how to match changes to released versions exactly, but I'll just wait for a release. Have to admit I never quite looked through your 'towncrier' system. @Diman0

PR #2568, opened 2022-11-29
Fix a bug preventing users without IMAP access to access the webmails

### LINKED ISSUE #2451: not allowing POP3/IMAP leads to infinite loop in webmail
v1.9.32

I noticed a small bug. If both are disabled, webmail is stuck in an infinite loop. I guess nobody ever tried it before since both are checked by default.

Not very consequential, but I figured you might want to know. Not sure about the use case either. I unchecked them because there was no need for this particular account and found it that way.

Cheers

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Fix a bug preventing users without IMAP access to access the webmails

### Related issue(s)
- close #2451

## Prerequisites

### DIFF
--- a/core/admin/mailu/internal/nginx.py
+++ b/core/admin/mailu/internal/nginx.py
@@ -26,12 +26,14 @@ STATUSES = {
     }),
 }
 
+WEBMAIL_PORTS = ['10143', '10025']
+
 def check_credentials(user, password, ip, protocol=None, auth_port=None):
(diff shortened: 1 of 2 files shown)


ghostwheel42 (commented):
Shouldn't we also add a feature "webmail access" to effectively split webmail from imap access?

nextgens:
We could but then it wouldn't be back-portable. IMHO it should be another PR if we do it.

ghostwheel42 (approved):
LGTM

But we should gather constants like WEBMAIL_PORTS in one central place (in another PR)

PR #2576, opened 2022-12-07
Add net_bind_service capability for python executable

fastlorenzo:
## What type of PR?

bug-fix

## What does this PR do?

Fixes capabilities needed to bind on privileged port.

### DIFF
--- a/core/base/Dockerfile
+++ b/core/base/Dockerfile
@@ -80,8 +80,7 @@ FROM system
 
 COPY --from=build /app/venv/ /app/venv/
 COPY --chown=root:root --from=build /app/snuffleupagus.so /usr/lib/php81/modules/
-RUN setcap 'cap_net_bind_service=+ep' /app/venv/bin/gunicorn \
-  ; setcap 'cap_net_bind_service=+ep' /usr/bin/python3.10
+RUN setcap 'cap_net_bind_service=+ep' /app/venv/bin/gunicorn 'cap_net_bind_service=+ep' /usr/bin/python3.10
 
 ENV VIRTUAL_ENV=/app/venv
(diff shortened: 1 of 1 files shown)


nextgens (dismissed):

core/base/Dockerfile
```diff
@@ -80,7 +80,8 @@ FROM system
 
 COPY --from=build /app/venv/ /app/venv/
 COPY --chown=root:root --from=build /app/snuffleupagus.so /usr/lib/php81/modules/
-RUN setcap 'cap_net_bind_service=+ep' /app/venv/bin/gunicorn
+RUN setcap 'cap_net_bind_service=+ep' /app/venv/bin/gunicorn \
```
Diman0, line 83: When multiple commands are run, we use the `set -euxo pipeline` construct. This ensures we see the actual command that fails and the full command will be printed. For more info see https://gist.github.com/mohanpedala/1e2ff5661761d3abd0385e8223e16425?permalink_comment_id=3945021

You could change it to:
```
RUN set -euxo pipefail \
  ; setcap 'cap_net_bind_service=+ep' /app/venv/bin/gunicorn \
  ; setcap 'cap_net_bind_service=+ep' /usr/bin/python3.10
```

Diman0 (approved):
see in-line comment

Diman0 (dismissed):
see in-line comments

core/base/Dockerfile
nextgens: No need for two commands even. [dfaba5b](https://github.com/Mailu/Mailu/pull/2576/commits/dfaba5bb17d9e460ce43cf12b9b7033900ac425d)

nextgens (approved):

PR #2580, opened 2022-12-12
Fixed roundcube carddav module

fastlorenzo:
## What type of PR?

bug-fix

## What does this PR do?

This makes the Carddav module of roundcube to work again.

Changes made:
- Add 2 missing packages in the container (`php81-xmlreader` and `php81-xmlwriter`)
- Disable one rule in snuffleupagus that blocked the web request needed from the plugin to interact with carddav

### DIFF
--- a/webmails/snuffleupagus.rules
+++ b/webmails/snuffleupagus.rules
@@ -84,6 +84,7 @@ sp.disable_function.function("ini_set").param("option").value("include_path").dr
 sp.disable_function.function("ini_set").param("option").value("open_basedir").drop();
 
 # Detect some backdoors via environment recon
+sp.disable_function.function("ini_get").filename("/var/www/roundcube/vendor/guzzlehttp/guzzle/src/functions.php").param("option").value("allow_url_fopen").allow();
 sp.disable_function.function("ini_get").param("option").value("allow_url_fopen").drop();
 sp.disable_function.function("ini_get").param("option").value("open_basedir").drop();
 sp.disable_function.function("ini_get").param("option").value_r("suhosin").drop();
@@ -97,7 +98,7 @@ sp.disable_function.function("is_callable").param("value").value("eval").drop();
 sp.disable_function.function("is_callable").param("value").value("exec").drop();
 sp.disable_function.function("is_callable").param("value").value("system").drop();
 sp.disable_function.function("is_callable").param("value").value("shell_exec").drop();
-sp.disable_function.function("is_callable").filename_r("/app/libraries/snappymail/pgp/gpg\.php$").param("value").value("proc_open").allow();
+sp.disable_function.function("is_callable").filename_r("^/var/www/snappymail/snappymail/v/\d+\.\d+\.\d+/app/libraries/snappymail/pgp/gpg\.php$").param("value").value("proc_open").allow();
 sp.disable_function.function("is_callable").param("value").value("proc_open").drop();
 sp.disable_function.function("is_callable").param("value").value("passthru").drop();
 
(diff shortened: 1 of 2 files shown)


webmails/snuffleupagus.rules
```diff
@@ -84,7 +84,7 @@ sp.disable_function.function("ini_set").param("option").value("include_path").dr
 sp.disable_function.function("ini_set").param("option").value("open_basedir").drop();
 
 # Detect some backdoors via environment recon
-sp.disable_function.function("ini_get").param("option").value("allow_url_fopen").drop();
+# sp.disable_function.function("ini_get").param("option").value("allow_url_fopen").drop(); # Disabled as this is necessary for carddav plugin to work
```
nextgens, line 87: Please enable it only for the required file instead of disabling it globally.
https://snuffleupagus.readthedocs.io/config.html#virtual-patching

nextgens (dismissed):

webmails/snuffleupagus.rules
fastlorenzo: Done ;)

fastlorenzo:
FYI this was the error: 
```
[12-Dec-2022 14:26:59] WARNING: [pool php] child 8465 said into stderr: "NOTICE: PHP message: PHP Fatal error:  [snuffleupagus][0.0.0.0][disabled_function][drop] Aborted execution on call of the function 'ini_get', because its argument '$option' content (allow_url_fopen) matched a rule in /var/www/roundcube/vendor/guzzlehttp/guzzle/src/functions.php on line 114"
```

webmails/snuffleupagus.rules
```diff
@@ -84,6 +84,7 @@ sp.disable_function.function("ini_set").param("option").value("include_path").dr
 sp.disable_function.function("ini_set").param("option").value("open_basedir").drop();
 
 # Detect some backdoors via environment recon
+sp.disable_function.function("ini_get").filename_r("/var/www/roundcube/vendor/guzzlehttp/guzzle/src/functions\.php$").param("option").value("allow_url_fopen").allow();
```
nextgens, line 87: Please use filename() instead; no need for the regexp engine here.

nextgens (changes requested):

webmails/snuffleupagus.rules
fastlorenzo: I did first, but then noticed on line 101 you used filename_r and wanted to keep consistent: https://github.com/Mailu/Mailu/blob/d0061b95e8dda06ee79d2337b15cbbb1d123ffd9/webmails/snuffleupagus.rules#L101

Any reason for this?

nextgens: That has an actual regexp; in snappymail the path changes for each version.

fastlorenzo: @nextgens sorry but I don't see the version check in the following string: `/app/libraries/snappymail/pgp/gpg\.php$`
Am I missing something or is it a mistake?

nextgens: The version is on the left side of the path... that's why this is a regexp.

The full path is:
``/var/www/snappymail/snappymail/v/2.21.3/app/libraries/snappymail/pgp/gpg.php``

nextgens: If you want to change it to be more specific, be my guest :)

``^/var/www/snappymail/snappymail/v/\d+\.\d+\.\d+/app/libraries/snappymail/pgp/gpg\.php$``

fastlorenzo: gotcha now, thanks :) I'll make it more specific then so it's more clear ;)

nextgens (approved):
LGTM, if you could confirm that snappymail's carddav also works that'd be great.

fastlorenzo:
> LGTM, if you could confirm that snappymail's carddav also works that'd be great.

let me check it as well

Diman0 (approved):

PR #2581, opened 2022-12-14
fix missing casting to int for SESSION_KEY_BITS

fastlorenzo:
## What type of PR?

bug-fix

## What does this PR do?

This PR adds a missing env var casting for the `SESSION_KEY_BITS` variable.
When trying to provide a different value via env var, the value is passed as a string and then compared to a int.
The following check then throws a cast error: https://github.com/Mailu/Mailu/blob/50c7fa882e0a2cea57cecdd4b9c89bba13362f5a/core/admin/mailu/utils.py#L309-L312

## Prerequisites


nextgens (approved):
LGTM

ghostwheel42 (approved):

PR #2587, opened 2022-12-19
fix roundcube/sieve

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Without this snuffleupagus is throwing a tantrum on ini_get(), when saving a sieve filter from roundcube.

```
[17-Dec-2022 13:44:08] WARNING: [pool php] child 21853 said into stderr: "NOTICE: PHP message: PHP Fatal error:  [snuffleupagus][0.0.0.0][disabled_function][drop] Ab
orted execution on call of the function 'ini_get', because its argument '$option' content (suhosin.request.max_vars) matched a rule in /var/www/roundcube/plugins/man
agesieve/lib/Roundcube/rcube_sieve_engine.php on line 532"
```

## Prerequisites

### DIFF
--- a/webmails/Dockerfile
+++ b/webmails/Dockerfile
@@ -7,10 +7,11 @@ LABEL version=$VERSION
 
 COPY snappymail/pubkey.asc /tmp/snappymail.asc
 COPY roundcube/pubkey.asc /tmp/roundcube.asc
+COPY roundcube/roundcube.diff /tmp/roundcube.diff
 
 RUN set -euxo pipefail \
   ; apk add --no-cache \
-    nginx gpg gpg-agent \
+    nginx gpg gpg-agent patch \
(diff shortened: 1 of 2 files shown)


Diman0:
@nextgens 
This is already (unintentionally?) part of PR 2562. I'm fine with keeping it in PR 2562 and closing this, or removing it in PR2562 and merging this PR.

Now I at least now the purpose of this patch. I was already wondering why this was added for PR2562 which is for DNS based resolution.

Diman0:
Nvm. You explained in the other PR. We will use this PR for the roundcube/sieve fix.

Diman0 (approved):
LGTM

I verified  that creating/changing sieve scripts work in roundcube. The Sieve filter is also correctly applied.

From php8 mbstring.func_overload is deprecated and removed (https://www.php.net/manual/en/mbstring.overload.php) . Once we update to the new roundcube, we must check if this part is still required in the diff patch. I expect it is removed in roundcube 1.6.x

PR #2588, opened 2022-12-19
IMAP folder names may contain characters outside of \w: [a-zA-Z0-9]

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

IMAP folder names may contain characters outside of \w: [a-zA-Z0-9]. Typically it may be subfolders...

I have also simplified the regexp since we strip spaces the line below.

This is used for "external accounts"/fetchmail.

## Prerequisites


Diman0 (approved):
LGTM

PR #2590, opened 2022-12-20
autodetect SUBNET

nextgens:
## What type of PR?

enhancement

## What does this PR do?

autodetect SUBNET; switch to unbound for name resolution.

The idea here is to reduce the number of tricky questions we ask the user during setup and minimize the opportunities for missconfiguration: "smart defaults". If you get it wrong you turn your instance into an open-relay.

### Related issue(s)
- #2614

## Prerequisites

### DIFF
--- a/core/base/libs/socrate/socrate/system.py
+++ b/core/base/libs/socrate/socrate/system.py
@@ -1,9 +1,12 @@
+import dns.resolver
 import hmac
 import logging as log
 import os
 from pwd import getpwnam
 import socket
+import sys
 import tenacity
+from textwrap import wrap
 
 @tenacity.retry(stop=tenacity.stop_after_attempt(100),
                 wait=tenacity.wait_random(min=2, max=5))
@@ -38,10 +41,23 @@ def set_env(required_secrets=[]):
     for secret in required_secrets:
         os.environ[f'{secret}_KEY'] = hmac.new(bytearray(secret_key, 'utf-8'), bytearray(secret, 'utf-8'), 'sha256').hexdigest()
 
-    return {
+    if not os.environ.get('SUBNET'):
+        os.environ['SUBNET'] = get_network_v4()
+
+    to_return={
             key: _coerce_value(os.environ.get(key, value))
             for key, value in os.environ.items()
            }
+
+    try:
+        ip_resolver = dns.resolver.query("resolver")[0].address
+        log.info(f'Setting DNS to {ip_resolver}')
+        with open('/etc/resolv.conf','w') as f:
+            f.write(f'nameserver {ip_resolver}\noptions ndots:0\n')
+    except:
+        pass
+
+    return to_return
 
 def clean_env():
     """ remove all secret keys """
@@ -53,3 +69,53 @@ def drop_privs_to(username='mailu'):
     os.setgid(pwnam.pw_gid)
     os.setuid(pwnam.pw_uid)
     os.environ['HOME'] = pwnam.pw_dir
+
+def get_network_v4():
+	dflgw = None
+	routes = []
+	for line in open('/proc/net/route', 'r'):
+		r = line.strip().split()
+		try:
+			route = [r[0]] + [int(c, 16) for c in r[1:]]
+		except ValueError:
+			pass
+		else:
+			if route[1] != 0:
+				routes.append(route)
+			elif dflgw is None:
+				dflgw = (route[0], route[2])
+
+	if dflgw is None:
+		return None
+
+	intf, ip = dflgw
+	for route in routes:
+		if route[0] == intf and route[1] == ip&route[7]:
+			net = route[1]
+			mask = bin(route[7]).count("1")
+			return f'{net&0xff}.{net>>8&0xff}.{net>>16&0xff}.{net>>24&0xff}/{mask}'
+
+def get_network_v6():
+	dflgw = None
+	routes = []
+	for line in open('/proc/net/ipv6_route', 'r'):
+		r = line.strip().split()
+		try:
+			route = [r[-1]] + [int(c, 16) for c in r[:-1]]
+		except ValueError:
+			pass
+		else:
+			if route[1] != 0 and route[2] != 0:
+				routes.append(route)
+			elif dflgw is None and route[9] == 3:
+				dflgw = (route[0], route[5])
+
+	if dflgw is None:
+		return None
+
+	intf, ip = dflgw
+	for route in routes:
+		if route[0] == intf and route[1] == ip&((2**route[2]-1)^0xffffffffffffffffffffffffffffffff):
+			prefix = ':'.join(wrap(f'{route[1]:032x}', 4))
+			mask = route[2]
+			return f'{prefix}/{mask}'
--- a/core/base/libs/socrate/socrate/system.py
+++ b/core/base/libs/socrate/socrate/system.py
@@ -47,7 +47,7 @@ def set_env(required_secrets=[]):
         with open('/etc/resolv.conf','w') as f:
             f.write(f'nameserver {ip_resolver}\noptions ndots:0\n')
     except:
-        pass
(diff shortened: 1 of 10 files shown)


core/base/libs/socrate/socrate/system.py
```diff
@@ -37,11 +40,74 @@ def set_env(required_secrets=[]):
     for secret in required_secrets:
         os.environ[f'{secret}_KEY'] = hmac.new(bytearray(secret_key, 'utf-8'), bytearray(secret, 'utf-8'), 'sha256').hexdigest()
 
-    return {
+    if not os.environ.get('SUBNET'):
+        os.environ['SUBNET'] = get_network_v4()
+
+    to_return={
```
ghostwheel42, line 46: Can't you just move this below the try-except block?

ghostwheel42 (commented):
Don't we need to call `set_env()` in every container when removing the resolv.conf entries?
At least radicale and clamav don't call it.

core/base/libs/socrate/socrate/system.py
nextgens: IMHO we want failures to be visible here... why would we silence them?

nextgens:
> Don't we need to call `set_env()` in every container when removing the resolv.conf entries? At least radicale and clamav don't call it.

I don't think we do; neither of them use any of the variables ... nor connect to other containers. They will get the system-provided resolver that should work

core/base/libs/socrate/socrate/system.py
ghostwheel42: I was just talking about the location. It shouldn't go _in_ the try/except block, but _after_ it.
Why store it in "to_return" and return it later when it can just be a `return { ... }` like before?

ghostwheel42:
> They will get the system-provided resolver that should work

But those containers wouldn't have a dnssec validation then, wouldn't they?
So Virus signatures for example could come from a modified source for example.

nextgens:
> > They will get the system-provided resolver that should work
> 
> But those containers wouldn't have a dnssec validation then, wouldn't they? So Virus signatures for example could come from a modified source for example.

Yes. I'm naively assuming that clamav is either signing them or fetching them securely.

core/base/libs/socrate/socrate/system.py
nextgens: Sure, that's d38ddfff

core/base/libs/socrate/socrate/system.py
```diff
@@ -38,6 +41,17 @@ def set_env(required_secrets=[]):
     for secret in required_secrets:
         os.environ[f'{secret}_KEY'] = hmac.new(bytearray(secret_key, 'utf-8'), bytearray(secret, 'utf-8'), 'sha256').hexdigest()
 
+    try:
+        ip_resolver = dns.resolver.query("resolver")[0].address
+        log.info(f'Setting DNS to {ip_resolver}')
+        with open('/etc/resolv.conf','w') as f:
+            f.write(f'nameserver {ip_resolver}\noptions ndots:0\n')
+    except:
+        pass
```
Diman0, line 50: Can this be enhanced with a log message as well?
Maybe: 
`
log.info('The unbound container with name resolver cannot be found or is not enabled. Falling back to docker embedded DNS server')
`

towncrier/newsfragments/1341.misc
```diff
@@ -2,3 +2,4 @@ Remove HOST_* variables, use *_ADDRESS everywhere instead. Please note that thos
 Derive a different key for admin/SECRET_KEY; this will invalidate existing sessions
 Ensure that rspamd starts after clamav
 Only display a single HOSTNAME on the client configuration page
+Autodetect SUBNET
```
Diman0, line 5: Please add that we also switch to unbound for local name resolution.

Diman0 (dismissed):
Two small changes are requested. I still need to test the PR.

core/base/libs/socrate/socrate/system.py
nextgens: Done in 1da7b28d

towncrier/newsfragments/1341.misc
nextgens: Done in 1da7b28d

optional/unbound/unbound.conf
```diff
@@ -8,6 +8,7 @@ server:
   do-udp: yes
   do-tcp: yes
   do-daemonize: no
+  do-not-query-localhost: no
   access-control: {{ SUBNET }} allow
 {%- if SUBNET6 %}
   access-control: {{ SUBNET6 }} allow
```
ghostwheel42, line 14: The `access-control SUBNET6` (no []) and the `interface: ::0` above do not work (no newline)
Can you rebase this - I think this was fixed in the meantime.

ghostwheel42 (changes requested):

nextgens:
@ghostwheel42 I've done the merge; I can't test the ipv6 stuff though.

optional/unbound/start.py
```diff
@@ -6,8 +6,23 @@
 from socrate import conf, system
 
 log.basicConfig(stream=sys.stderr, level=os.environ.get("LOG_LEVEL", "WARNING"))
-system.set_env()
 
-conf.jinja("/unbound.conf", os.environ, "/etc/unbound/unbound.conf")
+env = os.environ.copy()
```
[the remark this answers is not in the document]
nextgens: Hmm, we don't want it to override the resolver here

optional/unbound/unbound.conf
nextgens: Nothing was changed here... so if it's broken it was broken elsewhere.

optional/unbound/start.py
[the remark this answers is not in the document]
ghostwheel42: I don't understand - system.set_env normalizes the env values (true/false, etc.)

optional/unbound/unbound.conf
ghostwheel42: Ah, I've marked the wrong line, I was thinking about `{{- '  interface: ::0' if SUBNET6 }}`.
I'll fix it. It creates an unbound.conf like this:
```
server:
  verbosity: 1
  interface: 0.0.0.0  interface: ::0
  logfile: ""
  do-ip4: yes
```

ghostwheel42:
bors try

ghostwheel42:
bors retry

nextgens:
bors try

nextgens:
bors try

Diman0:
In the latest failed build, the filter test fails because rspamd cannot resolve the hostname oletools.
log fragment from rspamd ( https://github.com/Mailu/Mailu/actions/runs/4032768517/jobs/6932882387#step:11:368)
```
 2023-01-28 17:11:40 #1(main) <a911d4>; upstreams; rspamd_resolve_addrs: address resolution for oletools failed: Try again
2023-01-28 17:11:40 #1(main) <bhtbkb>; lua; oletools.lua:85: cannot parse servers oletools:11343
```

nextgens:
Agreed but I don't understand why. There is a ``depends_on:`` that covers both resolver and oletools.

Diman0:
Could this be a problem?
From https://docs.docker.com/config/containers/container-networking/#ip-address-and-hostname:
```
By default, the container gets an IP address for every Docker network it attaches to
```

Resolver is only connected to network default.
Antispam is connected to network default & noinet.
Oletools is connected to network noinet.

Could antispam get from the resolver the ip address for accessing oletools via network default (ip address) instead of network noinet?
Or does the error really indicate that the resolver did not return an ip address?

nextgens:
> Could antispam get from the resolver the ip address for accessing oletools via network default (ip address) instead of network noinet?

I don't think so since oletools is only connected to "noinet"

> Or does the error really indicate that the resolver did not return an ip address?

That's how it appears to me. The usual culprit is that the oletools hasn't started yet... but we do have a depend_on, making this something I don't understand yet.

nextgens:
bors try

PR #2603, opened 2022-12-28
Enable HAPROXY protocol on SUBNET

### LINKED ISSUE #894: Maximum number of connections from user+IP exceeded (mail_max_userip_connections=10):
Can't figure out how to increase the dovecot mail_max_userip_connections.
I have .env set with:

`AUTH_RATELIMIT=100/minute;5000/hour`

But still get this error below (so clearly it's still set to 10 instead of 100):

`Maximum number of connections from user+IP exceeded (mail_max_userip_connections=10):`

### LINKED ISSUE #1364: Maximum number of connections from user+IP exceeded
Hi, we have a problem... :-)
We have changed the original value of "AUTH_RATELIMIT" to "AUTH_RATELIMIT=100/minute;6000/hour", but logs continue to say " Maximum number of connections from user+IP exceeded (mail_max_userip_connections=20)" while reading response from upstream..."
We have made docker-compose dow and docker-compose up -d, but without result.
How can we change the default limit set during the installation?
Thanks in advance.

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

- Enable HAPROXY in between front and imap: With this we avoid running into the limitations of  ``mail_max_userip_connections`` and the logfiles reflect the real IP.
- Enable HAPROXY in between front and smtp: with this postfix and rspamd are aware of whether TLS was used or not on the last hop. In practice this won't work as nginx doesn't send PROTO yet.
- Discard redundant log messages from postfix

With all of this, not only are the logs easier to understand but ``doveadm who`` also works as one would expect.

### Related issue(s)
- closes #894
- #1328
- closes #1364
- #1705

## Prerequisites

### DIFF
--- a/core/admin/mailu/internal/views/dovecot.py
+++ b/core/admin/mailu/internal/views/dovecot.py
@@ -17,7 +17,7 @@ def dovecot_passdb_dict(user_email):
     return flask.jsonify({
         "password": None,
         "nopassword": "Y",
-        "allow_nets": ",".join(allow_nets)
+        "allow_real_nets": ",".join(allow_nets)
     })
 
 @internal.route("/dovecot/userdb/")
--- a/core/dovecot/conf/dovecot.conf
+++ b/core/dovecot/conf/dovecot.conf
@@ -11,6 +11,8 @@ default_internal_user = dovecot
 default_login_user = mail
 default_internal_group = dovecot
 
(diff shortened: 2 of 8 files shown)


nextgens:
bors try

nextgens:
bors try

ghostwheel42 (approved):
Lgtm

OdyX:
Could, or a variation of this be backported to 1.9 ?

PR #2605, opened 2022-12-29
Reduce the SSL session caches from 50m each to 3m each

nextgens:
## What type of PR?

enhancement

## What does this PR do?

Reduce the SSL session caches from 50m each to 3m each. This should be good for 12k sessions (within 1day, see http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_session_cache and our ssl_session_timeout) for each cache and will help reduce memory usage.

## Prerequisites

### DIFF
--- a/core/nginx/conf/nginx.conf
+++ b/core/nginx/conf/nginx.conf
@@ -100,7 +100,7 @@ http {
       include /etc/nginx/tls.conf;
       ssl_stapling on;
       ssl_stapling_verify on;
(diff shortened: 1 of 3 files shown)


nextgens:
3M may be a bit aggressive for the HTTPd part... but then again, the worst case scenario is that a full rekey is required (additional network roundtrip + CPU time). IMHO it's fine and doesn't need a new knob.

PR #2607, opened 2022-12-29
Update python dependencies as suggested by dependabot

ghostwheel42:
## What type of PR?

bug-fix

## What does this PR do?

Update dependencies to silence dependabot (vulnerabilities are probably not exploitable)
Only the certifi upgrade could be backported.


ghostwheel42:
bors try

nextgens (approved):
LGTM

PR #2612, opened 2023-01-03
Remove duplicated 'actionstart = ' in fail2ban conf.

lucassith:
In fail2ban example configuration for ipset `option #2`, there is a duplicated string which makes the ipset and fail2ban fail to create the set. 
Fail2ban will never ban any ip due to this error.

## What type of PR?

documentation


ghostwheel42 (approved):
Good catch. LGTM.

nextgens (approved):

PR #2613, opened 2023-01-04
Enhance network segregation

nextgens:
## What type of PR?

enhancement

## What does this PR do?

- put radicale and webmail on their own network: this is done for security: that way they have no privileged access anywhere (no access to redis, no access to XCLIENT, ...)
- remove the EXPOSE statements from the dockerfiles. These ports are for internal comms and are not meant to be exposed in any way to the outside world.

### Related issue(s)
- #2611

## Prerequisites

### DIFF
--- a/core/nginx/Dockerfile
+++ b/core/nginx/Dockerfile
@@ -17,7 +17,8 @@ ARG VERSION
 LABEL version=$VERSION
 
 RUN set -euxo pipefail \
-  ; apk add --no-cache certbot nginx nginx-mod-mail openssl
+  ; apk add --no-cache certbot nginx nginx-mod-http-brotli nginx-mod-stream nginx-mod-mail openssl \
+  ; rm /etc/nginx/conf.d/stream.conf
 
 COPY conf/ /conf/
 COPY --from=static /static/ /static/
@@ -25,7 +26,7 @@ COPY *.py /
 
 RUN echo $VERSION >/version
 
-EXPOSE 80/tcp 443/tcp 110/tcp 143/tcp 465/tcp 587/tcp 993/tcp 995/tcp 25/tcp 10025/tcp 10143/tcp
+EXPOSE 80/tcp 443/tcp 110/tcp 143/tcp 465/tcp 587/tcp 993/tcp 995/tcp 25/tcp 14190/tcp 10025/tcp 10143/tcp
 HEALTHCHECK --start-period=60s CMD curl -skfLo /dev/null http://localhost/health
 
 VOLUME ["/certs", "/overrides"]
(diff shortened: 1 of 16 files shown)


webmails/roundcube/config/config.inc.php
```diff
@@ -28,7 +28,7 @@
 $config['default_port'] = '10143';
 
 // Sieve script management
-$config['managesieve_host'] = '{{ IMAP_ADDRESS or "imap" }}';
+$config['managesieve_host'] = '{% if TLS_FLAVOR != "notls" %}ssl://{% endif %}{{ HOSTNAMES.split(",")[0] }}:4190';
```
ghostwheel42, line 31: What do we gain by connection through the public address here?
imap and smtp is done through front (unencrypted), so we could also du sieve through front unencrypted (port 14190)
Another (maybe better) way would be to do everything the webmails do through the public and encrypted endpoints like every client does.

core/nginx/conf/nginx.conf
```diff
@@ -275,6 +277,24 @@ http {
     include /etc/nginx/conf.d/*.conf;
 }
 
+stream {
+    # managesieve with implicit TLS.
+    # we can't do explicit since dovecot doesn't have the certificates
+    # It's arguable whether we should use the standardized port number for doing something non-standard
```
ghostwheel42, line 283: I think it's fine. Clients should show an error.
Other proxies can detect TLS on the fly, so we could add starttls when a proxy is implemented.

nextgens: OK cool.

webmails/roundcube/config/config.inc.php
nextgens: It bypasses the rate-limits and authentication is handled differently (with temporary tokens). Using the public ports would mean allowing that type of tokens on those ports and finding another way to rate-limit/exempt from rate limits.

nextgens: OK. Let's keep it simple.

nextgens:
bors try

ghostwheel42 (approved):
lgtm

ghostwheel42:
bors try

core/nginx/Dockerfile
```diff
@@ -17,15 +17,16 @@ ARG VERSION
...
+  ; rm /etc/nginx/conf.d/stream.conf
 
 COPY conf/ /conf/
 COPY --from=static /static/ /static/
 COPY *.py /
 
 RUN echo $VERSION >/version
 
-EXPOSE 80/tcp 443/tcp 110/tcp 143/tcp 465/tcp 587/tcp 993/tcp 995/tcp 25/tcp 10025/tcp 10143/tcp
+EXPOSE 80/tcp 443/tcp 110/tcp 143/tcp 465/tcp 587/tcp 993/tcp 995/tcp 25/tcp 14190/tcp 10025/tcp 10143/tcp
```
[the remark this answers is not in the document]
nextgens: This is an old revision

PR #2615, opened 2023-01-04
Fix snappymail

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

This was broken in #2611. The other is non-consequential security-wise as nginx filters XHOST... but it's worth fixing regardless.

### Related issue(s)
- #2580
- #2611

## Prerequisites


ghostwheel42 (approved):
Lgtm

PR #2622, opened 2023-01-11
Fix smtplib.LMTP wrong argument name: ip -> host

zozzz:
## What type of PR?

bug-fix

## What does this PR do?

Bug fix when sending welcome message.

### Related issue(s)
- Mention an issue like: -
- Auto close an issue like: closes -

## Prerequisites


nextgens (approved):
Good catch; not sure what I was thinking here.

ghostwheel42 (approved):

PR #2624, opened 2023-01-12
Move runtime environment variables to the end

ghostwheel42:
## What type of PR?

bug-fix

## What does this PR do?

This moves the environment variables used at runtime from the system to the base image.
It's a workaround for a strange build issue observed when building with hardened malloc enabled.


nextgens:
That has broken something; the tests are failing

ghostwheel42:
bors try

nextgens (approved):

PR #2628, opened 2023-01-25
Set default for FETCHMAIL_ENABLED

ghostwheel42:
## What type of PR?

bug-fix

## What does this PR do?

Set the default for FETCHMAIL_ENABLED to true in the admin container.
This keeps existing functionality for people upgrading without re-creating the `mailu.env`.

### DIFF
--- a/tests/compose/fetchmail/mailu.env
+++ b/tests/compose/fetchmail/mailu.env
@@ -83,8 +83,11 @@ RELAYNETS=
 # Will relay all outgoing mails if configured
 RELAYHOST=
 
+# Show fetchmail functionality in admin interface
+FETCHMAIL_ENABLED=true
+
 # Fetchmail delay
-FETCHMAIL_DELAY=600
+FETCHMAIL_DELAY=15
 
 # Recipient delimiter, character used to delimiter localpart from custom address part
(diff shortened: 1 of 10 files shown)


tests/compose/fetchmail/mailu.env
```diff
@@ -83,6 +83,9 @@ RELAYNETS=
 # Will relay all outgoing mails if configured
 RELAYHOST=
 
+# Show fetchmail functionality in admin interface
+FETCHMAIL_ENABLED=false
```
Diman0, line 87: Should fetchmail not be enabled in mailu.env (tests/compose/fetchmail/mailu.env)  for the fetchmail test?
Maybe the fetchmail_delay could also be lowered to 15 seconds. 
I know the fetchmail test only starts all containers, but this at least tests if the fetchmail container can start fine and connect to admin to retrieve the fetchmail pull settings.

ghostwheel42: Hmm. Don't know - where is the test code?
I'll change it. Maybe we should update all test cases to the new files created by the setup utility before the next release.

Diman0 (approved):
Lgtm

ghostwheel42:
bors retry

PR #2630, opened 2023-01-25
Improved IPv6 support #2

### LINKED ISSUE #1789: gunicorn not listening on IPv6

### LINKED ISSUE #2392: Docker container crashes if IPv6 is disabled at the system level.
If listen [::] is found somewhere in the configs, but IPv6 is disabled at the host system level and in the docker, then the process crashes, and, accordingly, the docker container also crashes.

This can be manually climbed into each container, corrected, but it is not very convenient.

docker exec mailu_front_1 sed -i '/listen \[/d' /conf/nginx.conf
docker exec mailu_front_1 sed -i '/listen \[/d' /etc/nginx/nginx.conf
docker exec mailu_front_1 sed -i '/listen \[/d' /etc/nginx/http.d/default.conf
docker restart mailu_front_1

docker restart mailu_webdav_1 && docker exec -it mailu_webdav_1 sed -i 's/hosts =.*\[::\].*/hosts = 0.0.0.0:5232/g' /radicale.conf && docker restart mailu_webdav_1

Can you add a container launch option to remove listen [::] from configs?

ghostwheel42:
## What type of PR?

enhancement, bug-fix, documentation

## What does this PR do?

This is based on #2272 and adds some more fixes.

### Related issue(s)
- closes #1789
- closes #2392

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -145,51 +145,24 @@ Your mail service will be reachable for IMAP, POP3, SMTP and Webmail at the addr
 How to make IPv6 work?
 ``````````````````````
 
-Docker currently does not expose the IPv6 ports properly, as it does not interface with ``ip6tables``.
-Lets start with quoting everything that's wrong:
+Docker IPv6 interfacing with ``ip6tables``, which is required for proper IPv6 support, is currently considered experimental.
 
-  Unfortunately, initially Docker was not created with IPv6 in mind.
-  It was added later and, while it has come a long way, is still not as usable as one would want.
-  Much discussion is still going on as to how IPv6 should be used in a containerized world;
-  See the various GitHub issues linked below:
+You can enable experimental IPv6 support in docker via a custom ``/etc/docker/daemon.json`` file like this one:
 
-  - Giving each container a publicly routable address means all ports (even unexposed / unpublished ports) are suddenly
-    reachable by everyone, if no additional filtering is done
-    (`docker/docker#21614 <https://github.com/docker/docker/issues/21614>`_)
-  - By default, each container gets a random IPv6, making it impossible to do properly do DNS;
-    the alternative is to assign a specific IPv6 address to each container,
-    still an administrative hassle (`docker/docker#13481 <https://github.com/docker/docker/issues/13481>`_)
-  - Published ports won't work on IPv6, unless you have the userland proxy enabled
-    (which, for now, is enabled by default in Docker)
-  - The userland proxy, however, seems to be on its way out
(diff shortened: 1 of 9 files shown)


docs/faq.rst
```diff
@@ -147,11 +147,30 @@ How to make IPv6 work?
...
+You can enable `docker-ipv6nat` like this:
+
+  docker run -d --name ipv6nat --privileged --network host --restart unless-stopped -v /var/run/docker.sock:/var/run/docker.sock:ro -v /lib/modules:/lib/modules:ro robbertkl/ipv6nat
+
+If you want to try docker's experimental IPv6 support, it can be enabled like this:
 
 .. code-block:: json
 
   {
+      "userland-proxy": false,
```
Niduroki, line 173: For using *only* docker ip6tables (i.e. *not* `docker-ipv6nat`), this line is unnecessary.
I have not tested if disabling the userland proxy will interfere with normal operation in any way, if done without the need to do so.

ghostwheel42: Ah, yes, you are correct.
Hmm. I think I'll leave it in there because it doesn't harm and it's is a big problem (open relay) when its _not_ in there and IPv6 is enabled without the iptables setting.

ghostwheel42:
bors try

ghostwheel42:
I'm not sure if I should add a detection of sysctl flags like bindipv6only. 
I couldn't get my docker stack to _not_ bind v4 and v6 (latest version on Ubuntu 22.04), so maybe adding some documentation will be sufficient. 

After merging this we will get feedback of corner cases and can add code and/or documentation as necessary.

nextgens:
> I'm not sure if I should add a detection of sysctl flags like bindipv6only. I couldn't get my docker stack to _not_ bind v4 and v6 (latest version on Ubuntu 22.04), so maybe adding some documentation will be sufficient.

Odds are this is something ubuntu has patched in their kernel; I'd try a different distro.

nextgens:
What about oletools?

ghostwheel42:
> Odds are this is something ubuntu has patched in their kernel; I'd try a different distro.

I don't think so - It's probably done by docker itself. But I'll try some other kernels. I have a gentoo system here using a fairly vanilla kernel.

ghostwheel42:
> What about oletools?

good catch. I've added a chnage for this.

ghostwheel42:
bors try

ghostwheel42:
bors try

nextgens (approved):

PR #2631, opened 2023-01-25
Restful api finishing touches

Diman0:
## What type of PR?

enhancement

## What does this PR do?
Some finishing touches for the restful api.

- Make the API configurable via the setup utility.  
  - Configured exactly the same as the ADMIN and WEBMAIL. 
- We have a single config (API) that configures whether it is exposed (via front). Just like ADMIN. The API is always reachable by directly connecting to the admin container.
- API_TOKEN does not enable/disable the API anymore. When it is not configured, an error is returned (via the internet browser) that the API_TOKEN must be configured in mailu.env.
- Fix some small bugs in the setup utility ( selecting none in the dropdown boxes, now correctly changes the config)
- Update Flask-RestX to 1.0.5. This resolves the deprecation warnings introduced by Flask-RestX.

## Prerequisites

### DIFF
--- a/core/admin/mailu/api/__init__.py
+++ b/core/admin/mailu/api/__init__.py
@@ -8,17 +8,25 @@ def register(app, web_api_root):
     # register api bluprint(s)
     apidoc.apidoc.url_prefix = f'{web_api_root}/v{int(APIv1.VERSION)}'
     APIv1.api_token = app.config['API_TOKEN']
-    app.register_blueprint(APIv1.blueprint, url_prefix=f'{web_api_root}/v{int(APIv1.VERSION)}')
+    if app.config['API_TOKEN'] != '':
+        app.register_blueprint(APIv1.blueprint, url_prefix=f'{web_api_root}/v{int(APIv1.VERSION)}')
 
-    # add redirect to current api version
-    redirect_api = Blueprint('redirect_api', __name__)
-    @redirect_api.route('/')
-    def redir():
-        return redirect(url_for(f'{APIv1.blueprint.name}.root'))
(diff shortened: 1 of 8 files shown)


ghostwheel42:
bors try

core/admin/mailu/api/__init__.py
```diff
@@ -8,17 +8,25 @@ def register(app, web_api_root):
     # register api bluprint(s)
     apidoc.apidoc.url_prefix = f'{web_api_root}/v{int(APIv1.VERSION)}'
     APIv1.api_token = app.config['API_TOKEN']
-    app.register_blueprint(APIv1.blueprint, url_prefix=f'{web_api_root}/v{int(APIv1.VERSION)}')
+    if app.config['API_TOKEN'] != '':
+        app.register_blueprint(APIv1.blueprint, url_prefix=f'{web_api_root}/v{int(APIv1.VERSION)}')
```
ghostwheel42, line 12: Using `int()` is not necessary in a format string

ghostwheel42:
bors try

ghostwheel42:
bors try

ghostwheel42 (changes requested):
When testing API was enabled by default in the setup container.
I think it should be disabled.

ghostwheel42:
ah, and I think there should be a small example hex token in the API_TOKEN text field
Otherwise this LGTM

Diman0:
@ghostwheel42 
I have made the proposed changes. For generating the token I simply picked the first JavaScript sample I could find for generating passwords. A static token felt too insecure for me. We don't want 50% of all Mailu installations to use the same token.

ghostwheel42 (approved):
LGTM

PR #2633, opened 2023-01-28
Don't apply antispoof rules on locally generated emails

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Don't apply antispoof rules on locally generated emails; This was breaking the auto-responder and sieve rules.

## Prerequisites


ghostwheel42 (approved):
LGTM

PR #2638, opened 2023-01-30
further finishing touches for restful api

Diman0:
- Fix setup utility setting correct value to env var API. It now also sets `false` when the API is disabled in the setup utility.
- Fix IF statement for enabling API in nginx.conf. Setting a different value than `API=true` in mailu.env now disabled the API endpoint in nginx.
- Use safer command for regenerating example API token. It uses crypto.getRandomValues() (as suggested by nextgens) which should be more random than the previously used method. 

## What type of PR?

bug-fix

## Prerequisites

### DIFF
--- a/core/nginx/conf/nginx.conf
+++ b/core/nginx/conf/nginx.conf
@@ -252,7 +252,7 @@ http {
       {% endif %}
       {% endif %}
 
-      {% if API %}
+      {% if API == 'true' %}
       location ~ {{ WEB_API or '/api' }} {
         include /etc/nginx/proxy.conf;
         proxy_pass http://$admin;
--- a/core/nginx/conf/nginx.conf
+++ b/core/nginx/conf/nginx.conf
@@ -252,7 +252,7 @@ http {
       {% endif %}
       {% endif %}
 
-      {% if API == 'true' %}
+      {% if API %}
       location ~ {{ WEB_API or '/api' }} {
         include /etc/nginx/proxy.conf;
         proxy_pass http://$admin;
--- a/core/nginx/config.py
+++ b/core/nginx/config.py
@@ -5,8 +5,7 @@ import logging as log
 import sys
 from socrate import system, conf
 
-system.set_env()
-args = os.environ.copy()
+args = system.set_env()
 log.basicConfig(stream=sys.stderr, level=args.get("LOG_LEVEL", "WARNING"))
 
 args['TLS_PERMISSIVE'] = str(args.get('TLS_PERMISSIVE')).lower() not in ('false', 'no')
--- a/core/nginx/config.py
+++ b/core/nginx/config.py
@@ -17,8 +17,8 @@ with open("/etc/resolv.conf") as handle:
     args["RESOLVER"] = f"[{resolver}]" if ":" in resolver else resolver
 
 # TLS configuration
-cert_name = os.getenv("TLS_CERT_FILENAME", default="cert.pem")
-keypair_name = os.getenv("TLS_KEYPAIR_FILENAME", default="key.pem")
+cert_name = args.get("TLS_CERT_FILENAME", "cert.pem")
+keypair_name = args.get("TLS_KEYPAIR_FILENAME", "key.pem")
 args["TLS"] = {
     "cert": ("/certs/%s" % cert_name, "/certs/%s" % keypair_name),
     "letsencrypt": ("/certs/letsencrypt/live/mailu/nginx-chain.pem",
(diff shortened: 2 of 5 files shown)


Diman0:
For testing all paths
- Run the setup utility. 
  - Create a configuration with the API disabled. `API=false` should be in mailu.env
  - Create a configuration with the API enabled. `API=true` should be in mailu.env 
- API endpoint exposed
  - set `API=true` in mailu.env. Confirm API is reachable.
  - set `API=false` in mailu.env. Confirm API is not reachable.

core/nginx/conf/nginx.conf
```diff
@@ -252,7 +252,7 @@ http {
       {% endif %}
       {% endif %}
 
-      {% if API %}
+      {% if API == 'true' %}
```
ghostwheel42, line 255: This works, but I think using socrate's output in `/config.py` would be better.
Instead of 
```
system.set_env()
args = os.environ.copy()
```
this should do the trick (API would the be normalized to boolean):
```
args = system.set_env()
```

setup/templates/steps/config.html
```diff
@@ -93,7 +93,7 @@
   It is not possible to use the API without an API token.</p>
 
   <div class="form-group">
-      <input type="checkbox" name="api_enabled" value="false" id="api" >
+      <input type="checkbox" name="api_enabled" value="true" id="api" >
```
ghostwheel42, line 96: This should be "false", to have the API disabled by default, shouldn't it?

setup/static/render.js
```diff
@@ -1,12 +1,7 @@
...
-var tokenLength = 12;
-var token = "";
-
-for (var i = 0; i <= tokenLength; i++) {
-	var randomNumber = Math.floor(Math.random() * chars.length);
-	token += chars.substring(randomNumber, randomNumber +1);
-   }
+var random_array = new Uint32Array(2);
+crypto.getRandomValues(random_array);
+var token =  random_array[0].toString() + random_array[1].toString();
```
ghostwheel42, line 4: Is concatenating two 32bit integers a "good" api token?
Shouldn't we use something longer, like in this function?
```
function generatePassword(length) {
  const charset = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789';
  let password = '';
  const array = new Uint8Array(length);
  crypto.getRandomValues(array);
  for (let i = 0; i < length; i++) {
    password += charset[array[i] % charset.length];
  }
  return password;
}
generatePassword(32)
```

What about creating a random token on the server side (like we do for the SECRET_KEY)?

ghostwheel42 (changes requested):

setup/static/render.js
Diman0: I could do it server side. I will also change the length to 32. Although 20 should be fine IMHO. The API is rate limited after all.

setup/templates/steps/config.html
Diman0: No. That is handled by the JavaScript. I don't understand it. When testing it again, I see that the value `API=false` is always written in the docker-compose.yml. I will check why this happens.

You can actually test the PR with this docker-compose.yml
```
# This file is used to run the mailu/setup utility

version: '3.6'

services:
  redis:
    image: redis:alpine
    restart: always

  development:
    image: ${DOCKER_ORG}/setup:${DEVELOPMENT}
    ports:
      - "0.0.0.0:1234:80"
    env_file: .env
    environment:
      stable_version: ${DEVELOPMENT}
      this_version: ${DEVELOPMENT}
      VERSIONS: ${DEVELOPMENT}
    depends_on:
      - redis
    restart: always
```
and this .env file
```
# Current versions
#DEVELOPMENT=local
DEVELOPMENT=pr-2638
#DOCKER_ORG=mailu
DOCKER_ORG=mailuci
```
You can switch the comments in the .env file to switch to a locally build setup. By default local images are build with docker organisation `mailu` and tag `local`.

core/nginx/conf/nginx.conf
Diman0: Ah. Yes. I also prefer this. It might be worthwhile to check if we also use this in the other images.

setup/templates/steps/config.html
Diman0: Found it. I forgot to add the attribute `id="api_enabled" ` to the checkbox (or I removed it by mistake.)

setup/static/render.js
Diman0: I had a look. It indeed makes sense to re-use the server side token/password generator. I will use this with length 32.

Diman0:
bors try

Diman0:
bors retry

core/nginx/config.py
```diff
@@ -5,8 +5,7 @@
 import sys
 from socrate import system, conf
 
-system.set_env()
-args = os.environ.copy()
+args = system.set_env()
```
ghostwheel42, line 8: I think you should also replace all "os.getenv()" with "args.get()" now.
Sorry, for not pointing this out.
This doesn't make a diference now, but will surely cause side-effects later.

ghostwheel42 (changes requested):

ghostwheel42 (approved):
Lgtm

PR #2640, opened 2023-01-31
Add env variable to set sieve_vacation_to_header_ignore_envelope

nwinkelstraeter:
When used with SRS the vacation plugin creates a reply with SRS in the To: header which does not look nice for the recipient. Setting sieve_vacation_to_header_ignore_envelope will use the headers from the original source message instead of potentially rewritten ones.

Without this option auto-replies are sent with a To header with SRS, e.g `SRS0=uetG=43=sender.com=user@autoresponder.com`
With this option they are sent with just `user@sender.com`

This option is for whatever reason not part of the [pigeonhole docs](https://doc.dovecot.org/configuration_manual/sieve/extensions/vacation/) but it is documented here: https://github.com/dovecot/pigeonhole/blob/34431d7a67474c00d5b0554f5d45b867509f4a7a/NEWS#L338

## What type of PR?

enhancement

## What does this PR do?
This PR adds an environment variable to the set the `sieve_vacation_to_header_ignore_envelope` configuration

## Prerequisites

### DIFF
--- a/core/dovecot/conf/dovecot.conf
+++ b/core/dovecot/conf/dovecot.conf
@@ -173,6 +173,9 @@ plugin {
   # Include the recipient in vacation replies so that DKIM applies
   sieve_vacation_send_from_recipient = yes
 
+  # Use To: header from original message becaus envelope has a SRS address
+  sieve_vacation_to_header_ignore_envelope = yes
+
   # extract spam score from headers
   sieve_spamtest_status_type = strlen
   sieve_spamtest_status_header = X-Spam-Level
--- a/towncrier/newsfragments/2640.bugfix
+++ b/towncrier/newsfragments/2640.bugfix
@@ -1,0 +1,1 @@
+Uses the correct From address (instead of an SRS alias) in the sieve/vacation module


nextgens:
I don't think it makes sense to have a setting for this. We should agree on what is the correct behavior and just use that... those that want something else can use overrides.

Why isn't this the default if using the header rather than the envelope a good idea?

nwinkelstraeter:
The documentation for this setting states that one use-case for it SRS enabled servers, which mailu is.
So enabling it is the correct default, i guess.
But I just noticed that mailu has a piece of code which is already supposed to reverse SRS in the To: header of bounce mails https://github.com/Mailu/Mailu/blob/0bde746610681ec7df7ba91dbaa5234954b4ce1b/core/admin/mailu/internal/views/postfix.py#L107

Could/Should this be used for this instead?
I'm not entirely sure where that function is used, i guess via `recipient_canonical_maps = ${podop}recipientmap`
But then SRS addresses in vacation replies should already be rewritten to its original form.

nextgens:
The envelope To is... that's why you get the ooo message.

The problem here is that the header To isn't (since that is from the SRS-ized version of the envelope From of the initial message).

nwinkelstraeter:
Okay, then i will just change this PR to set `sieve_vacation_to_header_ignore_envelope` to yes by default.

nextgens:
Please do.

I'd also like to reconsider whether it's safe to use ```sieve_vacation_dont_check_recipient = yes``` here. I don't think it is and would rather see us use:
```
sieve_vacation_use_original_recipient =yes
lda_original_recipient_header = XXX-something
```
with the appropriate tweak to postfix/lmtp to ensure we get the envelope To we are after.

Diman0:
> Please do.
> 
> I'd also like to reconsider whether it's safe to use `sieve_vacation_dont_check_recipient = yes` here. I don't think it is and would rather see us use:
> 
> ```
> sieve_vacation_dont_check_recipient = no
> sieve_vacation_use_original_recipient = yes
> lda_original_recipient_header = XXX-something
> ```
> 
> with the appropriate tweak to postfix/lmtp to ensure we get the envelope To we are after.

This blog provides more background information (https://www.dlford.io/sieve-configure-vacation-responses/)

Diman0:
bors try

nextgens:
bors try

Please give this a try; it seems to work for me.
The only downside I can see to doing it this way is that we leak recipients of aliases in the newly introduced header.

PR #2643, opened 2023-02-01
Fix nginx config

Diman0:
Forgot to adapt some IF statements. All config is normalized now for front.

So true/false now matches the boolean value True/False. Instead if {% IF X == 'true' %} we should now use {% IF X %}

## What type of PR?

bug-fix

## What does this PR do?
Fixes a bug in the nginx config. This bug prevents the /admin endpoint from being exposed.

## Prerequisites

- [ n/a] In case of feature or enhancement: documentation updated accordingly
- [ n/a] Unless it's docs or a minor change: add [changelog](https://mailu.io/master/contributors/workflow.html#changelog) entry file.

### DIFF
--- a/core/nginx/conf/nginx.conf
+++ b/core/nginx/conf/nginx.conf
@@ -91,7 +91,7 @@ http {
       client_max_body_size {{ MESSAGE_SIZE_LIMIT|int + 8388608 }};
 
       # Listen on HTTP only in kubernetes or behind reverse proxy
-      {% if KUBERNETES_INGRESS == 'true' or TLS_FLAVOR in [ 'mail-letsencrypt', 'notls', 'mail' ] %}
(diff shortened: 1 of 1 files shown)


Diman0:
to verify the bug.
1. Start mailu with master images (docker-compose up -d). of course ADMIN/WEB_ADMIN must be enabled in mailu.env.
2. docker-compose exec front cat /etc/nginx/nginx.conf
3. Note that the output of the above command does not show the /admin endpoint.
4. Start the mailu deployment for mailuci/admin:pr-2643 (images should be build in the next 10 minutes)
5. docker-compose exec front cat /etc/nginx/nginx.conf. of course ADMIN/WEB_ADMIN must be enabled in mailu.env.
6. Note that the output shows the /admin endpoint

ghostwheel42:
bors retry

ghostwheel42 (approved):
LGTM

PR #2646, opened 2023-02-04
Smarter ratelimit

nextgens:
## What type of PR?

enhancement

## What does this PR do?

Only account for **distinct** usernames in the IP rate-limiter.

This enables to have a much tighter default as a user with a misconfigured device will now only account for a single attempt.

The goal here is to make the rate-limiter more acceptable and to avoid people disabling it altogether.

## Prerequisites


nextgens:
bors cancel

bors try

docs/configuration.rst
```diff
@@ -40,11 +40,12 @@ address.
...
 that are allowed to send emails from any existing address (spoofing the sender).
 
-The ``AUTH_RATELIMIT_IP`` (default: 60/hour) holds a security setting for fighting
-attackers that waste server resources by trying to guess user passwords (typically
-using a password spraying attack). The value defines the limit of authentication
-attempts that will be processed on non-existing accounts for a specific IP subnet
-(as defined in ``AUTH_RATELIMIT_IP_V4_MASK`` and ``AUTH_RATELIMIT_IP_V6_MASK`` below).
+The ``AUTH_RATELIMIT_IP`` (default: 5/hour) holds a security setting for fighting
+attackers that attempt a password spraying attack. The value defines the limit of
+authentication attempts that will be processed on **distinct** non-existing
```
ghostwheel42, line 45: Isn't this also checked for existing accounts?

nextgens: That's only checked for non-existing accounts. The other rate limit applies for existing accounts (XXX wrong attempts, regardless of source)

ghostwheel42: Ah, yes. But we should also de-duplicate those.
The problem occuring when changing a password: lots of devices (not using a token) try to log-in using the old password which will get the source IP blocked, even though no-one is brute-forcing passwords.

nextgens: That's arguable... security wise it's not a great idea of storing a representation of attempted passwords (it may be near-misses); We could hash them but would have to store them unsalted.

The second ratelimit locks the account, not the IP address.

We have all kind of fancy ways of ensuring we don't block legitimate users: If a login is successful the IP address won't get into rate-limits; we also use a special type of cookies to ensure that devices that have already authenticated successfully get access even if the account is locked.

nextgens: IMHO the right thing to do is to ensure those devices use tokens. That may mean making the autodiscovery hand over tokens automatically.

ghostwheel42:
bors try

nextgens:
bors try

core/admin/mailu/configuration.py
```diff
@@ -40,9 +40,9 @@
     'TLS_FLAVOR': 'cert',
     'INBOUND_TLS_ENFORCE': False,
     'DEFER_ON_TLS_ERROR': True,
-    'AUTH_RATELIMIT_IP': '60/hour',
+    'AUTH_RATELIMIT_IP': '5/hour',
     'AUTH_RATELIMIT_IP_V4_MASK': 24,
-    'AUTH_RATELIMIT_IP_V6_MASK': 56,
+    'AUTH_RATELIMIT_IP_V6_MASK': 48,
```
ghostwheel42, line 45: You need to alter the comment in the mailu.env file to `/48` if you change this.
I'm not sure if blocking a /48 is a good idea. At least her in germany many provider give you the bare minimum of addresses (/64 or /60)
And I know of hosting setups using a single /64 for hundreds of servers from differnet customers.

ghostwheel42 (changes requested):

core/admin/mailu/configuration.py
nextgens: done in 83cc23a5

ghostwheel42 (approved):
lgtm

PR #2651, opened 2023-02-05
Allow + in localpart of addresses to forward to

### LINKED ISSUE #1236: plus mail in auto-forward destination
The email address in `admin -> User settings -> Auto-forward -> Destination` cannot contains a '+' character (error: Invalid email address). But the '+' in email address is valid and admin should accept it.

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Allow + in localpart of addresses to forward to

### Related issue(s)
- close #1236

## Prerequisites

### DIFF
--- a/core/admin/mailu/ui/forms.py
+++ b/core/admin/mailu/ui/forms.py
@@ -37,7 +37,7 @@ class MultipleEmailAddressesVerify(object):
         self.message = message
 
     def __call__(self, form, field):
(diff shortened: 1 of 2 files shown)


ghostwheel42 (commented):
Hmm, shouldn't we use sah like this to identify a valid email address?

https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address

nextgens:
@ghostwheel42 something like that? both look scary TBH; we may be better splitting the fields and testing each part separately.

nextgens:
The previous regexp was also incompatible with new TLDs

PR #2664, opened 2023-02-14
Fix the bug reported by fastlorenzo

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Fix the bug reported by fastlorenzo: when using proxy-auth, if the user doesn't exist you have to hit the URL twice.

## Prerequisites

### DIFF
--- a/core/admin/mailu/sso/views/base.py
+++ b/core/admin/mailu/sso/views/base.py
@@ -62,7 +62,6 @@ def logout():
         response.set_cookie(cookie, 'empty', expires=0)
     return response
 
-
 @sso.route('/proxy', methods=['GET'])
 @sso.route('/proxy/<target>', methods=['GET'])
 def proxy(target='webmail'):
@@ -95,6 +94,8 @@ def proxy(target='webmail'):
         return flask.abort(500, 'Too many users in (domain=%s)' % domain)
     user = models.User(localpart=localpart, domain=domain)
     user.set_password(secrets.token_urlsafe())
+    flask.session.regenerate()
+    flask_login.login_user(user)
     models.db.session.add(user)
     models.db.session.commit()
     user.send_welcome()


core/admin/mailu/sso/views/base.py
```diff
@@ -95,6 +94,8 @@ def proxy(target='webmail'):
         return flask.abort(500, 'Too many users in (domain=%s)' % domain)
     user = models.User(localpart=localpart, domain=domain)
     user.set_password(secrets.token_urlsafe())
+    flask.session.regenerate()
+    flask_login.login_user(user)
```
Diman0, line 98: Shouldn't this (lines 97-98) occur `after` committing the changes to the database (after `models.db.session.commit()`).
Now you are creating a session for a non-existing user.

Diman0 (dismissed):
See in-line comment. 
Can you also address lines 83 & 90 & 91? 
These lines have very childish error messages. How did we miss this when reviewing the original PR?

`return flask.abort(500, 'You don\'t exist. Go away! (%s)' % email)`
`return flask.abort(500, 'You don\'t exist. Go away! (%s)' % email)`
`domain = models.Domain.query.get(desireddomain) or flask.abort(500, 'You don\'t exist. Go away! (domain=%s)' % desireddomain)`

Maybe change it to:
`return flask.abort(500, 'The user does not exist (%s)' % email)`
`return flask.abort(500, 'There is a problem with the user name (%s)' % email)`
`domain = models.Domain.query.get(desireddomain) or flask.abort(500, 'The requested domain does not exist (domain=%s)' % desireddomain)`
```

core/admin/mailu/sso/views/base.py
nextgens: That's 6a4d8603

nextgens:
"You don't exist. Go away!" is not childish; it is a reference you don't have, that's all :)

For the record, we wouldn't be the first ones to make this choice. See https://github.com/openssh/libopenssh/blob/05dfdd5f54d9a1bae5544141a7ee65baa3313ecd/ssh/ssh.c#L299 for instance.

ghostwheel42 (approved):
Lgtm

PR #2681, opened 2023-03-01
Set rainloop autologout time according to SESSION_TIMEOUT

### LINKED ISSUE #2680: Rainloop autologout triggers too early in 1.9 (and possibly snappymail/master too)
### Environment
- [x] docker-compose
- [ ] kubernetes
- [ ] docker swarm

### Version

- Version: `1.9` (latest)

## Description
Despite setting a quite liberal
```
SESSION_TIMEOUT=64800
PERMANENT_SESSION_LIFETIME=345600
```
rainloop automatically logs out users after 30min. This seems to be because the rainloop `autologout` is still on default — 30min. Setting the value in `application.ini` according to the configured timeout seems to solve the problem.

I currently cannot test, but I strongly suspect this is also true for snappymail on `master`.

## Replication Steps
- Set big session timeout and use rainloop
- Log in and let the session be idle

## Observed behaviour
- Autologout and redirect to sso happens after ~30min.

## Expected behaviour
- Autologout/redirect should trigger according to configured timeout

Nebukadneza:
## What type of PR?
bug-fix

## What does this PR do?
Set the autologout variable in rainloop according with systemwide session configuration so that autologout does not trigger too early or too late, which confuses and unnerves users.

### Related issue(s)
- closes #2680 

## Prerequisites

### DIFF
--- a/webmails/rainloop/defaults/application.ini
+++ b/webmails/rainloop/defaults/application.ini
@@ -17,3 +17,6 @@ allow_sync = On
 
 [plugins]
 contacts_autosave = On
+
+[defaults]
+autologout = {{ (((PERMANENT_SESSION_LIFETIME | default(10800)) | int)/60) | int }}


Diman0:
bors try

webmails/rainloop/defaults/application.ini
```diff
@@ -17,3 +17,7 @@ allow_sync = On
 
 [plugins]
 contacts_autosave = On
+
+[defaults]
+autologout = {{ ((PERMANENT_SESSION_LIFETIME | int)/60) | int }}
```
Diman0, line 22: `PERMANENT_SESSION_LIFETIME` with default value 10800 is not set by default in mailu.env. Either you must provide a default via applications.ini or set this default in applications.ini in case the variable is not set in mailu.env.

Diman0 (dismissed):

webmails/rainloop/defaults/application.ini
Nebukadneza: Thanks for the hint, added a `default` filter.

Nebukadneza:
Sorry for the push spam. After some input in the matrix channel, I’ve now set the autologout time to the permanent session lifetime, after which the session cannot be extended, so a webmail logout ought to happen anyways. The requests from webmail keep the session alive for this time, tested with both short testing lifetimes and production values on my instance …

Diman0:
It looks good. The tests should succeed now. I still need to test the PR before I approve it.

bors try

nextgens (approved):
LGTM

nextgens:
bors cancel

nextgens (changes requested):
This should target the master branch... if we want it on 1.9 we should tag it backport.

Diman0:
I'd like to close this PR. I already copied the images from docker hub to ghcr.io.
This is also only a minor bug. We normally only do backports for high priority bugs. Within 2 weeks we will have the new Mailu release with the fix for snappymail.

nextgens:
Agreed.

PR #2682, opened 2023-03-01
Set snappymail autologout time according to SESSION_TIMEOUT

### LINKED ISSUE #2680: Rainloop autologout triggers too early in 1.9 (and possibly snappymail/master too)

Nebukadneza:
## What type of PR?
bug-fix

## What does this PR do?
Set the autologout variable in snappymail according with systemwide session configuration so that autologout does not trigger too early or too late, which confuses and unnerves users.

!!!!! Please note that I currently (due to very limited time resources) cannot test on snappymail yet, so this is a "blind" flight PR !!!!!
I know it’s a bit insolent to open PRs with untested code, deferring the testing work to somebody else, but that’s the best I can do ATM. Sorry!

### Related issue(s)
- closes #2680 
- 1.9 backport siebling: #

## Prerequisites

### DIFF
--- a/webmails/snappymail/defaults/application.ini
+++ b/webmails/snappymail/defaults/application.ini
@@ -20,6 +20,7 @@ allow_sync = On
 
 [defaults]
 contacts_autosave = On
+autologout = {{ (((PERMANENT_SESSION_LIFETIME | default(10800)) | int)/60) | int }}
 
 [cache]
 enable = On


webmails/snappymail/defaults/application.ini
```diff
@@ -20,6 +20,7 @@ allow_sync = On
 
 [defaults]
 contacts_autosave = On
+autologout = {{ ((PERMANENT_SESSION_LIFETIME | int)/60) | int }}
```
Diman0, line 23: `PERMANENT_SESSION_LIFETIME` with default value 10800 is not set by default in mailu.env. Either you must provide a default via applications.ini or set this default in applications.ini in case the variable is not set in mailu.env.

Diman0 (changes requested):
`PERMANENT_SESSION_LIFETIME` with default value 10800 is not set by default in mailu.env. Either you must provide a default via applications.ini or set this defaut in applications.ini in case the variable is not set.

webmails/snappymail/defaults/application.ini
Nebukadneza: Thanks for the hint, added a `default` filter.

Nebukadneza:
Sorry for the push spam. After some input in the matrix channel, I’ve now set the autologout time to the permanent session lifetime, after which the session cannot be extended, so a webmail logout ought to happen anyways.

Again, I’ve still only been able to test this with rainloop in #2681 , not with snappy yet. I’d appreciate a quick look from someone with a test system, sorry for that.

Diman0:
It looks good. The tests should succeed now. I still need to test the PR before I approve it.

bors try

nextgens (approved):
Works for me.

PR #2703, opened 2023-03-16
Paranoia: drop the headers we don't use

nextgens:
## What type of PR?

enhancement

## What does this PR do?

Paranoia: drop the headers we don't use. This ensures there is no misunderstanding in between front and the other containers.

## Prerequisites


Diman0 (approved):
LGTM

PR #2707, opened 2023-03-17
Fix build.hcl / CI.yml regarding labels

Diman0:
## What type of PR?

bug-fix

## What does this PR do?
The version label and versions passed to the docs image were based on the tag. Now we first build the images with -build appended to the tag, we cannot use the tag as version label anymore. 

E.g. the docs image now has the version master-build. This causes any requests to https://mailu.io/master to fail now. Due to the label master-build, it expects https://mailu.io/master-build

Two new env vars are introduced to pass the version to the build.hcl file. This will be used to set the VERSION label in the image, and pass as build arguments to the docs image.

### DIFF
--- a/.github/workflows/build_test_deploy.yml
+++ b/.github/workflows/build_test_deploy.yml
@@ -142,6 +142,8 @@ jobs:
           DOCKER_ORG: ghcr.io/${{ steps.string.outputs.lowercase }}
           MAILU_VERSION: ${{ env.MAILU_VERSION }}
           PINNED_MAILU_VERSION: ${{ env.PINNED_MAILU_VERSION }}
+          LABEL_VERSION: ${{ env.MAILU_VERSION }}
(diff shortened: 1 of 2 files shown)


Diman0:
To test via buildx. 
```
cd Mailu
MAILU_VERSION=master-build PINNED_MAILU_VERSION=master-build docker buildx bake --print -f tests/build.hcl docs
```
output
```
{
  "group": {
    "default": {
      "targets": [
        "docs"
      ]
    }
  },
  "target": {
    "docs": {
      "context": "docs",
      "dockerfile": "Dockerfile",
      "args": {
        "VERSION": "master-build",
        "pinned_version": "master-build",
        "version": "master-build"
      },
      "tags": [
        "mailu/docs:master-build",
        "mailu/docs:latest"
      ],
      "platforms": [
        "linux/amd64"
      ]
    }
  }
}

```

Note that the args contain master-build while we should pass master.  That is addressed by this PR.

nextgens (approved):

PR #2300, opened 2022-03-27
Feature HA Proxy protocol support

hitech95:
## What type of PR?

Enhancement

## What does this PR do?

This PR implements [PROXY](https://www.haproxy.org/download/2.0/doc/proxy-protocol.txt) protocol for email endpoint on `front`. This allow to have the `front` container behind another email proxy. With `PROXY` protocol we can forward the `Origin-IP/Client-IP` wich allows `admin` to properly apply throttling rules also behind a proxy. Enabling the `PROXY_PROTOCOL` all the mail endpoints expect it.

Custom endpoints have also been added to provide **auth only** endpoints. this allow to have SSL termination in the proxy and STARTLS available on port 25 without having to forward TLS traffic with the proxy. 
Those endpoins are not expected to be mapped to the host so they are not been added to the [`Dockerfile`](https://github.com/Mailu/Mailu/blob/master/core/nginx/Dockerfile#L27) `EXPOSE`  list.

### Related issue(s)
- Properly use traefik or any proxy for SSL termination: #1422
- Allow to have `front` behind a mail proxy (like another nginx instance) : closes #1472

PR #2717, opened 2023-03-25
nginx: Allow HTTP and/or TCP ports to accept the PROXY protocol

OdyX:
This is a feature proposal, as a followup to close #2300, with a cleaner split proposal.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -247,7 +247,7 @@ controls whether HTTP headers such as ``X-Forwarded-For`` or ``X-Real-IP`` shoul
 The former should be the name of the HTTP header to extract the client IP address from and the
 later a comma separated list of IP addresses designating which proxies to trust.
 If you are using Mailu behind a reverse proxy, you should set both. Setting the former without
-the later introduces a security vulnerability allowing a potential attacker to spoof his source address.
+the latter introduces a security vulnerability allowing a potential attacker to spoof his source address.
 
 The ``TZ`` sets the timezone Mailu will use. The timezone naming convention usually uses a ``Region/City`` format. See `TZ database name`_  for a list of valid timezones This defaults to ``Etc/UTC``. Warning: if you are observing different timestamps in your log files you should change your hosts timezone to UTC instead of changing TZ to your local timezone. Using UTC allows easy log correlation with remote MTAs.
 
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -253,6 +253,14 @@ The ``TZ`` sets the timezone Mailu will use. The timezone naming convention usua
 
 .. _`TZ database name`: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones
 
+The ``PROXY_PROTOCOL`` (default: unset) allows the the front container to receive TCP and HTTP connections with
+the PROXY protocol. It can be set to:
+* ``http`` to accept the ``PROXY`` protocol on nginx's HTTP proxy ports
+* ``mail`` to accept the ``PROXY`` protocol on nginx's mail proxy ports
+* ``all`` to accept the ``PROXY`` protocol on all nginx's HTTP and mail proxy ports
+
+This requires to have a valid ``REAL_IP_FROM``.
+
 Antivirus settings
 ------------------
 
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -254,12 +254,18 @@ The ``TZ`` sets the timezone Mailu will use. The timezone naming convention usua
 .. _`TZ database name`: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones
 
 The ``PROXY_PROTOCOL`` (default: unset) allows the the front container to receive TCP and HTTP connections with
-the PROXY protocol. It can be set to:
+the `PROXY protocol`_ (originally introduced in HAProxy, now also configurable in other proxy servers).
+It can be set to:
 * ``http`` to accept the ``PROXY`` protocol on nginx's HTTP proxy ports
 * ``mail`` to accept the ``PROXY`` protocol on nginx's mail proxy ports
 * ``all`` to accept the ``PROXY`` protocol on all nginx's HTTP and mail proxy ports
 
-This requires to have a valid ``REAL_IP_FROM``.
+.. _`PROXY protocol`: https://github.com/haproxy/haproxy/blob/master/doc/proxy-protocol.txt
+
+This requires to have a valid ``REAL_IP_FROM`` (default: unset). Setting ``PROXY_PROTOCOL`` without setting
+``REAL_IP_FROM`` *will not work*. Make sure to set a ``REAL_IP_FROM`` only pointing to IP addresses or networks
+that you trust; accepting the ``PROXY`` protocol from untrusted sources is a serious security vulnerability,
+allowing a potential attacker to spoof his source address.
 
 Antivirus settings
 ------------------
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -247,7 +247,7 @@ controls whether HTTP headers such as ``X-Forwarded-For`` or ``X-Real-IP`` shoul
 The former should be the name of the HTTP header to extract the client IP address from and the
 later a comma separated list of IP addresses designating which proxies to trust.
 If you are using Mailu behind a reverse proxy, you should set both. Setting the former without
-the latter introduces a security vulnerability allowing a potential attacker to spoof his source address.
+the latter introduces a security vulnerability allowing a potential attacker to spoof their source address.
 
 The ``TZ`` sets the timezone Mailu will use. The timezone naming convention usually uses a ``Region/City`` format. See `TZ database name`_  for a list of valid timezones This defaults to ``Etc/UTC``. Warning: if you are observing different timestamps in your log files you should change your hosts timezone to UTC instead of changing TZ to your local timezone. Using UTC allows easy log correlation with remote MTAs.
 
@@ -265,7 +265,7 @@ It can be set to:
 This requires to have a valid ``REAL_IP_FROM`` (default: unset). Setting ``PROXY_PROTOCOL`` without setting
 ``REAL_IP_FROM`` *will not work*. Make sure to set a ``REAL_IP_FROM`` only pointing to IP addresses or networks
 that you trust; accepting the ``PROXY`` protocol from untrusted sources is a serious security vulnerability,
-allowing a potential attacker to spoof his source address.
+allowing a potential attacker to spoof their source address.
 
 Antivirus settings
 ------------------
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -256,6 +256,7 @@ The ``TZ`` sets the timezone Mailu will use. The timezone naming convention usua
 The ``PROXY_PROTOCOL`` (default: unset) allows the the front container to receive TCP and HTTP connections with
 the `PROXY protocol`_ (originally introduced in HAProxy, now also configurable in other proxy servers).
 It can be set to:
+
 * ``http`` to accept the ``PROXY`` protocol on nginx's HTTP proxy ports
 * ``mail`` to accept the ``PROXY`` protocol on nginx's mail proxy ports
 * ``all`` to accept the ``PROXY`` protocol on all nginx's HTTP and mail proxy ports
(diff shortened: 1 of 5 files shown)


nextgens:
Thank you for your contribution. Yes, simpler has more chances of being accepted.

Have you tested it? How does it behave if an attacker send an ``X-Forwarded-For`` header to the proxy? Will mod_realip ignore it since there is no ``real_ip_header`` configured? If not, what takes precedence?

OdyX:
Thanks for your quick review.

You're absolutely right @nextgens ; this wasn't thoroughly tested. It now has (I needed to upgrade from `1.9` to `master`), and there were indeed some fixes needed. :angel: 

I've re-phrased the acceptable values to 'http' and 'mail', to match the nginx server blocks, and verified that the correct IPs get seen by the backend servers (imap, smtp, antispam and webmail at least). Yay.

As for the `X-Real-IP` conflict, you're right; that was a possibility in my first proposal. Now it becomes an alternative; either you set `REAL_IP_HEADER`, or `PROXY_PROTOCOL` to `all` or `http`; that closes that gap!

nextgens (approved):
Thank you for the revised patch.

LGTM.

I have given it some basic testing and it seems to work as expected.

docs/configuration.rst
```diff
@@ -253,6 +253,14 @@ The ``TZ`` sets the timezone Mailu will use. The timezone naming convention usua
 
 .. _`TZ database name`: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones
 
+The ``PROXY_PROTOCOL`` (default: unset) allows the the front container to receive TCP and HTTP connections with
+the PROXY protocol. It can be set to:
+* ``http`` to accept the ``PROXY`` protocol on nginx's HTTP proxy ports
+* ``mail`` to accept the ``PROXY`` protocol on nginx's mail proxy ports
+* ``all`` to accept the ``PROXY`` protocol on all nginx's HTTP and mail proxy ports
+
+This requires to have a valid ``REAL_IP_FROM``.
```
nextgens, line 262: We may want to rephrase this and put a big fat warning: setting a broad REAL_IP_FROM can turn you into an open-relay.

We may also want to clarify what the PROXY Protocol is; maybe link https://www.haproxy.org/download/2.6/doc/proxy-protocol.txt ?

micw:
### Affect/use to the helm chart and kubernetes deployment

(copied from matrix chat)

The chart already uses either a standard LB or a node port. the latter is safer in context of real/open relay but less portable (e.g. won't work on cloud servers whith no external IP, requires down-before-up redeploy).

Adding proxy protocol suport would probably work better with cloud provider's LB to maintain the original IP (but honestly I dont know who supports it and who not).

Adding haproxy as option would help to mitigate the "external traffic policy" issue: on kubernetes you either have the traffic distributed to available nodes/instances but without original ip or you have the original ip but the traffic is local on the node. With haproxy you can deploy it on every node (similar to nginx ingress) and let it distribute the traffic including the original IP over your nodes.

Diman0 (dismissed):
https://github.com/Mailu/Mailu/blob/b243ea084d51ba18e6703a04091c969d6697b408/core/nginx/conf/proxy.conf#L8
Must be updated for the situation when the proxy protocol is used and REAP_IP_HEADER is not set. Otherwise the admin container will not trust the real ip address in all scenario's.

You could change it to
```
{% if REAL_IP_HEADER and REAL_IP_FROM %}
proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
proxy_set_header X-Forwarded-By $realip_remote_addr;
{% elif PROXY_PROTOCOL == 'http' and REAL_IP_FROM %}
proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
proxy_set_header X-Forwarded-By $realip_remote_addr;
{% else %}
proxy_set_header X-Forwarded-For $remote_addr;
proxy_hide_header X-Forwarded-By;
{% endif %}
```

I'm also wondering if we should update the reverse proxy documentation page (reverse.rst) with instructions on using proxy_protocol?
@nextgens What do you think? Do we consider this a too advanced subject to mention on the reverse proxy documentation page?
Using a reverse proxy because multiple web servers on the host are used for serving web content is common. For me it is uncommon to have multiple mail servers on the same host. It is an area a novice administrator (our target audience) should not even consider.

nextgens:
Good catch. It's even, ``{% if (REAL_IP_HEADER or  PROXY_PROTOCOL in ['http', 'all']) and REAL_IP_FROM %}`` no?

nextgens:
@Diman0 I don't mind it being under-documented for now... the only thing we need to make clear is that a broad ``REAL_IP_FROM`` without other form of access control is a terrible idea.

Diman0:
> Good catch. It's even, `{% if (REAL_IP_HEADER or PROXY_PROTOCOL in ['http', 'all']) and REAL_IP_FROM %}` no?

Yes. You are correct. That is a shorter notation for the same. That would be better:
```
{% if (REAL_IP_HEADER or  PROXY_PROTOCOL in ['http', 'all']) ) and REAL_IP_FROM %}
proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
proxy_set_header X-Forwarded-By $realip_remote_addr;
{% else %}
proxy_set_header X-Forwarded-For $remote_addr;
proxy_hide_header X-Forwarded-By;
{% endif %}
```

So
`{% if REAL_IP_HEADER and REAL_IP_FROM %}`
must be changed to
`{% if (REAL_IP_HEADER or  PROXY_PROTOCOL in ['http', 'all']) ) and REAL_IP_FROM %}`

I agree we must mention in the configuration.rst that it is very important to set a specific IP for REAL_IP_FROM. We could add more wording, that if it is set to a broad range that IP addresses  not belonging to you (but within the range) can send email without authentication. The attacker can pass a client IP that is part of the Mailu subnet (or relaynets if you enabled this).

docs/configuration.rst
OdyX: [Added](https://github.com/Mailu/Mailu/pull/2717/files#diff-68a31565c7f34fbe6ac80bbc0b4287ee3711fa615ce5b973c1f83f285d0d28faR256-R263) the link to the PROXY protocol main document on Github:
> .. `PROXY protocol`: https://github.com/haproxy/haproxy/blob/master/doc/proxy-protocol.txt

And [I phrased this](https://github.com/Mailu/Mailu/pull/2717/files#diff-68a31565c7f34fbe6ac80bbc0b4287ee3711fa615ce5b973c1f83f285d0d28faR265-R268) a bit dfferently; 
> This requires to have a valid ``REAL_IP_FROM`` (default: unset). Setting ``PROXY_PROTOCOL`` without setting
``REAL_IP_FROM`` *will not work*. Make sure to set a ``REAL_IP_FROM`` only pointing to IP addresses or networks
that you trust; accepting the ``PROXY`` protocol from untrusted sources is a serious security vulnerability,
allowing a potential attacker to spoof his source address.

Diman0:
bors try

OdyX:
bors try

OdyX:
Eh. Sorry for that rst typo; I mostly wrote RST blindly. Hopefully fixed now.

Diman0 (dismissed):
The health check breaks when I enable proxy protocol for `http` or `all`.

```
mailu-front-1  | 2023/03/28 08:27:44 [error] 11#11: *1 broken header: "GET /health HTTP/1.1
mailu-front-1  | Host: localhost
mailu-front-1  | User-Agent: curl/7.88.1
mailu-front-1  | Accept: */*
mailu-front-1  | 
mailu-front-1  | " while reading PROXY protocol, client: 127.0.0.1, server: 0.0.0.0:80
```

See line 222 in nginx.conf for the relevant location block. I think we need an extra server  block with the /health location which is only used when proxy_protocol is `http` or `all`.

nextgens:
Or we could change the healtcheck to "speak" proxy protocol. v1 looks easy enough with netcat
see https://seriousben.com/posts/2020-02-exploring-the-proxy-protocol/

nextgens:
I'd rather we used something like ``kill -0 $(pidof nginx)`` tbh

OdyX:
@nextgens debatable; nginx can be up, but not serving on the right ports, etc. Checking that its server is actually up is better (but your call).

I propose https://github.com/Mailu/Mailu/pull/2717/commits/21883ffc5c0b963781ad20b4307571f7b871c39b to add a port serving just a 204 health endpoint, and leave the other health endpoint, as it still can be useful for end-to-end check;
```
    # Healthcheck over localhost, for docker
    server {
      listen 127.0.0.1:10204;
      location /health {
        return 204;
      }
    }
```

nextgens:
Hmm, OK.

But then again, this will fail if v6 is enabled as localhost may resolve to ::1. Can you force v4 use in the healtcheck by spelling out 127.0.0.1 please?

nextgens:
bors try

Diman0:
I confirmed that proxy protocol works for `http` and `all`.

nextgens (approved):
same here; works for me.

Diman0:
bors try

Diman0:
bors try-

Diman0:
bors try

Diman0:
After bors finished, I will double check the errors are gone for the health check and proof read the docs. After that I will approve the PR.

Diman0 (changes requested):
I do not have write permission, otherwise I would have pushed this as a commit to the branch Raksha-ch:feat/haproxy-protocol-split.

The indendation of the bullet points does not occur. This occurs due to the missing line break.

The documentation does also not state that the `REAL_IP_HEADER` must be unset. If `REAL_IP_HEADER` is set, then nginx will not use the remote client IP from the proxy and all connections appear to come from the proxy (resulting in the proxy being rate limited and even banned if fail2ban is used).

Maybe change the documentation to:
```
The ``PROXY_PROTOCOL`` (default: unset) allows the the front container to receive TCP and HTTP connections with
the `PROXY protocol`_ (originally introduced in HAProxy, now also configurable in other proxy servers).
It can be set to:

* ``http`` to accept the ``PROXY`` protocol on nginx's HTTP proxy ports
* ``mail`` to accept the ``PROXY`` protocol on nginx's mail proxy ports
* ``all`` to accept the ``PROXY`` protocol on all nginx's HTTP and mail proxy ports

.. _`PROXY protocol`: https://github.com/haproxy/haproxy/blob/master/doc/proxy-protocol.txt

This requires to have a valid ``REAL_IP_FROM`` (default: unset). Setting ``PROXY_PROTOCOL`` without setting
``REAL_IP_FROM`` *will not work*. The ``REAL_IP_HEADER`` **must be unset**. Otherwise Mailu will not accept
the IP address from the remote client specified by the proxy. This results in the proxy being rate limited
or even banned (when fail2ban is used).
Make sure to set a ``REAL_IP_FROM`` only pointing to IP addresses or networks
that you trust; accepting the ``PROXY`` protocol from untrusted sources is a serious security vulnerability,
allowing a potential attacker to spoof their source address.
```

docs/configuration.rst
```diff
@@ -247,12 +247,26 @@ controls whether HTTP headers such as ``X-Forwarded-For`` or ``X-Real-IP`` shoul
...
+* ``mail`` to accept the ``PROXY`` protocol on nginx's mail proxy ports
+* ``all`` to accept the ``PROXY`` protocol on all nginx's HTTP and mail proxy ports
+
+.. _`PROXY protocol`: https://github.com/haproxy/haproxy/blob/master/doc/proxy-protocol.txt
+
+This requires to have a valid ``REAL_IP_FROM`` (default: unset). Setting ``PROXY_PROTOCOL`` without setting
+``REAL_IP_FROM`` *will not work*. Make sure to set a ``REAL_IP_FROM`` only pointing to IP addresses or networks
+that you trust; accepting the ``PROXY`` protocol from untrusted sources is a serious security vulnerability,
+allowing a potential attacker to spoof their source address.
+
```
OdyX, line 269:
```suggestion
The ``PROXY_PROTOCOL`` (default: unset) allows the the front container to receive TCP and HTTP connections with
the `PROXY protocol`_ (originally introduced in HAProxy, now also configurable in other proxy servers).
It can be set to:

* ``http`` to accept the ``PROXY`` protocol on nginx's HTTP proxy ports
* ``mail`` to accept the ``PROXY`` protocol on nginx's mail proxy ports
* ``all`` to accept the ``PROXY`` protocol on all nginx's HTTP and mail proxy ports

.. _`PROXY protocol`: https://github.com/haproxy/haproxy/blob/master/doc/proxy-protocol.txt

This requires to have a valid ``REAL_IP_FROM`` (default: unset). Setting ``PROXY_PROTOCOL`` without setting
``REAL_IP_FROM`` *will not work*. The ``REAL_IP_HEADER`` **must be unset**. Otherwise Mailu will not accept
the IP address from the remote client specified by the proxy. This results in the proxy being rate limited
or even banned (when fail2ban is used).
Make sure to set a ``REAL_IP_FROM`` only pointing to IP addresses or networks
that you trust; accepting the ``PROXY`` protocol from untrusted sources is a serious security vulnerability,
allowing a potential attacker to spoof their source address.

```

As suggested by @Diman0 .

Diman0 (approved):
LGTM. It took some effort, but everything looks fine now. The proxy protocol works in all three situations. The documentation is clear enough now with warnings to prevent misconfiguration.

nextgens (approved):
LGTM

Issue #2533, opened 2022-11-14
database - general connection string support (feature request)


ZsBT:
Most languages support _DB URL_ recently. Host, dbname, user, password do not cover the required connection parameters in all cases.
E.g., even `psycopg2.connect` can accept something like this:

`postgresql://user:password@eu-central-1.cockroachlabs.cloud:15757/roundcube?sslmode=verify-full&options=--cluster%3Ddeep-beast-666
`

We could give a new environment variable like this:

`DB_CONNSTRING=postgresql://xyz...`

nextgens:
I am fairly sure that if you get rid of ``DB_FLAVOR`` and set ``SQLALCHEMY_DATABASE_URI`` instead it will do the right thing. Granted this should be documented somewhere.

PR #2719, opened 2023-03-26
Introduce connection string (database url) for roundcube & simplify setup

Diman0:
## What type of PR?

enhancement

## What does this PR do?
As discussed in earlier Mailu meetings (#1582), we want to simplify configuring Mailu and make it more user-friendly. Especially the last part is an important mission statement of the Mailu project. 

This PR will remove the choice of what DB to use from setup. New users are guided now to make the correct choice of using SQLite.

For simplifying the configuration, all the database environment variables have been removed and replaced with a single connection string environment variable. 

For backwards compatibility, the old *DB_* setting can still be used. This is to make sure that master does not immediately break for all users. After X months after the next Mailu release, we can remove the old settings from the software. This provides a transition period. 

### Related issue(s)
- #2533

## Prerequisites


Diman0:
I only tested SQLite and MariaDB. So I can confrm that the updated mysql connector now also works with MariaDB.

Diman0:
bors try

nextgens (approved):
LGTM

PR #2721, opened 2023-03-27
Fix ci arm buildcache

Diman0:
## What type of PR?

Fix/enhancement CI

## What does this PR do?
Make sure that the arm build also uses build-cache. Remove the step of building the base image. This should not be required. When a main image is build, it will first try to build the base target. This build cache is stored in build cache of the image itself. This cache is re-used on the next workflow.

We can confirm this by doing a second bors try after the first bors try has succeeded.


nextgens (approved):
LGTM

PR #2725, opened 2023-03-28
Fix access to radicale

### LINKED ISSUE #2723: webdav/radicale is broken in master
1) setup is producing a ``docker-compose.yml`` where front is not connected to the radicale network whereas it should be
2) authentication via X-Remote-User seems broken to me: I get presented with a login prompt after basic auth
3) there is no good reason for it to run as root

I think this should be addressed before the release.

Diman0:
## What type of PR?

bug-fix

## What does this PR do?

Fix the access issue to radicale. I did not create a newsfragment, because this works fine on 1.9. This was only broken on master. 

### Related issue(s)
- closes #2723

## Prerequisites


nextgens (approved):
LGTM

PR #2732, opened 2023-04-01
Only account for distinct attempts in rate limits

nextgens:
## What type of PR?

enhancement

## What does this PR do?

Only account for distinct attempts in rate limits. This is solving the problem related to users changing their passwords and having their client hammer the old credentials.

Reduce the default to 50 distinct passwords per day

## Prerequisites


core/admin/mailu/limiter.py
```diff
@@ -68,9 +68,13 @@ def should_rate_limit_user(self, username, ip, device_cookie=None, device_cookie
             app.logger.warn(f'Authentication attempt from {ip} for {username} has been rate-limited.')
         return is_rate_limited
 
-    def rate_limit_user(self, username, ip, device_cookie=None, device_cookie_name=None):
+    def rate_limit_user(self, username, ip, device_cookie=None, device_cookie_name=None, password=''):
         limiter = self.get_limiter(app.config["AUTH_RATELIMIT_USER"], 'auth-user')
         if self.is_subject_to_rate_limits(ip):
+            truncated_password = hmac.new(bytearray(username, 'utf-8'), bytearray(password, 'utf-8'), 'sha256').hexdigest()[-6:]
+            if password and self.storage.get(f'dedup2-{username}-{truncated_password}') > 0:
```
Diman0, line 75:
```suggestion
            if password and (self.storage.get(f'dedup2-{username}-{truncated_password}') > 0):
```

Make it explicitly clear that `and` takes precedence over `>`>

Diman0 (dismissed):
It looks good to me. I only request that you add extra parentheses in the IF statement, to make it visible what condition operator should be evaluated first.

Diman0:
It looks good to me. I only need to test it.

bors try

Diman0 (approved):
LGTM
Tested and it works as expected.

PR #2734, opened 2023-04-02
Fix setup

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Remove LOG_DRIVER which never worked and replace it with journald by default
Fix a bug where front may get attached to networks that don't exist (#2725 was incomplete)

### Related issue(s)
- #2725

## Prerequisites

### DIFF
--- a/docs/compose/.env
+++ b/docs/compose/.env
@@ -135,12 +135,6 @@ WEBSITE=https://mailu.io
 # Advanced settings
 ###################################
 
-# Log driver for front service. Possible values:
-# json-file (default)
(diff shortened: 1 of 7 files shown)


Diman0:
@nextgens 
The fail2ban docs use container tag mailu-front.
Either change the fail2ban doc to use docker-front and docker-admin as container tags. 
Or change in the docker-compose file all docker-service tags to mailu-service tags (mailu-front, mailu-admin)

I personally prefer the tag `mailu-service` (`mailu-front`). We already use this in the fail2ban docs and this is more specific to mailu.

Diman0 (approved):
LGTM

PR #2741, opened 2023-04-04
Discuss reverting "STARTTLS is a bad idea" 

OdyX:
This reverts commit 0bccb5045c1f2ccd2f8ac13fc7e03faa43049912, which was incorporated directly in `main` without much (public? ) discussion that I've seen.

https://en.wikipedia.org/wiki/Opportunistic_TLS

The point is to enforce TLS in all cases, why still providing the StartTLS endpoints at all then?

### DIFF
--- a/core/admin/mailu/ui/templates/client.html
+++ b/core/admin/mailu/ui/templates/client.html
@@ -17,7 +17,7 @@
     </tr>
     <tr>
       <th>{% trans %}TCP port{% endtrans %}</th>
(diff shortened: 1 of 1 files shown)


Diman0:
We only advertise implicit TLS due to the vulnerabilities mentioned in your linked article and described in this article
https://nostarttls.secvuln.info/

OdyX:
Excellent, thanks for the detailed answer.

PR #2748, opened 2023-04-07
Improve ci by using intermediate images

Diman0:
## What type of PR?

enhancement

## What does this PR do?
As discussed in the mailu-dev chat, to improve build times we should
1. First build the base image and push this to ghcr.io
2. Secondly at the same time build the assets images and push this to ghcr.io.
3. Simultanuously build all other images which use the base/assets image from ghcr.io.

I've made these changes to the build_test_deploy.yml file and build-ci.yml file.
https://github.com/Mailu/Mailu/actions/runs/4637399505
In the worst case the build time appears to be around ~13 minutes now.

- build-ci.hcl will be used only by the github workflow. This file is modified to pull the base/assets image from the registry.
- build.hcl is unmodified. This file can be used to build all images locally.

I changed the name of the build jobs to make sure they fit in the overview list when viewing the action on github. You can now actually identify what job is run for what matrix target.

## Prerequisites

### DIFF
--- a/.github/workflows/build_test_deploy.yml
+++ b/.github/workflows/build_test_deploy.yml
@@ -101,13 +101,162 @@ jobs:
         run: |
           echo ${{ steps.targets.outputs.matrix }}
 
-# This job builds all the images. The build cache is stored in the github actions cache.
-# In further jobs, this cache is used to quickly rebuild the images.
+# This job builds the base and assets images. The build cache is stored in the ghcr.io registry.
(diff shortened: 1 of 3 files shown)


Diman0:
I think it is a good idea to backport this to the 2.0 branch. This ensures the images are also build more quickly for the 2.0 branch.

It looks like that build times are now between ~6 to ~14 minutes.

nextgens (approved):

Issue #2747, opened 2023-04-07
config-import failes with error '[TypeError] encoding without a string argument'


4athan:
### Environment

- [x] docker compose
- [ ] kubernetes
- [ ] docker swarm

### Version

- Version: `master and 2.0`

## Description
An exported config yaml or JSON with secrets can not be imported on another similar system. Every try produces: 
...
Error: [TypeError] encoding without a string argument
...

## Replication Steps
Export a config file on the cli with config-export and --secrets option. Then try to import it back in.

## Observed behaviour
...
Error: [TypeError] encoding without a string argument
...

## Expected behaviour
Config is imported.

## Logs

I have not seen any output related to this in the logs.

Diman0:
I can replicate this using:
```
docker compose exec -T admin flask mailu config-export --secrets > test.yml
docker compose exec -T admin flask mailu config-import -nv < test.yml
Error: [TypeError] encoding without a string argument
```

But it also occurs when secrets are not included:
```
docker compose exec -T admin flask mailu config-export > test2.yml
docker compose exec -T admin flask mailu config-import -nv < test2.yml
Error: [TypeError] encoding without a string argument
```

This appears to be a defect.

Diman0:
@4athan Please run the same import command again with the `--debug` switch and add the output to this issue.  I want ot make sure you have the same stack trace.

```
docker compose exec -T admin flask mailu config-import --debug --verbose --dry-run /data/test2.yml
INFO:sqlalchemy.engine.Engine:BEGIN (implicit)
INFO:sqlalchemy.engine.Engine:DELETE FROM alias
INFO:sqlalchemy.engine.Engine:[generated in 0.00020s] ()
INFO:sqlalchemy.engine.Engine:DELETE FROM domain
INFO:sqlalchemy.engine.Engine:[generated in 0.00023s] ()
INFO:sqlalchemy.engine.Engine:DELETE FROM relay
INFO:sqlalchemy.engine.Engine:[generated in 0.00014s] ()
INFO:sqlalchemy.engine.Engine:DELETE FROM user
INFO:sqlalchemy.engine.Engine:[generated in 0.00015s] ()
INFO:sqlalchemy.engine.Engine:ROLLBACK
Traceback (most recent call last):
  File "/app/venv/bin/flask", line 8, in <module>
    sys.exit(main())
  File "/app/venv/lib/python3.10/site-packages/flask/cli.py", line 1047, in main
    cli.main()
[...]

PR #2752, opened 2023-04-07
Fix config-import. Config with dkim key could not be imported.

Diman0:
## What type of PR?

bug-fix

## What does this PR do?
Imports with dkim keys failed due to two bugs
- The situation where the DKIM key was included as \<hidden\> in the config.yml was not handled
- An attempt was made to byte encode a value that was already a byte encoded value

This PR fixes both issues. Now config yml's with valid dkim keys and dkim keys with value \<hidden\> can be imported again using config-import.

### Related issue(s)
- #2747 

## Prerequisites


ghostwheel42 (approved):
Lgtm

PR #428, opened 2018-03-28
Add managesieve proxy

mildred:
Add managesieve proxy so it can be accessed from the outside. See #308

Built on top of #429

Issue #113, opened 2016-11-11
Implement Managesieve in dovecot


fabiorauber:
Brainstorming and comparing against my current solution (dovecot + roundcube based), would it be easy to enable managesieve protocol in dovecot, and enable it in rainloop or roundcube, so that the user has control of his own filters?

kaiyou:
I believe it is enabled already and available from roundcube. Some work might be required for rainloop.

kaiyou:
4e84fd1f6ae17e40f625fc6e4f177a4134fb1412 did fix this for Rainloop, should be working on ``master``.

Issue #81, opened 2016-10-18
Add proper managesieve support


kaiyou:
Currently Managesieve support is broken and unusable (not even exposed). We should make sure that:

- managesieve is properly enabled and users can modify rules
- STARTTLS is enabled and enforced
- the Roundcube filtering plugin is enabled

PR #2773, opened 2023-04-14
Expose managesieve to the outside world

nextgens:
## What type of PR?

Feature

## What does this PR do?

Expose managesieve to the outside world.

### Related issue(s)
- #428
- #113
- #81

## Prerequisites


nextgens:
It's a little bit quirky right now:
- only implicit TLS is supported (should we use the standard port?)
- dovecot doesn't know about the real client IP
```
c5ab2744bdbc:/app# doveadm who
username          # proto (pids) (ips)                                           
admin@example.com 1 sieve (454)  (192.168.203.8)
```
- authentication never fails permanently; we always return a temporary error:
```
"IMPLEMENTATION" "Dovecot Pigeonhole"
"SIEVE" "fileinto reject envelope encoded-character vacation subaddress comparator-i;ascii-numeric relational regex imap4flags copy include variables body enotify environment mailbox date index ihave duplicate mime foreverypart extracttext"
"NOTIFY" "mailto"
"SASL" "PLAIN LOGIN"
"VERSION" "1.0"
OK "Dovecot ready."
AUTHENTICATE "PLAIN" "AGFkbWluQGV4YW1wbGUuY29tAGxldG1laW4y"
NO (TRYLATER) "Temporary authentication failure. [c5ab2744bdbc:2023-04-14 17:02:45]"
```

@fastlorenzo let me know what you think.

nextgens:
The temporary auth fail is now solved; we can adjust the message as required.
```
"IMPLEMENTATION" "Dovecot Pigeonhole"
"SIEVE" "fileinto reject envelope encoded-character vacation subaddress comparator-i;ascii-numeric relational regex imap4flags copy include variables body enotify environment mailbox date index ihave duplicate mime foreverypart extracttext"
"NOTIFY" "mailto"
"SASL" "PLAIN LOGIN"
"VERSION" "1.0"
OK "Dovecot ready."
AUTHENTICATE "PLAIN" "AGFkbWluQGV4YW1wbGUuY29tAGxldG1laW4y"
NO "AuthFailed"
```

fastlorenzo:
Will test it out asap

nextgens:
bors try

nextgens:
bors try

nextgens:
bors try

Diman0 (dismissed):
- Changelog entry is missing.
- Proxy protocol documentation in Configuration.rst must be updated with the new `sieve` value.

Does the rate limiting functionality also work for managesieve? It will log the source ip address that is rate limited?

nextgens:
> Does the rate limiting functionality also work for managesieve?

No, it won't if we implement it the way currently proposed in this PR.

docs/configuration.rst
```diff
@@ -260,6 +260,7 @@ It can be set to:
 
 * ``http`` to accept the ``PROXY`` protocol on nginx's HTTP proxy ports
 * ``mail`` to accept the ``PROXY`` protocol on nginx's mail proxy ports
+* ``sieve`` to accept the ``PROXY`` protocol on nginx's managesieve proxy port
```
OdyX, line 263: Shall I suggest merging that with `mail` ? Having two made configuration easy (all combinations from "none, only mail, only http, both" are configurable).

With a third option, you either need 7 option names (to allow "mail + sieve" or "http + sieve", or complex string comparison.

fastlorenzo:
@nextgens it doesn't seem to work for me for an obscure reason...

Here are packet captures of front and dovecot containers when accessing sieve from the internet:
image


I can confirm connectivity between front and dovecot works on port 4190: 
image

I've checked the front logs but I don't see anything logged when I attempt to reach sieve from the outside world.

docs/configuration.rst
nextgens: Ok, changed in 801e6fe0

nextgens:
bors try

nextgens:
I think that #2790 is a better approach; I'll close this.

PR #2772, opened 2023-04-14
Always exempt app-tokens from rate limits

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Always exempt app-tokens from rate limits

## Prerequisites
Before we can consider review and merge, please make sure the following list is done and checked.
If an entry in not applicable, you can check it or remove it from the list.

- [ ] In case of feature or enhancement: documentation updated accordingly
- [x] Unless it's docs or a minor change: add [changelog](https://mailu.io/master/contributors/workflow.html#changelog) entry file.

PR #2785, opened 2023-04-20
comment regarding new rate limiting mechanism

elandorr:
I was confused why you lowered it this dramatically, then saw this one https://github.com/Mailu/Mailu/commit/d20c217ae68213aa0d5236e3189d758322f77c05.

This just changes the comment to make it clear in the generated .env.

I'm still unsure, if such a low number makes sense, though. What about people on VPN/CGNAT?

From the changelog:

> Now the rate limiter will only take distinct attempts into account. We have two different types of checks:
> 
>     to prevent crendential bruteforce (an attacker trying to guess a password), we limit the maximal amount of attempts an attacker has for a given account (from any IP address).
> 
>     to prevent password spraying (an attacker trying the same common password on all accounts he can enumerate), we limit the maximal number of non-existing accounts an attacker can attempt to authenticate against from a given network subnet.

Maybe the wording is still wrong, and this does *not* blacklist an IP for trying 5 random accounts, but only the combination of IP+account. Then I can understand why you lowered it so much I suppose. Still seems like that would lock out VPN/TOR/similar users right away. The bots hit all day, thousands of times.

## What type of PR?

documentation

## What does this PR do?

clarify

### DIFF
--- a/setup/flavors/compose/mailu.env
+++ b/setup/flavors/compose/mailu.env
@@ -29,7 +29,7 @@ POSTMASTER={{ postmaster }}
 # Choose how secure connections will behave (value: letsencrypt, cert, notls, mail, mail-letsencrypt)
 TLS_FLAVOR={{ tls_flavor }}
 
-# Authentication rate limit per IP (per /24 on ipv4 and /48 on ipv6)
+# Authentication rate limit per IP for failed login attempts or non-existing accounts (per /24 on ipv4 and /48 on ipv6)
 {% if auth_ratelimit_ip > '0' %}
 AUTH_RATELIMIT_IP={{ auth_ratelimit_ip }}/hour
 {% endif %}


setup/flavors/compose/mailu.env
```diff
@@ -29,7 +29,7 @@ POSTMASTER={{ postmaster }}
 # Choose how secure connections will behave (value: letsencrypt, cert, notls, mail, mail-letsencrypt)
 TLS_FLAVOR={{ tls_flavor }}
 
-# Authentication rate limit per IP (per /24 on ipv4 and /48 on ipv6)
+# Authentication rate limit per IP for failed login attempts or non-existing accounts (per /24 on ipv4 and /48 on ipv6)
```
nextgens, line 32: This does not accurately reflect what's currently in master; it may if/once #2772 is merged.

Diman0: @nextgens 
#2772 has been merged now. I think this is correct now?

The text itself contains a typo
```suggestion
# Authentication rate limit per IP for failed login attempts on non-existing accounts (per /24 on ipv4 and /48 on ipv6)
```

elandorr (commented):
I took it from the commit in OP. But https://github.com/Mailu/Mailu/commit/bb127d15ffccfbfeb3e9c61851105fdd07562465 came right after, my bad. Replaced.

elandorr (commented):
or with range

What consequences does this have for CGNAT by the way?

setup/flavors/compose/mailu.env
```diff
@@ -29,7 +29,7 @@ POSTMASTER={{ postmaster }}
 # Choose how secure connections will behave (value: letsencrypt, cert, notls, mail, mail-letsencrypt)
 TLS_FLAVOR={{ tls_flavor }}
 
-# Authentication rate limit per IP (per /24 on ipv4 and /48 on ipv6)
+# Authentication rate limit per IP for failed login attempts on unique non-existing accounts (password spraying counter-measure, per /24 on ipv4 and /48 on ipv6)
```
ghostwheel42, line 32: We could also mention `AUTH_RATELIMIT_IP_V4_MASK` and `AUTH_RATELIMIT_IP_V6_MASK`.
What about this?

```suggestion
# Authentication rate limit per network for failed login attempts on unique non-existing accounts (password spraying counter-measure)
# Network size defaults to /24 on IPv4 and to /48 on IPv6 (see also AUTH_RATELIMIT_IP_V4_MASK and AUTH_RATELIMIT_IP_V6_MASK)
```

PR #2786, opened 2023-04-20
adapted to v2 release and a docker change

elandorr:
- v2 changed the path
- docker deprecated/removed the scale command, you have to do it like this now

## What type of PR?
documentation

### DIFF
--- a/docs/antispam.rst
+++ b/docs/antispam.rst
@@ -83,8 +83,8 @@ Via the multimap filter it is possible to block emails from a sender domain. See
 
 The following steps have to be taken to configure an additional symbol (rule) that uses the multimap filter to block emails from sender domain.
 
-1. In the overrides folder create a configuration file for the multimap filter. This configuration is included by Rspamd in the main multimap configuration file. This means you do not have to use the "multimap {}" element. Files in the /mailu/overrides/rspamd/ folder are mapped to /etc/rspamd/override.d.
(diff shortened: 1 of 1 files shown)


docs/antispam.rst
```diff
@@ -129,8 +129,8 @@ The following steps have to be taken to configure an additional symbol (rule) th
 
    .. code-block:: bash
 
-     docker compose scale antispam=0
-     docker compose scale antispam=1
+     docker compose up -d --scale antispam=0
+     docker compose up -d --scale antispam=1
```
nextgens, line 133: The new command to use is:
``docker compose up antispam --force-recreate -d``

nextgens (changes requested):

docs/antispam.rst
elandorr: Both work, I followed the existing docs. Pick whichever.

nextgens (approved):
LGTM

Diman0 (approved):
LGTM

Good catch. We indeed missed this when updating the docs. I forgot we also mentioned this in antispam.rst.

enginefeeder101:
Would it be possible to backport this to 2.0?
I took me triple checking and eventually looking through the changelog and Github history to figure out why the [example](https://mailu.io/2.0/antispam.html#how-can-i-block-emails-from-a-domain) wasn't working.

PR #2799, opened 2023-04-27
Fix roundcube's spellchecker

### LINKED ISSUE #2798: roundcube - spell check - only English available
### Environment

- [x] docker compose
- [ ] kubernetes
- [ ] docker swarm

### Version
- Version: `2.0.10`

## Description
Since the 2.x release the German spell check is missing, in fact it looks like as if only English spell check is available. 

Based on my understanding it is using pspell and for this additional language files (dictionary) would have to be deployed, all thought I have not figured out how yet.

There are a lot of dictionaries available at http://ftp.gnu.org/gnu/aspell/dict/0index.html which I think should work with pspell.

## Replication Steps
I opened roundcube wrote an email and clicked on spell check but it only offered me English. My roundcube configuration was English and I thought maybe changing the UI language will change the spell check language too 
image
but despite changing it to German
image
the spell check language is still English only

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Without this only 'English' is offered.

### Related issue(s)
- close #2798 

## Prerequisites

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -888,7 +888,7 @@ $config['spellcheck_ignore_nums'] = true;
 $config['spellcheck_dictionary'] = true;
 ```
 
-If not, you can download the `aspell dictionnary`_ you require and place it in ``/usr/share/aspell/``. And then enable it by tweaking the following in the configuration file:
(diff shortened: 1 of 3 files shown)


ghostwheel42 (approved):
Lgtm

PR #2808, opened 2023-05-06
Ensure SCAN_MACROS works

### LINKED ISSUE #2805: SCAN_MACROS is incorrect case in Rspamd configs
In Mailu/core/rspamd/conf/external_services.conf, the first line is '{% if SCAN_MACROS == 'True' %}'. It is also the same in external_services_group.conf, note the capital 'T' in 'True'. When the mailu.env is generated, it generates SCAN_MACROS=true, i.e . all lowercase. Thus, in order to enable oletools, one must put SCAN_MACROS=True in environment file. I'm not sure what other items are reliant on SCAN_MACROS, but they should be checked as well.

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Ensure SCAN_MACROS works

### Related issue(s)
- closes #2805

## Prerequisites

### DIFF
--- a/core/rspamd/conf/composites.conf
+++ b/core/rspamd/conf/composites.conf
@@ -1,4 +1,4 @@
-{% if SCAN_MACROS == 'True' %}
+{% if SCAN_MACROS %}
 OLETOOLS_MACRO_MRAPTOR {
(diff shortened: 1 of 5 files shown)


ghostwheel42 (approved):
Lgtm

PR #2818, opened 2023-05-10
Improve auth-related logging

### LINKED ISSUE #2803: Improve auth-related logging
At the very least we should log:
- success/failure
- which service?
- which credentials were used if successful
- where from (including source port info)

arguably:
- a truncated hash of the failed creds
- where we are at rate-limiting wise

nextgens:
## What type of PR?

enhancement

## What does this PR do?

Improve auth-related logging

### Related issue(s)
- closes #2803 

## Prerequisites

### DIFF
--- a/core/admin/mailu/internal/nginx.py
+++ b/core/admin/mailu/internal/nginx.py
@@ -27,23 +27,26 @@ STATUSES = {
 
 WEBMAIL_PORTS = ['10143', '10025']
 
-def check_credentials(user, password, ip, protocol=None, auth_port=None):
+def check_credentials(user, password, ip, protocol=None, auth_port=None, source_port=None):
     if not user or not user.enabled or (protocol == "imap" and not user.enable_imap and not auth_port in WEBMAIL_PORTS) or (protocol == "pop3" and not user.enable_pop):
+        app.logger.info(f'Login attempt for: {user}/{protocol}/{auth_port} from: {ip}/{source_port}: failed: accound disabled')
         return False
-    is_ok = False
     # webmails
     if auth_port in WEBMAIL_PORTS and password.startswith('token-'):
         if utils.verify_temp_token(user.get_id(), password):
-            is_ok = True
-    if not is_ok and utils.is_app_token(password):
+            app.logger.debug(f'Login attempt for: {user}/{protocol}/{auth_port} from: {ip}/{source_port}: success: webmail-token')
+            return True
+    if utils.is_app_token(password):
         for token in user.tokens:
             if (token.check_password(password) and
                 (not token.ip or token.ip == ip)):
-                    is_ok = True
-                    break
-    if not is_ok and user.check_password(password):
-        is_ok = True
-    return is_ok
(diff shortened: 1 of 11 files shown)


nextgens:
bors try-
bors try

core/admin/mailu/ui/templates/token/list.html
```diff
@@ -16,6 +16,7 @@
 {%- call macros.table(order='[[1,"asc"]]') %}
 <thead>
   <tr>
+    <th>{% trans %}ID{% endtrans %}</th>
```
ghostwheel42, line 19: I think you should add the ID in the second column and change `1,"asc"` to `2,"asc"` in line 16.
I have started a PR for this "ordering problem", but unfortunately didn't finish it yet.

core/admin/mailu/sso/views/base.py
```diff
@@ -96,10 +97,12 @@ def _proxy():
     proxy_ip = flask.request.headers.get('X-Forwarded-By', flask.request.remote_addr)
     ip = ipaddress.ip_address(proxy_ip)
     if not any(ip in cidr for cidr in app.config['PROXY_AUTH_WHITELIST']):
+        flask.current_app.logger.error('Login failed by proxy - not on whitelist: from {client_ip} through {flask.request.remote_addr}.')
```
ghostwheel42, line 100: I think there is the `f""` missing in the four log messages, isn't it?

core/admin/mailu/internal/nginx.py
```diff
@@ -27,23 +27,26 @@
 
 WEBMAIL_PORTS = ['10143', '10025']
 
-def check_credentials(user, password, ip, protocol=None, auth_port=None):
+def check_credentials(user, password, ip, protocol=None, auth_port=None, source_port=None):
     if not user or not user.enabled or (protocol == "imap" and not user.enable_imap and not auth_port in WEBMAIL_PORTS) or (protocol == "pop3" and not user.enable_pop):
+        app.logger.info(f'Login attempt for: {user}/{protocol}/{auth_port} from: {ip}/{source_port}: failed: accound disabled')
```
ghostwheel42, line 32: Hmm, splitting the port with "/" is strange...
It should be "account disabled"

ghostwheel42 (changes requested):

core/admin/mailu/sso/views/base.py
nextgens: Good catch. that's bd398613

core/admin/mailu/internal/nginx.py
nextgens: The idea was to be able to ``grep|cut -d: |cut -d/ `` easily ... What would you rather use? ``!``? ``|``?

nextgens:
bors try

nextgens:
bord try

nextgens:
bors try

nextgens:
bors try

Diman0 (approved):
Looks good to me. I tested everything but the header proxy stuff.

PR #2823, opened 2023-05-16
Don't send ooo messages to noreply@

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Don't send ooo messages to noreply@

## Prerequisites

### DIFF
--- a/core/admin/mailu/internal/templates/default.sieve
+++ b/core/admin/mailu/internal/templates/default.sieve
@@ -35,5 +35,7 @@ if exists "X-Virus" {
 }
 
 {% if user.reply_active %}
-vacation :days 1 {% if user.displayed_name != "" %}:from "{{ user.displayed_name }} <{{ user.email }}>"{% endif %} :subject "{{ user.reply_subject }}" "{{ user.reply_body }}";
+if not address :localpart :contains ["From","Reply-To"] ["noreply","no-reply"]{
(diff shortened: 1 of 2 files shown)


core/admin/mailu/internal/templates/default.sieve
```diff
@@ -35,5 +35,7 @@ if exists "X-Virus" {
 }
 
 {% if user.reply_active %}
-vacation :days 1 {% if user.displayed_name != "" %}:from "{{ user.displayed_name }} <{{ user.email }}>"{% endif %} :subject "{{ user.reply_subject }}" "{{ user.reply_body }}";
+if not address :localpart :contains ["From","Reply-To"] ["noreply","no-reply"]{
```
ghostwheel42, line 38: Is this correct? Isn't the vacation message sent to the "envelope sender"?
Looks like this is the part where the vacation responses recipient is determined:
https://github.com/dovecot/pigeonhole/blob/87ea4d74a051399222b6e0ca0e6f59700c89f923/src/lib-sieve/plugins/vacation/cmd-vacation.c#L864

nextgens: If by correct you mean RFC compliant, the answer is no... But it matches real-world usage, where there is no point in sending ooo to addresses that won't accept them or won't be read.

ghostwheel42: But the ooo message isn't sent to the address mentioned in "reply-to", isnt it?
Aren't there other headers we should look at when adding an exception like "X-Auto-Response-Suppress" or mailing list headers?

PR #2838, opened 2023-05-30
Authentication failed for email clients when the password contained a non latin-1 character.

### LINKED ISSUE #2837: No special characters in passwords allowed
### Environment

- [x] docker compose
- [ ] kubernetes
- [ ] docker swarm

### Version

- Version: `2.0`

## Description

Authentication fails when using special characters in a user's password.

I cannot use special characters in a password, e.g. '€'. It worked in version 1.9

## Replication Steps

Use for example '€' in a user's password.

## Observed behaviour

Clients are not able to login to IMAP server.

## Expected behaviour

Clients should be able to login to IMAP server.

## Logs

Logs from `admin` container:

```
192.168.203.1 - - [30/May/2023:08:39:07 +0000] "GET /internal/auth/email HTTP/1.0" 500 0 "-" "-"
[2023-05-30 08:39:07 +0000] [18] [ERROR] Error handling request /internal/auth/email
  File "/app/venv/lib/python3.10/site-packages/gunicorn/workers/gthread.py", line 271, in handle
    keepalive = self.handle_request(req, conn)
  File "/app/venv/lib/python3.10/site-packages/gunicorn/workers/gthread.py", line 331, in handle_request
    resp.close()
  File "/app/venv/lib/python3.10/site-packages/gunicorn/http/wsgi.py", line 391, in close
    self.send_headers()
  File "/app/venv/lib/python3.10/site-packages/gunicorn/http/wsgi.py", line 322, in send_headers
    util.write(self.sock, util.to_bytestring(header_str, "latin-1"))
  File "/app/venv/lib/python3.10/site-packages/gunicorn/util.py", line 565, in to_bytestring
    return value.encode(encoding)
UnicodeEncodeError: 'latin-1' codec can't encode character '\u20ac' in position 267: ordinal not in range(256)
```

Diman0:
## What type of PR?

bug fix

## What does this PR do?
Fixes a bug that results in authentication failing for email clients when the password contains a non latin-1 character.
Issue was caused by the header Auth-Password being returned with non latin-1 characters. Headers must always be latin-1 encoded. Resolved the issue by url encoding the password.

Since the returned password is only used as a partial hash for the rate limiter, I did not add code for un-quoting the password in the /internal/email endpoint.

### Related issue(s)
- closes #2837 

## Prerequisites

### DIFF
--- a/core/admin/mailu/internal/views/auth.py
+++ b/core/admin/mailu/internal/views/auth.py
@@ -58,7 +58,7 @@ def nginx_authentication():
         try:
             password = raw_password.encode("iso8859-1").decode("utf8")
         except:
-            app.logger.warn(f'Received undecodable password from nginx: {raw_password!r}')
+            app.logger.warn(f'Received undecodable password for {username} from nginx: {raw_password!r}')
             utils.limiter.rate_limit_user(username, client_ip, password=None)
         else:
             utils.limiter.rate_limit_user(username, client_ip, password=password)
--- a/core/admin/mailu/internal/views/auth.py
+++ b/core/admin/mailu/internal/views/auth.py
@@ -6,6 +6,7 @@ import flask
 import flask_login
 import base64
 import sqlalchemy.exc
+import urllib
 
 @internal.route("/auth/email")
 def nginx_authentication():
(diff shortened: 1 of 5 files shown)


nextgens (dismissed):
Check line 123 out; there is another instance of the same.

We need to ensure that the decoded and non-decoded versions match the same hash

Diman0:
bors try

core/admin/mailu/internal/views/auth.py
```diff
@@ -52,7 +54,14 @@ def nginx_authentication():
         if not is_port_25:
             utils.limiter.exempt_ip_from_ratelimits(client_ip)
     elif is_valid_user:
-        utils.limiter.rate_limit_user(username, client_ip, password=response.headers.get('Auth-Password', None))
+        password = None
+        try:
+            password = raw_password.encode("iso8859-1").decode("utf8")
+        except:
+            app.logger.warn(f'Received undecodable password from nginx: {raw_password!r}')
```
nextgens, line 61:
```suggestion
            app.logger.warn(f'Received undecodable password for {username} from nginx: {raw_password!r}')
```

Issue #2819, opened 2023-05-10
Postfix log file not working


Lex999:
### Environment

- [x] docker compose
- [ ] kubernetes
- [ ] docker swarm

### Version

- Version: 2.0.12

## Description

Postfix logs are not written to file after setting the POSTFIX_LOG_FILE environment variable.

## Replication Steps
Set the POSTFIX_LOG_FILE environment variable of the Postfix container to some path which is mounted into the container.

## Observed behaviour
The follwing lines are written to the log file:

```
WARNING:root:Disabling hardened-malloc on this CPU
INFO:MAIN:MTA-STS daemon starting...
2023-05-10 20:16:07 INFO     MAIN: MTA-STS daemon starting...
INFO:MAIN:Starting eventloop...
INFO:MAIN:uvloop is not available. Falling back to built-in event loop.
INFO:MAIN:Eventloop started.
2023-05-10 20:16:07 INFO     MAIN: Starting eventloop...
2023-05-10 20:16:07 INFO     MAIN: uvloop is not available. Falling back to built-in event loop.
2023-05-10 20:16:07 INFO     MAIN: Eventloop started.
2023-05-10 20:16:07 INFO     MAIN: Server started.
INFO:MAIN:Server started.
2023-05-10 20:16:07 INFO     MAIN: Proactive policy fetching is disabled.
INFO:MAIN:Proactive policy fetching is disabled.
```

## Expected behaviour

The Postfix logs should follow after the above lines. 

```
WARNING:root:Disabling hardened-malloc on this CPU
INFO:MAIN:MTA-STS daemon starting...
2023-05-10 20:16:07 INFO     MAIN: MTA-STS daemon starting...
INFO:MAIN:Starting eventloop...
INFO:MAIN:uvloop is not available. Falling back to built-in event loop.
INFO:MAIN:Eventloop started.
2023-05-10 20:16:07 INFO     MAIN: Starting eventloop...
2023-05-10 20:16:07 INFO     MAIN: uvloop is not available. Falling back to built-in event loop.
2023-05-10 20:16:07 INFO     MAIN: Eventloop started.
2023-05-10 20:16:07 INFO     MAIN: Server started.
INFO:MAIN:Server started.
2023-05-10 20:16:07 INFO     MAIN: Proactive policy fetching is disabled.
INFO:MAIN:Proactive policy fetching is disabled.
May 10 20:16:08 mail postfix/postfix-script[347]: starting the Postfix mail system
[...]

PR #2839, opened 2023-05-30
Fix postfix logging to file

Lex999:
## What type of PR?

Bug-fix

## What does this PR do?

Postfix standard/error output is polled by the startup script and forwarded to the standard/error output of the script. Consequently, this output is passed through the LogFilter class which allows logging to a file. 

### Related issue(s)

Solves issue #2819 .

## Prerequisites

### DIFF
--- a/core/postfix/start.py
+++ b/core/postfix/start.py
@@ -6,6 +6,8 @@ import shutil
 import multiprocessing
 import sys
 import re
+import subprocess
+import threading
 
 from podop import run_server
 from socrate import system, conf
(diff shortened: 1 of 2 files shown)


nextgens:
I'd rather drop ``POSTFIX_LOG_FILE`` altogether and document how to to do the same thing with journald

Lex999:
That would be fine for me as well. I'd like to note, that the fiters of the LogFilter class are not applied to Postfix output without this fix. Maybe this is no problem. I don't know what the filters are good for.

redispade:
Is there any progress with this?

PR #2844, opened 2023-06-02
Ensure we log which account is invalid

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Ensure we log which account is invalid

## Prerequisites

### DIFF
--- a/core/admin/mailu/internal/nginx.py
+++ b/core/admin/mailu/internal/nginx.py
@@ -28,9 +28,9 @@ STATUSES = {
 
 WEBMAIL_PORTS = ['14190', '10143', '10025']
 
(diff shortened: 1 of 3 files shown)


ghostwheel42 (commented):
Lgtm. But we also have a traceback when the username is not a valid email address. I get errors like this when people try to @: 
```
Out: 250 CHUNKING
In:  MAIL FROM: <fjfafsfp@@outlook.com>
Out: 250 2.1.0 Ok
In:  RCPT TO: 
```
(There's also a traceback somewhere in the logs from model.py) 
Maybe a fix for this belongs here, too?

nextgens:
bors try

@ghostwheel42 that doesn't fix it completely (the email won't be rejected early) but at least it makes the stacktrace go away

Issue #2846, opened 2023-06-03
Suggestion: allow 'send-only' users


OdyX:
## Version

- Version: `2.0`

## Description

In order to avoid copying the same password in multiple places, I've resorted to creating machine-specific users, to send emails from servers or laptops: let's say `user@example.com` is my main email, setup correctly in Mailu. I have setup a `machines.example.com` subdomain, with an `ironfist@machines.example.com`; I can setup a no-copy redirect for that email to `user@example.com` and allow `ironfist@machines.example.com` with the allow-spoofing setup, so that it can send emails `From: user@example.com` in my name. The usual case I have is my laptop's exim config.

So. It would be nice to avoid allowing that `ironfist@machines.example.com` user to be able to receive email _at all_. (One easy workaround is not setting up an MX record, but depending on the A/AAAA record, it could still happen).

Of course, another recourse would be to allow multiple passwords for a single user; such as device-specific passwords. GMail has some sort of setup for that; basically you can have (and generate) multiple independent addresses, limited to doing specific things: in the above scenario, if I could add a laptop-specific password for my `user@example.com` account, that would also work, of course.

nextgens:
We already have app-specific passwords, we call them tokens and you can configure them on ``/admin/token/list``

There is no way to prevent them from accessing and sending emails but by default they are not allowed to login to the webmail.

OdyX:
Oh, indeed. That solves the "device-specific password" part. "send-only users" would still be useful (so not closing).

nextgens:
I am fairly sure that this can be achieved with the existing code and a sieve script.

That said, having the ability to limit some tokens to send-only/receive-only may be worth it

Diman0:
You can simply disable pop3 & imap for the user access if the user may only send email. It is documented here:
https://mailu.io/2.0/webadministration.html#webadministration-add-user

OdyX:
@Diman0 sure, but an email sent to that email would still be either accepted or forwarded. I'd rather have them rejected.

nextgens:
Sounds like you haven't heard of rfc5429 :)

I still see some value in having permissions "per token" rather than "per user". in OAUTH2 parlance, that'd be "scope".

Diman0:
This can already be achieved via sieve script or rspamd multimap. But I can imagine some people would want to have a checkbox per user that enables/disables the ability to receive email for that account. No-reply email accounts are the primary example I can think off.

nextgens:
I have implemnted the token part in #2852

PR #2852, opened 2023-06-08
Improve tokens

nextgens:
## What type of PR?

enhancement

## What does this PR do?

- Allow multiple IP addresses/networks to be set for tokens.
- Allow for token rights to be curtailed (full, receive-only, send-only)

### Related issue(s)
- #2846

## Prerequisites


nextgens:
Let's see if the tests like the migration.

bors try

nextgens:
bors try

nextgens:
There is no consensus on what rights we want on the tokens... so I've submitted #2866 with the non-contentious bits and the rest is deferred to another PR.

PR #2855, opened 2023-06-12
Update reverse.rst to include TLS_FLAVOR=mail for nginx

TheStachelfisch:
## What type of PR?

documentation

## What does this PR do?
Changes the documentation about using external proxies.
If Nginx is used to upgrade HTTP to HTTPS connections and the TLS_FLAVOR isn't set to `mail`, then you will be indefinitely redirected. This PR adds a warning in the documentation about that.

This option is already mentioned for Traefik in a later section, but not for Nginx.

### Related issue(s)
#1021

## Prerequisites

### DIFF
--- a/docs/reverse.rst
+++ b/docs/reverse.rst
@@ -100,7 +100,7 @@ Because the admin interface is served as ``/admin``, the RESTful API as ``/api``
     }
   }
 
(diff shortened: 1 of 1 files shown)


ghostwheel42 (approved):
Thanks. LGTM.

ghostwheel42:
This PR is obsolete now, as there is only a traefik setup in the docs now.

PR #2866, opened 2023-06-23
Improve tokens (add ipranges)

nextgens:
## What type of PR?

enhancement

## What does this PR do?

Allow multiple IP addresses/networks to be set for tokens.

## Prerequisites


Diman0 (approved):
I tested that it indeed works when using as notation an IP address, or CIDR (xx.yy.zz.vv/nn). Using a comma separated list of IP addresses also works fine.

LGTM

Issue #2861, opened 2023-06-20
Postfix times out when I connect to it


dzervas:
### Environment

- [x ] docker compose
- [ ] kubernetes
- [ ] docker swarm

### Version

- Version: `2.0.14`

## Description
The problem is that I can't receive e-mails.

## Replication Steps
So I've traced it:
- tried to connect to port 25 over the internet with telnet - works till `rcpt to` command (and after ~10-20s it responds with `451 4.3.2 Internal server error`
- same when I connect from host
- same when I connect within the container (port 25)

## Observed behaviour
E-mails do not get delivered
Logs from the smtpd container:
```

smtp_1       | Jun 20 23:40:26 e-restaurants postfix/smtpd[347]: connect from mailu_front_1.mailu_default[192.168.203.3]
smtp_1       | Jun 20 23:40:26 e-restaurants postfix/smtpd[347]: lost connection after CONNECT from mailu_front_1.mailu_default[192.168.203.3]
smtp_1       | Jun 20 23:40:26 e-restaurants postfix/smtpd[347]: disconnect from mailu_front_1.mailu_default[192.168.203.3] commands=0/0
```

## Expected behaviour
Not timeout and receive the email.

---

My domain is e-restaurants.gr and a valid inbox is `thisisabeautifultest`
Maybe something's wrong with my DNS records?

EDIT: `mailu.env` https://pastebin.com/UBxwkMqJ and `docker-compose.yml` https://pastebin.com/BhDM1D47
EDIT2: I also found these logs in the `front` container:

```

front_1      | 2023/06/20 23:48:06 [info] 24#24: *213 client <client ip>:53369 connected to 0.0.0.0:25
front_1      | 2023/06/20 23:48:08 [info] 23#23: *205 upstream timed out (110: Operation timed out) while connecting to upstream, client: 87.253.233.105 using starttls, server: 0.0.0.0:25
front_1      | 2023/06/20 23:48:08 [crit] 23#23: *205 SSL_shutdown() failed (SSL: error:0A000123:SSL routines::application data after close notify) while connecting to upstream, client: <client ip> using starttls, server: 0.0.0.0:25
```

dzervas:
I also see:
```
smtp_1       | DEBUG:root:Received bytearray(b'domain e-restaurants.gr')
smtp_1       | DEBUG:root:Request domain/e-restaurants.gr
smtp_1       | DEBUG:root:Table get e-restaurants.gr
[...]

PR #2869, opened 2023-06-24
increase the number of postfix workers

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

increase the number of postfix workers. It was reduced when we were using postscreen ... but this specific part hasn't been reverted properly.

The practical effect is that we should see less ``451 4.3.2 Internal server error``.

### Related issue(s)
- #2861

## Prerequisites

### DIFF
--- a/core/postfix/conf/master.cf
+++ b/core/postfix/conf/master.cf
@@ -2,10 +2,10 @@
 #               (yes)   (yes)   (yes)   (never) (100)
 
 # Exposed SMTP service
-smtp      inet  n       -       n       -       1       smtpd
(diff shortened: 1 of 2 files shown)


ghostwheel42 (approved):
Lgtm

floor12:
I lost 3 days to fix this issue on big number of users... I had `204 upstream timed out (110: Operation timed out` in nginx  logs and queue of clients. `docker-compose pull && docker-compose down && docker-compose up -d ` fixed this issue.

Thank you guys a lot.

PR #2901, opened 2023-08-10
Force pw change

### LINKED ISSUE #2877: Force the user to change their password at next login
### Environment

- [x] docker compose
- [ ] kubernetes
- [ ] docker `swarm`

### Version

- Version: `2.0`

## Description

Instead of directly changing the password by the admin, the admin can create a one-time password reset link for users who have forgotten their passwords, with the condition that the link is connected to the user's email.
## Expected behaviour

First, admin creates a password reset link by accessing the user's email edit menu, where the admin typically resets the password by directly entering a new one. Then, the admin clicks on the option to reset the selected user's email, and the system generates the one-time use link for the user. Subsequently, the admin provides the link to the user, and when the user opens the link, a page appears for them to change their password.

nextgens:
## What type of PR?

Feature

## What does this PR do?

Allow administrators to force a user to change his password. Prune web-sessions on password change.

### Related issue(s)
- closes #2877 

## Prerequisites

### DIFF
--- a/core/admin/mailu/api/v1/user.py
+++ b/core/admin/mailu/api/v1/user.py
@@ -209,7 +209,7 @@ class User(Resource):
         if 'enabled' in data:
             user_found.enabled = data['enabled']
         if 'change_pw_next_login' in data:
-            user_new.change_pw_next_login = data['change_pw_next_login']
+            user_found.change_pw_next_login = data['change_pw_next_login']
         if 'enable_imap' in data:
             user_found.enable_imap = data['enable_imap']
         if 'enable_pop' in data:
--- a/core/admin/mailu/api/v1/user.py
+++ b/core/admin/mailu/api/v1/user.py
@@ -63,6 +63,7 @@ user_fields_put = api.model('UserUpdate', {
     'quota_bytes': fields.Integer(description='The maximum quota for the user’s email box in bytes', example='1000000000'),
     'global_admin': fields.Boolean(description='Make the user a global administrator'),
     'enabled': fields.Boolean(description='Enable the user. When an user is disabled, the user is unable to login to the Admin GUI or webmail or access his email via IMAP/POP3 or send mail'),
+    'change_pw_next_login': fields.Boolean(description='Force the user to change their password at next login'),
     'enable_imap': fields.Boolean(description='Allow email retrieval via IMAP'),
     'enable_pop': fields.Boolean(description='Allow email retrieval via POP3'),
     'allow_spoofing': fields.Boolean(description='Allow the user to spoof the sender (send email as anyone)'),
@@ -119,6 +120,8 @@ class Users(Resource):
             user_new.global_admin = data['global_admin']
         if 'enabled' in data:
             user_new.enabled = data['enabled']
+        if 'change_pw_next_login' in data:
+            user_new.change_pw_next_login = data['change_pw_next_login']
         if 'enable_imap' in data:
             user_new.enable_imap = data['enable_imap']
         if 'enable_pop' in data:
--- a/core/admin/mailu/api/v1/user.py
+++ b/core/admin/mailu/api/v1/user.py
@@ -194,7 +194,7 @@ class User(Resource):
             return {'code': 404, 'message': f'User {email} cannot be found'}, 404
 
         if 'raw_password' in data:
-            user_found.set_password(data['raw_password'])
+            user_found.set_password(data['raw_password'], '')
         if 'comment' in data:
             user_found.comment = data['comment']
         if 'quota_bytes' in data:
--- a/core/admin/mailu/api/v1/user.py
+++ b/core/admin/mailu/api/v1/user.py
@@ -16,6 +16,7 @@ user_fields_get = api.model('UserGet', {
     'quota_bytes': fields.Integer(description='The maximum quota for the user’s email box in bytes', example='1000000000'),
     'global_admin': fields.Boolean(description='Make the user a global administrator'),
     'enabled': fields.Boolean(description='Enable the user. When an user is disabled, the user is unable to login to the Admin GUI or webmail or access his email via IMAP/POP3 or send mail'),
+    'change_pw_next_login': fields.Boolean(description='Force the user to change their password at next login'),
     'enable_imap': fields.Boolean(description='Allow email retrieval via IMAP'),
     'enable_pop': fields.Boolean(description='Allow email retrieval via POP3'),
     'allow_spoofing': fields.Boolean(description='Allow the user to spoof the sender (send email as anyone)'),
@@ -40,6 +41,7 @@ user_fields_post = api.model('UserCreate', {
     'quota_bytes': fields.Integer(description='The maximum quota for the user’s email box in bytes', example='1000000000'),
     'global_admin': fields.Boolean(description='Make the user a global administrator'),
     'enabled': fields.Boolean(description='Enable the user. When an user is disabled, the user is unable to login to the Admin GUI or webmail or access his email via IMAP/POP3 or send mail'),
+    'change_pw_next_login': fields.Boolean(description='Force the user to change their password at next login'),
     'enable_imap': fields.Boolean(description='Allow email retrieval via IMAP'),
     'enable_pop': fields.Boolean(description='Allow email retrieval via POP3'),
     'allow_spoofing': fields.Boolean(description='Allow the user to spoof the sender (send email as anyone)'),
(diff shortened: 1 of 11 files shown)


core/admin/migrations/versions/0ba45693748d_.py
```diff
@@ -0,0 +1,22 @@
...
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.add_column('user', sa.Column('change_pw_next_login', sa.Boolean(), nullable=False, server_default=sa.sql.expression.false()))
+
+
+def downgrade():
+    pass
```
ghostwheel42, line 22: Why not remove the column here?

core/admin/mailu/manage.py
```diff
@@ -121,7 +121,7 @@ def password(localpart, domain_name, password):
     email = f'{localpart}@{domain_name}'
     user  = models.User.query.get(email)
     if user:
-        user.set_password(password)
+        user.set_password(password, '')
```
ghostwheel42, line 124: This change is not necessary

core/admin/mailu/api/v1/user.py
```diff
@@ -194,7 +194,7 @@ def patch(self, email):
             return {'code': 404, 'message': f'User {email} cannot be found'}, 404
 
         if 'raw_password' in data:
-            user_found.set_password(data['raw_password'])
+            user_found.set_password(data['raw_password'], '')
```
ghostwheel42, line 197: This change is not necessary

core/admin/mailu/models.py
```diff
@@ -623,11 +624,13 @@ def check_password(self, password):
             self._credential_cache[self.get_id()] = (self.password.split('$')[3], passlib.hash.pbkdf2_sha256.using(rounds=1).hash(password))
         return result
 
-    def set_password(self, password, raw=False):
+    def set_password(self, password, raw=False, session=None):
         """ Set password for user
             @password: plain text password to encrypt (or, if raw is True: the hash itself)
         """
         self.password = password if raw else User.get_password_context().hash(password)
+        if session:
```
ghostwheel42, line 632: This is okay, but adding `is not None` would be more precise.

core/admin/mailu/sso/views/base.py
```diff
@@ -66,6 +69,41 @@ def login():
...
+    client_port = flask.request.headers.get('X-Real-Port', None)
+    form = forms.PWChangeForm()
+
+    if form.validate_on_submit():
+        if msg := utils.isBadOrPwned(form):
+            flask.flash(msg, "error")
+            return flask.redirect(flask.url_for('sso.pw_change'))
+        if form.oldpw.data == form.pw2.data:
+            # TODO: fuzzy match?
+            flask.flash("The new password can't be the same as the old password", "error")
```
ghostwheel42, line 85: I think we'll need some translations here...

ghostwheel42 (commented):
Looks good (code review), but I didn't have time to test it yet.

core/admin/migrations/versions/0ba45693748d_.py
nextgens: Other migrations in Mailu don't. I guess the rationale is that we don't want to loose data.

Do you want me to implement it here?

core/admin/mailu/manage.py
nextgens: I would like to the sessions for that user to be destroyed when the admin resets the password. Maybe I should change the interface to make it clearer

nextgens: clarified in 786da5e0

core/admin/mailu/api/v1/user.py
nextgens: clarified in 786da5e0

core/admin/mailu/models.py
nextgens: done in 786da5e0

core/admin/mailu/sso/views/base.py
nextgens: Done in 193b3d52

core/admin/migrations/versions/0ba45693748d_.py
nextgens: Done in a5f3c7ea

nextgens:
bors try

nextgens:
bors try

nextgens:
bors try

Diman0 (dismissed):
Via the admin it works. When it is ticked (change password on next login), then I need to change my password on next login.

Settings this via the RESTful API does not work.
```
curl -X 'PATCH' \
  'https://test.berichthier.nl/api/v1/user/test%40example.com' \
  -H 'accept: application/json' \
  -H 'Authorization: secret' \
  -H 'Content-Type: application/json' \
  -d '{

  "change_pw_next_login": true

}'
```
Returns that it updated the user, but I can login without needing to change my password.

When I look up the user via the RESTful API, the field is also missing in the results. 


I hope to have time tomorrow to look into why the RESTful API part does not work.

nextgens:
> Settings this via the RESTful API does not work.

I think I have fixed it in [a338e9c](https://github.com/Mailu/Mailu/pull/2901/commits/a338e9c75e0568638be90fb388c75dbd7a1a9a5a)

Diman0:
bors try

core/admin/mailu/api/v1/user.py
```diff
@@ -203,6 +208,8 @@ def patch(self, email):
             user_found.global_admin = data['global_admin']
         if 'enabled' in data:
             user_found.enabled = data['enabled']
+        if 'change_pw_next_login' in data:
+            user_new.change_pw_next_login = data['change_pw_next_login']
```
Diman0, line 212: The wrong variable name is used. user_new should be user_found.

```suggestion
            user_found.change_pw_next_login = data['change_pw_next_login']
```

Diman0 (dismissed):
I tested the API again:
* Creating user via API works with force password change.
* List user also shows the property now
*  Updating user does not work yet. See in-line comment with proposed fix. You mixed up two variable names.

core/admin/mailu/api/v1/user.py
```diff
@@ -194,7 +199,7 @@ def patch(self, email):
             return {'code': 404, 'message': f'User {email} cannot be found'}, 404
 
         if 'raw_password' in data:
-            user_found.set_password(data['raw_password'])
+            user_found.set_password(data['raw_password'], keep_only_session='')
```
ghostwheel42, line 202: I don't think there should be an empty string passed to set_password.

core/admin/mailu/manage.py
```diff
@@ -121,7 +122,8 @@ def password(localpart, domain_name, password):
     email = f'{localpart}@{domain_name}'
     user  = models.User.query.get(email)
     if user:
-        user.set_password(password)
+        user.set_password(password, keep_only_session='')
```
ghostwheel42, line 125: Same here

ghostwheel42 (changes requested):

core/admin/mailu/api/v1/user.py
nextgens: As in we don't want to nuke the open sessions? Or it should be something that is more explicit like [] ?

It's fairly common for existing sessions to be destroyed upon password change, it ensures that if an attacker has one it will be revoked.

Diman0: I concur it is fairly common to destroy all sessions when the user's password is changed. Isn't this even a best practice?

ghostwheel42: Yes, sessions should be cleared. It was just confusing to me how this is done.
Having an empty string as argument didn't occur to me as "clear all sessions".

What about
a) having the method "set_password" only do what it's name says and clearing sessions explicitely where we want to do it.
b) having the method "set_password" _always_ clear all sessions unless the argument keep_sessions is set to True (which will keep all sessions) or a set() which will keep the sessions in the set.

I think this will be more clear then.

nextgens: Check [0e4c29c](https://github.com/Mailu/Mailu/pull/2901/commits/0e4c29c26957f65e4ba776349a05d07e01fa200d) out

nextgens:
bors try

core/admin/mailu/api/v1/user.py
ghostwheel42: 👍

ghostwheel42 (approved):
LGTM

Issue #1502, opened 2020-05-12
Proposal: Test for correct IP configuration (avoid open relay)


micw:
It seems that users accidentally create open relays with Mailu, especially in setups where load balancers or proxies are involved. Reason in most cases is that the requests seems to come from $SUBNET because the originating IP get lost at load balancer level.
What do you think about some automatic check that prevents this and enables email only after the check has passed? The check could be repeated at an interval, blocking any email activity if it fails.

Such a check would require external "help", e.g. a kind of service that is called by mailu and connects from a known IP to any of the email services (like SMTP). On nginx we wolud see if the originating IP or any other IP occurs here. Instead of SMTP, we could also use a deticated port.

The check could of course be configurable, enabled by default but easy to switch off via config variable.

A similar check could also be created to verify that the IP address of outgoing mails matches the configured IP for a domain.

djerfy:
I have the same situation (Kubernetes with Traefik Ingress Controller). For the moment I build my image without `{{ SUBNET }}` in `mynetworks` (https://github.com/Mailu/Mailu/blob/master/core/postfix/conf/main.cf#L20).

Nebukadneza:
Hi There,

The `Mailu`-Project is currently in a bit of a bind! We are short on man-power, and we need to judge if it is possible for us to put in some work on this issue.

To help with that, we are currently trying to find out which issues are actively keeping users from using `Mailu`, which issues have someone who want to work on them — and which issues may be less important. These a less important ones could be discarded for the time being, until the project is in a more stable and regular state once again.

In order for us to better assess this, it would be helpful if you could put a **reaction on this post** (use the :smiley: icon to the top-right).
- 👍️ if you **need this** to be able to use Mailu. Ideally, you’d also be able to test this on your installation, and provide feedback …
- 🎉 if you find it a **nice bonus**, but no deal-breaker
- 🚀 if you want to **work on it yourself**!
We want to keep this **voting open for 2 weeks** from now, so please help out!

PR #2903, opened 2023-08-12
Make traefik work

### LINKED ISSUE #1038: Traefik guide needs more love
Hi,

first of all - thank you for this amazing project. I'm seeing lots of room for improvement, but at the same time I'm :clap: all of what you've accomplished so far.

I'm trying to use mailu as a mail-sender/receiver for a dockerized GitLab server setup so I've started following `mailus` guide on how to use traefik.

https://mailu.io/master/reverse.html#traefik-as-reverse-proxy

I found it quite hard to follow. I've just found the example for traefik.toml and the docker-compose.yml.
So the actual suggested way is to launch the dumper with the same docker-compose file. I don't feel good about storing the certificates for other traefik services in the same mount as the mail-server lives.

---
1. starting traefik (with the certificate dumper service as well)
2. mailu_frontend bind mounts the `DOMAIN.{key,crt}` from the traefik dump
3. mailu starts all services up => creates a directory for the mount point since traefik didn't request a certifciate yet and the dumper not having dumped the not yet generated certificate
4. mailu stops frontend => traefik doesn't see frontend healthy => doesn't request a certificate

I'm trying to find a solution for this problem.

---

Also I'm trying to figure out how much everything is going to change with traefik 2.0 (and tcp support).

I think with traefik 2.0 it would be nice to use traefik as TLS termination and to expose TLS for imap/pop/smtp through traefik instead of doing it within mailu.

### LINKED ISSUE #1422: Mailu behind Traefik 2.2
Hello,

i am loving the mailu system, but i struggle implementing it in my environment.
I am running a Traefik2.2 instance as reverse proxy and to handle all SSL stuff in one place.
Starting with the webmailer everything went well, sending and receiving mails worked out of the box.
Extending the usage to an imap client (Thunderbird) worked also, connecting via 993 to Traefik and Traefik sends all requests unencrypted to the nginx container.
I am struggling to get SMTP to work.
Traefik is in a different subnet than Mailu, to encapsulate different types of services.
When i add the Traefik net to the Relaynets in Mailu.env, i get an open relay, due to the fact that Traefik tunnels all traffic to the respective nginx container. Doing so is a bad idea, cause everyone could use the server for spam.

I tried to find out, why mailu does not authenticate my mail address and my user, though i am using port 465, but i simply failed.
Could you provide any hints, how to configure mailu to either authenticate everything coming from a specific host (Traefik) via SASL, or any idea, how i could change my setup to get the stuff rolling?

Is there any way to let Traefik handle all the ports and simply redirect them?

Kind regards,
LarsTi

### LINKED ISSUE #2683: Certdump: Make acme.json as a variable - enhancement
### Environment

- [x] docker-compose
- [ ] kubernetes
- [ ] docker swarm

### Version

- Version: `1.9`

## Description
I am using Traefik as reverse proxy and Traefik can handle multiple domain names for one host. For this reason i setup Traefik to store for each domain the certificates in different files, e. g. `acme-<domain>.json`
Certdumps script has a hardcoded point to acme.json in a specific folder. I mounted the volume of Traefik which contains the acme files.
Without editing the script in the running Certdump container i can't get the updated certificate that is done by Traefik

## Replication Steps
Store cert in a separate file.
traefik.yml:
```
certificatesResolvers:
  letsencrypt-<domain>:
    acme:
      email: certificate@example.com
      storage: /data/acme-<domain>.json
      tlsChallenge: {}
```
Using the cert resolver for Mailu.
Mailu docker-compose:
```
traefik.http.routers.mailu.tls.certresolver=letsencrypt-<domain>
```
Mount the acme.json folder to certdump
Mailu docker-compose:
```
certdumper:
    image: ${DOCKER_REGISTRY}/${DOCKER_ORG:-mailu}/${DOCKER_PREFIX:-}traefik-certdumper:${MAILU_VERSION:-1.9}
    container_name: mailu_certdump-1.9
    restart: always
    environment:
      - DOMAIN=mail.<fqdn>
      - TRAEFIK_VERSION=${TRAEFIK_VERSION}
    volumes:
      - /etc/localtime:/etc/localtime:ro
      # Folder, which contains the acme.json
      - "traefik_data:/traefik:ro"
      # Folder, where cert.pem and key.pem will be written
      - "certdumper-data-1.9:/output"
```
## Observed behaviour
In the Traefik folder is still an old acme.json file with expired certificates. Certdump always takes acme.json until I change it manually in `run.sh` script. Of course this works only until the container will be deployed again.
## Expected behaviour
Certdump should not take hardcoded acme.json
## Logs
Not necessary.

nextgens:
## What type of PR?

enhancement

## What does this PR do?

Document how traefik is meant to be configured: only Mailu needs certificates, we don't need certdumper.

Add a check to ensure we don't "burn" credit with letsencrypt because of a reverse proxy misconfiguration (where port 80 isn't redirected to Mailu)

### Related issue(s)
- closes #1422
- closes #1038
- #2855
- closes #2683
- #1502

## Prerequisites


nextgens:
bors try

nextgens:
bors try

core/nginx/letsencrypt.py
```diff
@@ -39,8 +45,30 @@
 # Wait for nginx to start
 time.sleep(5)
 
+class MyRequestHandler(SimpleHTTPRequestHandler):
+    def do_GET(self):
+        self.path = '/var/empty/'
+        return SimpleHTTPRequestHandler.do_GET(self)
+
+def serve_one_request():
+    with HTTPServer(("0.0.0.0", 8008), MyRequestHandler) as server:
```
ghostwheel42, line 54: Looks like port should be 8080, shouldn't it?

core/nginx/letsencrypt.py
```diff
@@ -39,8 +45,30 @@
...
+        server.handle_request()
+
 # Run certbot every day
 while True:
+    while True:
+        hostname = os.environ['HOSTNAMES'].split(' ')[0]
+        target = f'http://{hostname}/.well-known/acme-challenge/testing'
```
ghostwheel42, line 61: Shouldn't we test with a known existing file served by this conatiner?

core/nginx/letsencrypt.py
nextgens: No, that was done on purpose to mirror what certbot's standalone uses

core/nginx/letsencrypt.py
nextgens: That's cf090b42 ; 204 should be unique enough to avoid false positives.

nextgens:
bors try

ghostwheel42 (approved):

docs/reverse.rst
```diff
@@ -2,258 +2,136 @@ Using an external reverse proxy
...
-
-If your Traefik is configured to automatically request certificates from *letsencrypt*, then you’ll have a certificate
-for ``mail.your.example.com`` now. However, ``mail.your.example.com`` might only be the location where you want the Mailu web-interfaces
-to live — your mail should be sent/received from ``your.example.com``, and this is the ``DOMAIN`` in your ``.env``?
-To support that use-case, Traefik can request ``SANs`` for your domain. The configuration for this will depend on your Traefik version.
 
-Mailu must also be configured with the information what header is used by the reverse proxy for passing the remote
-client IP.  This is configured in mailu.env:
+      # the second part is important to ensure Mailu can get certificates from letsencrypt
+      - "traefik.http.routers.web.rule=Host(`mail.example.com`) || Path(`/.well-known/acme-challenge/`)"
```
witten, line 65: I know I'm a little late to the party, but shouldn't this be `PathPrefix` instead of `Path` so that it actually matches a challenge? And maybe `||` should be `&&` here as well?

nextgens: Ah yes, PathPrefix it is; I will send a PR tomorrow unless you beat me to it.

The OR is to ensure we catch all hostnames ( client auto configuration, mta-sts, ...)

witten: It's all yours. I haven't successfully gotten Traefik working with this method yet, so I wouldn't have a way to test any changes!

nextgens: You need to run master for it to work

witten: Ah okay, that explains a lot. I thought I saw some of this PR's changes make its way into 2.0.36 containers, but maybe it's not everything.

PR #2923, opened 2023-08-25
Delete the dovecot PID file to avoid race conditions

### LINKED ISSUE #2917: IMAP container error on reboot
### Environment

- [x] docker compose
- [ ] kubernetes
- [ ] docker swarm

### Version

- Version: `2.0`

## Description

Occasionally, when performing a server reboot for routine maintenance, Mailu does not start correctly. I have narrowed it down to the imap container. The only error I see in the imap container logs is: Fatal: Dovecot is already running with PID 9 (read from /run/dovecot/master.pid). This is on my production machine, so I cannot keep the service down too long to troubleshoot. Before, I would restart the entire stack, but since I looked into it more on this last occurrence, I simply restarted the IMAP container and everything works fine.  I am apprehensive about even posting this bug, since I do not have more information to offer. I can spin up a test environment and keep rebooting until the error occurs if necessary. I was running 1.9 for about a year, and it seems like this only started once I moved to 2.0 back in April, but I'm not sure.

Thanks for all of the hard work you all do.

## Replication Steps

It is very difficult to reproduce. It happens maybe 10% of the time.  

## Logs

```
Fatal: Dovecot is already running with PID 9 (read from /run/dovecot/master.pid)
```

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Apparently sometimes the podop subprocess can be re-assigned the same PID when the container is restarted, causing havoc.

This may explain some of the "I can't access the webmail" tickets.

I wonder if the same can happen in other containers too (front for instance).

### Related issue(s)
- closes #2917 

## Prerequisites

### DIFF
--- a/core/dovecot/start.py
+++ b/core/dovecot/start.py
@@ -33,6 +33,7 @@ for script_file in glob.glob("/conf/*.script"):
 # Run Podop, then postfix
 os.system("chown mail:mail /mail")
 os.system("chown -R mail:mail /var/lib/dovecot /conf")
+os.system("rm -rf /run/dovecot/master.pid")
 
 multiprocessing.Process(target=start_podop).start()
 os.system("dovecot -c /etc/dovecot/dovecot.conf -F")
--- a/towncrier/newsfragments/2917.bugfix
+++ b/towncrier/newsfragments/2917.bugfix
@@ -1,0 +1,1 @@
+Ensure that we delete any pre-exising PID file before starting dovecot


nextgens:
bors try

nextgens:
bors try

core/base/libs/socrate/socrate/system.py
```diff
@@ -97,6 +97,8 @@ def set_env(required_secrets=[], log_filters=[], log_file=None):
     for secret in required_secrets:
         os.environ[f'{secret}_KEY'] = hmac.new(bytearray(secret_key, 'utf-8'), bytearray(secret, 'utf-8'), 'sha256').hexdigest()
 
+    os.system('find / -xdev -type f -name \*.pid -print -delete')
```
ghostwheel42, line 100: Shouldn't we know where the pid files are located other than /?

nextgens: The idea was to make it future proof. By adding -xdev we ensure we only crawl "the image" so it's reasonably fast... and only done once at startup anyway.

nextgens: Oh and another thing is that you can't list "non existing" directories as find arguments, it will error out

ghostwheel42: Yeah. Sure. But I really like to stick to /run here without xdev.
This is where pid files have to reside according to LFS.
Deleting files outside doesn't feel right.

nextgens: As seen on the PR, postfix is at /queue/pid/master.pid ... so do you want me to restrict this to /run and revert the change on postfix?

ghostwheel42: Ah that's bad. But /queue is not scanned when using `-xdev` so it won't be found either.
```
b9cf64850518:/app# ls /queue/pid/
inet.10025     master.pid     unix.cleanup   unix.flush     unix.outclean  unix.smtp
inet.smtp      unix.bounce    unix.defer     unix.lmtp      unix.retry     unix.trace
b9cf64850518:/app# find / -xdev -name '*.pid' 2>/dev/null
b9cf64850518:/app# find / -name '*.pid' 2>/dev/null
/queue/pid/master.pid
```

nextgens: Good point, that's bbed508e

nextgens:
bors try

ghostwheel42 (approved):
LGTM

Issue #2824, opened 2023-05-16
Add quota_bytes_used on UserGet object API


0pc0deFR:
### Environment

- [x] docker compose
- [ ] kubernetes
- [ ] docker swarm

### Version

- Version: `2.0`

## Description
Hello,

Can you add quota_bytes_used on UserGet object API ?

PR #2951, opened 2023-09-20
Quota bytes used

ctrl-i:
## What type of PR?

Enhancement

## What does this PR do?

Allows the mailbox size of the users to be queried by the api.

Previously quota_bytes_used key was excluded from the api when you queried for a list of users.  This is useful to allow a user to monitor the volume of emails being used.

### Related issue(s)
- Auto close an issue: [closes 2824](https://github.com/Mailu/Mailu/issues/2824)

## Prerequisites

### DIFF
--- a/core/admin/mailu/schemas.py
+++ b/core/admin/mailu/schemas.py
@@ -1176,7 +1176,7 @@ class UserSchema(BaseSchema):
         model = models.User
         load_instance = True
         include_relationships = True
-        exclude = ['_email', 'domain', 'localpart', 'domain_name']
+        exclude = ['_email', 'domain', 'localpart', 'domain_name', 'quota_bytes_used']
 
         primary_keys = ['email']
         exclude_by_value = {
--- a/core/admin/mailu/schemas.py
+++ b/core/admin/mailu/schemas.py
@@ -1176,7 +1176,7 @@ class UserSchema(BaseSchema):
         model = models.User
         load_instance = True
         include_relationships = True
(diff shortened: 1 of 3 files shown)


core/admin/mailu/schemas.py
```diff
@@ -1176,7 +1176,7 @@ class Meta:
         model = models.User
         load_instance = True
         include_relationships = True
-        exclude = ['_email', 'domain', 'localpart', 'domain_name', 'quota_bytes_used']
```
Diman0, line 1179: This breaks the command line [mailu config-export](https://mailu.io/2.0/cli.html#config-export)
This cli command now outputs this value (quota_bytes_used) and that is not allowed. 
The output from mailu config-export must be a valid yml file that can be imported again.

Diman0 (changes requested):

See [the contribution documentation](https://mailu.io/2.0/contributors/workflow.html#changelog). A changelog entry must be added.

See in-line comment. These changes break the mailu cli config-export. That needs to be resolved before this is mergeable.

core/admin/mailu/schemas.py
ghostwheel42: The exclude line must not be removed.
The config export is designed to only export configuration data, not runtime data.

Diman0: Why was this change added at all?
If I only update `user_fields_get = api.model('UserGet',` with the quota_bytes_used it already works for me.

ctrl-i:
If quota_bytes_used is removed from the exclude line it is no longer reported in API.

Issue #2949, opened 2023-09-19
Store DKMI keys in database instead of filesystem


Deltachaos:
Related to #2017 it would be much easier for management of DKMI keys if the data is stored in database instead of the filesystem. This would remove the requirement for shared storage in the Admin.

Issue #2687, opened 2023-03-08
Cannot generate DKIM keys for long domain names


GodSaveTheDoge:
### Environment

- [x] docker-compose
- [ ] kubernetes
- [ ] docker swarm

### Version

- Version: `1.9` (just updated with `docker-compose pull`)

## Description
Mailu fails to generate DKIM keys for domains that are too long and, instead, fails with a 500 internal server error

## Replication Steps
- Go to the admin panel > mail domains
- Add `llllllllllllllllllllllllllllllllllllllllllllllllllllllll.lllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllll.lllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllll.lllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllll.com`
- Try to generate new dkim keys

## Observed behaviour
It fails to generate them

## Expected behaviour
It should either generate them or tell the user that the domain name is too long (and maybe acknowledge it when the domain is added too)

## Logs
```
admin_1     | Traceback (most recent call last):
admin_1     |   File "/usr/lib/python3.9/site-packages/flask/app.py", line 2073, in wsgi_app
admin_1     |     response = self.full_dispatch_request()
admin_1     |   File "/usr/lib/python3.9/site-packages/flask/app.py", line 1518, in full_dispatch_request
admin_1     |     rv = self.handle_user_exception(e)
admin_1     |   File "/usr/lib/python3.9/site-packages/flask/app.py", line 1516, in full_dispatch_request
admin_1     |     rv = self.dispatch_request()
admin_1     |   File "/usr/lib/python3.9/site-packages/flask/app.py", line 1502, in dispatch_request
admin_1     |     return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
admin_1     |   File "/usr/lib/python3.9/site-packages/flask_login/utils.py", line 272, in decorated_view
admin_1     |     return func(*args, **kwargs)
admin_1     |   File "/app/mailu/ui/access.py", line 28, in wrapper
admin_1     |     return callback(function, args, kwargs, dargs, dkwargs)
admin_1     |   File "/app/mailu/ui/access.py", line 19, in callback
admin_1     |     return function(*args, **kwargs)
admin_1     |   File "/app/mailu/ui/access.py", line 105, in wrapper
admin_1     |     return function(*args, **kwargs)
[...]

Issue #2616, opened 2023-01-05
How do I use external redis with a password?


fruworg:
### Environment

- [x] docker-compose

### Version

- Version: `master`

## Description

Hi. I couldn't find a line in the documentation for specifying a password. In nextcloud, it is implemented like this:
REDIS_HOST=\<redis-ip\>
REDIS_HOST_PASSWORD=\<redis-pass\>

nextgens:
Currently this is not an option. Mailu uses several redis databases on the same instance, if we were to support external instances we would have to provide a configuration mechanism too.

Overall a lot of complexity for little gain. If you do want this as a feature request, please spell out the usecase.

Issue #2053, opened 2021-11-15
Make trusted ARC signers configurable


nextgens:
ARC support has been implemented in #495 but no provisions have been made to mark signers as trusted.

Rspamd has a configuration map ``whitelisted_signers_map`` (see https://github.com/rspamd/rspamd/commit/1fa88bcd56301f2f41319b987ac89909c87b6d0b) that we could either:
- pre-populate with "smart defaults"  (add the major players)
- document (at least how to set an override)
- make configurable statically (a configuration option)
- make configurable dynamically. It's a map: it could be fetched from admin and there might be an argument to be made that every user could contribute to it). We could also "learn" from existing emails if we wanted to.

Without it emails receive ``ARC_ALLOW`` that applies -1.0; ``ARC_ALLOW_TRUSTED`` is an additional -2.0.
For reference, ``R_DKIM_REJECT`` is -1.0, ``DMARC_POLICY_QUARANTINE`` is -1.5 and ``BLACKLIST_DMARC`` is -6

nextgens:
To be clear: ARC is useful when an email is "forwarded". The typical usecase is an email sent by valid_reject_DMARC@example.com received at example@gmail.com that's configured to be forwarded to example@test.mailu.io

If we don't change anything odds are Mailu will mark it spam and may even reject the said email.

PR #2952, opened 2023-09-23
Move DKIM keys to database

Deltachaos:
This PR solves issue #2949.

Moving the DKIM keys to database removes the requirement for shared storage in the admin. 

Currently a migration path from files to database is missing. My idea is it to add this to the migration file, but i am not to deeply into the python framework here to know if this is a good idea or not. If you have some tips i happily add this feature so that this PR can get merged.

### DIFF
--- a/core/admin/mailu/api/v1/domains.py
+++ b/core/admin/mailu/api/v1/domains.py
@@ -223,7 +223,8 @@ class Domain(Resource):
         if not domain_found:
             return { 'code': 404, 'message': f'Domain {domain} does not exist'}, 404
         domain_found.generate_dkim_key()
-        domain_found.save_dkim_key()
+        db.session.add(domain_found)
+        db.session.commit()
         return {'code': 200, 'message': f'DKIM/DMARC keys have been generated for domain {domain}'}, 200
 
 @dom.route('/<domain>/manager')
--- a/core/admin/mailu/configuration.py
+++ b/core/admin/mailu/configuration.py
@@ -55,7 +55,6 @@ DEFAULT_CONFIG = {
     'WELCOME_SUBJECT': 'Dummy welcome topic',
     'WELCOME_BODY': 'Dummy welcome body',
     'DKIM_SELECTOR': 'dkim',
-    'DKIM_PATH': '/dkim/{domain}.{selector}.key',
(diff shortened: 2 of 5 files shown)


nextgens:
Ah; I have code for doing this on a branch somewhere...

IMHO we can't merge it as is: we want to address the other related DKIM tickets too: #2687 #2616 #2053 #2050 #1700 maybe even #1519

We need a separate table (as the same keys could be used cross-domains -like for ARC-), we probably want additional fields (keytype, usage, generation timestamp, status, ...)

Deltachaos:
I see. But wouldn’t having the keys i the database even without changing the current behaviour that much, make a migration to the changes you mentioned easier?

nextgens:
> I see. But wouldn’t having the keys i the database even without changing the current behaviour that much, make a migration to the changes you mentioned easier?

I don't mind incremental improvements... but the database schema should be as close to the end-state as possible. The last thing we want is to have to lost information in the process and make migration more complex as a result.

At the very least, we want a separate table with a column for the selector, a timestamp it was generated/imported on, and a way to link it to the domain it was enabled for (keeping in mind that there may be multiple DKIM keys)

nextgens:
Here is what I had on my branch if you need inspiration:

[0001-V1.txt](https://github.com/Mailu/Mailu/files/12746951/0001-V1.txt)

```
class Key(Base):
    """ A Key used for DKIM or ARC """
    __tablename__ = 'key'

    name = db.Column(db.String, primary_key=True, nullable=False)
    key_type = db.Column(db.Enum(KEY_TYPE), nullable=False)
    key_size = db.Column(db.Integer, nullable=False)
    private = db.Column(db.String, unique=True, nullable=False)
    public = db.Column(db.String, unique=True, nullable=False)
    generated = db.Column(db.DateTime, nullable=True, default=datetime.utcnow())
    active = db.Column(db.DateTime, nullable=True)
    retired = db.Column(db.DateTime, nullable=True)
    use_for_dkim = db.Column(db.Boolean, nullable=True, default=False)
    use_for_arc = db.Column(db.Boolean, nullable=True, default=False)
    """ a misnamer: Domain or Alias that makes use of it """
    domains = db.relationship('Domain', secondary=key_mapping,
            lazy='dynamic', back_populates="keys")
```

PR #2961, opened 2023-10-06
Hardened malloc was not disabled for oletools when an CPU with missing flags is used

### LINKED ISSUE #2959: 2.0.25 gives Restarting (132) issue in docker
### Environment

- [x ] docker compose
- [ ] kubernetes
- [ ] docker swarm

### Version

- Version: `2.0.25`

## Description

pulled the latest build (2.0.25) and the mailu docker containers would no longer start. They would go to a status of "Restarting (132)" and continuously just restart

## Expected behaviour

Docker containers to start up and run

Diman0:
## What type of PR?
bug fix

## What does this PR do?
Updates oletools to also disable hardened malloc when used CPU misses flags

### Related issue(s)
- closes #2959 

## Prerequisites

### DIFF
--- a/core/base/libs/socrate/socrate/system.py
+++ b/core/base/libs/socrate/socrate/system.py
@@ -66,7 +66,7 @@ def _is_compatible_with_hardened_malloc():
         lines = f.readlines()
         for line in lines:
             # See #2764, we need vmovdqu
-            if line.startswith('flags') and ' avx ' not in line:
+            if line.startswith('flags') and ' avx ' not in line and ' avx2 ' not in line:
                 return False
             # See #2541
             if line.startswith('Features') and ' lrcpc ' not in line:
--- a/core/base/libs/socrate/socrate/system.py
+++ b/core/base/libs/socrate/socrate/system.py
@@ -66,7 +66,7 @@ def _is_compatible_with_hardened_malloc():
         lines = f.readlines()
         for line in lines:
             # See #2764, we need vmovdqu
-            if line.startswith('flags') and ' avx ' not in line and ' avx2 ' not in line:
+            if line.startswith('flags') and ' avx2 ' not in line:
                 return False
             # See #2541
             if line.startswith('Features') and ' lrcpc ' not in line:
--- a/core/base/libs/socrate/socrate/system.py
+++ b/core/base/libs/socrate/socrate/system.py
@@ -66,6 +66,7 @@ def _is_compatible_with_hardened_malloc():
         lines = f.readlines()
         for line in lines:
             # See #2764, we need vmovdqu
+            # See #2959, we need vpunpckldq
             if line.startswith('flags') and ' avx2 ' not in line:
                 return False
             # See #2541
@@ -80,7 +81,7 @@ def set_env(required_secrets=[], log_filters=[], log_file=None):
     log.basicConfig(stream=sys.stderr, level=os.environ.get("LOG_LEVEL", 'WARNING'))
 
     if 'LD_PRELOAD' in os.environ and not _is_compatible_with_hardened_malloc():
-        log.warning('Disabling hardened-malloc on this CPU')
+        log.warning('Disabling hardened-malloc on this CPU: it requires Advanced Vector Extensions.')
         del os.environ['LD_PRELOAD']
 
     """ This will set all the environment variables and retains only the secrets we need """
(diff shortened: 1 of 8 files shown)


Diman0:
I will set the WIP flag until I have replicated the issue locally and confirmed this PR resolves it.

nextgens:
[x86_64.txt](https://github.com/Mailu/Mailu/files/12829585/x86_64.txt)
[arm64.txt](https://github.com/Mailu/Mailu/files/12829586/arm64.txt)

Here are the instructions in use... so we need 'avx2' for vpunpckldq, what else?

core/base/libs/socrate/socrate/system.py
```diff
@@ -66,7 +66,7 @@ def _is_compatible_with_hardened_malloc():
         lines = f.readlines()
         for line in lines:
             # See #2764, we need vmovdqu
-            if line.startswith('flags') and ' avx ' not in line:
+            if line.startswith('flags') and ' avx ' not in line and ' avx2 ' not in line:
```
nextgens, line 69:
```suggestion
            if line.startswith('flags') and ' avx2 ' not in line:
```
You won't find a CPU that does the later but not the former.

As currently spelt out, if either is present the test will skip. This is not what we want, we want both to be present.

nextgens: We may want to explain in the warning line 83 that we are missing vectorized instructions

nextgens (changes requested):

nextgens (dismissed):
LGTM

Diman0:
Please reject the PR. It does not work yet.
I think I figured out the problem.
In the Dockerfile of the base image we enable the hardened malloc:
https://github.com/Mailu/Mailu/blob/dd1de62eb0e3e0f7a6dec1de5c57c3b64bb54d9b/core/base/Dockerfile#L82

As a result, the test in system.py to enable/disable hardened malloc cannot be even reached. Start.py is already run with the hardened malloc due to LD_PRELOAD=LD_PRELOAD="/usr/lib/libhardened_malloc.so being set. So incompatible systems will have the problem that start.py immediately crashes.

I propose to 
1. set `LD_PRELOAD=` in the base image. 
2. Adapt the hardened malloc test in system.py to enable the hardened malloc via LD_PRELOAD= if requirements are met (cpu flags/features  are present)

nextgens:
bors try

nextgens:
bors try

nextgens:
bors try

nextgens:
bors try

nextgens:
@Diman0 are we happy with this now?

We may want to double check that it works in practice: use lsof to see if all PIDs of all containers are using the library

Dennis14e:
Just my 2 cents: Wouldn't it be nicer to have a custom environment variable like "ENABLE_HARDENED_MALLOC" instead of the user setting LD_PRELOAD themselves?
Maybe LD_PRELOAD is needed for something else too at some point or an Alpine update changes the path or something similar.

Greetings & thanks for Mailu!

Diman0:
> Just my 2 cents: Wouldn't it be nicer to have a custom environment variable like "ENABLE_HARDENED_MALLOC" instead of the user setting LD_PRELOAD themselves? Maybe LD_PRELOAD is needed for something else too at some point or an Alpine update changes the path or something similar.
> 
> Greetings & thanks for Mailu!

The point is that hardened malloc can only be made available for the entrypoint process if LD_PRELOAD is already set when the container is started. 

@nextgens It looks good to me. I will double check that it runs as expected on the VM with an outdated cpu (qemu64) and on a VM with a modern cpu with hardened malloc enabled.

Diman0:
@nextgens It looks good to me.
On an outdated CPU (e.g. with CPU type qemu64), I do not see the warnings in the logs and Mailu starts fine.
On a modern CPU, I do get the warning to enable hardened malloc.
When running with LD_PRELOAD (hardened malloc enabled), the output of lsof confirms this.

nextgens (approved):

Diman0:
bors retry

Diman0:
bors retry

nextgens:
bors retry

nextgens:
bors try

nextgens (approved):

nextgens:
@mergifyio backport

PR #2972, opened 2023-10-09
switch to fts-flatcurve

### LINKED ISSUE #2184: Full text search for attachments
Hello,
is the full text search feature available for the message attachments, at least for the most common mediatype like pdf, docx, txt, etc.?
I've read something about xapian and solr but I didn't catch if the full-text is limited only to text fields (from, to, subject, body) or if there's a way to search also in the attachments.

### LINKED ISSUE #2390: Feature Request: Add xapian compact script, options, documentation
### Environment
 - [X] docker-compose
 - [ ] kubernetes
 - [ ] docker swarm

### Versions
1.9

## Description
Feature Request

Xapian (full text search) databases can grow to be huge, in my case some mail accounts which have 10gb of mail have xabian databases of 5gb. Xapian has a program called xapian-compact that will make them smaller, some of my databases were reduced by 50% after running this. it would be great to have a command that we could run that would filter through all the mail accounts and compact the xapian databases. i was able to do this running inside of docker but it's just a complex one liner and doesn't do all the mail accounts, only one at a time. A full compact script could be configured to run automatically via cron or manually.

```
for file in ./* ; do xapian-compact $file "${file}-new" && chown -R mail:mail "${file}-new" && mv $file "${file}-old" && mv "${file}-new" $file && rm -rf "${file}-old"; done
```

Also, xapain doesn't appear to be documented much, although there should be options to disable indexing of certain binary files like email attachments, and possibly even filter by type. for example, maybe scanning PDFs is good but scanning JPGs and MOVs is not. I think this depends mostly on the imap server configuration and how it sends data to xapian.

Example compacting rations (first column is new old size, second column is new size)
image

### LINKED ISSUE #2976: FTS doesn't perform OCR on attachments for keyword extraction
If one send a PDF composed of images, no keywords will be extracted/indexed.

To fix that, we could do some OCR:
- https://github.com/tesseract-ocr/tesseract

or if we want to get fancy, something like
- https://github.com/JaidedAI/EasyOCR

Or the built-in integration with
- https://tika.apache.org/
There is a security argument going for that option: it's written in java (memory  safe language) and would remove the need for decode2text and friends.

This is a follow-up to #2184 and #2972

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Switch from fts-xapian to fts-flatcurve. This should address the problem with indexes getting too big and will be the default in dovecot 2.4.

Don't forget to nuke old indexes to reclaim space.

### Related issue(s)
- close #2390
- close #2184
- close #2976

## Prerequisites


Diman0 (dismissed):
See in-line comment

core/dovecot/conf/dovecot.conf
```diff
@@ -57,8 +57,9 @@ plugin {
   quota_clone_dict = proxy:/tmp/podop.socket:quota
 
   {% if (FULL_TEXT_SEARCH or '').lower() not in ['off', 'false', '0'] %}
-  fts = xapian
-  fts_xapian = partial=2 full=30
+  fts = flatcurve
+  fts_languages = {% FULL_TEXT_SEARCH or 'en' %}
```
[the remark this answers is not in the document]
nextgens: Yeah, I've added it in a subsequent commit below

nextgens:
bors try

nextgens:
bors try

nextgens:
Stuff that could be improved include:

- add the snowball filter (pending on https://gitlab.alpinelinux.org/alpine/aports/-/issues/15347)
- switch from catdoc to iowriter (would be slower and make the image much bigger but would improve the parsing of office documents)

Diman0 (changes requested):
I hate to ask this, but did you actually test setup and the docker-compose.yml & mailu.env file created with it? It is good practice for me (not missing things with code reviewing)

  I miss the changes to the mailu.env file. Shouldn't mailu.env must be enhanced with:
```
{% if tika_enabled %}
FULL_TEXT_SEARCH_ATTACHMENTS=true
{% endif %}
```
Otherwise the setting is missing.

docs/configuration.rst shows that FULL_TEXT_SEARCH is set to `en` by default, but configuration.py is not updated to reflect this. It is currently not set.

Also the commented out FULL_TEXT_SEARCH=off in mailu.env should be changed to reflect the new default 
`FULL_TEXT_SEARCH=en`

`FTS_ATTACHMENTS_ADDRESS` is not added to the Dockerfile of base. So this env var is currently not know.

nextgens:
@Diman0 sorry about this; I have pushed too early.

It should be ready for review now.

I've identified two potential problems:
- there is no arm64 image for tika yet (qemu installs fine/transparently here)
- I am not sure how much memory tika will take and where it will stop

Diman0:
bors try

towncrier/newsfragments/2971.bugfix
```diff
@@ -0,0 +1,12 @@
+- Switch from fts-xapian to fts-flatcurve. This should address the problem with indexes getting too big and will be the default in dovecot 2.4
+- Enable full-text search of email attachments if configured (via Tika: you'll need to re-run setup)
```
Diman0, line 2: Please add that currently only x64 is supported or that ARM is not supported.

setup/templates/steps/compose/02_services.html
```diff
@@ -64,6 +64,15 @@
   <i>Oletools scans documents in email attachements for malicious macros. It has a much lower memory footprint than a full-fledged anti-virus.</i>
 </div>
 
+<div class="form-check form-check-inline">
+  <label class="form-check-label">
+    <input class="form-check-input" type="checkbox" name="tika_enabled" value="true">
+    Enable Tika
+  </label>
+
+  <i>Tika scans documents in email attachments, process (OCR, keyword extraction) and then index them in a way they can be efficiently searched. This requires significant ressources (RAM, CPU and storage).</i>
```
Diman0, line 73: Please make more clear that you really need this for the ability to search through attachments.
It must also mention that currently only amd64 / x64 is supported.
e.g.
`Tika enables the functionality for searching through attachments. Tika scans documents in email attachments, process (OCR, keyword extraction) and then index them in a way they can be efficiently searched. This requires significant ressources (RAM, CPU and storage). This is currently not supported on ARM.
`

setup/flavors/compose/mailu.env
```diff
@@ -186,3 +188,5 @@ DEFAULT_SPAM_THRESHOLD=80
 # This is a mandatory setting for using the RESTful API.
 API_TOKEN={{ api_token }}
 
+# Whether tika should be enabled (scan/OCR email attachements)
```
Diman0, line 191:
```suggestion
# Whether tika should be enabled (scan/OCR email attachements). To enable this feature, recreate the docker-compose.yml file via setup. 
```

Diman0 (changes requested):
Code-wise and functionality wise everything looks good. Only some requests to update wordings
* warn that ARM is not supported.
* Make more clear in setup that tika enables searching through attachments.

setup/templates/steps/compose/02_services.html
nextgens: That's 0a7797e9

setup/flavors/compose/mailu.env
nextgens: That's 0a7797e9

Diman0 (approved):
LGTM 
Mr Burns Excellent

PR #2975, opened 2023-10-10
Forbidden_file_extension.map could not be overridden.

### LINKED ISSUE #2937: Overriding rspamd forbidden_file_extension.map doesn't work as documented
## Environment & Version
The host is Ubuntu 22.04
### Environment

- [x] docker compose
- [ ] kubernetes
- [ ] docker swarm

### Version

- Version: `2.0`

## Description

My forbidden_file_extension.map placed in `overrides/rspamd/` is ignored and the default file is used instead.

## Replication Steps

With a Mailu 2.0 instance, follow the [instructions](https://mailu.io/2.0/antispam.html#can-i-change-the-list-of-authorized-file-attachments) for changing allowed attachments.  Change `overrides/rspamd/forbidden_file_extension.map` and add/remove some lines, then run `docker compose restart antispam`.  Re-run the `docker compose exec antispam cat /etc/rspamd/local.d/forbidden_file_extension.map` command to see the file used by rspamd.

## Observed behaviour

I get the output of `core/rspamd/conf/forbidden_file_extension.map` from the Mailu distribution.  Indeed, attachments allowed are still in line with this file, rather than the new changed file.

## Expected behaviour

I'd expect the modified forbidden_file_extension.map in the `overrides/rspamd` directory is taken into account.  For instance, if I remove js from the list, I should be able to send an attachment with the js extension.

## My diagnosis
The overridden file is available at `/overrides/forbidden_file_extension.map` in the antispam container, as expected.  However, `core/rspamd/start.py` adds `forbidden_file_extension.map` to the `config_files` variable so the override file at `/overrides/forbidden_file_extension.map` is ignored and not copied over.  The start script runs Jinja on `/conf/forbidden_file_extension.map` but since there's no directives it's just copied verbatim to `/etc/rspamd/local.d/`.  This default file is what rspamd reads.  The `multimap.conf` file references the map file, but that has no directive for overriding the path of the map file so overriding `multimap.conf` doesn't help.

## Possible solution
To get the expected behaviour (my override file is copied to `/etc/rspamd/local.d/` instead of the default file), I suspect the start script needs to be altered.  I'm unsure how this was intended to be done originally, but I'd probably modify the start script so an administrator can specify whether certain override configuration files should be copied despite being on the `config_files` list.  A pleasant way to still keep Jinja flexibility is for the copied override configuration files to be processed as templates too.  For instance I could copy the `multimap.conf` template to a file in the overrides directory, change it and let the start script process the new template.  This is nicely used in Kolla Ansible for instance.  If it's something relatively straightforward like this I'd be happy to make a PR for it, but might take some time since it's my first issue/PR.

Diman0:
## What type of PR?

bug fix

## What does this PR do?

Forbidden_file_extension.map could not be overridden. Start.py has been modified to allow this map to be overridden.
Now the instructions on https://mailu.io/master/antispam.html#can-i-change-the-list-of-authorized-file-attachments are valid again.

### Related issue(s)
- closes #2937 

## Prerequisites

### DIFF
--- a/core/rspamd/start.py
+++ b/core/rspamd/start.py
@@ -16,7 +16,8 @@ env = system.set_env()
 config_files = []
 for rspamd_file in glob.glob("/conf/*"):
     conf.jinja(rspamd_file, env, os.path.join("/etc/rspamd/local.d", os.path.basename(rspamd_file)))
-    config_files.append(os.path.basename(rspamd_file))
+    if rspamd_file != '/conf/forbidden_file_extension.map':
+        config_files.append(os.path.basename(rspamd_file))
 
 for override_file in glob.glob("/overrides/*"):
     if os.path.basename(override_file) not in config_files:
--- a/towncrier/newsfragments/2937.bugfix
+++ b/towncrier/newsfragments/2937.bugfix
@@ -1,0 +1,2 @@
+forbidden_file_extension.map could not be overridden. This file can be overriden to tweak with file extensions are allowed.
+The instructions on https://mailu.io/master/antispam.html#can-i-change-the-list-of-authorized-file-attachments work again.


nextgens (approved):
LGTM

PR #2981, opened 2023-10-13
Further improvements to fts

nextgens:
## What type of PR?

enhancement

## What does this PR do?

- update instructions
- set a timeout on search: if the indexes are not up to date, it will take at most 50s to fail
- limit parallelism to 1 (is that what we want?)
- switch to a different image that has ARM support
- ensure that journald logs the output of tika and oletools
- update the upgrade instructions to ensure one doesn't DoS his server while re-indexing
- add a healthcheck for tika

## Prerequisites


ghostwheel42 (approved):
LGTM

PR #2982, opened 2023-10-13
Switch to upstream's clamav image

### LINKED ISSUE #2059: Use official clamav docker image for Mailu clamav image
With https://github.com/Cisco-Talos/clamav having official docker support https://hub.docker.com/r/clamav/clamav it might be worth considering referring or preferring that container in the future?

nextgens:
## What type of PR?

enhancement

## What does this PR do?

Switch to upstream's clamav image. The only downside is that it doesn't support armv7... but odds are it won't have enough RAM anyway.

Why do we want this?
- faster build times
- will be easier to upgrade
- less maintenance required for us
- it may fix the CI tests failing because of clamav/EICAR not getting caught
- it will save bandwidth (as we won't re-download the definitions from fresh: they are in the base image)
- it puts clamav on its own network (security)

### Related issue(s)
- close #2059

## Prerequisites


nextgens:
bors try

nextgens:
bors try

nextgens:
bors try

ghostwheel42 (approved):

PR #2991, opened 2023-10-18
Remove IMAP ID capability from nginx

### LINKED ISSUE #2938: [2.0 only] Nginx IMAP proxy does not support ID command but advertises it
### Environment

- [X] docker compose
- [ ] kubernetes
- [ ] docker swarm

### Version

- Version: `2.0.23`

## Description
The 2.0 branch of Mailu uses Nginx as an IMAP proxy in the front container. This proxy contains this configuration for IMAP capabilities:
```
imap_capabilities IMAP4 IMAP4rev1 UIDPLUS SASL-IR LOGIN-REFERRALS ID ENABLE IDLE LITERAL+;
```
This includes the `ID` command which can be issued before authentication and thus before the connection is proxied to Dovecot, but the Nginx IMAP proxy does not support this command. It responds with:
```
BAD [b'invalid command']. Data: b'DPJO2 ID ("name" "OfflineIMAP" "version" "8.0.0")
```
when connecting with a client like OfflineIMAP which sends an ID command before authenticating.

Sending `ID` after authentication would work as then the command is handled by Dovecot.

This can be fixed by removing the `ID` from the IMAP capabilities list (manually tested).

This bug is not present in the master branch because the Nginx IMAP proxy was replaced by the Dovecot proxy.

If you want I'm willing submit a pull request to fix this issue. But your docs say to submit PRs against master and I'm unsure how to proceed because the bug is only present in the 2.0 branch.

I'm also unsure if other capabilities should be removed too like `IDLE` and `LITERAL+` as after the authentication the Dovecot server answers with a new list of actually supported capabilities anyway. But not sure if all clients out there then update their list of supported capabilities.

## Replication Steps

Try to connect to Mailu with a client like OfflineIMAP which sends an ID command before authentication.

## Logs

Unfortunately not much is logged about this.

gaudenz:
The ID capability is not supported by nginx. Advertizing it leads to problems with client which send an ID command before authenticating. Before authentication commands are not forwarded to dovecot, but directly processed by nginx.

## What type of PR?

bugfix PR

### Related issue(s)

Closes #2938

## Prerequisites

### DIFF
--- a/core/nginx/conf/nginx.conf
+++ b/core/nginx/conf/nginx.conf
@@ -340,7 +340,7 @@ mail {
     # Advertise real capabilities of backends (postfix/dovecot)
     smtp_capabilities PIPELINING "SIZE {{ MESSAGE_SIZE_LIMIT }}" ETRN ENHANCEDSTATUSCODES 8BITMIME DSN;
     pop3_capabilities TOP UIDL RESP-CODES PIPELINING AUTH-RESP-CODE USER;
-    imap_capabilities IMAP4 IMAP4rev1 UIDPLUS SASL-IR LOGIN-REFERRALS ID ENABLE IDLE LITERAL+;
+    imap_capabilities IMAP4 IMAP4rev1 UIDPLUS SASL-IR LOGIN-REFERRALS ENABLE IDLE LITERAL+;
 
     # Default SMTP server for the webmail (no encryption, but authentication)
     server {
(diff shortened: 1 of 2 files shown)


nextgens (approved):
Thank you for your contribution

We usually do not accept PRs against our stable branch... but in this instance master is not affected and it looks simple enough.

ghostwheel42 (approved):

PR #3025, opened 2023-10-29
Fix SETUP generating invalid files.

### LINKED ISSUE #1524: [SUGG] Little verification in Mailu setup
Hi thanks for Mailu it is a good project.
I submit this suggestion because i made the error and take many time to find it.
In step 4 of Mailu Setup for Docker compose : 
**Subnet of the docker network** it could be nice to verify if the last octet of the IP4 address is equal to 0 because if it is not the SMTP wont work.

mailu

Regards

### LINKED ISSUE #2396: setup.mailu.io: add server validation for configuration?
## Description
I'm using umatrix that block scripts from third-party domains.
So when I open https://setup.mailu.io, jquery from third party CDN was blocked, but web page looks good and I did not notice anything suspicious.

But without JQuery there is no input for WEB_ADMIN variable. And in result mailu.env is incorrect. Example: https://setup.mailu.io/master/file/048a14d0-9220-4468-8ac6-2b9bf524ff18/mailu.env
```
ADMIN=true
...
WEB_ADMIN=
```
So backend could generate unhealthy configuration without any notice.

### LINKED ISSUE #2890: Setup should double check that it doesn't produce a broken config even if browsers don't send it the data it requires
### Environment

- [x ] docker compose
- [ ] kubernetes
- [ ] docker swarm

### Version

- Version: `master`

## Description

My setup mailu with TLS_FLAVOUR mail. Mailu is setup to listen on 8080 & 8443. My reverse nginx proxy listening on 80/443 is forwarding all requests to mail on port :8080. Something's weird with the mailu front nginx server. All requests to the mail domain with / for webmail end up in a loop of 302 but /admin is working fine but no way to login to webmail. Even if I connect directly to the mailu IP and port it doesn't work at all.
## Replication Steps

Setup a mailu server with the docker-compose and mailu.env file enclosed.
docker-compose.yml
```
# This file is auto-generated by the Mailu configuration wizard.
# Please read the documentation before attempting any change.
# Generated for compose flavor

version: '2.2'

services:

  # External dependencies
  redis:
    image: redis:alpine
    restart: always
    volumes:
      - "/opt/docker/mailu/redis:/data"
    depends_on:
      - resolver
    dns:
      - 192.168.203.254

  # Core services
  front:
    image: ${DOCKER_ORG:-ghcr.io/mailu}/${DOCKER_PREFIX:-}nginx:${MAILU_VERSION:-master}
    restart: always
    env_file: mailu.env
    environment:
      - LETSENCRYPT_RESTART_CONTAINER=true
    logging:
      driver: journald
      options:
        tag: mailu-front
    ports:
      - "10.0.55.85:8080:80"
      - "10.0.55.85:8443:443"
      - "10.0.55.85:25:25"
      - "10.0.55.85:465:465"
      - "10.0.55.85:587:587"
      - "10.0.55.85:110:110"
      - "10.0.55.85:995:995"
      - "10.0.55.85:143:143"
      - "10.0.55.85:993:993"
    networks:
      - default
      - webmail
    volumes:
      - "/opt/docker/mailu/certs:/certs"
      - "/opt/docker/mailu/overrides/nginx:/overrides:ro"
    depends_on:
      - resolver
    dns:
      - 192.168.203.254

  resolver:
    image: ${DOCKER_ORG:-ghcr.io/mailu}/${DOCKER_PREFIX:-}unbound:${MAILU_VERSION:-master}
    env_file: mailu.env
    restart: always
    networks:
      default:
        ipv4_address: 192.168.203.254

  admin:
    image: ${DOCKER_ORG:-ghcr.io/mailu}/${DOCKER_PREFIX:-}admin:${MAILU_VERSION:-master}
    restart: always
    env_file: mailu.env
    logging:
      driver: journald
      options:
        tag: mailu-admin
    volumes:
      - "/opt/docker/mailu/data:/data"
      - "/opt/docker/mailu/dkim:/dkim"
    depends_on:
      - redis
      - resolver
    dns:
      - 192.168.203.254

  imap:
    image: ${DOCKER_ORG:-ghcr.io/mailu}/${DOCKER_PREFIX:-}dovecot:${MAILU_VERSION:-master}
    restart: always
    env_file: mailu.env
    logging:
      driver: journald
      options:
        tag: mailu-imap
    volumes:
      - "/opt/docker/mailu/mail:/mail"
      - "/opt/docker/mailu/overrides/dovecot:/overrides:ro"
    depends_on:
      - front
      - resolver
    dns:
      - 192.168.203.254

  smtp:
    image: ${DOCKER_ORG:-ghcr.io/mailu}/${DOCKER_PREFIX:-}postfix:${MAILU_VERSION:-master}
    restart: always
    env_file: mailu.env
    logging:
      driver: journald
      options:
        tag: mailu-smtp
    volumes:
      - "/opt/docker/mailu/mailqueue:/queue"
      - "/opt/docker/mailu/overrides/postfix:/overrides:ro"
    depends_on:
      - front
      - resolver
    dns:
      - 192.168.203.254

  oletools:
    image: ${DOCKER_ORG:-ghcr.io/mailu}/${DOCKER_PREFIX:-}oletools:${MAILU_VERSION:-master}
    hostname: oletools
    restart: always
    networks:
      - noinet
    depends_on:
      - resolver
    dns:
      - 192.168.203.254

  antispam:
    image: ${DOCKER_ORG:-ghcr.io/mailu}/${DOCKER_PREFIX:-}rspamd:${MAILU_VERSION:-master}
    hostname: antispam
    restart: always
    env_file: mailu.env
    logging:
      driver: journald
      options:
        tag: mailu-antispam
    networks:
      - default
      - noinet
    volumes:
      - "/opt/docker/mailu/filter:/var/lib/rspamd"
      - "/opt/docker/mailu/overrides/rspamd:/overrides:ro"
    depends_on:
      - front
      - redis
      - oletools
      - antivirus
      - resolver
    dns:
      - 192.168.203.254

  # Optional services
  antivirus:
    image: ${DOCKER_ORG:-ghcr.io/mailu}/${DOCKER_PREFIX:-}clamav:${MAILU_VERSION:-master}
    restart: always
    env_file: mailu.env
    volumes:
      - "/opt/docker/mailu/filter:/data"
    depends_on:
      - resolver
    dns:
      - 192.168.203.254

  # Webmail
  webmail:
    image: ${DOCKER_ORG:-ghcr.io/mailu}/${DOCKER_PREFIX:-}webmail:${MAILU_VERSION:-master}
    restart: always
    env_file: mailu.env
    volumes:
      - "/opt/docker/mailu/webmail:/data"
      - "/opt/docker/mailu/overrides/roundcube:/overrides:ro"
    networks:
      - webmail
    depends_on:
      - front

networks:
  default:
    driver: bridge
    ipam:
      driver: default
      config:
        - subnet: 192.168.203.0/24
  webmail:
    driver: bridge
  noinet:
    driver: bridge
    internal: true
```

mailu.env
```
# Mailu main configuration file
#
# This file is autogenerated by the configuration management wizard for compose flavor.
# For a detailed list of configuration variables, see the documentation at
# https://mailu.io

###################################
# Common configuration variables
###################################

# Set to a randomly generated 16 bytes string
SECRET_KEY=XXXXXXXXXXXXXXXX

# Subnet of the docker network. This should not conflict with any networks to which your system is connected. (Internal and external!)
SUBNET=192.168.203.0/24

# Main mail domain
DOMAIN=example.com

# Hostnames for this server, separated with commas
HOSTNAMES= mail.example.com

# Postmaster local part (will append the main mail domain)
POSTMASTER=admin

# Choose how secure connections will behave (value: letsencrypt, cert, notls, mail, mail-letsencrypt)
TLS_FLAVOR=mail

# Authentication rate limit per IP (per /24 on ipv4 and /48 on ipv6)
AUTH_RATELIMIT_IP=5/hour

# Authentication rate limit per user (regardless of the source-IP)
AUTH_RATELIMIT_USER=50/day

# Opt-out of statistics, replace with "True" to opt out
DISABLE_STATISTICS=False

###################################
# Optional features
###################################

# Expose the admin interface (value: true, false)
ADMIN=true

# Choose which webmail to run if any (values: roundcube, snappymail, none)
WEBMAIL=roundcube

# Expose the API interface (value: true, false)
API=false

# Dav server implementation (value: radicale, none)
WEBDAV=none

# Antivirus solution (value: clamav, none)
ANTIVIRUS=none

# Scan Macros solution (value: true, false)
SCAN_MACROS=false

###################################
# Mail settings
###################################

# Message size limit in bytes
# Default: accept messages up to 50MB
# Max attachment size will be 33% smaller
MESSAGE_SIZE_LIMIT=50000000

# Message rate limit (per user)
MESSAGE_RATELIMIT=200/day

# Networks granted relay permissions
# Use this with care, all hosts in this networks will be able to send mail without authentication!
RELAYNETS=

# Will relay all outgoing mails if configured
RELAYHOST=

# Enable fetchmail
FETCHMAIL_ENABLED=False

# Fetchmail delay
FETCHMAIL_DELAY=600

# Recipient delimiter, character used to delimiter localpart from custom address part
RECIPIENT_DELIMITER=+

# DMARC rua and ruf email
DMARC_RUA=admin
DMARC_RUF=admin

# Welcome email, enable and set a topic and body if you wish to send welcome
# emails to all users.
WELCOME=false
WELCOME_SUBJECT=Welcome to your new email account
WELCOME_BODY=Welcome to your new email account, if you can read this, then it is configured properly!

# Maildir Compression
# choose compression-method, default: none (value: gz, bz2, zstd)
COMPRESSION=
# change compression-level, default: 6 (value: 1-9)
COMPRESSION_LEVEL=

# IMAP full-text search is enabled by default. Set the following variable to off in order to disable the feature.
# FULL_TEXT_SEARCH=off

###################################
# Web settings
###################################

# Path to redirect / to
WEBROOT_REDIRECT=/

# Path to the admin interface if enabled
WEB_ADMIN=/admin

# Path to the webmail if enabled
WEB_WEBMAIL=/

# Path to the API interface if enabled
WEB_API=

# Website name
SITENAME=Capere Mail

# Linked Website URL
WEBSITE=https://mail.example.com


###################################
# Advanced settings
###################################

# Docker-compose project name, this will prepended to containers names.
COMPOSE_PROJECT_NAME=mailu

# Number of rounds used by the password hashing scheme
CREDENTIAL_ROUNDS=12

# Header to take the real ip from
REAL_IP_HEADER=X-Real-IP

# IPs for nginx set_real_ip_from (CIDR list separated by commas)
REAL_IP_FROM=0.0.0.0/0, ::0

# choose wether mailu bounces (no) or rejects (yes) mail when recipient is unknown (value: yes, no)
REJECT_UNLISTED_RECIPIENT=yes

# Log level threshold in start.py (value: CRITICAL, ERROR, WARNING, INFO, DEBUG, NOTSET)
LOG_LEVEL=WARNING

# Timezone for the Mailu containers. See this link for all possible values https://en.wikipedia.org/wiki/List_of_tz_database_time_zones
TZ=Europe/Berlin

# Default spam threshold used for new users
DEFAULT_SPAM_THRESHOLD=80

# API token required for authenticating to the RESTful API.
# This is a mandatory setting for using the RESTful API.
API_TOKEN=

# SESSION COOKIE DISABLED TO AVOID ENDLESS 302/REDIRECTS
SESSION_COOKIE_SECURE=False
``` 
## Observed behaviour

Login to https://mail.example.com/ ending up in a HTTP 302 loop
## Expected behaviour

Webmail should be accessible and open once requested via /
## Logs

```
- - mail.example.com - 10.0.55.1 forwarded for - - - [28/Jul/2023:14:17:20 +0000]  "POST /sso/login?url=/ HTTP/2.0" 302 231 "https://mail.example.com/sso/login?url=/" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/115.0.0.0 Safari/537.36"
- - mail.example.com - 10.0.55.1 forwarded for - - - [28/Jul/2023:14:17:20 +0000]  "GET / HTTP/2.0" 302 0 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/115.0.0.0 Safari/537.36"
- - mail.example.com - 10.0.55.1 forwarded for - - - [28/Jul/2023:14:17:22 +0000]  "GET /sso.php HTTP/2.0" 302 0 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/115.0.0.0 Safari/537.36"
- - mail.example.com - 10.0.55.1 forwarded for - - - [28/Jul/2023:14:17:23 +0000]  "GET /sso.php HTTP/2.0" 302 0 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/115.0.0.0 Safari/537.36"
- - mail.example.com - 10.0.55.1 forwarded for - - - [28/Jul/2023:14:17:24 +0000]  "GET /sso.php HTTP/2.0" 302 0 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/115.0.0.0 Safari/537.36"
- - mail.example.com - 10.0.55.1 forwarded for - - - [28/Jul/2023:14:17:25 +0000]  "GET /sso.php HTTP/2.0" 302 0 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/115.0.0.0 Safari/537.36"
- - mail.example.com - 10.0.55.1 forwarded for - - - [28/Jul/2023:14:17:26 +0000]  "GET /sso.php HTTP/2.0" 302 0 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/115.0.0.0 Safari/537.36"
- - mail.example.com - 10.0.55.1 forwarded for - - - [28/Jul/2023:14:17:27 +0000]  "GET /sso.php HTTP/2.0" 302 0 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/115.0.0.0 Safari/537.36"
- - mail.example.com - 10.0.55.1 forwarded for - - - [28/Jul/2023:14:17:28 +0000]  "GET /sso.php HTTP/2.0" 302 0 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/115.0.0.0 Safari/537.36"

```

Diman0:
## What type of PR?

bug-fix

## What does this PR do?
If JavaScript script is disabled or JS files are blocked from the setup website, the SETUP website would generate invalid configuration. Checks were added to prevent this situation from occurring.

Bootstrap loaded an outdated jquery version 1. We already use Jquery 2. I have disabled this and configured bootstrap to load our jquery js file.

Also fixed a regular expression that makes sure the specified Mailu storage path is valid.

### Related issue(s)
- closes #2396
- closes #2890
- closes #1524  

## Prerequisites

- [n/a] In case of feature or enhancement: documentation updated accordingly


nextgens:
Can we get #1524 nailed too? What about doing the check server side and erroring out if we don't get parameters instead?

Diman0:
> Can we get #1524 nailed too? What about doing the check server side and erroring out if we don't get parameters instead?

I wasn't aware of that issue. Yes. That should be included as well.

Diman0:
@nextgens I have updated the hint for setting the subnet and changed the regular expression so it only accepts a 0 for the fourth parameter. So this format is accepted *.*.*.0/** (where * is a number 0-9).

How would server side validation look like? I don't have a lot experience with web development. Does it offer advantages?
 I  don't want to spend too much time on this. In the future we want to get rid of setup after all.

nextgens:
> @nextgens I have updated the hint for setting the subnet and changed the regular expression so it only accepts a 0 for the fourth parameter. So this format is accepted _._.*.0/** (where * is a number 0-9).
> 

That may break things for those that do not use a /24 ... we could just switch strict to True here
 https://github.com/Mailu/Mailu/blob/master/setup/server.py#L94 ... and return a proper error.

> How would server side validation look like? I don't have a lot experience with web development. Does it offer advantages? I don't want to spend too much time on this. In the future we want to get rid of setup after all.

The main advantage would be that it would work in all scenarios (including when part of the JS is disabled because of some browser extension).

Well we would double-check if we do get the parameters we expect... and if we don't we flash() an error message.

Diman0:
> > @nextgens I have updated the hint for setting the subnet and changed the regular expression so it only accepts a 0 for the fourth parameter. So this format is accepted _._.*.0/** (where * is a number 0-9).
> 
> That may break things for those that do not use a /24 ... we could just implement something like https://github.com/Mailu/Mailu/blob/master/setup/server.py#L94 ... and return a proper error.
> 

I'm probably missing something, what would exactly break? The reg exp only enforces a 0 for the 4th parameter. So you can use:

* 192.168.12.0/24
* 192.168.0.0/16
* 192.0.0.0/8

nextgens:
You can't use 192.168.0.128/25 for instance. Subnets may have less than 256 addresses and enforcing that there is a zero as the 4th parameter is technically the wrong test. What you want to do is ensure that the "host bits" are set to zero... but you can't do it easily using a regexp.

Mind you maybe this is advanced-enough usage that we don't mind it being broken.

Diman0:
I didn't know that you could create such (limited) subnets. I think it is too advanced for Mailu. The suggested subnet should only be changed if it already exists. The main goal is to prevent less experienced users from making a mistake by picking 192.111.12.1/24 as subnet.

nextgens:
Another way to deal with this would be to read it server-side and output the right thing in the config (fix it for the user).

https://docs.python.org/3/library/ipaddress.html#ipaddress.ip_network

Diman0:
Ah that function checks if the provided IP is a valid subnet. I did not realize that python offered this functionality. This makes it easy to check it server-side and show a human readable message that the subnet is invalid. I will adapt setup to do this.

Diman0:
bors try

nextgens:
let's do the same for SUBNET6 below, no?

nextgens:
I'd get rid of the scary regexp at https://github.com/Mailu/Mailu/pull/3025/files#diff-11e701060cb3c1f0e208aa5ad066ac140e32551bcd8ab794d795669c4d40941bL38 too

Diman0:
bors try

Diman0:
discussed during meeting.
By default make div container hidden. The text you need javascript is shown by default. Via onloaded event use javascript (something from render.js maybe) to make the container div visible and the warning invisible. So if javascript is blocked you still see the warning.

Do all validation server side. So add checks for everything in `def submit()`

Diman0:
bors try

setup/Dockerfile
```diff
@@ -14,6 +14,10 @@ COPY main.py ./main.py
 
 RUN echo $VERSION >> /version
 
+#Note: This is appended so we can explicitly check if this JS file has been loaded
+#by the user's internet browser when accessing the setup site.
+RUN echo var jQueryMailu=\'loaded\'\; >> ./static/jquery.min.js
```
nextgens, line 19: Not required anymore

setup/templates/steps/config.html
```diff
@@ -58,13 +58,6 @@
   </p>
 </div>
 
-<div class="form-check form-check-inline">
-  <label class="form-check-label">
-    <input class="form-check-input" type="checkbox" name="disable_statistics" value="True">
-    Opt-out of statistics
```
nextgens, line 64: Hmm, I wouldn't remove this without removing the code in admin. Can't we fix it instead?

setup/flavors/compose/mailu.env
```diff
@@ -39,9 +39,6 @@ AUTH_RATELIMIT_IP={{ auth_ratelimit_ip }}/hour
 AUTH_RATELIMIT_USER={{ auth_ratelimit_user }}/day
 {% endif %}
 
-# Opt-out of statistics, replace with "True" to opt out
-DISABLE_STATISTICS={{ disable_statistics or 'False' }}
```
nextgens, line 43: Same thing, I wouldn't remove it if we are still making the request

nextgens (changes requested):

setup/templates/steps/config.html
Diman0: It has never worked. It requires a separate webserver to run (https://github.com/Mailu/Stats). This webserver has never been deployed.  

But I agree, if this is removed then the code in admin should be removed as well. I didn't think about admin.

Diman0: I could reinstate the code, but will simply hide the element and set it to disabled in mailu.env. 
That way we can still choose if we want to implement it in the future by hosting the webserver required for this.

Diman0:
bors try

setup/server.py
```diff
@@ -90,12 +94,65 @@ def wizard():
     def submit():
         data = flask.request.form.copy()
         data['uid'] = str(uuid.uuid4())
+        valid = True
+        try:
+            ipaddress.ip_address(data['bind4'])
```
nextgens, line 99: ipaddress.IPv4Address would be a better choice here

setup/server.py
```diff
@@ -90,12 +94,65 @@ def wizard():
...
+        except:
+            flask.flash('Configured IPv4 address is invalid', 'error')
+            valid = False
+        try:
+            ipaddress.ip_address(data['bind6'])
```
nextgens, line 104: same here and below for ip_network

setup/static/render.js
```diff
@@ -1,21 +1,30 @@
...
+$(document).ready(function() {
+	$("#no_java_script").hide();
+	$("#container").show();
+});
+
 $(document).ready(function() {
 	if ($("#webmail").val() == 'none') {
 		$("#webmail_path").hide();
 		$("#webmail_path").val("");
+		$("#webmail_path").prop('required',false);
```
nextgens, line 13: I would rather get rid of all ``WEB_`` variables tbh. We agreed Mailu should have it's own FQDN, let's simplify and hard-code paths on it.

setup/templates/steps/compose/03_expose.html
```diff
@@ -16,11 +16,10 @@
...
   <label>IPv4 listen address</label>
 <!--   Validates IPv4 address -->
-  <input class="form-control" type="text" name="bind4" value="127.0.0.1" 
-  		pattern="^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$">
-  <label>Subnet of the docker network. This should not conflict with any networks to which your system is connected. (Internal and external!)</label>
-  <input class="form-control" type="text" name="subnet" required pattern="^([0-9]{1,3}\.){3}[0-9]{1,3}(\/([0-9]|[1-2][0-9]|3[0-2]))$"
-    value="192.168.203.0/24">
+  <input class="form-control" type="text" name="bind4" value="127.0.0.1" >
+  <label>Subnet of the docker network. This should not conflict with any networks to which your system is connected. (Internal and external!)
+    Fourth number is always 0. Normally the format is '*.*.*.0/24'</label>
```
nextgens, line 21:
```suggestion
    Usually the format is '*.*.*.0/24'</label>
```

nextgens (dismissed):

setup/static/render.js
Diman0: I'm fine with removing it from setup only. Setup should be simplified and help the end-user not making any mistakes. A typical user should not have to change the `WEB_` settings. 

Advanced users can still tweak mailu.env.

Diman0:
bors try

Diman0:
bors try-

Diman0:
bors try

nextgens (approved):

nextgens:
@Diman0 should we backport this?

Diman0:
> @Diman0 should we backport this?

That is not a bad idea. The issue was reported for mailu 2.0 after all.

nextgens:
@mergifyio backport

nextgens:
@mergifyio backport 2.0

PR #3050, opened 2023-11-12
[WIP] Migrate to Flask-SQLAlchemy 3.1.1 and SQLAlchemy 2.0.23

DjVinnii:
## What type of PR?

Enhancement

## What does this PR do?
Upgrade to Flask-SQLAlchemy 3.1.1 and SQLAlchemy 2.0.23.

## Prerequisites

- [N/A] In case of feature or enhancement: documentation updated accordingly

### DIFF
--- a/core/admin/mailu/internal/views/dovecot.py
+++ b/core/admin/mailu/internal/views/dovecot.py
@@ -3,13 +3,14 @@ from mailu.internal import internal
 from flask import current_app as app
 
 import flask
-import socket
-import os
 import sqlalchemy.exc
+
+db = models.db
+
 
 @internal.route("/dovecot/passdb/<path:user_email>")
 def dovecot_passdb_dict(user_email):
-    user = models.User.query.get(user_email) or flask.abort(404)
+    user = db.get_or_404(models.User, user_email)
     allow_nets = []
     allow_nets.append(app.config["SUBNET"])
     if app.config["SUBNET6"]:
@@ -20,16 +21,18 @@ def dovecot_passdb_dict(user_email):
         "allow_real_nets": ",".join(allow_nets)
     })
 
+
 @internal.route("/dovecot/userdb/")
 def dovecot_userdb_dict_list():
     return flask.jsonify([
-        user[0] for user in models.User.query.filter(models.User.enabled.is_(True)).with_entities(models.User.email).all()
+        user[0] for user in db.session.execute(db.select(models.User).filter_by(enabled=True).with_only_columns(models.User.email)).all()
     ])
+
 
 @internal.route("/dovecot/userdb/<path:user_email>")
 def dovecot_userdb_dict(user_email):
     try:
-        quota = models.User.query.filter(models.User.email==user_email).with_entities(models.User.quota_bytes).one_or_none() or flask.abort(404)
+        quota = db.one_or_404(db.select(models.User).filter_by(email=user_email).with_only_columns(models.User.quota_bytes))
     except sqlalchemy.exc.StatementError as exc:
         flask.abort(404)
     return flask.jsonify({
@@ -39,11 +42,11 @@ def dovecot_userdb_dict(user_email):
 
 @internal.route("/dovecot/quota/<ns>/<path:user_email>", methods=["POST"])
 def dovecot_quota(ns, user_email):
-    user = models.User.query.get(user_email) or flask.abort(404)
+    user = db.get_or_404(models.User, user_email)
     if ns == "storage":
         user.quota_bytes_used = flask.request.get_json()
         user.dont_change_updated_at()
-        models.db.session.commit()
+        db.session.commit()
     return flask.jsonify(None)
 
 
@@ -54,5 +57,5 @@ def dovecot_sieve_name(script, user_email):
 
 @internal.route("/dovecot/sieve/data/default/<path:user_email>")
 def dovecot_sieve_data(user_email):
-    user = models.User.query.get(user_email) or flask.abort(404)
+    user = db.get_or_404(models.User, user_email)
     return flask.jsonify(flask.render_template("default.sieve", user=user))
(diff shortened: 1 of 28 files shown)


DjVinnii:
I think I tested most of this PR, but it is possible I forgot something. Maybe it is good to improve local testing experience with proper unit/feature tests for at least the admin part.  For now, it would be great if someone can take a look for initial review.

nextgens:
You are more than welcome to add tests :)

bors try

Diman0:
bors try

core/admin/mailu/manage.py
```diff
@@ -96,7 +97,7 @@ def admin(localpart, domain_name, password, mode):
 def user(localpart, domain_name, password):
     """ Create a user
     """
-    domain = models.Domain.query.get(domain_name)
+    domain = db.session.get(models.domain, domain_name)
```
Diman0, line 100: This fails during the automatic test with the error:
```
   File "/app/mailu/manage.py", line 100, in user
    domain = db.session.get(models.domain, domain_name)
                            ^^^^^^^^^^^^^
AttributeError: module 'mailu.models' has no attribute 'domain'. Did you mean: 'Domain'?
```

full stack trace at: https://github.com/Mailu/Mailu/actions/runs/6868358963/job/18678819073#step:11:1197

Diman0 (changes requested):
Tests do not pass. See in-line comment

DjVinnii:
bors try

core/admin/mailu/manage.py
DjVinnii: This should be `models.Domain` indeed, fixed.

nextgens:
bors retry

nextgens:
bors try

core/admin/mailu/internal/views/dovecot.py
```diff
@@ -20,16 +21,18 @@ def dovecot_passdb_dict(user_email):
...
 
+
 @internal.route("/dovecot/userdb/<path:user_email>")
 def dovecot_userdb_dict(user_email):
     try:
-        quota = models.User.query.filter(models.User.email==user_email).with_entities(models.User.quota_bytes).one_or_none() or flask.abort(404)
+        quota = db.one_or_404(db.select(models.User.quota_bytes).filter_by(email=user_email))
     except sqlalchemy.exc.StatementError as exc:
         flask.abort(404)
     return flask.jsonify({
```
Diman0, line 38: Latest CI error is for this function. It is used when someone logs in via imap. 


https://github.com/Mailu/Mailu/actions/runs/6870165755/job/18684540112#step:11:1755
```
 [2023-11-14 22:19:48,356] INFO in nginx: Login attempt for: admin@mailu.io/submission/587 from: 192.168.203.1/36240: success: password
[2023-11-14 22:19:49,176] ERROR in app: Exception on /internal/dovecot/userdb/user@mailu.io [GET]
  File "/app/venv/lib/python3.11/site-packages/flask/app.py", line 2190, in wsgi_app
    response = self.full_dispatch_request()
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/venv/lib/python3.11/site-packages/flask/app.py", line 1486, in full_dispatch_request
    rv = self.handle_user_exception(e)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/venv/lib/python3.11/site-packages/flask_restx/api.py", line 674, in error_router
    return original_handler(e)
           ^^^^^^^^^^^^^^^^^^^
  File "/app/venv/lib/python3.11/site-packages/flask/app.py", line 1484, in full_dispatch_request
    rv = self.dispatch_request()
         ^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/venv/lib/python3.11/site-packages/flask/app.py", line 1469, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/mailu/internal/views/dovecot.py", line 39, in dovecot_userdb_dict
    "quota_rule": f"*:bytes={quota[0]}"
                             ~~~~~^^^
TypeError: 'int' object is not subscriptable
```

Diman0 (changes requested):
CI caught another error. See in-line comment. I'm not sure what the problem is. This function should be called when you access imap (dovecot). Dovecot should use this to lookup the user.

core/admin/mailu/internal/views/dovecot.py
nextgens: That query is a relatively "hot path", we should keep the .with_entities() equivalent to (a) reduce the size of what is returned and (b) make it easier on the DBM.

@DjVinnii please ensure you "scope" the selects everywhere we do in the existing code.

DjVinnii: This is my mistake, I assumed the  `db.select(models.User.quota_bytes)` wouldn't only return the quota_bytes but an object.

According to the SQLAlchemy documentation the `.with_entities()` is part of the legacy query API and replaces the SELECT with the given entities. I think there are 2 way to tackle this:
1. Only select the `quota_bytes` and not the whole model so there is no need to use an index to select the quota bytes.
2. Select the whole model and use `.with_only_columns()` and use the index to target the quota bytes.

For now I went with option 2 in order to keep the changes as close as to the original. I also did the same for other queries with `.with_entities()`

nextgens:
bors retry

nextgens:
#3138 is just a lot less invasive... so let's go for that instead

DjVinnii:
@nextgens I fully agree with you. Unfortunately I haven't had the time to fully complete this PR. I still think it is good to move to [the 2.0 style ORM queries](https://docs.sqlalchemy.org/en/20/changelog/migration_20.html#migration-orm-usage). If you think so as well, I'd still like to work on this (probably in smaller PR's so it is more manageable). What do you think?

PR #3057, opened 2023-11-15
Retry up to 5 times if not ready

nextgens:
Add retry logic to ensure we don't always pass if clamav is not ready; return a 25 error only if it's a permanent rejection code so that we can differentiate in between "definitely reject" and "was rejected because of a glitch"


nextgens:
bors try

nextgens:
Looks like it works and is much better; I'll add those commits to the backport too

Diman0 (approved):
lgtm

Issue #1040, opened 2019-06-05
Feature: (optionally) include Dovecot encryption plugin


alexanderadam:
I just stumbled [over this Dovecot encryption plugin](https://github.com/posteo/scrambler-plugin) which allows to have encrypted mailboxes.

It might be especially useful for the folks that focuses on security and issues like #391, #506 and #703 show that there is some interest in encryption in general.

blinkeye:
May I ask your opinion in using this plugin vs. an encrypted maildata share (like [Cryptsetup LUKS)](https://gitlab.com/cryptsetup/cryptsetup/)?

alexanderadam:
Sure: if I understood this correctly, the Dovecot encryption plugin encrypts the mailbox for each user individually. Meaning: even admins aren't able to read any email from another user.

Using just an encrypted mountpoint (like LUKS) however doesn't provides this at all, because admins can naturally access all data on it.

paulolieuthier:
There's also [MailCrypt](https://wiki2.dovecot.org/Plugins/MailCrypt).

alexanderadam:
> There's also MailCrypt.

nice. I wouldn't care which variant would be integrated. Both sound suitable :+1:

Angedestenebres:
Hey,

Maybe we can also suggest Zeyple? See: https://github.com/infertux/zeyple

It was suggested by one of our user.

alexanderadam:
@Angedestenebres this is something different though: Zeyple is about _outgoing_ emails whereas Dovecot encryption / MailCrypt are about _stored_ emails.

I agree that those might be used together but I guess this should be a separate ticket.

kaiyou:
After some discussion, zeyple is deemed both useless and possibly largely insecure.

MailCrypt might still be included if someone wishes to make a PR.

Nebukadneza:
Hi There,

The `Mailu`-Project is currently in a bit of a bind! We are short on man-power, and we need to judge if it is possible for us to put in some work on this issue.

To help with that, we are currently trying to find out which issues are actively keeping users from using `Mailu`, which issues have someone who want to work on them — and which issues may be less important. These a less important ones could be discarded for the time being, until the project is in a more stable and regular state once again.

In order for us to better assess this, it would be helpful if you could put a **reaction on this post** (use the :smiley: icon to the top-right).
- 👍️ if you **need this** to be able to use Mailu. Ideally, you’d also be able to test this on your installation, and provide feedback …
[...]

PR #3058, opened 2023-11-15
Enable IMAP-METADATA

nextgens:
## What type of PR?

Feature

## What does this PR do?

Enable IMAP METADATA; that will enable IMAP clients to store information server-side

### Related issue(s)
- #1040

## Prerequisites


nextgens:
That seems to work... but we may want to do it differently so that front can "see" the attributes too.

nextgens:
Yeah that needs to be done in a way "front" gets sight of it.

PR #3060, opened 2023-11-16
nginx: Add PROXY_PROTOCOL configure as HTTPS, not for HTTP

756yang:
## What type of PR?

This submission is used to solve the problem of TCP proxy forwarding directly to the port monitored by Mailu through SNI diversion and maintaining the normal application of the letsencrypt certificate.

When 'proxy_protocol' is enabled, it only applies to HTTPS listening ports and not to HTTP ports.

## What does this PR do?

I found that at the beginning of the mailu configuration, the letsencrypt certificate can only be applied for through port 80. If the public IP port 80 is not properly connected to the mailu container, the certificate cannot be applied for.

If the mailu container uses "proxy_protocol", it will also enable this protocol for port 80. This way, the mailu inside the container needs an external anti proxy tool to handle the request correctly. If it is nginx, the HTTP request needs to be configured with "proxy_protocol on" in the stream module and forwarded, which is actually not elegant because it has to be forwarded twice with a new port.

The solution is to make the HTTP port that Mailu listens to not do "proxy_protocol".

## Prerequisites

### DIFF
--- a/core/nginx/conf/nginx.conf
+++ b/core/nginx/conf/nginx.conf
@@ -22,7 +22,7 @@ http {
 
     {% if REAL_IP_HEADER %}
     real_ip_header {{ REAL_IP_HEADER }};
-    {% elif PROXY_PROTOCOL in ['all', 'all-but-http', 'http'] %}
+    {% elif PROXY_PROTOCOL in ['all', 'all-but-http', 'http', 'https'] %}
     real_ip_header proxy_protocol;
     {% endif %}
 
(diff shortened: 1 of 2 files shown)


nextgens:
I am sorry but no.

``proxy_protocol on`` is all you need with nginx to reverse proxy things in a way Mailu can handle. Using the stream module is not required for port 80.

I don't think we should support the scenario where all ports but port 80 are behind a proxy.

If you want to document how nginx should be configured and send a PR with that, we can accept it.

756yang:
> `proxy_protocol on` is all you need with nginx to reverse proxy things in a way Mailu can handle. Using the stream module is not required for port 80.

@nextgens If the 80 port of Mailu does not use an external nginx stream module proxy, using an HTTP service proxy will directly report an error:  **[emerg] 2234#2234: "proxy_protocol" directive is not allowed here**

The nginx conf is like:
``` conf
http {
...
	server {
		listen 80;
		server_name zm.xx.yy; 
		location ^~ /.well-known/acme-challenge/ {
			proxy_protocol on;
			proxy_pass http://127.0.0.1:8080;
		}
	}
}
```

PR #3061, opened 2023-11-18
Allow PROXY_PROTOCOL to be per-port customizable.

alex4108:
## What type of PR?

Enhancement

## What does this PR do?

Allow `PROXY_PROTOCOL` env variable to be set on only a specific port.  Eg, in my case, only tcp/25 needs to have PROXY_PROTOCOL enabled.

## Prerequisites

- [ x ] In case of feature or enhancement: documentation updated accordingly


nextgens:
#3060 was asking for something similar

PR #3076, opened 2023-11-30
Slow down the turtle policy

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Slow down the turtle policy

### Related issue(s)
- close #3075 

## Prerequisites


nextgens:
The next major release: 2.1 that was due last month

PR #3083, opened 2023-12-07
Check SPF/DKIM/DMARC record on dns like MX records

Riscue:
## What type of PR?

Enhancement

## What does this PR do?

Checks TXT records for SPF, DKIM, DMARC same as MX record checker

### DIFF
--- a/core/admin/mailu/models.py
+++ b/core/admin/mailu/models.py
@@ -384,7 +384,7 @@ class Domain(Base):
         app_config_domain = app.config['DOMAIN']
         try:
             return any(
-                rset.to_text() == '"v=DMARC1"'
+                rset.to_text().startswith('"v=DMARC1;')
                 for rset in dns.resolver.resolve(f'{self.name}._report._dmarc.{app_config_domain}.', 'TXT')
             )
         except dns.exception.DNSException:
@@ -492,7 +492,7 @@ class Alternative(Base):
         app_config_domain = app.config['DOMAIN']
         try:
             return any(
-                rset.to_text() == '"v=DMARC1"'
+                rset.to_text().startswith('"v=DMARC1;')
                 for rset in dns.resolver.resolve(f'{self.name}._report._dmarc.{app_config_domain}.', 'TXT')
             )
         except dns.exception.DNSException:


core/admin/mailu/models.py
```diff
@@ -334,6 +335,46 @@ def check_mx(self):
         except dns.exception.DNSException:
             return False
 
+    def check_spf(self):
+        """ checks if SPF record for domain points to mailu host """
+        try:
+            hostnames = app.config['HOSTNAMES'].replace(',', '|')
+            return any(
+                re.search(f'^"v=spf1 mx a:{hostnames} [?+-~]all"$', rset.to_text())
```
ghostwheel42, line 343: I think hostnames should be grouped: `f'^"v=spf1 mx a:({hostnames}) [?+-~]all"$'`

ghostwheel42 (changes requested):
Hi. Thanks for your contribution.
I was thinking about doing the checks asynchronously (by having a "spinner" image and requesting the check via src="" returning a good or bad image), but this is a good start.
I didn't run in yet, but I think the regex should be changed.

core/admin/mailu/models.py
```diff
@@ -334,6 +335,46 @@ def check_mx(self):
...
+        try:
+            selector = app.config['DKIM_SELECTOR']
+            return any(
+                rset.to_text() == f'"v=DKIM1; k=rsa; p={self.dkim_publickey}"'
+                for rset in dns.resolver.resolve(f'{selector}._domainkey.{self.name}.', 'TXT')
+            )
+        except dns.exception.DNSException:
+            return False
+
+    def check_dmarc(self):
```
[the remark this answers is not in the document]
Riscue: `Dmarc record` and `DMARC report record` was grouped, check mark icon is in first column per row. That's why i checked both record in one method. And both record has only one icon in their first column. I can split and put two check mark next to record besides row

core/admin/mailu/models.py
Riscue: We can improve all regexes if there is expert about dns records. Yeah my mistake, i will update this.

ghostwheel42:
Sorry for the delay. At first I thought the simplest way to implement this would be a route to check dns entries like `.../check/dkim` returning a single location header pointing to an image (good or bad).
The web page would have a spinning circle as lowsrc and the check as src. 
Another way would be to do it via JS/jquery and a jsonp response. This could return more precise error messages and the values of the actual records. 

Alex

core/admin/mailu/models.py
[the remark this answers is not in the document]
Riscue: Done

core/admin/mailu/models.py
Riscue: Done

Riscue:
> Sorry for the delay. At first I thought the simplest way to implement this would be a route to check dns entries like `.../check/dkim` returning a single location header pointing to an image (good or bad). The web page would have a spinning circle as lowsrc and the check as src. Another way would be to do it via JS/jquery and a jsonp response. This could return more precise error messages and the values of the actual records.
> 
> Alex

Done via async image loading. Please check the routes for images and css changes

Riscue:
bors try

Riscue:
@ghostwheel42 Hey alex bors stuck, could you please force for retry and review new commits

ghostwheel42:
bors try

Riscue:
@ghostwheel42 Hey alex fixed a bug, please retry

ghostwheel42:
bors try

Riscue:
Hey Alex @ghostwheel42 it is done actually, what should i do next?

ErwanMAS:
@Riscue  can you update your P.R.  toward the last version of the code 
you have now some conflicts on *core/admin/mailu/ui/templates/domain/details.html*

core/admin/mailu/models.py
```diff
@@ -343,6 +344,53 @@ def check_mx(self):
...
+        except dns.exception.DNSException:
+            return False
+
+
+    def check_dmarc_report(self):
+        """ checks if DMARC Report record for domain points to mailu host """
+        app_config_domain = app.config['DOMAIN']
+        try:
+            return any(
+                rset.to_text() == '"v=DMARC1"'
```
ErwanMAS, line 387: @Riscue missing here the semi-column , must be `'"V=DMARC1;"'`

Riscue: Thanks, I checked it with the `startswith` method, it seems to be more accurate this way.

core/admin/mailu/models.py
```diff
@@ -343,6 +344,53 @@ def check_mx(self):
...
+            )
+        except dns.exception.DNSException:
+            return False
+
+
+    def check_dmarc(self):
+        """ checks if DMARC record for domain points to mailu host """
+        try:
+            return any(
+                rset.to_text().startswith('"v=DMARC1;')
```
ghostwheel42, line 375:
```suggestion
                rset.to_text().strip('"').split(';')[0].strip() == 'v=DMARC1'
```
This is more robust - also use this for the report check, as there do not need to be any other tags defined.

core/admin/mailu/models.py
```diff
@@ -343,6 +344,53 @@ def check_mx(self):
...
+        except dns.exception.DNSException:
+            return False
+
+
+    def check_dmarc_report(self):
+        """ checks if DMARC Report record for domain points to mailu host """
+        app_config_domain = app.config['DOMAIN']
+        try:
+            return any(
+                rset.to_text().startswith('"v=DMARC1;')
```
ghostwheel42, line 387:
```suggestion
                rset.to_text().strip('"').split(';')[0].strip() == 'v=DMARC1'
```

core/admin/mailu/models.py
```diff
@@ -404,6 +452,53 @@ def check_mx(self):
...
+            )
+        except dns.exception.DNSException:
+            return False
+
+
+    def check_dmarc(self):
+        """ checks if DMARC record for domain points to mailu host """
+        try:
+            return any(
+                rset.to_text().startswith('"v=DMARC1;')
```
ghostwheel42, line 483:
```suggestion
                rset.to_text().strip('"').split(';')[0].strip() == 'v=DMARC1'
```

core/admin/mailu/models.py
```diff
@@ -404,6 +452,53 @@ def check_mx(self):
...
+        except dns.exception.DNSException:
+            return False
+
+
+    def check_dmarc_report(self):
+        """ checks if DMARC Report record for domain points to mailu host """
+        app_config_domain = app.config['DOMAIN']
+        try:
+            return any(
+                rset.to_text().startswith('"v=DMARC1;')
```
ghostwheel42, line 495:
```suggestion
                rset.to_text().strip('"').split(';')[0].strip() == 'v=DMARC1'
```

ghostwheel42 (commented):
Hi @Riscue and sorry for being quiet for so long.

I think this is a good start at checking if the DNS config is like the one proposed.

Of course it is not a 100% check as some mx tools do and it will probably yield false-positives and false negatives, but the people using their own records should know what they are doing (for example when using "include:" in the spf records).

I would move the routes below the 'details' path, so instead of '/domain/<domain_name>/check/mx' I would use '/domain/details/<domain_name>/check/mx'.
This will allow to strip the '../../domain/' and '../../alternative/' from the templates.

@nextgens Do you think we need to do some kind of caching or rate limiting on the check routes to avoid abuse by doing lots of dns queries?

I think after the changes are done we could start testing this in the master branch.

nextgens (changes requested):
@ghostwheel42 I'm not worried about it. DNS has a built-in cache mechanism with TTL, this will go no further than the local resolver.

What needs to be checked is how timeouts are handled. Browsers only allow so many parallel connections... if we block awaiting for a DNS reply for several items on the same page, we may exceed the limit. It may be better to timeout earlier and return some "unknown" png faster... refreshing the page (which could also be done automatically) would then display the right results

PR #3096, opened 2023-12-19
Fix 3095: fix the form validator of forwarding addresses

### LINKED ISSUE #3095: Validation of redirect address in UI
There are several CMSs that require the forwarding address to be specified partially in uppercase. Validation does not allow this.

изображение

изображение

```
grep MAILU_VERSION docker-compose.yml mailu.env
docker-compose.yml:    image: ${DOCKER_ORG:-ghcr.io/mailu}/${DOCKER_PREFIX:-}nginx:${MAILU_VERSION:-2.0}
docker-compose.yml:    image: ${DOCKER_ORG:-ghcr.io/mailu}/${DOCKER_PREFIX:-}unbound:${MAILU_VERSION:-2.0}
docker-compose.yml:    image: ${DOCKER_ORG:-ghcr.io/mailu}/${DOCKER_PREFIX:-}admin:${MAILU_VERSION:-2.0}
docker-compose.yml:    image: ${DOCKER_ORG:-ghcr.io/mailu}/${DOCKER_PREFIX:-}dovecot:${MAILU_VERSION:-2.0}
docker-compose.yml:    image: ${DOCKER_ORG:-ghcr.io/mailu}/${DOCKER_PREFIX:-}postfix:${MAILU_VERSION:-2.0}
docker-compose.yml:    image: ${DOCKER_ORG:-ghcr.io/mailu}/${DOCKER_PREFIX:-}oletools:${MAILU_VERSION:-2.0}
docker-compose.yml:    image: ${DOCKER_ORG:-ghcr.io/mailu}/${DOCKER_PREFIX:-}rspamd:${MAILU_VERSION:-2.0}
docker-compose.yml:    image: ${DOCKER_ORG:-ghcr.io/mailu}/${DOCKER_PREFIX:-}radicale:${MAILU_VERSION:-2.0}
docker-compose.yml:    image: ${DOCKER_ORG:-ghcr.io/mailu}/${DOCKER_PREFIX:-}fetchmail:${MAILU_VERSION:-2.0}
docker-compose.yml:    image: ${DOCKER_ORG:-ghcr.io/mailu}/${DOCKER_PREFIX:-}webmail:${MAILU_VERSION:-2.0}
```

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Fix the form validator of forwarding addresses (allow uppercase).

I don't think this warrants a backport but if you think it does add the label.

### Related issue(s)
- closes #3095

## Prerequisites

### DIFF
--- a/core/admin/mailu/ui/forms.py
+++ b/core/admin/mailu/ui/forms.py
@@ -39,7 +39,7 @@ class MultipleEmailAddressesVerify(object):
 
     def __call__(self, form, field):
         pattern = re.compile(r'^([_a-z0-9\-\+]+)(\.[_a-z0-9\-\+]+)*@([a-z0-9\-]{1,}\.)*([a-z]{1,})(,([_a-z0-9\-\+]+)(\.[_a-z0-9\-\+]+)*@([a-z0-9\-]{1,}\.)*([a-z]{2,}))*$')
-        if not pattern.match(field.data.replace(" ", "")):
+        if not pattern.match(field.data.lower().replace(" ", "")):
             raise validators.ValidationError(self.message)
 
 class MultipleFoldersVerify(object):
--- a/core/admin/mailu/ui/forms.py
+++ b/core/admin/mailu/ui/forms.py
@@ -38,8 +38,8 @@ class MultipleEmailAddressesVerify(object):
         self.message = message
 
     def __call__(self, form, field):
-        pattern = re.compile(r'^([_a-z0-9\-\+]+)(\.[_a-z0-9\-\+]+)*@([a-z0-9\-]{1,}\.)*([a-z]{1,})(,([_a-z0-9\-\+]+)(\.[_a-z0-9\-\+]+)*@([a-z0-9\-]{1,}\.)*([a-z]{2,}))*$')
-        if not pattern.match(field.data.lower().replace(" ", "")):
+        pattern = re.compile(r"^(?:\s*\S+@([a-z0-9\-]{1,}\.)*([a-z]{1,})[\s|,]*)+$", re.IGNORECASE)
+        if not pattern.match(field.data.replace(" ", "")):
             raise validators.ValidationError(self.message)
 
 class MultipleFoldersVerify(object):
--- a/core/admin/mailu/ui/forms.py
+++ b/core/admin/mailu/ui/forms.py
@@ -38,7 +38,7 @@ class MultipleEmailAddressesVerify(object):
         self.message = message
 
     def __call__(self, form, field):
-        pattern = re.compile(r"^(?:\s*\S+@([a-z0-9\-]{1,}\.)*([a-z]{1,})[\s|,]*)+$", re.IGNORECASE)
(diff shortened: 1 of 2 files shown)


towncrier/newsfragments/3095.bugfix
```diff
@@ -0,0 +1 @@
+Ensure that the form validator related to forwarding addresses allows for uppercase
```
ghostwheel42, line 1: Filename should be 3096

core/admin/mailu/ui/forms.py
```diff
@@ -39,7 +39,7 @@ def __init__(self,message=_('Invalid email address.')):
 
     def __call__(self, form, field):
         pattern = re.compile(r'^([_a-z0-9\-\+]+)(\.[_a-z0-9\-\+]+)*@([a-z0-9\-]{1,}\.)*([a-z]{1,})(,([_a-z0-9\-\+]+)(\.[_a-z0-9\-\+]+)*@([a-z0-9\-]{1,}\.)*([a-z]{2,}))*$')
-        if not pattern.match(field.data.replace(" ", "")):
+        if not pattern.match(field.data.lower().replace(" ", "")):
```
ghostwheel42, line 42: I'd rather do the check in the regular expression. Removing all spaces is not exactly correct. 
```

        pattern = re.compile(r'^ *([_a-z0-9\-\+]+)(\.[_a-z0-9\-\+]+)*@([a-z0-9\-]{1,}\.)*([a-z]{1,})( *, *([_a-z0-9\-\+]+)(\.[_a-z0-9\-\+]+)*@([a-z0-9\-]{1,}\.)*([a-z]{2,}))* *$', re.I)
        if not pattern.match(field.data):
```

This will not validate idn findings of course. Only punycode encoded domains will be correct.

ghostwheel42 (dismissed):
I think the whole test should happen in the regular expression.

core/admin/mailu/ui/forms.py
nextgens: Well, adding them is not correct either; they may or may not be present... so yes we could have ``\s*`` everywhere. I am sure that none of it matters since ultimately we do store it as a CSV field where we do strip the spaces (so why would we care about them in validation?)

As for the validity of the regexp, it is obviously wrong and not just for IDNs. The localpart can contain all kind of things we currently do not allow. See https://en.wikipedia.org/wiki/Email_address#Local-part ; Here I was trying to fix the reported problem with the simplest patch possible.

ghostwheel42: Sure. I'd say adding re.I would be even simpler.

nextgens: Ok. What about 3057444 then? Or do we want to fix LOCALPART and friends too?

towncrier/newsfragments/3095.bugfix
nextgens: https://mailu.io/2.0/contributors/workflow.html#changelog says that it should be the ticket number... why do you want it to be the PR number?

ghostwheel42 (approved):

PR #3102, opened 2023-12-21
Prevent SMTP smuggling

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Prevent SMTP smuggling; see https://www.mail-archive.com/postfix-users@postfix.org/msg100901.html and https://sec-consult.com/blog/detail/smtp-smuggling-spoofing-e-mails-worldwide/

## Prerequisites


ghostwheel42 (approved):

OdyX:
Could this be released as stable? :pray:

PR #3150, opened 2024-01-26
Add trailing semicolon for DMARC authorisation record

su-ex:
## What type of PR?

enhancement, bug-fix

## What does this PR do?

Add trailing semicolon for DMARC authorisation records, which seems to be necessary: https://stackoverflow.com/a/72463456
I ran into this when testing my domain with internet.nl.

## Prerequisites


ghostwheel42 (approved):
Good catch. LGTM

nextgens:
Hmm, we need a towncrier entry, especially if we backport it

nextgens (approved):

Issue #3162, opened 2024-02-12
snappymail integration broken with TLS_FLAVOR=notls in master


acran:
- Version: `master`

## Description
Using snappymail as webmail is broken in latest master. Trying to login to the webmail results in a redirect loop.
Even if logged in sending new mails fails. Also managing sieve rules is not possible due to a server error.

## Replication Steps
1. Create a clean new mailu instance using stable 2.0 images using snappymail as webmail (image `ghcr.io/mailu/webmail:2.0`)
2. Login to webmail and verify everthing works as expected
3. Change webmail image from 2.0 tag to master (i.e. use image `ghcr.io/mailu/webmail:master`)
4. Try to login to webmail again and observe described errors

## Possible Fix
I narrowed the origin of the regression down to f143aa3dc8d8ac6d9b8ca34d13b22d90318c4523: changing `secure` back to `0` in `default.json` fixes the described issues:
~~~diff
diff --git a/webmails/snappymail/defaults/default.json b/webmails/snappymail/defaults/default.json
index 12252aee..3d8efaa2 100644
--- a/webmails/snappymail/defaults/default.json
+++ b/webmails/snappymail/defaults/default.json
@@ -3,7 +3,7 @@
     "IMAP": {
         "host": "{{ FRONT_ADDRESS }}",
         "port": 10143,
-        "secure": 2,
+        "secure": 0,
         "shortLogin": false,
         "ssl": {
             "verify_peer": false,
@@ -20,7 +20,7 @@
     "SMTP": {
         "host": "{{ FRONT_ADDRESS }}",
         "port": 10025,
-        "secure": 2,
+        "secure": 0,
         "shortLogin": false,
         "ssl": {
             "verify_peer": false,
@@ -37,7 +37,7 @@
     "Sieve": {
         "host": "{{ FRONT_ADDRESS }}",
         "port": 14190,
-        "type": 2,
+        "secure": 0,
         "shortLogin": false,
         "ssl": {
             "verify_peer": false,
~~~

But I am not sure about the implications of this change.

nextgens:
The problem is that you are mixing master and 2.0; If you want to run master, re-run setup and use the master version of all images, that will fix the problem you are encountering.

acran:
nope, thats not it. I tested this with all latest images.

Just to be sure, I just retried it with a fresh config from the setup wizard: https://setup.mailu.io/master/setup/38bd828c-130e-4e91-bd11-867d0ddfd86b

<details>
[...]

PR #3163, opened 2024-02-12
Add impersonation feature for admins to webmail

acran:
## What type of PR?

Feature

## What does this PR do?

This adds a button to the user list for admins to open the webmailer as another user.

### Related issue(s)
- Discussion #3106 

## Prerequisites

- [ ] ~In case of feature or enhancement: documentation updated accordingly~

### DIFF
--- a/core/admin/mailu/internal/views/auth.py
+++ b/core/admin/mailu/internal/views/auth.py
@@ -73,7 +73,11 @@ def user_authentication():
         and flask_login.current_user.enabled):
         response = flask.Response()
         original_email = flask_login.current_user.get_id()
-        email = flask.session.get('webmail_user_email', original_email)
+
+        email = flask.request.headers.get('X-User-Email', original_email)
+        if (email != original_email and not utils.is_owner(original_email, email)):
+            return flask.abort(403)
+
         response.headers["X-User"] = models.IdnaEmail.process_bind_param(flask_login, email, "")
         response.headers["X-User-Token"] = utils.gen_temp_token(email, flask.session)
         response.headers["X-Original-User"] = models.IdnaEmail.process_bind_param(flask_login, original_email, "")
--- a/core/admin/mailu/ui/templates/sidebar.html
+++ b/core/admin/mailu/ui/templates/sidebar.html
@@ -96,7 +96,7 @@
       <li class="nav-header text-uppercase text-primary" role="none">{% trans %}Go to{% endtrans %}</li>
       {%- if config["WEBMAIL"] != "none" and current_user.is_authenticated %}
       <li class="nav-item" role="none">
-        <a href="{{ url_for('.user_webmail') }}" class="nav-link" role="menuitem">
+        <a href="{{ config["WEB_WEBMAIL"] }}" class="nav-link" role="menuitem">
         <i class="nav-icon far fa-envelope"></i>
         <p>{% trans %}Webmail{% endtrans %}</p>
         </a>
--- a/core/admin/mailu/ui/templates/user/list.html
+++ b/core/admin/mailu/ui/templates/user/list.html
@@ -40,7 +40,7 @@
       <a href="{{ url_for('.fetch_list', user_email=user.email) }}" title="{% trans %}Fetched accounts{% endtrans %}"><i class="fa fa-download"></i></a>&nbsp;
       {%- endif -%}
       {%- if config["WEBMAIL"] != "none" -%}
-      <a href="{{ url_for('.user_webmail', user_email=user.email) }}" title="{% trans %}Webmail{% endtrans %}" target="_blank"><i class="fa fa-envelope"></i></a>&nbsp;
+      <a href="{{ config["WEB_ADMIN"] }}/webmail-as/{{ user.email }}/" title="{% trans %}Webmail{% endtrans %}" target="_blank"><i class="fa fa-envelope"></i></a>&nbsp;
       {%- endif -%}
     </td>
     <td>{{ user }}</td>
--- a/core/admin/mailu/ui/views/users.py
+++ b/core/admin/mailu/ui/views/users.py
@@ -159,17 +159,6 @@ def user_reply(user_email):
                 flask.url_for('.user_list', domain_name=user.domain.name))
     return flask.render_template('user/reply.html', form=form, user=user)
 
-@ui.route('/user/webmail', methods=['GET'], defaults={'user_email': None})
-@ui.route('/user/webmail/<path:user_email>', methods=['GET'])
-@access.owner(models.User, 'user_email')
-def user_webmail(user_email):
-    user_email_or_current = user_email or flask_login.current_user.email
-    user = models.User.query.get(user_email_or_current) or flask.abort(404)
-
-    flask.session['webmail_user_email'] = user.get_id()
-
-    return flask.redirect(f'{app.config["WEB_WEBMAIL"]}/sso.php')
-
 
 @ui.route('/user/signup', methods=['GET', 'POST'])
 @ui.route('/user/signup/<domain_name>', methods=['GET', 'POST'])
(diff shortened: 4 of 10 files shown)


Diman0 (changes requested):
This PR looks interesting. I can imagine that many people find this an interesting feature. 

The functionality appears to work fine. When trying to visit the webmail/<user> url when logged in with a non-admin user, you receive a 403 error (As expected).

Since new elements are added to the UI (extra buttons for a user on the user list), it must be added to the documentation.
webadminstration.rst. More specifically here:
https://github.com/Mailu/Mailu/blob/efb3892b09655ee8890571ba828ad51fa16f04be/docs/webadministration.rst?plain=1#L315

When using roundcube a banner is displayed in top with the information that you are impersonating an user. However this is missing for snappymaill.
If you don't know how to add a banner for SnappyMail, then I recommend creating a discussion here asking how to do this:
https://github.com/the-djmaze/snappymail/discussions

acran:
> Since new elements are added to the UI (extra buttons for a user on the user list), it must be added to the documentation.
`webadminstration.rst`.

Thank you for the suggestion. I know added the information on the new buttons to the documentation there.
(I also amended the a previous commit to only show the webmail button only if webmail is actually enabled)

> When using roundcube a banner is displayed in top with the information that you are impersonating an user. However this is missing for snappymaill.

Yes, I had skipped snappymail since I was unable to correctly configure it for local testing (see #3162).

Having had a look again now into snappymail, it seems to me that snappymail does not provide a straight-forward API to easily add a simple banner when logged in as another user the same way as with roundcibe. So I'm a bit skeptical whether the effort required to implement this analogously to roundcube is really worth it...

@nextgens is adding the same banner to snappymail a hard requirement for this PR? Would there be maybe another way than a banner to notify the user which may be easier to implement in snappymail?

And besides this: is there anything else blocking this PR from being merged?

nextgens:
Can't we just frame the webmails when impersonation is in use?

acran:
@nextgens I (rebased and) added a PoC for a framed webmail with impersonation in a8dbc5c5a1525faf850aeb00367481933bf927d3.

But in my opinion framing does not provide the best user experience and can be confusing. Because the webmailer will use the same session cookie with and without framing. So having an open tab with the admin user's normal mailbox while opening another user's mailbox with impersonation will also switch the session in the open tab to the other user's mailbox - but without the alert.

Also, having the impersonated mailbox opened in a frame (especially when keeping the mailu admin navigation as in this PoC, but that can easily be removed, though) could imply to the user that impersonation is only active in that window...

So, I'd rather let the webmailer itself handle everything instead.

But let me know which route you want to go here, and I can update the PR accordingly.

nextgens:
bors try

acran:
> Would it be doable/better to have a separate 'virtual' path for each webmail instance? So that one could open several tabs with different inboxes at the same time without having things breaking.
> 
> Say redirect to /webmail-as/test@example.com/ and have nginx proxy that too?

I implemented this with https://github.com/Mailu/Mailu/pull/3163/commits/2eea2a44f19e18853a8a4fa861316ae93664f6ed. It is working but I'm not sure if it really is worth it because it requires a lot of additional complexity for the integration:

First, for the webmail to be available under multiple URLs we not only need to make it available with different base URLs but even under _dynamic_ base URLs to include the e-mail address.
Since the `front` nginx rewrites the request URI the original/dynamic URL gets lost. I "solved" this by passing the correct base URL as an additional `X-Remote-Base-Url` header to the webmail nginx:

https://github.com/Mailu/Mailu/blob/2eea2a44f19e18853a8a4fa861316ae93664f6ed/core/nginx/conf/nginx.conf#L205
https://github.com/Mailu/Mailu/blob/2eea2a44f19e18853a8a4fa861316ae93664f6ed/core/nginx/conf/nginx.conf#L275

This then has to be handled by nginx _as well as_ the PHP code:

https://github.com/Mailu/Mailu/blob/2eea2a44f19e18853a8a4fa861316ae93664f6ed/webmails/nginx-webmail.conf#L53
https://github.com/Mailu/Mailu/blob/2eea2a44f19e18853a8a4fa861316ae93664f6ed/webmails/roundcube/config/config.inc.php#L18

> If the path is different we may have different cookies too (so completely different parallel webmail sessions).

Yes, having different path for each mailbox this is possible by using `proxy_cookie_path` to limit the cookies set by the webmail to the proxied path in `front`.
But this may cause unexpected issues if the path overlap, e.g., with `WEB_WEBMAIL=/`.

Another problem was that `auth_request` does not allow to use variables in the auth URL. So to pass the impersonated user e-mail from the request I again had to add an additional `X-User-Email` header:

https://github.com/Mailu/Mailu/blob/2eea2a44f19e18853a8a4fa861316ae93664f6ed/core/nginx/conf/nginx.conf#L261-L262
https://github.com/Mailu/Mailu/blob/2eea2a44f19e18853a8a4fa861316ae93664f6ed/core/nginx/conf/nginx.conf#L329-L332
https://github.com/Mailu/Mailu/blob/2eea2a44f19e18853a8a4fa861316ae93664f6ed/core/admin/mailu/internal/views/auth.py#L77

All of this already adds much complexity. Additionally it probably also still requires some refactoring on how the valid webmail token(s) are stored in the session. Because up until now there is always only one valid token in the session and previously we differentiated the user (original user or impersonated user) by storing the e-mail in the session as well. But now there could be multiple valid e-mail/token combinations.
Should we re-use the same token in the session for all e-mails? Or should we rather use a new token when impersonating? I guess, both would work...

If this solution should be used, I can try to polish up the commits again, so what do you think?

nextgens:
> @nextgens after thinking about it a bit more: having a dedicated proxied path for impersonated webmail is actually quite neat and provides the best UX be even allowing to have two different mailboxes open side by side in the same browser session. Implementing that cleanly though still requires some refactoring of the admin service in how the tokens are stored in the session. I can take a shot at it if wanted.
> 
> The issue with conflicting cookies in case when `WEB_WEBMAIL` overlaps with the impersonating path can not really be resolved though (it would probably required some kind of cookie-rewriting middleware, which would blow up complexity...).
> 

I am not sure I am following regarding overlapping paths; We could use [proxy_cookie_path](
http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cookie_path) to ensure we maintain independent cookies with the same names but different paths, no?

> So it stand to reason whether the added complexity is really worth the result since it'll probably still be a rarely used feature...
> 
> How best to proceed here?

I am not sure either

acran:
> I am not sure I am following regarding overlapping paths; We could use [proxy_cookie_path](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cookie_path) to ensure we maintain independent cookies with the same names but different paths, no?

Yes, this works and is utilized by this PR, see e.g.
https://github.com/Mailu/Mailu/blob/2eea2a44f19e18853a8a4fa861316ae93664f6ed/core/nginx/conf/nginx.conf#L274-L277

But there is an edge case with `WEB_WEBMAIL=/` and e.g. `WEB_ADMIN=/admin`. With that the regular webmail will be available at path `/` while an impersonated webmail is at path `/admin/webmail-as/$user_email`. Since the latter is a subpath of the former the browser can mix up the session cookies for the normal and the impersonated webmails.

With something like `WEB_WEBMAIL=/webmail` and `WEB_ADMIN=/admin` this works without issues, though.

nextgens:
I don't see that as a problem: we won't be backporting the feature and personally I am okay with removing ``WEB_*`` variables in the name of simplicity for next release.

We are already instructing those that do run reverse proxies to forward **everything** for Mailu's FQDNs to Mailu.

acran:
> personally I am okay with removing `WEB_*` variables in the name of simplicity for next release.

Is this already on the roadmap? Because I see these config options as a valuable feature and would want to see them gone - especially not for this :sweat_smile: 

> I don't see that as a problem

Aside from this single edge case me neither. Even keeping the `WEB_*` variables this could be "fixed" by adding a warning to the documentation or maybe a sanity check into the admin scripts...

So anyway, should I then proceed with the necessary changes to cleanup the storage of tokens in the session so we can bring this PR to a merge?

PR #3184, opened 2024-03-07
Remove redundant variable assignment

strugee:
## What type of PR?

Bugfix

## What does this PR do?

See diff; this variable is set again two lines down

### Related issue(s)

None

## Prerequisites


nextgens (approved):
LGTM

PR #3194, opened 2024-03-16
Fix 3113

### LINKED ISSUE #3113: Wrong schema for certain api endpoints in swagger docs
## Environment & Version

- `Docker Compose version v2.21.0`
- Version: `2.0.36`

## Description
I am currently trying out the mailu rest API and I found that the generated swagger.json file doesn't seem to be correct. For the endpoint `GET /alias/destination/{domain}` the response type is of the type

```json
"schema": {
    "$ref": "#/definitions/Alias"
}
```

For the endpoint `GET /alias/` the response type is:
```json
"schema": {
    "type": "array",
    "items": {
        "$ref": "#/definitions/Alias"
    }
}
```

Clearly both api endpoints can return multiple aliases. Still only `GET /alias/` (per definition) returns an array and `GET /alias/destination/{domain}` should only return a single alias. This is of course not the case as both return arrays.

I assume the get function lacks the decorator `@alias.doc('list_alias')`:

https://github.com/Mailu/Mailu/blob/cc59be41e15c085270b52f49739398236c131827/core/admin/mailu/api/v1/alias.py#L112-L117

## Replication Steps
1. Open the swagger docs at `yourMailuInstance/api/v1/swagger.json`
2. Check the return types of certain api endpoints

## Observed behaviour
There are wrong types for different api endpoints in the swagger docs. Specifically the `/alias/destination/{domain}` endpoint defines a `#/definitions/Alias` instead of an array of said type.

## Expected behaviour
The `/alias/destination/{domain}` endpoint should define an array return type. This is especially necessary for code generators that generate api clients based on the swagger documentation.

## Logs
No logs needed

Diman0:
## What type of PR?

bug-fix

## What does this PR do?
Fixes swaggerui documentation of all RESTful api end points. The API documentation should now be valid for each endpoint. 

### Related issue(s)
- close #3113 

## Prerequisites

### DIFF
--- a/.github/workflows/build_test_deploy.yml
+++ b/.github/workflows/build_test_deploy.yml
@@ -418,7 +418,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target: ["core", "fetchmail", "filters", "webmail", "webdav"]
+        target: ["api", "core", "fetchmail", "filters", "webmail", "webdav"]
         time: ["2"]
         include:
           - target: "filters"
--- a/.github/workflows/build_test_deploy.yml
+++ b/.github/workflows/build_test_deploy.yml
@@ -418,7 +418,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target: ["api", "core", "fetchmail", "filters", "webmail", "webdav"]
+        target: ["core", "fetchmail", "filters", "webmail", "webdav"]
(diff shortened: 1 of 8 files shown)


Diman0:
bors retry

Diman0:
bors retry

nextgens (approved):
LGTM

image

PR #3197, opened 2024-03-18
Address CVE-2024-23829 (CVE for aiohttp)

Diman0:
## What type of PR?

security update

## What does this PR do?

Updates library to patch CVE-2024-23829.

## Prerequisites


nextgens (approved):
It's podop, not directly exposed that requires it... not something we need to be concerned about IMHO

PR #3200, opened 2024-03-20
Fix highligting disabled users

nwinkelstraeter:
This just changes the class on the rows of disabled users from `warning` to `bg-warning` . As warning is not available in AdminLTE 3 disabled users where not highlighted anymore-

image

## What type of PR?
bug-fix

### Related issue(s)
#3166

### DIFF
--- a/core/admin/mailu/ui/templates/user/list.html
+++ b/core/admin/mailu/ui/templates/user/list.html
@@ -28,7 +28,7 @@
 </thead>
 <tbody>
   {%- for user in domain.users %}
-  <tr{% if not user.enabled %} class="warning"{% endif %}>
+  <tr{% if not user.enabled %} class="text-muted"{% endif %}>
     <td>
       <a href="{{ url_for('.user_edit', user_email=user.email) }}" title="{% trans %}Edit{% endtrans %}"><i class="fas fa-pencil-alt"></i></a>&nbsp;
     </td>


nextgens:
Can we get them greyed out instead? with some light-grey

nwinkelstraeter:
There is `bg-light`, which I think is too light
image

And `bg-gray`, which is maybe a little too dark but from the two I would prefer this one
image

Looking at the examples adding the `disabled` class would probably look good, but for me this does not do anything.
https://adminlte.io/themes/v3/pages/UI/general.html

nextgens:
@Diman0 @ghostwheel42 thoughts?

I guess the alternative is to add a column for it... and change the default sort keys.

ghostwheel42:
Maybe there's a style for strikethrough?

nwinkelstraeter:
There does not seem to be a pre-made style for strikethrough. I found `text-muted` which also looks good I think.
image

If we want strikethrough we need to add some custom stlyes. I tried doing something similar to what adminlte already has for todo-lists. To make this pretty I would also need to add same classes to the columns so i do not have to strikethrough all of them.
image

```css
tr.done {
  color:#697582;
  font-weight:500;
  text-decoration:line-through
}
tr.done .badge {
  background-color:#adb5bd!important
}
```

nextgens:
I think I like ``text-muted`` too

nextgens (approved):
LGTM

ghostwheel42 (approved):
LGTM

PR #3290, opened 2024-06-14
Support multiple "internal" subnets

lel-amri:
## What type of PR?

Enhancement

## What does this PR do?

It adds support for multiple "internal" subnets as defined in configurations SUBNET and SUBNET6.

## TODO

- [ ] Test it
- [ ] Document the new syntax

### DIFF
--- a/core/admin/mailu/internal/views/dovecot.py
+++ b/core/admin/mailu/internal/views/dovecot.py
@@ -11,9 +11,9 @@ import sqlalchemy.exc
 def dovecot_passdb_dict(user_email):
     user = models.User.query.get(user_email) or flask.abort(404)
     allow_nets = []
-    allow_nets.append(app.config["SUBNET"])
+    allow_nets.extend(app.config["SUBNET"].split(","))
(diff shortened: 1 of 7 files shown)


nextgens:
I am not following, what is the use case for this as opposed to using multiple RELAYNETS (which is supported)?

lel-amri:
Hello @nextgens,
I don't know of the details of the internal use of SUBNET/RELAYNETS, but here is what I "naïvely" think when reading [the documentation for the configuration](https://github.com/Mailu/Mailu/blob/c5d246e08c160064a45b3ab91d38985405085124/docs/configuration.rst): `SUBNET`/`SUBNET6` is for the sub-networks of the components of Mailu, while `RELAYNETS` is only useful for the SMTP server.
In my case, I'm running Mailu on K8S and have a dis-joint IPv6-only sub-network for pods, theoretically making ::/0 the only subnet that is sure to cover all pods CIDRS.

nextgens:
We are not going to merge this until we have settled what we want to do ipv6-wise.

PR #3307, opened 2024-06-22
ports not in PORTS should not be bound

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Ensure we only bind ports that do feature in PORTS.
Previously we would bind ports 110 and 143 even though we shouldn't have.

## Prerequisites


nextgens:
bors try

nextgens:
bors try

nextgens:
bors try

core/admin/mailu/models.py
```diff
@@ -259,16 +259,16 @@ def dns_autoconfig(self):
         """ return list of auto configuration records (RFC6186) """
         hostname = app.config['HOSTNAME']
         protocols = [
-            ('imap', 143, 20),
-            ('pop3', 110, 20),
-            ('submission', 587, 20),
+            ('imap', 143, 20 if '143' in app.config['PORTS'].split(',') else 0),
```
ghostwheel42, line 262: I think you should normalize ports before (line 161) like this
`ports = {port.strip() for port in app.config['PORTS'].split(",")}`
And use `if '143' in ports else 0` here and below.

We really should start normalizing the options at a central place (in the Config class).

core/nginx/dovecot/proxy.conf
```diff
@@ -148,17 +144,12 @@ service lmtp {
...
+    port = {% if PORT_587 %}587{% else %}0{% endif %}
 {%- if PROXY_PROTOCOL_587 %}
     haproxy = yes
-{% endif %}
-{%- else %}
-# if the section is unset the port is bound anyways
-   port = 0
 {% endif %}
    }
-{%- if TLS_465 and PORT_465 %}
```
ghostwheel42, line 161: I think it is okay to disallow disabling the necessary ports, but maybe it should throw an error, so people know we do not support disabling the ports by design (and use a firewall ar change the Dockerfile to do so if they really want to).

core/admin/mailu/models.py
nextgens: done in cdbfa4ee01e87ed02e9e0e5e129a6b19c8901f20

nextgens:
bors try

core/admin/mailu/models.py
```diff
@@ -257,18 +257,19 @@ def dns_dmarc_report(self):
...
+            ('submission', 587, 20 if '587' in ports else 0),
         ]
         if app.config['TLS_FLAVOR'] != 'notls':
             protocols.extend([
-                ('autodiscover', 443, 10),
+                ('autodiscover', 443, 10 if '443' in ports else 0),
                 ('submissions', 465, 10),
                 ('imaps', 993, 10),
-                ('pop3s', 995, 10),
+                ('pop3s', 995, 10 if '995' in ports else 0),
```
ghostwheel42, line 272: I think there is an `if prio` missing in the loop below (line 278)
`for proto, port, prio in protocols if prio`
for this to work.

ghostwheel42 (dismissed):

core/admin/mailu/models.py
nextgens: I don't get why: we always define ``prio`` and that part of the code has not changed

ghostwheel42: What do you want to achive by setting prio to 0?
I think it should lead to the DNS records not being generated. This can be done by add inf "if prio".
Why should we add DNS records for disabled services like pop3 with prio 0?

nextgens: I meant to disable it... but apparently the right way to do that is to change the target. I'll update the PR.

nextgens:
bors try

ghostwheel42 (approved):
LGTM

Issue #2050, opened 2021-11-10
rfc8463: DKIM with Ed25519 keys


nextgens:
It would be great if Mailu could double-sign using ed25519 keys as described in [RFC8463](https://datatracker.ietf.org/doc/html/rfc8463).

Most of the code is already there: rspamd can do it, we just need to figure out how it would work UI-wise.

nextgens:
We've talked about it yesterday at the dev meeting; The consensus was to create a new entry in the menubar that would read "Keys"... that would lead to a new view displaying a table with all ARC/DKIM keys available for signing.

We've also talked about whether it would make sense to have start/expiry dates there and UI-wise we thought it would be easier to have a simpler create/promote/revoke lifecycle instead.

Each domain (including those not handled by the local Mailu instance) could have multiple keys associated (including of different types). All promoted keys would be used for signing (dual-signing as required).

PR #3350, opened 2024-07-22
Feature: dkim for alternative domains

### LINKED ISSUE #1519: Sent emails from alternate domains don't support DKIM signing
Additionaly I'd love headers sent from alternate domains to only mention the alternate domain, with no mention of the "original" domain. My use is for alternate domains to just merge incoming emails.

I know postfix allows that, but do we want this feature available in mailu ?

I'd be happy with a workaround. For instance, the admin UI blocks creating domains and alternates with similar names, but I haven't tested yet what happens if I just alter tables in the db.

Jumper78:
## What type of PR?

feature

#### use same DKIM key of main domain for signing

Instead of dealing with key creation for each alternative domain, this implementation of the solution uses one key for all domains, the main domain and all alternative domains. Upon Rspamd requesting the DKIM key of a domain, it is not only checked if the domain is in the list of main domains, it also checked if it part of the alternative domains. If it is in this list, it sends the DKIM key of the connected main domain together with the name of the alternative domain.

#### show needed entries in the domain detailed view of the main domain

To make it easier for the admin to create the DKIM and DMARC entries (and the MX and SPF entries) for the alternative domains, we go through all alternative domains and print the entries.

### missing (and currently not planned to be added)

The zonefile at the top of the detail page will still only cover the main domain.

### Related issue(s)
- DKIM signing of the alternative domains is a requested feature; it closes #1519
- it keeps the original file based handling of DKIM keys; it does not implement #2952

## Prerequisites

### DIFF
--- a/core/admin/mailu/internal/views/rspamd.py
+++ b/core/admin/mailu/internal/views/rspamd.py
@@ -24,6 +24,15 @@ def rspamd_dkim_key(domain_name):
                     'selector': flask.current_app.config.get('DKIM_SELECTOR', 'dkim'),
                 }
             )
+    elif domain := models.Alternative.query.get(domain_name):
+        if key := domain.domain.dkim_key:
+            selectors.append(
+                {
+                    'domain'  : domain.name,
+                    'key'     : key.decode('utf8'),
+                    'selector': flask.current_app.config.get('DKIM_SELECTOR', 'dkim'),
+                }
+            )
     return flask.jsonify({'data': {'selectors': selectors}})
 
 @internal.route("/rspamd/local_domains", methods=['GET'])
--- a/core/admin/mailu/models.py
+++ b/core/admin/mailu/models.py
@@ -355,6 +355,54 @@ class Alternative(Base):
     domain = db.relationship(Domain,
         backref=db.backref('alternatives', cascade='all, delete-orphan'))
 
+    @property
+    def dns_dkim(self):
+        """ return DKIM record for domain """
+        if self.domain.dkim_key:
+            selector = app.config['DKIM_SELECTOR']
+            return f'{selector}._domainkey.{self.name}. 600 IN TXT "v=DKIM1; k=rsa; p={self.domain.dkim_publickey}"'
+
+    @cached_property
+    def dns_dmarc(self):
+        """ return DMARC record for domain """
+        if self.domain.dkim_key:
+            domain = app.config['DOMAIN']
+            rua = app.config['DMARC_RUA']
+            rua = f' rua=mailto:{rua}@{domain};' if rua else ''
+            ruf = app.config['DMARC_RUF']
+            ruf = f' ruf=mailto:{ruf}@{domain};' if ruf else ''
+            return f'_dmarc.{self.name}. 600 IN TXT "v=DMARC1; p=reject;{rua}{ruf} adkim=s; aspf=s"'
+
+    @cached_property
+    def dns_dmarc_report(self):
+        """ return DMARC report record for mailu server """
+        if self.domain.dkim_key:
+            domain = app.config['DOMAIN']
+            return f'{self.name}._report._dmarc.{domain}. 600 IN TXT "v=DMARC1;"'
+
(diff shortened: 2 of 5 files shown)


Jumper78:
screenshot-alternative-domain-details
This is how the part for the alternative domains looks like on the domain details page.

nextgens:
The hunk at ``core/admin/mailu/internal/views/rspamd.py`` looks reasonable to me... and is all that's required to get the feature to work. I'm not sure we want the UI bits until the keys can be changed (which would make this a much bigger PR).

@Diman0 thoughts?

Jumper78:
Thank you very much for the comment! Maybe a brief comment on what I thought when I wrote the UI change:
If the key for the alternative domain is the same as the key of the main domain, there is no way to change it separately. The "Regenerate Keys" button at the top of the page is sufficient. This is why (in my thinking) the UI change, which is only _showing_ the DNS entries for alternative domains, is adequate and should go with the change in `rspam.py`.
If you have in mind to have the alternative domain keys changeable separately from the other key, we should think of a completely different way to implement the feature "DKMI keys for alternative domains". Then the change in `rspam.py` should not be implemented at all.
I also thought about implementing separate keys for all alternative domains, so one individual one per domain. The implementation effort would be significantly higher, it would be more complicated to use DKMI and I currently see no advantage to have separate keys… except the fact that having one key per domain _feels_ like the “correct way” to implement it. In my opinion, if your server gets compromised, you will have to change all of the keys at the same time anyway and there is no advantage to have individual keys.
Let me know what you think about it.

ser:
IMHO it's a reasonable solution to quickly move things forward, as currently it's impossible to send any mail from alternative domains due to DMARC enforcements on receiving sides.

nextgens (approved):
LGTM

nextgens:
We can do it afterwards.

IMHO you want the ability to have separate keys to cover two main usecases:
- signing with more than one key (see #2050 for why that may be useful)
- allowing the import of keys from other systems (so that you don't have to wait for DNS propagation... or because you don't control the DNS zone of the said domain)

ghostwheel42 (approved):

Aaron-Ritter:
> screenshot-alternative-domain-details This is how the part for the alternative domains looks like on the domain details page.

@Jumper78 @nextgens Is the display for Alternatives implemented? Do we have to regenerate the keys for it to work?

nextgens:
@Aaron-Ritter  Do not highlight me without a good reason.

Yes and no.

There is no such convention outside of the BIND world; no one cares about the wire format. This PR does what is done for the main DKIM records; If you had suggestions to make the right time was before it got merged... and the right format was to submit a review.

Aaron-Ritter:
> @Aaron-Ritter Do not highlight me without a good reason.
> 
> Yes and no.
> 
> There is no such convention outside of the BIND world; no one cares about the wire format. This PR does what is done for the main DKIM records; If you had suggestions to make the right time was before it got merged... and the right format was to submit a review.

I've highlighted you because i was not sure if it was actually released before creating an issue for it (done now) as the latest release notes where a copy from the previous one.

The [255 TXT limit](https://support.google.com/a/answer/11613097?hl=en#:~:text=For%20TXT%20records%20that%20include,with%20a%20255%2Dcharacter%20limit.) is something i noticed while testing and added it as a note but will create a issue accordingly.

Jumper78:
Just for completeness (the question was already answered in the issue #3375): the code change was merged into the master branch. Since nearly everyone is currently still running 2024.06 branch releases (these are the latest releases at the point of writing), the feature is not visible unless you:
- either create the docker images on your own (see ...)
- or use the images created during this pull request (e.g. for the admin image you use `image: ${DOCKER_ORG:-ghcr.io/mailu}/${DOCKER_PREFIX:-}admin:pr-3350`)
- or you wait for the next major release, maybe something like 2024.XX

I explicitly do not recommend to combine the docker images from the 2024.06 releases with the `admin:pr-3350` docker image. It does work for me on my productive system ... but it is not a safe way to do it.

PR #3376, opened 2024-08-15
fix: PATCH /user will overwrite existing "forward_destination"

### LINKED ISSUE #3374: PATCH for /user does not update `forward_destination` addresses
## Environment & Version

- k8s `1.27.3`
- Version: `2024.06.3`

## Description
Using the Mailu API, when trying to update a user object, changing the `forward_destination` list of emails the change is not applied.

Current solution: delete and recreate the user.

## Replication Steps
1. Create a user through API with `POST`
```shell
data='{
  "email": "test@example.com",
  "raw_password": "s3cr3t!",
  "comment": "test API",
  "quota_bytes": 1000000000,
  "global_admin": false,
  "enabled": true,
  "change_pw_next_login": false,
  "enable_imap": true,
  "enable_pop": true,
  "allow_spoofing": true,
  "forward_enabled": true,
  "forward_destination": [
    "test2@example.com"
  ],
  "forward_keep": false,
  "reply_enabled": false,
  "displayed_name": "Test user",
  "spam_enabled": true,
  "spam_mark_as_read": true,
  "spam_threshold": 80
}'

curl -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" -X POST -d "$data" $URL/v1/user
```

1. Update the user `forward_destination` list (adding or changing entries) with `PATCH`
```shell
# elements removed for readability, but for PATCH only changed values and email are required
data='{
  "email": "test@example.com",
  "forward_enabled": true,
  "forward_destination": [
    "test2@example.com",
    "test3@example.com"
  ],
  "forward_keep": false,
}'

curl -k -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" -X PATCH -d "$data" $URL/v1/user/test@example.com
```

## Observed behaviour
The `forward_destination` field is not updated

## Expected behaviour
The `forward_destination` list is replaced with what was given in the `PATCH` body

## Logs
There is no log, even in DEBUG, except that the request was received and used a valid token.
```shell
[DEBUG] PATCH /api/v1/user/test@example.com
INFO in common: Valid API token provided by 172.18.0.10.
```

DrPsychick:
## What type of PR?

bug-fix for #3374 

### Related issue(s)
- closes #3374 

## Prerequisites

### DIFF
--- a/core/admin/mailu/api/v1/user.py
+++ b/core/admin/mailu/api/v1/user.py
@@ -242,8 +242,7 @@ class User(Resource):
         if 'forward_enabled' in data:
             user_found.forward_enabled = data['forward_enabled']
         if 'forward_destination' in data and len(data['forward_destination']) > 0:
-            if len(data['forward_destination']) == 0:
-                user_found.forward_destination = data['forward_destination']
+            user_found.forward_destination = data['forward_destination']
         if 'forward_keep' in data:
             user_found.forward_keep = data['forward_keep']
         if 'reply_enabled' in data:
--- a/towncrier/newsfragments/3374.bugfix
+++ b/towncrier/newsfragments/3374.bugfix
@@ -1,0 +1,1 @@
+Fix a bug preventing user "forward_destination" to be updated with a PATCH request


nextgens (approved):
LGTM.

Is it important enough to warrant a backport?

core/admin/mailu/api/v1/user.py
```diff
@@ -242,8 +242,7 @@ def patch(self, email):
         if 'forward_enabled' in data:
             user_found.forward_enabled = data['forward_enabled']
         if 'forward_destination' in data and len(data['forward_destination']) > 0:
```
ghostwheel42, line 244: Why do we not change forward destination when the length is zero?

Shouldn't the zero test also be removed?

Is it possible to delete the forward destination with this test in place?

DrPsychick: What purpose would have setting the `forward_destination` to an empty list? Disable forwarding could be achieved by disabling it. So the only reason I can think of right now is to remove a previously configured destination to "hide" where it was forwarded to when it was active.

Ofc, the data is not needed when forwarding is disabled, so technically, we could also remove the zero-length check, but I'm not involved enough to know why it was added.

ghostwheel42 (approved):
Okay. LGTM then.

PR #3382, opened 2024-08-25
chore: Add return type hints

DjVinnii:
## What type of PR?

Enhancement

## What does this PR do?
Add return type hints for the models. 

Type hinting in Python helps with the readability of the code as well as better IDE support. However, due to the nature of Python it is not enforcing the types as Python is not a type safe language.


DjVinnii:
Btw, I'd like to open more PRs to add type hinting but will make sure it will be in smaller chunks to keep it easier to review.

PR #3410, opened 2024-09-07
Update main.cf

DeveloperRaju23:
Got lots of delayed emails which looks like a hard bounce in newsletter software.

## What type of PR?

### Related issue(s)
- Mention an issue like: #001

## Prerequisites


nextgens:
This looks like a workaround we do not want upstream.

It's unclear why you are not fixing the newsletter software instead... or why you think the delay messages are inappropriate.

PR #3440, opened 2024-09-10
Update reverse proxy documentation for using Traefik on a different host

Diman0:
## What type of PR?

documentation

## What does this PR do?

It adds an extra section to the reverse proxy documentation. It provides an example on how to use Traefik on a different host than the host running Mailu. Now we will have documented both use cases where the reverse proxy is on the same host or a different host than Mailu.

### Related issue(s)
n/a

## Prerequisites

### DIFF
--- a/docs/reverse.rst
+++ b/docs/reverse.rst
@@ -7,10 +7,31 @@ other Web services are available on other FQDNs served from the same IP address.
 In such a configuration, one would usually run a front-end reverse proxy to serve all
 Web contents based on criteria like the requested hostname (virtual hosts).
 
+The Mailu team recommends to use Traefik as reverse proxy. This program is solely designed
+to act as reverse proxy. It is easy to configure and offers a streamlined experienced when
+used with other docker containers on the same host.
+
+When using a reverse proxy, Mailu should still be used for requesting certificates.
+This prevents duplicate certificates.
+Other reasons are:
+
+- Mailu controls the ciphers and keys used for certificates (Mailu has both RSA and ECDSA certs)
+- Mailu cerbot client is configured to prevent hitting ratelimits of the CA
+- Due to usage of proxy protocol it is less likely that Mailu becomes an open relay due to misconfiguration of special headers.
+- No special scripting is required to copy over certs from the proxy to Mailu
+
+When using Mailu with Traefik as reverse proxy. Traefik is configured to use proxy protocol for all ports, but port 80.
+Port 80 is configured to let through the letsencrypt HTTP challenges for Mailu.
+This means that if Traefik must request certificates for other services than Mailu, Traefik should use the TLS (TLS-ALPN-01)
+challenge method. The HTTP challenge (HTTP-01) cannot be used since this is already used by Mailu.
+
 .. _traefik_proxy:
 
-Traefik as reverse proxy
-------------------------
+Traefik as reverse proxy (same host)
+------------------------------------
+
+This example is for when Traefik (reverse proxy) runs on the same host as Mailu.
+It makes use of the docker configuration method for Traefik.
 
 In your docker-compose.yml, remove the `ports` section of the `front` container
 and add a section like follows:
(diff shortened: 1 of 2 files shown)


docs/reverse.rst
```diff
@@ -7,10 +7,31 @@ other Web services are available on other FQDNs served from the same IP address.
...
+
+- Mailu controls the ciphers and keys used for certificates (Mailu has both RSA and ECDSA certs)
+- Mailu cerbot client is configured to prevent hitting ratelimits of the CA
+- Due to usage of proxy protocol it is less likely that Mailu becomes an open relay due to misconfiguration of special headers.
+- No special scripting is required to copy over certs from the proxy to Mailu
+
+When using Mailu with Traefik as reverse proxy. Traefik is configured to use proxy protocol for all ports, but port 80.
+Port 80 is configured to let through the letsencrypt HTTP challenges for Mailu.
+This means that if Traefik must request certificates for other services than Mailu, Traefik should use the TLS (TLS-ALPN-01)
+challenge method. The HTTP challenge (HTTP-01) cannot be used since this is already used by Mailu.
```
nextgens, line 26: That's not true: it cannot be used for the **same** hostnames but it can be used for others

docs/reverse.rst
```diff
@@ -111,3 +132,164 @@ in mailu.env:
...
+          - mailu-pop3s
+        rule: "HostSNI(`*`)"
+        service: "mailu-pop3s"
+        tls:
+          passthrough: true
+      mailu-smtps:
+        entryPoints:
+          - mailu-smtps
+        rule: "HostSNI(`*`)"
+        service: "mailu-smtps"
```
nextgens, line 236: Let's call this 'submissions' instead of 'smtps'

docs/reverse.rst
```diff
@@ -111,3 +132,164 @@ in mailu.env:
...
+
+
+.. code-block:: yaml
+
+  http:
+    routers:
+      mailu-web:
+        entryPoints:
+          - web
+        rule: "Host(`mail.example.com) || PathPrefix(`/.well-known/acme-challenge/`))"
```
nextgens, line 195: This is too broad if we don't want to break it for other vhosts... is there a fancy way to declare all a variable containing all the hostnames we want to match so that we could reuse the same in `mailu-websecure`?

Diman0: You also used this?
https://github.com/Mailu/Mailu/blob/341afeb57d266c1d990d1c41a6c57f0d2852117a/docs/reverse.rst?plain=1#L59

With Traefik V3 you cannot define multiple hosts in Host in the rule. All matchers take a single value:
https://doc.traefik.io/traefik/v3.0/migration/v2-to-v3-details/#new-v3-syntax-notable-changes

So you would simply add a lot of `Host('yet another hostname from HOSTNAMES')` to your rule.

I don't see a problem with this. For all other vhosts Traefik can use the TLS challenge (TLS-ALPN-01) which uses port 443. So this will not lead to any issues. It's as easy as configuring:

```
certificatesResolvers:
  myresolver:
    acme:
      email: redacted@redacted.redacted
      storage: /etc/traefik/acme/acme.json
      caServer: "https://acme-v02.api.letsencrypt.org/directory"
      #caServer: "https://acme-staging-v02.api.letsencrypt.org/directory"
      tlsChallenge: {}
#      httpChallenge:
#        # used during the challenge
#        entryPoint: web
```
Below in comment you see the HTTP challenge which cannot be used because Mailu uses it. With above config, Traefik uses TLS challenge via port 443 and Mailu uses port 80 for the HTTP challenge.

I've this running on my homelab.

docs/reverse.rst
Diman0: This is due to only using PathPrefix(`/.well-known/acme-challenge/`)" for port 80.
This makes sure that all certbot calls for Mailu succeed. Otherwise you need to many rules for all hosts in HOSTNAMES in mailu.env.

Diman0:
bors try

docs/reverse.rst
nextgens: OK, let's add a comment to spell out that we need the rule to let the requests through for all hostnames Mailu will use (the OR is not an accident)

nextgens (approved):
Let's go by steps, this is an improvement regardless.

PR #3465, opened 2024-09-12
Maybe fix #3402

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Fix a potential problem with SO_REUSEADDR that may prevent admin from starting up

### Related issue(s)
- close #3402 

## Prerequisites

### DIFF
--- a/core/nginx/letsencrypt.py
+++ b/core/nginx/letsencrypt.py
@@ -6,8 +6,6 @@ import requests
 import sys
 import subprocess
 import time
-from threading import Thread
-from http.server import HTTPServer, SimpleHTTPRequestHandler
 
 log.basicConfig(stream=sys.stderr, level="WARNING")
 hostnames = ','.join(set(host.strip() for host in os.environ['HOSTNAMES'].split(',')))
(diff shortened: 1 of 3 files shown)


core/nginx/letsencrypt.py
```diff
@@ -45,33 +43,20 @@
...
-        thread.join()
+        try:
+            r = requests.get(target)
+            if r.status_code != 204:
+                log.critical(f"Can't reach {target}!, please ensure it's fixed or change the TLS_FLAVOR.")
+                time.sleep(5)
+            else:
+                break
+        except Exception as e:
+            log.error(f"Exception while fetching {target}!", exc_info = e)
```
ghostwheel42, line 59: we need some delay here:
```suggestion
            log.error(f"Exception while fetching {target}!", exc_info = e)
            time.sleep(15)
```

ghostwheel42 (dismissed):

core/nginx/letsencrypt.py
nextgens: That's c11815b3

ghostwheel42 (approved):

PR #3558, opened 2024-09-22
Only enable HARDENED_MALLOC with kernel 6.1+

nextgens:
## What type of PR?

enhancement

## What does this PR do?

Disable HARDENED_MALLOC unless the requirements are met

### Related issue(s)
- #3398

## Prerequisites

### DIFF
--- a/core/base/libs/socrate/socrate/system.py
+++ b/core/base/libs/socrate/socrate/system.py
@@ -58,6 +58,11 @@ class LogFilter(object):
         self.stream.flush()
 
 def _is_compatible_with_hardened_malloc():
+    with open('/proc/version', 'r') as f:
+        major, minor = f.readline().split()[2].split('.')[:2]
+        if int(major) < 6 or int(major) == 6 and int(minor) < 1:
+            return False
+
     with open('/proc/cpuinfo', 'r') as f:
         lines = f.readlines()
         for line in lines:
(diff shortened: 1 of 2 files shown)


core/base/libs/socrate/socrate/system.py
```diff
@@ -58,6 +58,11 @@ def flush(self):
         self.stream.flush()
 
 def _is_compatible_with_hardened_malloc():
+    with open('/proc/version', 'r') as f:
```
DjVinnii, line 61: It is also possible to get the kernel version from the[ `platform` module](https://docs.python.org/3/library/platform.html#platform.release).

DjVinnii: Same goes for the checking the CPU features on line 66:

```python
import platform

cpu_info = platform.processor()

if 'avx2' not in cpu_info:
    return False
if 'lrcpc' not in cpu_info:
    return False
```

nextgens:
```
root@a1-full:/mailu# docker compose exec admin python3
Python 3.12.3 (main, Apr 18 2024, 07:52:31) [GCC 13.2.1 20240309] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import platform
>>> 'lrcpc' in platform.processor()
False
>>> 
root@a1-full:/mailu# grep -c1 lrcpc /proc/cpuinfo 
4
```

DjVinnii: Oh whoops. That is indeed not the case, my bad. However `platform.release()` will work for the kernel version.

Diman0 (approved):
lgtm

PR #3599, opened 2024-09-29
fail2ban: improve limiter regex, add failed logins regex

spomata:
## What type of PR?

Minor FAQ update

## What does this PR do?

Improving fail2ban FAQ section. Namely, to:

- improve existing regex for rate limiter, matching the approach used for the failregex example of mailu-front
- add a new regex to match failed logins independent of rate-limiting.

### Related issue(s)
- None

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -671,6 +671,7 @@ Please note that the provided example will block the subnet from sending any ema
   # Fail2Ban configuration file
   [Definition]
   failregex = ^\s?\S+ mailu\-admin\[\d+\]: \[\S+ \S+\] WARNING in limiter: Authentication attempt from <HOST>(?: for (?:[^ ]+@[^ ]+))? has been rate-limited\.$
+              ^\s?\S+ mailu\-admin\[\d+\]: \[\S+ \S+\] INFO in base: Login attempt for:(?: [^ ]+@[^ ]+) from: <HOST>\/None: failed: badauth:.*$
   ignoreregex =
   journalmatch = CONTAINER_TAG=mailu-admin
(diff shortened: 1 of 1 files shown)


nextgens (changes requested):
> match failed logins independent of rate-limiting

That's exactly what we don't want!

The rate limiter deduplicates attempts, ensuring that when a user changes their password their account doesn't get rate/limited or banned while they update the config of their MUAs.

spomata:
> That's exactly what we don't want!
> 
> The rate limiter deduplicates attempts, ensuring that when a user change their password their account doesn't get rate/limited or banned while they update the config of their MUAs.

I struggle to see how a string of botched login attempts in a `findtime` time span is compatible with some benign activity.

nextgens:
The average user has several devices... Let's assume phone, tablet and laptop. Each of these is likely to have a fat client installed that will poll every 10 minutes or so. Most will establish parallel sessions to speed things up ; I believe that the default Apple client uses 7 or so.

Do the maths and you will realise that it is fairly easy to get banned when they change their password (as they are unlikely to have all devices handy and update them concurrently) unless the server deduplicates attempts. Mailu's rate limiter does it the smart way: it will catch attackers with very few false positives.

In an ideal world users would read the documentation and use a token per fat client. These are high entropy (not bruteforceable) and are therefore exempt from rate limits.

spomata:
My use case is completely different hence why the proposed regex.

In any case fail2ban configuration is strictly subjective so feel free to salvage the first commit which adds no extra restrictions (simply a rewrite) or drop this PR altogether.

PR #3655, opened 2024-11-18
feat(webmail): customize php memory limit

GoldenBookCover:
## What type of PR?

Feature: Customize php setting: `memory_limit` for webmail container.

### Related issue(s)

When there are a large number of emails, the webmail container fails to boot because of insufficent memory.

## Prerequisites

Add a new variable `WEBMAIL_MEMORY_LIMIT` in `mailu.env`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ pip-selfcheck.json
 /.idea
 /.vscode
 qemu-arm-static
+/build/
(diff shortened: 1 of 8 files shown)


GoldenBookCover:
OK, Let me fix that.

GoldenBookCover:
I have updated the default value and the template.

Issue #3643, opened 2024-11-12
Running into issues when running traefik already on a host and trying to follow new proxy setup


moekify:
## Environment & Version

- `docker compose version`
- Version: `2024.06.24`

If you are not using docker compose do not file any new issue here.
Kubernetes related issues belong to https://github.com/Mailu/helm-charts/issues
If you are not using docker compose or kubernetes, create a new thread on user support in [disussions](https://github.com/Mailu/Mailu/discussions/categories/user-support).
Non-bug reports (or bug reports that do not follow the template) are moved to [disussions](https://github.com/Mailu/Mailu/discussions).

## Description

I'm trying to setup mailu `2024.06` using a traefik instance that is defined in a different docker-compose file as it is already on my server and handles a lot of other services that are needing certificates etc. supplied via traffic. 

I tried to already change that traefik is using the tls challenge instead of http to retrieve certificates, however, mailu still fails to get certificates, presumably because it cannot reach the acme challenge. 

I saw additional documentation written here: https://github.com/Mailu/Mailu/pull/3440, but think this still doesn't really cover it. It's unfortunately not obvious what has to be done to get it to work. Appreciate any help

Issue #3052, opened 2023-11-14
Reverse proxy documentation: PROXY_PROTOCOL setting is wrong


liquidat:
## Environment & Version

- `docker compose version`: v2.20.2
- Version: `2.0`

## Description

At least for me, Traefik v2 as a TCP reverse proxy to IMAP and SMTP only work when `PROXY_PROTOCOL` is set to `all`.
However, the [documentation](https://mailu.io/master/reverse.html) says `PROXY_PROTOCOL` has to be set to `all-but-http`.  But with that configuration, I am not able to make a successful connection.

## Replication Steps

1. Set up mailu with Traefik as reverse proxy according to the [documentation](https://mailu.io/master/reverse.html). In my case, certificates are provided via mailu certdumper
2. Ensure that TCP ports are also proxied, like port 993 is listened by Traefik, and imap is listening on 20993
3. Try to connect to imaps from the outside via `openssl s_client -crlf -connect firun.bayz.de:993`
4. Observe the error messages

## Observed behaviour

A connection is not possible, the SSL handshake fails.

## Expected behaviour

If I change the `PROXY_PROTOCOL` setting to `all`, the connection gets through.

## Logs

With `all-but-http` I get various errors:

client:
```
CONNECTED(00000003)
001E966D237F0000:error:0A000126:SSL routines:ssl3_read_n:unexpected eof while reading:ssl/record/rec_layer_s3.c:303:
---
no peer certificate available
---
No client certificate CA names sent
---
SSL handshake has read 0 bytes and written 324 bytes
Verification: OK
---
New, (NONE), Cipher is (NONE)
This TLS version forbids renegotiation.
Compression: NONE
Expansion: NONE
No ALPN negotiated
Early data was not sent
Verify return code: 0 (ok)
---
```

Traefik:
```
traefik       | time="2023-11-14T00:40:54Z" level=debug msg="Handling TCP connection from 192.168.0.1:35256 to firun.bayz.de:20993"
traefik       | time="2023-11-14T00:40:54Z" level=debug msg="Dial with lookup to address firun.bayz.de:20993"
traefik       | time="2023-11-14T00:40:54Z" level=debug msg="Error while setting TCP connection deadline: set tcp 192.168.0.2:41770: use of closed network connection"
traefik       | time="2023-11-14T00:41:30Z" level=debug msg="Authentication failed" middlewareName=auth@file middlewareType=BasicAuth
```

front:
```
[...]

Issue #2700, opened 2023-03-14
[Traefik] All containers except front/nginx are not showing the correct IP address


sephentos:
### Environment

- [X] docker compose
- [ ] kubernetes
- [ ] docker swarm

### Version

- Version: `1.9`

## Description

I noticed that I suddenly no longer had access to IMAP via Thunderbird. In the logs I then saw that the IP address `172.18.0.4` was affected by the rate limit and could no longer log in. So far so good. Just some random bots tried their luck.

However, this is not my client IP address but, I assume, the IP from which Traefik accesses IMAP/SMTP. Anyway, my WAN IP is different.
So everyone currently can lockout literaly everyone by sending a specific amount of requests.

Oddly enough, the logs from mailu-front (NGINX) show the correct IP address.
Not surprising given the following settings for NGINX:

```
REAL_IP_HEADER=X-Real-Ip
REAL_IP_FROM=172.18.0.4
```

But how can I make sure that the other containers get the right IP address?

Traefik has host access to all ports and apparently it works for HTTPS since mailu-front shows the correct IP address in all logs.

```docker-compose.yml (traefik)```
```
networks:
  traefik:
    external: true

[...]
  traefik:
    image: "traefik:latest"
    container_name: traefik
    ports:
      - target: 80
        published: 80
        protocol: tcp
        mode: host
      - target: 443
        published: 443
        protocol: tcp
        mode: host
      - target: 25
        published: 25
        protocol: tcp
        mode: host
      - target: 143
        published: 143
        protocol: tcp
        mode: host
      - target: 465
        published: 465
        protocol: tcp
        mode: host
      - target: 587
        published: 587
        protocol: tcp
        mode: host
      - target: 993
        published: 993
        protocol: tcp
        mode: host
```

`traefik conf`
```
  smtp:
    address: :25
  smtps:
    address: :465
  submission:
    address: :587
  imap:
    address: :143
  imaps:
    address: :993
```

```docker-compose.yml (Mailu)```
```
networks:
  default:
    driver: bridge
    ipam:
      driver: default
      config:
        - subnet: 192.168.203.0/24
  traefik:
    external: true

[...]
  front:
    image: ${DOCKER_ORG:-mailu}/${DOCKER_PREFIX:-}nginx:${MAILU_VERSION:-1.9}
    restart: always
    env_file: mailu.env
    logging:
      driver: json-file
    volumes:
      - "/d/mailu/data/certs:/certs"
[...]

PR #3684, opened 2024-12-17
feat: added Caddy proxy setup instructions for Wordpress integration

ksylvan:
### CHANGES

- Added Caddy as a reverse proxy documentation section.
- Provided setup for serving Wordpress alongside Mailu.
- Detailed creating a standalone docker-compose file for Caddy.
- Included instructions to modify `docker-compose.yml` for Caddy usage.
- Explained SSL certificate handling with Let's Encrypt and Caddy.

## What type of PR?

documentation

## What does this PR do?

This PR adds comprehensive documentation for using Caddy as a reverse proxy with Mailu, providing an alternative to the existing Traefik setup. The new documentation explains how to configure Caddy to handle web-facing applications while allowing Mailu to manage mail-related services.

This also avoids the complexity of managing proxy-protocol when using the Traefik as a proxy setup.

### Related issue(s)

- #3643 
- #3052 
- #2700 

## Prerequisites

N/A

### DIFF
--- a/docs/reverse.rst
+++ b/docs/reverse.rst
@@ -111,3 +111,87 @@ in mailu.env:
   WEBROOT_REDIRECT=/sso/login
 
 Using the above configuration, Traefik will proxy all the traffic related to Mailu's FQDNs without requiring duplicate certificates.
+
+.. _caddy_proxy:
+
+Caddy as reverse proxy
+----------------------
+
+This setup is suitable for when you want to serve Wordpress or other web-facing applications (on port 80 and 443)
+on your domains from the same machine as Mailu.
+
+This setup is simpler than using Traefik, and only deals with the reverse proxying of the web-facing applications,
+leaving MailU to serve the ports related to sendmail and its related services (25, 110, 143, 465, 587, 993, 995, 4190).
+
+.. note::
+  In the following example, I assume that MailU is set up in `/opt/mailu` and Wordpress in `/opt/wordpress`.
+
+In your docker-compose.yml, remove 80 and 443 from the `ports` section of the `front` container.
+
+Create a standalone docker-compose file for Caddy, in `/opt/caddy` with something similar to the following:
+
+.. code-block:: yaml
+
+  services:
+  caddy:
+    image: caddy:2
+    restart: unless-stopped
+    volumes:
+      - "/opt/caddy/conf:/etc/caddy"
+      - "/opt/caddy/data:/data"
+      - /opt/mailu/certs/letsencrypt:/certs:ro
+    ports:
+      - "80:80"
+      - "443:443"
+      - "443:443/udp"
+    dns:
+      - 192.168.203.254
+    networks:
+      - mailu_default
+      - wordpress_default
+  networks:
+  mailu_default:
+    external: true
+  wordpress_default:
+    external: true
+
+The volumes section is important, as it mounts the Let's Encrypt certificates into the container (readonly), so
+that Caddy can use them to serve the SSL certificates. Also note that the caddy container is on the same network
+as the front and wordpress containers, so that it can reach them.
+
+In `/opt/caddy/conf/`, create a file called `Caddyfile` with the following contents:
+
+.. code-block:: text
+
+  # Handle ACME challenge requests before applying the HTTP to HTTPS redirection
+  http:// {
+    handle /.well-known/acme-challenge/* {
+      reverse_proxy front:80
+    }
+    # Redirect all other HTTP traffic to HTTPS
+    handle {
+      redir https://{host}{uri}
+    }
+  }
+
+  mail.your_domain_name.com {
+    tls /certs/live/mailu/fullchain.pem /certs/live/mailu/privkey.pem
+    reverse_proxy front:80
+  }
+  www.your_domain_name.com {
+    tls /certs/live/mailu/fullchain.pem /certs/live/mailu/privkey.pem
+    reverse_proxy wordpress:80
+  }
+  your_domain_name.com {
+    tls /certs/live/mailu/fullchain.pem /certs/live/mailu/privkey.pem
+    reverse_proxy wordpress:80
+  }
+
+The first few lines are for handling the ACME challenge requests before applying the HTTP to HTTPS redirection.
+Note that this setup leaves the responsibility of managing and refreshing the Let's Encrypt certificates to Mailu,
+and Caddy will only be responsible for serving the SSL certificates given to it.
+
+If you are serving multiple web-facing applications using different names, you can add thoose names
+to the `Caddyfile` as well, reverse-proxying them to the correct container and port for each application,
+and make sure that the container is accessible from the Caddy container by adding its network to the
+networks section of the Caddy container's Docker Compose file.


nextgens (dismissed):
- This breaks Mailu's rate-limiter as it masquerades the IP address of the request. At the very least `X-Forwarded-For` should be set and Mailu should be configured to accept it (and that's where things stop being simple).
- This will bind a specific type of SSL certificate (rather than using RSA/ECDSA as appropriate based on client capabilities)
- the software is called Mailu, nothing else.

More generally speaking, I'm not sure we want this at all. There are good reasons why we want SSL termination to be done by Mailu.

ksylvan:
> * This breaks Mailu's rate-limiter as it masquerades the IP address of the request. At the very least `X-Forwarded-For` should be set and Mailu should be configured to accept it (and that's where things stop being simple).
> * This will bind a specific type of SSL certificate (rather than using RSA/ECDSA as appropriate based on client capabilities)
> * the software is called Mailu, nothing else.
> 
> More generally speaking, I'm not sure we want this at all. There are good reasons why we want SSL termination to be done by Mailu.

The rate limiter seems to be working as expected. I see lines like this in my logs:

```
admin-1            | [2024-12-18 17:26:41,467] WARNING in limiter: Authentication attempt from 87.120.93.10 has been rate-limited.
```

I'll see what I can figure out about the SSL RSA/ECDSA issue. The SSL certs are still handled by Mailu and served on every port except for https, which

nextgens:
> The rate limiter seems to be working as expected.

Try logging in on the web-interface using different, non-existing usernames up to ``AUTH_RATELIMIT_IP`` times (5/hour by default) through the proxy. Odds are this will prevent any login from any IP as the proxy itself will end up in the rate-limit.

It can't work any other way if you have not set ``REAL_IP_FROM``

ksylvan:
Thanks, @nextgens for that feedback.

I changed the configuration slightly, and changed the documentation to add `REAL_IP_FROM` and `REAL_IP_HEADER` and now I see this when attempting to do a bad login:

```
admin-1            | [2024-12-18 20:24:04,489] INFO in base: Login attempt for: foo@gmail.com/sso/https from: 67.180.199.110/None: failed: badauth: f4861e
```

instead of:

```
admin-1            | [2024-12-18 20:28:49,776] INFO in base: Login attempt for: foo@gmail.com/sso/https from: 192.168.203.8/47138: failed: badauth: f4861e
```

ksylvan:
I updated the setup and it addresses most of the questions.

> Hmm, it's still missing source port information (see the None?).

Yes, and I'm not exactly sure where to fix that. When I do an invalid login, I see this in the admin logs:

```text
admin-1            | [2024-12-22 00:02:03,317] INFO in base: Login attempt for: foo@gmail.com/sso/https from: 67.180.199.110/None: failed: badauth: 67b797
```

In my caddy debug logs, I see the following headers are being set. You can see the `X-Real-IP` and `X-Real-Port` headers being sent.

```json
{
  "remote_ip": "67.180.199.110",
  "remote_port": "55970",
  "client_ip": "67.180.199.110",
  "method": "GET",
  "host": "mail.example.com",
  "headers": {
    "Sec-Fetch-Mode": [
      "cors"
    ],
    "Sec-Fetch-Dest": [
      "font"
    ],
    "X-Real-Ip": [
      "67.180.199.110"
    ],
    "X-Forwarded-Proto": [
      "https"
    ],
    "X-Forwarded-Host": [
      "mail.example.com"
    ],
    "X-Forwarded-For": [
      "67.180.199.110"
    ],
    "User-Agent": [
      "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36"
    ],
    "X-Real-Port": [
      "55970"
    ]
  },
  "tls": {
    "resumed": false,
    "version": 772,
    "cipher_suite": 4865,
    "proto": "h2",
    "server_name": "mail.example.com"
  }
}
```

> As you have now realised, it's a lot trickier to get right than it looks.

Yes. 💯 I appreciate the work and thought that has gone into Mailu.

> I have also noticed that what you are proposing will only suggest a single FQDN. This will break client auto configuration, mta-sts, ...

You can do multiple FQDN and mta-sts and auto-configuration (my new update shows all that).

One nit: The default MailU configuration (with or without using an external reversproxy) serves the `apple.mobileconfig` with a `text/html` content type. That causes the page to simply be displayed and no auto-config happens.

I fixed that in my `Caddyfile` by having it change the content type to be:

```
Content-Type application/x-apple-aspen-config
```

And now it behaves as you would expect:

image

nextgens:
> One nit: The default MailU configuration (with or without using an external reversproxy) serves the `apple.mobileconfig` with a `text/html` content type. That causes the page to simply be displayed and no auto-config happens.
> 
> I fixed that in my `Caddyfile` by having it change the content type to be:
> 
> ```
> Content-Type application/x-apple-aspen-config
> ```
>

[...]

Good catch!

It's the wrong place where to fix it though. I've submitted #3691

nextgens:
> > Hmm, it's still missing source port information (see the None?).
> 
> Yes, and I'm not exactly sure where to fix that. When I do an invalid login, I see this in the admin logs:

I suggest you read http://nginx.org/en/docs/http/ngx_http_realip_module.html

nextgens:
`X-Forwarded-For` needs to be formatted as per [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986).

`X-Real-Port` and `X-Real-IP` are internal things that we use that caddy should not set

ghostwheel42:
I think adding [X-Forwarded-For](https://github.com/Mailu/Mailu/pull/3684/files#diff-a9bb069c3b466936ce6291330aa58be5be8a4962da65a0df87565062868aea5bR214) will help.
Keeping X-Real-IP and X-Real-Port doesn't hurt, but the backends need to be configured accordingly (like the nginx serving the wordpress for example).

nextgens:
agreed

bors try

digitalsigi:
Hi,
thx for your work, it at least pointed me in the right direction. It tried your solution, but I get this error:

```
Attaching to caddy-1
Error response from daemon: failed to set up container networking: driver failed programming external connectivity on endpoint caddy-caddy-1 (0245ec740b66f281b9b499197ea7ce3efe901d533333a559ab033b659f03a1eb): Bind for xxx.xxx.xxx.xxx:80 failed: port is already allocated
```
I am using the unmodified docker-compose.yml from mailu configuration. Did I miss something?

digitalsigi:
> Hi, thx for your work, it at least pointed me in the right direction. It tried your solution, but I get this error:
> 
> ```
> Attaching to caddy-1
> Error response from daemon: failed to set up container networking: driver failed programming external connectivity on endpoint caddy-caddy-1 (0245ec740b66f281b9b499197ea7ce3efe901d533333a559ab033b659f03a1eb): Bind for xxx.xxx.xxx.xxx:80 failed: port is already allocated
> ```
> 
> I am using the unmodified docker-compose.yml from mailu configuration. Did I miss something?

OK, fixed that. I overlooked to remove lines referencing port 80 and 443.
But still not working. It looks that I now cannot reach webmail and there is a problem with letsencrypt for wordpress container.

PR #3699, opened 2024-12-30
Add the mariadb connector as per 3449

### LINKED ISSUE #3449: Admin container fails to connect to external MariaDB database
## Environment & Version

- Docker Compose version `v2.12.0`
- Version: `2024.06`

## Description

After upgrading to the latest 2024.06 version the admin container is unable to connect to an external MariaDB database due to incompatible collation.

## Replication Steps

- Use MariaDB as an external database backend. In my case version 11.5.2.
- Update mailu to the latest version.

## Observed behaviour

The admin container failes with 
```
sqlalchemy.exc.DatabaseError: (mysql.connector.errors.DatabaseError) 1273 (HY000): Unknown collation: 'utf8mb4_0900_ai_ci'
```

## Expected behaviour

Admin container should be able to connect to either mysql or mariadb backends.

## Cause
MariaDB has no support for utf8mb4_0900_ai_ci which is the new default since MySQL version 8.0. In the current sqlalchemy version shipped with mailu, the mysqlconnector sets utf8mb4_0900_ai_ci as the collation to use when connecting. This causes all MariaDB connections to fail.

## Workaround
Overwrite the default collation for mysqlconnector in the connection string in mailu.env:
```
SQLALCHEMY_DATABASE_URI=mysql+mysqlconnector://<user>:<passwd>@<host>/<database>?collation=utf8mb4_unicode_ci
```

## Possible Solution
Include the "mariadbconnector" module for sqlalchemy and use it in the sqlalchemy connection string. Currently it is not included with the version shipped by mailu.

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Add the mariadb connector as per #3449.

MariaDB has no support for utf8mb4_0900_ai_ci which is the new default since MySQL version 8.0. In the current sqlalchemy version shipped with mailu, the mysqlconnector sets utf8mb4_0900_ai_ci as the collation to use when connecting. This causes all MariaDB connections to fail.

To fix the issue, either use the right connector or ensure it's configured with the right collation:
```
SQLALCHEMY_DATABASE_URI=mysql+mysqlconnector://<user>:<passwd>@<host>/<database>?collation=utf8mb4_unicode_ci
```

### Related issue(s)
- closes #3449

## Prerequisites

### DIFF
--- a/core/base/Dockerfile
+++ b/core/base/Dockerfile
@@ -53,7 +53,7 @@ ENV \
 
 RUN set -euxo pipefail \
   ;  machine="$(uname -m)" \
-  ; deps="build-base gcc libffi-dev python3-dev" \
+  ; deps="build-base gcc libffi-dev python3-dev mariadb-dev" \
   ; [[ "${machine}" != x86_64 ]] && \
       deps="${deps} cargo git libretls-dev mariadb-connector-c-dev postgresql-dev" \
   ; apk add --virtual .build-deps ${deps} \
--- a/core/base/requirements-dev.txt
+++ b/core/base/requirements-dev.txt
@@ -22,6 +22,7 @@ gunicorn
 idna
 itsdangerous
 limits
+mariadb
 marshmallow
 marshmallow-sqlalchemy
 mysql-connector-python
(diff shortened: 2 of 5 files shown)


nextgens:
bors try

ghostwheel42 (approved):

NIronwolf:
As also suggested in 3449, can this get added to the documentation to use it?  I had to keep searching around until I finally found that issue.  I can confirm just updating my mailu.env to mysql+mariadbconnector did resolve my admin container not starting up.  Many thanks for the fix, but I think it should get included in the docs in the Database config.

Thank you!

PR #3729, opened 2025-01-18
Allow setting collation via env variable and add uvloop

Grennith:
## What type of PR?

Enhancement / Bugfix

## What does this PR do?

As of https://github.com/Mailu/Mailu/pull/3701, the collation shall be set by overwriting `SQLALCHEMY_DATABASE_URI` to contain the collation of the related DB. However, this is currently not possible in the Helm chart of Mailu at all. It's statically set there and would also require not setting DB_NAME etc. to not have it overwritten, see https://github.com/Mailu/Mailu/blob/master/core/admin/mailu/configuration.py#L144

Additionally, uvloop is added to the prod requirements of which postfix-mta-sts-resolver makes use of.

### Related issue(s)
- Mention an issue like: #3449 

## Prerequisites

- [ x ] In case of feature or enhancement: documentation updated accordingly

### DIFF
--- a/core/base/requirements-prod.txt
+++ b/core/base/requirements-prod.txt
@@ -78,6 +78,7 @@ tabulate==0.9.0
 tenacity==8.2.3
 typing_extensions==4.11.0
 urllib3==2.2.1
+uvloop==0.21.0
 validators==0.28.1
 visitor==0.1.3
 vobject==0.9.7
--- a/core/admin/mailu/configuration.py
+++ b/core/admin/mailu/configuration.py
@@ -25,6 +25,7 @@ DEFAULT_CONFIG = {
     'DB_PW': None,
     'DB_HOST': 'database',
     'DB_NAME': 'mailu',
+    'DB_APPENDIX': '',
     'SQLITE_DATABASE_FILE': 'data/main.db',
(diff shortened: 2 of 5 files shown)


core/base/requirements-prod.txt
```diff
@@ -78,6 +78,7 @@ tabulate==0.9.0
 tenacity==8.2.3
 typing_extensions==4.11.0
 urllib3==2.2.1
+uvloop==0.21.0
```
nextgens, line 81: If you add it there you need to add it to requirements-dev.txt too

We didn't add it as we don't need great performance out of mta-sts-resolver

nextgens (changes requested):

core/base/requirements-prod.txt
Grennith: Added

nextgens:
Please add a towncrier entry describing the changes

Otherwise it looks good to me.

Grennith:
Hi @nextgens , thanks for the review. I've added a brief towncrier entry as requested

nextgens (approved):
LGTM

ghostwheel42 (approved):

Issue #3673, opened 2024-12-06
Default Docker setup disables ClamAV updates


homeserverhq:
Update: The title was changed to reflect the actual problem. This is not an urgent security issue, but it is something that should be addressed nonetheless.

Review this Issue: https://github.com/Mailu/Mailu/issues/3612, as it describes the exact thing that I just ran across. The issue was closed because it was believed to be an upstream ClamAV issue, but it is in fact due to Mailu's setup. I would have responded to that issue, but the thread was already closed out.

The problem is very simple. The clamav/clamav-debian:1.2.3-45 container operates as user clamav (1000:1000):

```
root@8648903efaef:/# cat /etc/passwd

root:x:0:0:root:/root:/bin/bash
daemon:x:1:1:daemon:/usr/sbin:/usr/sbin/nologin
bin:x:2:2:bin:/bin:/usr/sbin/nologin
sys:x:3:3:sys:/dev:/usr/sbin/nologin
sync:x:4:65534:sync:/bin:/bin/sync
games:x:5:60:games:/usr/games:/usr/sbin/nologin
man:x:6:12:man:/var/cache/man:/usr/sbin/nologin
lp:x:7:7:lp:/var/spool/lpd:/usr/sbin/nologin
mail:x:8:8:mail:/var/mail:/usr/sbin/nologin
news:x:9:9:news:/var/spool/news:/usr/sbin/nologin
uucp:x:10:10:uucp:/var/spool/uucp:/usr/sbin/nologin
proxy:x:13:13:proxy:/bin:/usr/sbin/nologin
www-data:x:33:33:www-data:/var/www:/usr/sbin/nologin
backup:x:34:34:backup:/var/backups:/usr/sbin/nologin
list:x:38:38:Mailing List Manager:/var/list:/usr/sbin/nologin
irc:x:39:39:ircd:/run/ircd:/usr/sbin/nologin
gnats:x:41:41:Gnats Bug-Reporting System (admin):/var/lib/gnats:/usr/sbin/nologin
nobody:x:65534:65534:nobody:/nonexistent:/usr/sbin/nologin
_apt:x:100:65534::/nonexistent:/usr/sbin/nologin
clamav:x:1000:1000:Clam Antivirus:/var/lib/clamav:/bin/false
```

The volume is mounted as a subdirectory of filter in the docker-compose (as per https://setup.mailu.io/):

```
    volumes:
      - "/mailu/filter/clamav:/var/lib/clamav"
```

[...]

PR #3735, opened 2025-01-26
Fix `clamav` path to allow for updates

nazar-pc:
## What type of PR?

Bug-fix, documentation

### Related issue(s)
Closes https://github.com/Mailu/Mailu/issues/3673 by placing ClamAV files under `mailu/clamav` instead of `mailu/filter/clamav`.

Users will want to change their `docker-compose.yml` accordingly and remove `mailu/filter/clamav` after upgrade.

I also updated ClamAV version while I was at it (I didn't find any breaking changes in the changelog), though [the latest release is not pushed to this image yet](https://github.com/Cisco-Talos/clamav/issues/1442). Also I'm wondering why is it using exact version instead of `:1` or `:1.4` for example, but decided to not change that to make it less controversial.

## Prerequisites
This will not affect existing setups, though it would be nice to notify users somehow.

### DIFF
--- a/setup/flavors/compose/docker-compose.yml
+++ b/setup/flavors/compose/docker-compose.yml
@@ -217,7 +217,7 @@ services:
   # Optional services
   {% if antivirus_enabled %}
   antivirus:
-    image: clamav/clamav-debian:1.2.3-45
+    image: clamav/clamav-debian:1.4.1-16
     restart: always
     logging:
       driver: journald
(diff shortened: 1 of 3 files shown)


nextgens (dismissed):
Add a towncrier entry please; pinning 1.4 is fine by me

nazar-pc:
There is also `/mailu/filter/clamav` path in `tests/compose/filters/docker-compose.yml`, but I didn't touch it to not break tests that I do not understand.

nextgens:
Feel free to update the path there too.

Is there anything in the documentation that also need updating?

nazar-pc:
I searched the repo and didn't find other mentions, so not exactly sure. It only seems to be used for ClamAV database updates and probably some other internal stuff, so shouldn't really make any difference beyond path to anyone.

nextgens (approved):
LGTM. Thanks!

ghostwheel42 (approved):
LGTM. Maybe we can add a rm in the other container to automatically remove the now unused folder. 
Moving it is probably not necessary, correct?

PR #3739, opened 2025-01-29
Fix the webpack build due to dependOn issue

### LINKED ISSUE #3738: Broken asset build due to webpack dependency management
Hello everyone (long time no see 🤗 ),

We just upgraded to Mailu 2024.06.32, skipping a couple versions, and find ourselves with a bug due to the admin generated JS failing to load. The issue originates in the assets code, which builds a broken output JS at the moment. We are able to reproduce on the latest Mailu code, possibly due to a recent Webpack minor or patch upgrade.

I tries to isolate the issue very specifically, and it looks like an interaction between static assets (`mailu.png` in the case of Mailu) and inter-entry dependencies (between `app.js` and `vendor.js`). I was able to reproduce based on the minimal following example : 

```js
const path = require('path');
const webpack = require('webpack');

module.exports = {
    mode: 'production',
    entry: {
        app: {
            import: ['./app.js', './logo.png'],
            dependOn: 'vendor',
        },
        vendor: './vendor.js',
    },
    output: {
        path: path.resolve(__dirname, 'static/'),
        filename: '[name].js',
        assetModuleFilename: '[name][ext]',
    },
    module: {
        rules: [
            {
                test: /\.js$/,
                use: ['babel-loader'],
            },
            {
                test: /\.png$/,
                type: 'asset/resource',
            },
        ],
    },
};
```

All `app.js`, `vendor.js` and `logo.png` are empty files, here are the installed versions, aligned with Mailu :

```
test-webpack@1.0.0
│
├── babel-loader@8.4.1
│
└── webpack-cli@4.10.0
```

When loading the assets in the following page :

```html
<!doctype html>
<html>
  <head>
    <meta charset="utf8">
  </head>
  <body>
    <script src="/static/vendor.js"></script>
    <script src="/static/app.js"></script>
  </body>
</html>
```

This results in :

```
Uncaught TypeError: e[r] is not a function
    o http://localhost:8000/static/vendor.js:1
    k http://localhost:8000/static/app.js:1
    <anonymous> http://localhost:8000/static/app.js:1
    e http://localhost:8000/static/vendor.js:1
    <anonymous> http://localhost:8000/static/app.js:1
[vendor.js:1:114](http://localhost:8000/static/vendor.js)
```

The error is located in Webpack dynamic loading code, which kicks in when enabling dependencies. It tries to load a chunk for the static asset, which does not exist because it is generated by the `asset/resource` rule that generates a separate static file.

I'll try and fix this today and provide a pull request.

kaiyou:
## What type of PR?

Bug-fix

## What does this PR do?

As stated in Webpack documentation, when using multiple entrypoints and dependencies, it is recommended to export the runtime as a single separated chunk.

See: https://webpack.js.org/guides/code-splitting/#entry-dependencies

### Related issue(s)
- closes #3738 

## Prerequisites

- [x] Unless it's docs or a minor change: add [changelog](https://mailu.io/master/contributors/workflow.html#changelog) entry file. -> this is a minor build change

### DIFF
--- a/core/admin/assets/webpack.config.js
+++ b/core/admin/assets/webpack.config.js
@@ -60,6 +60,7 @@ module.exports = {
     ],
     optimization: {
         minimize: true,
+        runtimeChunk: 'single',
         minimizer: [
             new terse(),
             new mini({
--- a/core/admin/assets/webpack.config.js
+++ b/core/admin/assets/webpack.config.js
@@ -9,10 +9,11 @@ module.exports = {
     mode: 'production',
     entry: {
         app: {
-            import: ['./assets/app.css', './assets/mailu.png', './assets/app.js'],
+            import: ['./assets/app.css', './assets/app.js'],
             dependOn: 'vendor',
         },
         vendor: './assets/vendor.js',
+        logo: './assets/mailu.png',
     },
     output: {
         path: path.resolve(__dirname, 'static/'),
@@ -60,7 +61,6 @@ module.exports = {
     ],
     optimization: {
         minimize: true,
-        runtimeChunk: 'single',
         minimizer: [
             new terse(),
             new mini({


kaiyou:
After proper testing, I was only partially right: this fixes the error but not the bug, since the code simplydoes not load anymore. Tagging as WIP, I'll provide an updated PR asap.

kaiyou:
After I found the issue was not the way we load the runtime, but a change related to Webpack runtime itself, as explained in the related issue, I updated the PR, which now does fix the bug.

nextgens (approved):
LGTM.

Welcome back!

PR #3758, opened 2025-02-14
added idna function to perform puny encoding on IDN domains

### LINKED ISSUE #3743: DKIM signing from internationalized domains (IDN-domains)
**Expected behaviour:**
When creating a domain with an internationalized domain name (IDN) (see RFC 5890) in Mailu and adding a DKIM-key, send out emails should get a valid DKIM signature recognised by all other mail providers.

**Observed behaviour:**
Authentification results vary, depending on the mail service on the receiving end. Gmail and Mailu accept the DKIM signature as `dkim=pass`, web.de and Microsoft based mail-services reject the signature (stating `dkim=fail`) according to the lines in the e-mail headers. The same result is also reflected in the DMARC reports sent out by the services the next day.

**Initial root-cause analysis:**
E-mail headers on the receiving side state the following:
web.de: `Dkim-Signature: v=1; a=rsa-sha256; c=relaxed/relaxed; d=test2.xxxxxÃ¼xxxx.de; s=dkim; t=...`
Gmail: `Dkim-Signature: v=1; a=rsa-sha256; c=relaxed/relaxed; d=test2.xxxxxÃ¼xxxx.de; s=dkim; t=...`
The "Ã¼" in the domain name should probably be a German umlaut "ü".

According to section 3.5 of [RFC 6376](https://datatracker.ietf.org/doc/html/rfc6376#section-3.5), the "Internationalized domain names MUST be encoded as A-labels, as described in Section 2.3 of [RFC5890]". A-labels according to [RFC 5890](https://datatracker.ietf.org/doc/html/rfc5890#section-2.3) are ASCII-compatible encodings starting with `xn--`. Probably, some of the email providers are stricter (web.de, Microsoft) than others (Gmail, Mailu) when it comes to handling IDN domains in DKIM signatures. Anyway, I think the correct `d=`-entry in the DKIM-signature header field should look like:
`Dkim-Signature: v=1; a=rsa-sha256; c=relaxed/relaxed; d=test2.xn--xxxxxxxxx-xxa.de; s=dkim; t=...`

Since DKIM signing in Mailu is done by the [DKIM signing module of rspamd](https://rspamd.com/doc/modules/dkim_signing.html), I am still not sure where to find the problem in the code:
- is it a problem with mailu handing over the email-domain to rspamd
- or is rspamd ignoring that IDN-domains need a special treatment.

I would be grateful for any hints.

Jumper78:
## What type of PR?

bug-fix

### Related issue(s)
- fixes issue where DKIM signatures from domains with IDN are not accepted by some mail servers: closes #3743

## Prerequisites

### DIFF
--- a/core/admin/mailu/internal/views/rspamd.py
+++ b/core/admin/mailu/internal/views/rspamd.py
@@ -2,6 +2,7 @@ from mailu import models
 from mailu.internal import internal
 
 import flask
+import idna
 
 def vault_error(*messages, status=404):
     return flask.make_response(flask.jsonify({'errors':messages}), status)
@@ -19,7 +20,7 @@ def rspamd_dkim_key(domain_name):
         if key := domain.dkim_key:
             selectors.append(
                 {
-                    'domain'  : domain.name,
+                    'domain'  : idna.encode(domain.name.lower()).decode('ascii'),
                     'key'     : key.decode('utf8'),
                     'selector': flask.current_app.config.get('DKIM_SELECTOR', 'dkim'),
                 }
@@ -28,7 +29,7 @@ def rspamd_dkim_key(domain_name):
         if key := domain.domain.dkim_key:
             selectors.append(
                 {
-                    'domain'  : domain.name,
+                    'domain'  : idna.encode(domain.name.lower()).decode('ascii'),
                     'key'     : key.decode('utf8'),
                     'selector': flask.current_app.config.get('DKIM_SELECTOR', 'dkim'),
                 }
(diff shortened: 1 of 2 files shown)


nextgens (approved):
LGTM.

I'd prefer if we had a test to go with it though...

Jumper78:
> I'd prefer if we had a test to go with it though...

A proper test would involve sending an e-mail from an IDN domain, e.g., admin@xn--4ca0bs.mailu.io, to another mail server that checks the validity of the DKIM signature. Unfortunately, Mailu itself is not picky enough. It accepts also DKIM signatures if they are not following section 3.5 of [RFC 6376](https://datatracker.ietf.org/doc/html/rfc6376#section-3.5).
If we would use a Mailu-server on the receiving end (and to my understanding your current tests are doing it that way), you can only check if the `d=`-field of the DKIM signature has a puny encoded domain in it.

ghostwheel42 (approved):

Issue #3757, opened 2025-02-14
Remove Nixspam from rspamd config


Pinkbyte:
According to https://www.nixspam.net/ their RBL is discontinued from 2025-01-16, so i suggest removing it from current Mailu docker image of rspamd

mdartmann:
The operators of nixspam have started to send out abuse requests to the infrastructure providers of people hosting rspamd versions that are still accessing nixspam, so I looked into this some more.

This rbl was removed from rspamd in https://github.com/rspamd/rspamd/pull/5300 , however, no new rspamd version has been released with this change. 

mailu uses Alpine in its `antispam` image, so I [filed a bug upstream in the `aports` repository](https://gitlab.alpinelinux.org/alpine/aports/-/issues/16915) and [already prepared a patch for them to apply for the `edge` Alpine version](https://gitlab.alpinelinux.org/mae/aports/-/commit/f0f2d33ae4eaf4ecc596f81df765538c80b2151c), so now it's up to the Alpine maintainers to resolve this.

Until Alpine resolves this, Mailu can patch the config files locally by adding a patch in the [rspamd Dockerfile](https://github.com/Mailu/Mailu/blob/master/core/rspamd/Dockerfile). I will work on that and submit a patch later today.

Borks:
How would i go about fixing this for an older version of mailu? Can I override it somehow?

nextgens:
The behaviour of those sending those abuse request is unacceptable. Lack of planning on their part does not make it an emergency for everyone else... Nor should it give an opportunity to do PR.

The correct procedure to shutdown a public RBL is documented at https://www.rfc-editor.org/rfc/rfc6471.html#section-3.4 and none of it involves (ab)using the abuse process less than a month after shutdown.

I will oppose any PR proposed by principle and encourage others to deny them the attention they seek.

packet23:
I believe the RBL can be disabled by adding the following override:

```
rbls {
  nixspam {
    disabled = true;
  }
}
```

as `overrides/rspamd/rbl.conf`.

nimaps:
@packet23 Thank you, I disabled nixspam and can confirm that I see the following line:

```
rbl.lua:1199: disable rbl "nixspam"
```

I'll see if I receive anymore complaints...

This was ridiculous and not professional at all. To discontinue the service without any planning and in a completely rushed manner is one thing but sending abuse complaints to hetzner as a means of informing users??? WTF?

Aaron-Ritter:
[...]

PR #3759, opened 2025-02-14
rspamd: Remove obsolete Nixspam RBL

mdartmann:
## What type of PR?

Bugfix / Workaround

## What does this PR do?
The Nixspam RBL has been shut down in mid-January and has been removed from rspamds since, but no release with this removal was created. Since Nixspams' provider has started sending abuse reports to the hosting providers of servers who are still issuing requests to this RBL, remove it from the default config here until Alpine does it in the upstream package.

This change should be rolled back once the upstream package was fixed.

### Related issue(s)
Closes: #3757
See-Also: https://gitlab.alpinelinux.org/alpine/aports/-/issues/16914

## Prerequisites

### DIFF
--- a/core/rspamd/Dockerfile
+++ b/core/rspamd/Dockerfile
@@ -14,6 +14,10 @@ RUN set -euxo pipefail \
 COPY conf/ /conf/
 COPY start.py /
 
+# To be removed once https://gitlab.alpinelinux.org/alpine/aports/-/issues/16914 is resolved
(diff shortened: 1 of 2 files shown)


nextgens:
I oppose to us merging this.

Both at the technical level (blindly deleting random parts of the config without ensuring that the file is what we expect is a foot-gun I want us to stay well clear off) and the moral level (see my comments on the issue).

PR #3779, opened 2025-02-20
Update Dockerfile

### LINKED ISSUE #3777: 2024.06 [front] nginx emerg: Error loading shared library gx_stream_module.so
right after docker compose upgrade from 2.0 to 2024.06, front fails to load with this error in log:

```
[emerg] 17#17: dlopen() "/var/lib/nginx/modules/ngx_stream_module.so" failed (Error loading shared library /var/lib/nginx/modules/ngx_stream_module.so: No such file or directory) in /etc/nginx/nginx.conf:9
```

I don't have a test environment to try and debug this, only a production server for myself and my friends. Cannot let them down while I investigate.
rolled back to 2.0

The image with this error is this one
```
REPOSITORY                             TAG                  IMAGE ID       CREATED         SIZE
ghcr.io/mailu/nginx                    2024.06              b1fffb6d5f72   4 days ago      285MB
```

When I load this image as is, the library indeed does not exist:
```
ls -l /var/lib/nginx/modules/

-rwxr-xr-x    1 root     root         18984 Aug 14  2024 ngx_http_brotli_filter_module.so
-rwxr-xr-x    1 root     root         14440 Aug 14  2024 ngx_http_brotli_static_module.so
-rwxr-xr-x    1 root     root         95920 Aug 14  2024 ngx_mail_module.so
```

During startup, config.py replaces /etc/nginx/nginx.conf by /conf/nginx.conf and indeed on line 9 there is `load_module "modules/ngx_mail_module.so";`
For the life of me I can't see anywhere what could be downloading this library. Therefore it's missing plain and simple.

audioscavenger:
missing nginx-mod-stream on line 20 causes container to fail to start completely as described in #3777

## What type of PR?
bug-fix

## What does this PR do?
missing nginx-mod-stream on line 20 causes container to fail to start completely as described in https://github.com/Mailu/Mailu/issues/3777

### Related issue(s)
closes #3777 

## Prerequisites
none

### DIFF
--- a/core/nginx/Dockerfile
+++ b/core/nginx/Dockerfile
@@ -17,7 +17,7 @@ ARG VERSION
 LABEL version=$VERSION
 
 RUN set -euxo pipefail \
-  ; apk add --no-cache certbot nginx nginx-mod-http-brotli nginx-mod-mail openssl dovecot-lua dovecot-pigeonhole-plugin 'dovecot-lmtpd<2.4' dovecot-pop3d dovecot-submissiond
+  ; apk add --no-cache certbot nginx nginx-mod-http-brotli nginx-mod-mail nginx-mod-stream openssl dovecot-lua dovecot-pigeonhole-plugin 'dovecot-lmtpd<2.4' dovecot-pop3d dovecot-submissiond
 
 COPY conf/ /conf/
 COPY --from=static /static/ /static/


audioscavenger:
i'm sorry, it's a PR against master as required by https://mailu.io/2024.06/contributors/workflow.html#pr-workflow why am i blocked?

audioscavenger:
okay I can see that master nginx.conf will instead reference ngx_mail_module.so so that's the file in error, got it.

Issue #3262, opened 2024-05-17
I get antivirus errors that I cannot interpret


cebor:
## Environment & Version

- `docker compose version`
- Version: `master`

If you are not using docker compose do not file any new issue here.
Kubernetes related issues belong to https://github.com/Mailu/helm-charts/issues
If you are not using docker compose or kubernetes, create a new thread on user support in [disussions](https://github.com/Mailu/Mailu/discussions/categories/user-support).
Non-bug reports (or bug reports that do not follow the template) are moved to [disussions](https://github.com/Mailu/Mailu/discussions).

## Description

Lately I have noticed that the antispam container throws strange errors. They look like in the following.

I have disabled antivirus for some time now for performance reasons and can't figure out what these errors mean. Are there possibly still a few leftovers that are interfering here?

Is there any way I can clear the cache to fix this?

```log
antispam-1  | 2024-05-17 10:55:55 #1(main) <frreun>; symcache; add_augmentation: added implied flags (010000000) for symbol FORCE_ACTION_ANTISPOOF_DMARC_ENFORCE_LOCAL as it has important augmentation
antispam-1  | 2024-05-17 10:55:55 #1(main) <frreun>; symcache; add_augmentation: added implied flags (010000000) for symbol FORCE_ACTION_ANTISPOOF_NOAUTH as it has important augmentation
antispam-1  | 2024-05-17 10:55:55 #1(main) <frreun>; symcache; add_augmentation: added implied flags (010000000) for symbol FORCE_ACTION_ANTIVIRUS_FAILED as it has important augmentation
antispam-1  | 2024-05-17 10:55:55 #1(main) <frreun>; symcache; add_augmentation: added implied flags (010000000) for symbol FORCE_ACTION_ANTIVIRUS_FLAGGED as it has important augmentation
antispam-1  | 2024-05-17 10:55:55 #1(main) <frreun>; symcache; add_augmentation: added implied flags (010000000) for symbol FORCE_ACTION_ANTISPOOF_AUTH_FAILED as it has important augmentation
antispam-1  | 2024-05-17 10:55:55 #1(main) <frreun>; symcache; add_augmentation: added implied flags (0100000000000000000) for symbol FORBIDDEN_FILE_EXTENSION as it has passthrough augmentation
antispam-1  | 2024-05-17 10:56:48 #1(main) <frreun>; symcache; load_items: cannot open file /var/lib/rspamd/symbols.cache: No such file or directory
[...]

PR #3784, opened 2025-03-04
Bugfix rspamd config errors

Jumper78:
## What type of PR?

This **bug-fix** addresses configuration errors that show up in the rspamd module.

## What does this PR do?

 When I performed a configdump in the antispam module (e.g., by executing `docker exec mailu-antispam-1 rspamadm configdump mx_check`), I received the following error messages:

```
unknown worker attribute: type; worker type: normal
unknown worker attribute: type; worker type: controller
cannot add dependency from OLETOOLS_MACRO_MRAPTOR on FORCE_ACTION_ANTIVIRUS_FLAGGED: invalid symbol types
cannot add dependency from OLETOOLS_MACRO_SUSPICIOUS on FORCE_ACTION_ANTIVIRUS_FLAGGED: invalid symbol types
cannot add dependency from OLETOOLS_FAIL on FORCE_ACTION_ANTIVIRUS_FAILED: invalid symbol types
```

### Related issue(s)
- the errors were shown as well in issue #3262.
- the bug-fix should also close some of the errors in discussion #3263

## Prerequisites

### DIFF
--- a/core/rspamd/conf/external_services.conf
+++ b/core/rspamd/conf/external_services.conf
@@ -3,9 +3,6 @@ oletools {
   # default olefy settings
   servers = "{{ OLETOOLS_ADDRESS }}:11343";
 
-  # symbol to add (add it to metric if you want non-zero weight)
-  symbol = "OLETOOLS";
-
   # needs to be set explicitly for Rspamd < 1.9.5
   scan_mime_parts = true;
   extended = true;
--- a/core/rspamd/conf/external_services.conf
+++ b/core/rspamd/conf/external_services.conf
@@ -1,7 +1,10 @@
 {% if SCAN_MACROS %}
 oletools {
   # default olefy settings
-  servers = "{{ OLETOOLS_ADDRESS }}:11343"
+  servers = "{{ OLETOOLS_ADDRESS }}:11343";
+
+  # symbol to add (add it to metric if you want non-zero weight)
+  symbol = "OLETOOLS";
 
   # needs to be set explicitly for Rspamd < 1.9.5
   scan_mime_parts = true;
--- a/core/rspamd/conf/external_services.conf
+++ b/core/rspamd/conf/external_services.conf
@@ -3,6 +3,9 @@ oletools {
   # default olefy settings
   servers = "{{ OLETOOLS_ADDRESS }}:11343";
 
+  # symbol to add (add it to metric if you want non-zero weight)
+  symbol = "OLETOOLS";
+
   # needs to be set explicitly for Rspamd < 1.9.5
   scan_mime_parts = true;
   extended = true;
--- a/core/rspamd/conf/force_actions.conf
+++ b/core/rspamd/conf/force_actions.conf
@@ -18,13 +18,13 @@ rules {
     action = "reject";
     expression = "CLAM_VIRUS | OLETOOLS_MACRO_MRAPTOR | OLETOOLS_MACRO_SUSPICIOUS";
     message = "Rejected (dangerous/malicious code detected)";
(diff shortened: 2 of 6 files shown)


Jumper78:
Short description what I changed and why I think that is correct:
- in `core/rspamd/conf/external_services.conf` there was a semi colon missing
- `core/rspamd/conf/force_actions.conf` caused the "cannot add dependency" errors
  - the `honor_action` lines make sure that these force actions do not overwrite a higher ranking action
  - with the `honor_action` lines there are no more dependency error messages
- the attribute `type` is not needed anymore in the worker configurations
  - it causes the described errors
  - the attribute is already set in a higher ranking config file (https://github.com/rspamd/rspamd/blob/22b1d712b36654514db359dbebe1d096d828fb05/conf/rspamd.conf#L67)

Jumper78:
bors try

nextgens:
bors try

Jumper78:
Unfortunately, the tests are right. I still have a problem with my bugfix.
I confirmed on my private test-server that Mailu now accepts mails with `2003x32_word_msgbox_stomped_fakecode.doc` as attachment ... and that is not good.

nextgens:
FYI the example given at https://rspamd.com/doc/modules/external_services.html#oletools-default-mode is also missing the semicolon

nextgens:
@Jumper78 can you get to the bottom of this please?

Jumper78:
@nextgens , I will try my best to find a solution.
Could you do me a favour? Could you start another `bors try`? The old logs are gone and I would like to compare the test results with the results on my private test server.

nextgens:
bors try

Jumper78:
Just a quick comment about the current status of my analysis. I changed three things in this PR:
- I deleted the lines with `type = {...}` --> these changes do not destroy the functionality
- I added a semicolon --> this change also does not destroy the functionality
- I added `honor_action` lines to `force_actions.conf` --> those are the ones leading to the file `2003x32_word_msgbox_stomped_fakecode.doc` not being blocked by the mail server

I also checked the functionality of the force actions. Unfortunately, the force actions based on the OLE-Tools seem to not work properly in the current master tree. Mails with `2003x32_word_msgbox_stomped_fakecode.doc` as attachment are rejected with `554 5.7.1 Spam message rejected\r\n` while they should be rejected with the comment `Rejected (dangerous/malicious code detected)`. So, I am still convinced that the force actions are not work as intended and that should be fixed.
I already have a solution where the force actions are working, but there are still error-messages that should not be there. I will continue working on it.

Jumper78:
Unfortunately, I am not able to get rid of all of the error messages. Two errors remain when I run `docker exec mailu-antispam-1 rspamadm configdump mx_check`, I still get:
`cannot add dependency from OLETOOLS_FAIL on FORCE_ACTION_ANTIVIRUS_FAILED: invalid symbol types`
`cannot add dependency from OLETOOLS_A on FORCE_ACTION_OLETOOLS_DANGEROUS: invalid symbol types`
But with the current solution the force_action rule OLETOOLS_DANGEROUS works properly and E-Mails detected by the rules are rejected due to force_actions and not only due to the spam score.
@nextgens, could you retrigger `bors try`?

nextgens:
bors try

Jumper78:
The build succeded and, as expected, the file 2003x32_word_msgbox_stomped_fakecode.doc was rejected with the error message:
`reply: retcode (554); Msg: b'5.7.1 Rejected (dangerous macro detected)'`

nextgens (approved):
LGTM

ghostwheel42 (approved):

PR #3790, opened 2025-03-14
Replace marshmallow missing with load_default

DjVinnii:
## What type of PR?

Enhancement

## What does this PR do?
Replace the `missing` parameter with `load_default`. This changed in version 3.13.0 of marshmallow (Mailu currently uses 3.21.2) and `missing` will be removed in marshmallow 4.

## Prerequisites


nextgens (approved):
https://marshmallow.readthedocs.io/en/stable/marshmallow.fields.html#base-field-class

PR #3887, opened 2025-08-12
Remove OCSP stapling by Nginx

### LINKED ISSUE #3886: Let's Encrypt dropped OCSP - remove INOP stapling
Let's Encrypt no longer features OCSP and therefore Let's Encrypt certs don't support stapling anymore. This triggers a warning when the frontend container Nginx is (re)started or reloaded which may cause problems when this is run as part of i.e. a Certbot post cert renew hook

Since most people use Let's Encrypt certs, please remove the stapling from the frontend container Nginx config which has been introduced with https://github.com/Mailu/Mailu/issues/1616

svoop:
Let's Encrypt has dropped OCSP support and the resulting warning may cause problems when restarting or reloading the frontend Nginx e.g. as part of a post cert update hook.

## What type of PR?

Feature removal

### Related issue(s)
closes #3886

## Prerequisites


svoop:
@nextgens No idea why my git client does this, I don't add this removal to the commit, but it keeps popping up anyways. I've force pushed a commit without from git CLI.

nextgens (approved):
LGTM

PR #3904, opened 2025-08-21
Upgrade Tika to fix CVE-2025-54988

### LINKED ISSUE #3903: fix CVE-2025-54988 / Tika XXE
The docker deployments should be safe; the containers have no access to internet nor do they have the environment variables...

The same cannot be said of helm deployments.

The fix is in Tika 3.2.2 (https://github.com/apache/tika/commit/bfee6d5569fe9197c4ea947a96e212825184ca33); it won't be upgraded to automatically as it requires a change in docker-compose.yml (running setup).

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Upgrade Tika to fix CVE-2025-54988 ; Current stable is v3.2.2 ... we pin latest as we have a poor record of keeping up with upstream and the interface we use ought to be stable.

You will need to run setup again for the fix to be applied.

This is defence in depth (very low risk on docker deployments).

### Related issue(s)
- closes #3903

## Prerequisites

### DIFF
--- a/setup/flavors/compose/docker-compose.yml
+++ b/setup/flavors/compose/docker-compose.yml
@@ -156,7 +156,7 @@ services:
 
 {% if tika_enabled %}
   fts_attachments:
-    image: apache/tika:2.9.2.1-full
+    image: apache/tika:latest-full
     hostname: tika
     logging:
       driver: journald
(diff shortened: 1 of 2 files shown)


ghostwheel42 (approved):
Using "latest" is dangerous and may break.
Maybe we need to have our own images (which are just a copy of the upstream ones) but pinned.

PR #3915, opened 2025-09-07
Fix PR-3864 (missing persistence)

DaCHack:
## What type of PR?

 Bug-fix

## What does this PR do?
Fixes missing persistence of the user Option „invisible“ from PR3864

Sorry for Spamming PRs. Just noticed that the user settings are Not persistent without adapting the data Model accordingly as well.


DaCHack:
@nextgens Unclear to me how to adapt the Database for the test. Thought that adjusting the Model in Models.py should suffice…

nextgens:
You need a migration too; see https://mailu.io/2024.06/contributors/database.html

You should really test this out locally before sending a PR for review; If you need assistance join #mailu-dev on https://app.element.io/#/room/#mailu-dev:matrix.org

DaCHack:
bors retry

DaCHack:
> You need a migration too; see https://mailu.io/2024.06/contributors/database.html
> 
> You should really test this out locally before sending a PR for review; If you need assistance join #mailu-dev on https://app.element.io/#/room/#mailu-dev:matrix.org

Added the migration. In the dev environment it works fine now. Can you please trigger the CI so I can test it with a live full stack using `${MAILU_VERSION:-pr-3915}`, please @nextgens ?

nextgens:
bors try

DaCHack:
@nextgens I noticed that the --invisible option seems to be a global option only and cannot be applied to individual polls in the fetchmailrc file. I hope I did understand this correct since documentation is very limited on this option.
Thus, I reverted the changes of the PR and introduced a switch via environment variables in fetchmail.py
I remember we did not want to introduce additional ENV, but there are no global settings in the UI which could include a switch only available to the administrator.

Suggest to run `bors try` again for me to test the ENV in a full stack again and create a documentation entry afterwards if this works.

DaCHack:
bors try

optional/fetchmail/fetchmail.py
```diff
@@ -66,10 +67,9 @@ def run(debug):
         for fetch in fetches:
             fetchmailrc = ""
             options = "options antispam 501, 504, 550, 553, 554"
-            if "FETCHMAIL_OPTIONS" in os.environ: options += f'{ os.environ["FETCHMAIL_OPTIONS"]}'
+            if "FETCHMAIL_POLL_OPTIONS" in os.environ: options += f'{ os.environ["FETCHMAIL_POLL_OPTIONS"]}'
```
ghostwheel42, line 70:
```suggestion
            if "FETCHMAIL_POLL_OPTIONS" in os.environ: options += f' {os.environ["FETCHMAIL_POLL_OPTIONS"]}'
```
Space is in wrong position.

optional/fetchmail/fetchmail.py
```diff
@@ -55,7 +55,8 @@ def fetchmail(fetchmailrc):
     with tempfile.NamedTemporaryFile() as handler:
         handler.write(fetchmailrc.encode("utf8"))
         handler.flush()
-        command = FETCHMAIL.format(shlex.quote(handler.name))
+        if "FETCHMAIL_OPTIONS" in os.environ: fetchmail_custom_options = f'{ os.environ["FETCHMAIL_OPTIONS"]}' else fetchmail_custom_options = ""
```
ghostwheel42, line 58:
```suggestion
        fetchmail_custom_options = os.environ.get("FETCHMAIL_OPTIONS", "")
```
Just shortened this a bit.

ghostwheel42 (dismissed):
Hi, I have two small changes.

ghostwheel42:
bors try

ghostwheel42 (dismissed):

DaCHack:
@ghostwheel42 Thanks. Would You do another Bord try for me please? Testing revealed a remaining reference to the old invisible flag and Makes the admin Container Crash. Fixed now

ghostwheel42:
bors try

DaCHack:
@ghostwheel42 @nextgens I tried another approach to enable both invisible mode and IMAP IDLE using mailu's fetchmail.
This approach is now working for me like a charm for 2 weeks. Feel free to test my container image on docker hub: dachack/fetchmail.

You should be able to use it with any existing configuration with the only effect that fetchmail will do the first fetch right after 20sec and then looping according to the user setting in the environment variable.

If you use it like me with the additional environment variable `FETCHMAIL_OPTIONS = --invisible --idle` then it will not poll the remote server regularly but just wait for new mail and deliver it immediately to dovecot.

Please you and the mailu team review the code, let me know in case of any suggestions, and merge it to master at your liking.

Thanks for all your hints before!

ghostwheel42:
@DaCHack What do you think about fixing it like I did in #4022?
I know this PR also adds "idle mode", but I think it needs more work.

Especially when fetches are edited at runtime fetchmail instances would need to be stopped/started.

It's probably better to do idle in a separate PR and maybe not start an instance per fetch but a single one (or two: one with, one without the invisible mode) and reconfigure those when fetches change.

nextgens:
Agreed, the right way to handle this is to have two separate instances of fetchmail, one with and one without the invisible mode.

Each user request would then be passed to the right one.

As @ghostwheel42 said, if you want IDLE support (so long running fetchmail instances) you also need to ensure that the config is reloaded when things change.

PR #3922, opened 2025-09-15
Feature: send dmarc reports

Jumper78:
## What type of PR?

Feature

## What does this PR do?

Adds automated DMARC aggregate reporting support using Rspamd 3.0+ features. This enables Mailu to automatically generate and send daily DMARC aggregate reports to domains that request them via their DMARC policy records.

Key features:
- Daily automated report generation via Alpine's periodic system (= crond)
- Configurable via DMARC_SEND_REPORTS environment variable (turning on and off)
- Uses existing Mailu variables (POSTMASTER, DOMAIN, SITENAME) for configuration
- Includes setup tool integration with new checkbox
- Supports configuration overrides via overrides/dmarc.conf
- Reports sent at 2 AM daily for previous day's email traffic

### Related issue(s)
- Discussion in #3780 

## Prerequisites

## Additional implementation notes:

- Requires Redis for DMARC data storage (already present in Mailu)
- Uses rspamadm dmarc_report command for report generation
- Reports are sent from POSTMASTER@DOMAIN to requesting domains
- Feature is disabled by default and must be explicitly enabled


Jumper78:
I found the reason why the tests did fail and hopefully corrected it.

jackmaninov:
bors retry

ghostwheel42:
bors try

core/rspamd/start.py
```diff
@@ -34,6 +34,18 @@
         pass
     log.warning("Admin is not up just yet, retrying in 1 second")
 
+# Configure DMARC reporting cron job
+if os.environ.get('DMARC_SEND_REPORTS', 'false').lower() == 'true':
+    # Create cron job for daily DMARC reporting
+    with open('/etc/periodic/daily/dmarc-reports', 'w') as f:
+        f.write('#!/bin/sh\n')
+        f.write('# Send DMARC reports for yesterday\n')
+        f.write('su rspamd -s /bin/sh -c "/usr/bin/rspamadm dmarc_report $(date --date yesterday +%Y%m%d)" >/var/log/dmarc-reports.log 2>&1\n')
```
firstred: Doesn't work (anymore?) on alpine because of the missing `--date yesterday`, but it can easily be replaced by:
```
f.write('su rspamd -s /bin/sh -c "/usr/bin/rspamadm dmarc_report $(date -d "@$(( $(date -u +%s) - 86400 ))" +%Y%m%d)" >/var/log/dmarc-reports.log 2>&1\n')
```

Jumper78: Very good catch! I will correct the code. Thank you very much!

Jumper78: Hi @firstred,

Thank you again for catching the Alpine compatibility issue! I've been testing and researching the date parameter behavior, and I'd like to propose removing the `-u` flag from the final `date` command.

## Current implementation:
```bash
date -u -d @$(($(date -u +%s)-86400)) +%Y%m%d
```

## Proposed change:
```bash
date -d @$(($(date +%s)-86400)) +%Y%m%d
```

## Reasoning:

**Rspamd explicitly uses local timezone for DMARC report keys**, not UTC. This was changed in [commit c6a5be1](https://github.com/rspamd/rspamd/commit/c6a5be1c1d08b4e4948a215a27c6406b51f48c8a) on May 8, 2022:

> [Fix] Dmarc report: Use local timezone instead of GMT

The commit changed Rspamd's internal key generation from:
```lua
local period = os.date('!%Y%m%d',  -- GMT
    task:get_date({format = 'connect', gmt = true}))
```

To:
```lua
local period = os.date('%Y%m%d',  -- Local time
    task:get_date({format = 'connect', gmt = false}))
```

This means Rspamd stores DMARC data in Redis with keys based on **local time**, not UTC. When we call `rspamadm dmarc_report 20251224`, it looks up Redis keys that were created using the container's local timezone.

## Why local time is correct:

Using local container time ensures:
- The cron execution time matches Rspamd's key generation timezone
- Reports are sent for the correct 24-hour period as stored in Redis
- No timezone mismatches between data collection and report generation

If we use UTC calculation in a container running in JST (UTC+9), executing at 02:00 JST would calculate a different date than what Rspamd used when storing the data, potentially missing reports or sending duplicates.

What do you think about this approach? Did I miss anything?

delacor:
any news here? :)

Jumper78:
> any news here? :)

From my point of view, everything is working fine. I have this version up and running on my test system sind months without problems.

nextgens (approved):
LGTM

I'd have preferred if it used a user crontab and saved the su though... but I guess that's personal taste

ghostwheel42 (approved):

PR #3927, opened 2025-09-22
Fixes around non-blocking I/O in the thread manager

### LINKED ISSUE #3852: webmail container stops responding - UnicodeDecodeError
My mailu server (docker compose setup on a small vhost) works perfectly for a couple of days, then suddenly webmail stops responding. Restarting `webmail` doesn't help, but restarting `front` does.

The logs of `front` show the following error:

```
front-1  | Exception in thread Thread-2 (forward_text_lines):
front-1  | Traceback (most recent call last):
front-1  |   File "/usr/lib/python3.12/threading.py", line 1073, in _bootstrap_inner
front-1  |     self.run()  File "/usr/lib/python3.12/threading.py", line 1010, in run
front-1  |     self._target(*self._args, **self._kwargs)
front-1  |   File "/app/venv/lib/python3.12/site-packages/socrate/system.py", line 157, in forward_text_lines
front-1  |     current_line = src.readline()
front-1  |                    ^^^^^^^^^^^^^^
front-1  |   File "<frozen codecs>", line 322, in decode
front-1  | UnicodeDecodeError: 'utf-8' codec can't decode byte 0xac in position 174: invalid start byte
```

Any help and hints how to proceed debugging the problem greatly appreciated.

sdomi:
## What type of PR?

bug-fix

## What does this PR do?

Attempts to fix an issue around blocking I/O in `start.py`. Currently, if a process crashes leaving a zombie thread, `start.py` will use 100% CPU on two threads (trying to read stdout/stderr). This can be reproduced by starting a vanilla Mailu, and running `killall dovecot` (can be outside the container); Dovecot always leaves one zombie thread on crash/exit, so the container *never* restarts. My patch fixes this.

This is related to (closes #3852), although I haven't been able to create a repro for that exact case. Nonetheless, I added `errors='replace'` to the popen invocation, as it should prevent mangling the log output. I've seen a couple more issues that are likely related, but can't find them now.

**Important note**: Despite my efforts, `mailu-front` still seems to connect to the old dovecot process, which renders the UI unusable (webmail throws me into a redirect loop). Restarting `mailu-front` frees up the defunct dovecot process and fixes the application - I suppose someone should look into detecting this condition and restarting `mailu-front` if it happens. Unfortunately, I'm not sure how to approach this.

### DIFF
--- a/core/base/libs/socrate/socrate/system.py
+++ b/core/base/libs/socrate/socrate/system.py
@@ -9,6 +9,7 @@ import socket
 import tenacity
 import subprocess
 import threading
+import time
 
 @tenacity.retry(stop=tenacity.stop_after_attempt(100),
                 wait=tenacity.wait_random(min=2, max=5))
@@ -163,12 +164,14 @@ def drop_privs_to(username='mailu'):
 def forward_text_lines(src, dst):
     while True:
         current_line = src.readline()
+        if not current_line:
+            return
         dst.write(current_line)
 
 
 # runs a process and passes its standard/error output to the standard/error output of the current python script
 def run_process_and_forward_output(cmd):
-    process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, errors='replace')
 
     stdout_thread = threading.Thread(target=forward_text_lines, args=(process.stdout, sys.stdout))
     stdout_thread.daemon = True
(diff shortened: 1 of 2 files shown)


ghostwheel42 (dismissed):
LGTM

sdomi:
are there any blockers preventing us from merging this sometime soon? :)

aviallon:
@sdomi @ghostwheel42 this fixed an issue I had for ages where this combined with another (much harder to fix) issue on my setup which forced me to manually restart the containers daily.
THANK YOU.

nextgens (approved):

nextgens (changes requested):
Please add a newsfragment so that it can be released.

sdomi:
@nextgens done - sorry for the delay :)

ghostwheel42 (approved):

PR #3938, opened 2025-10-18
Switch to clamav's base image

### LINKED ISSUE #3937: Persistence and image of clamav don't match the developers recommandation
Hi all,

in https://github.com/Mailu/Mailu/blob/0ce884df29938134e6868c8e55c1079f4aefcbbd/setup/flavors/compose/docker-compose.yml#L228-L229 persistence is enabled for the ClamAV container. According to the [ClamAV documentation](https://docs.clamav.net/manual/Installing/Docker.html#image-selection-recommendations) you should either use the _base image and add persistent storage to the container or use the normal image without persistent storage. 

The volume should be removed or the _base image selected if there is a good reason to have persistence enabled.

ref Mailu/helm-charts#496

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Use clamav's lighter base-image

### Related issue(s)
- closes #3937

## Prerequisites

### DIFF
--- a/setup/flavors/compose/docker-compose.yml
+++ b/setup/flavors/compose/docker-compose.yml
@@ -217,7 +217,7 @@ services:
   # Optional services
   {% if antivirus_enabled %}
   antivirus:
-    image: clamav/clamav-debian:1.5_base
+    image: clamav/clamav-debian:1.4_base
     restart: always
     logging:
       driver: journald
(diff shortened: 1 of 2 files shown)


WebSpider:
As mentioned [here](https://github.com/Mailu/helm-charts/pull/488#issuecomment-3416534529): Is 1.5 the best image for us to run on?

nextgens:
OK, we'll just switch to the base image then... and there is no good reason to backport it.

PR #3945, opened 2025-11-03
Implement user deletion form in edit template

cruvzadmin:
Added a form to delete a user with confirmation.

## What type of PR?

### Related issue(s)
- Mention an issue like: #001

## Prerequisites


nextgens:
Thank you but this is not something we want; we need to prevent username/email re-use.

We have an entry in the FAQ on how to really delete users if you really want to.

PR #3966, opened 2025-12-17
Decouple rspamd version from alpine

### LINKED ISSUE #3959: rspamd Update
Been using mailu privately for years now and it works well, **thank you** all that maintain it.

In recent months, the amount of spam that makes it through to my mailbox has been increasing. Taking a quick look I noticed that the "antispam" section in mailu claims that I am using rspamd 3.8.4. 

Image

This appears to be a version from February 2024 - while the current version of rpamd is 3.14.0 - https://github.com/rspamd/rspamd/tags

checking my docker compose images, I appear to be using the "newest" tag suggested by the documentation here: https://mailu.io/2024.06/setup.html

```
antispam-1          ghcr.io/mailu/rspamd                  2024.06             1352d223d224        295MB
```

Am I doing something wrong or is 3.8.4 the latest version available via mailu? If 3.8.4 is truely the latest version, could it be updated? My hope would be that newer versions would reduce the amounts of spam that make it through to my inbox.

firstred:
## What type of PR?

Feature

## What does this PR do?

It decouples the version of Rspamd from Alpine by building the package instead. The version is currently set to v3.14.2.

### Related issue(s)
- closes #3959
- #3960 

## Prerequisites

### DIFF
--- a/towncrier/newsfragments/3966.misc
+++ b/towncrier/newsfragments/3966.misc
@@ -1,0 +1,1 @@
+Upgrade rspamd to 3.14.2


nextgens:
I am sorry but no.

The way forward is to upgrade master to the newer alpine image. 3.23 has rspamd 3.14.
https://pkgs.alpinelinux.org/packages?name=rspamd&branch=v3.23&repo=&arch=x86_64&origin=&flagged=&maintainer=

Please note that upgrading will likely break dovecot unless changes are made elsewhere in the codebase.

firstred:
Thanks for your reply and all the work you've done on Mailu!

Is the difficult upgrade from dovecot 2.3 to 2.4 holding the project back?
I've been able to push it all the way to alpine v3.23.0 and PHP 8.4, except for dovecot.

I've migrated the authentication step in the dovecot container (not the proxy) from the dict driver to lua-based authentication (basically translating the calls to `{( ADMIN_ADDRESS}}:8080/internal/dovecot/$` to lua auth). That seems to be a first big step towards 2.4, since the dict driver has been removed. It now works with both 2.3 and 2.4. Would you be interested in such a PR?
Is there a place where we can find all the changes that need to be made for a successful migration to dovecot 2.4?

nextgens:
> Is the difficult upgrade from dovecot 2.3 to 2.4 holding the project back?

It's not difficult, it just needs to be done :)

> Would you be interested in such a PR?

Yes! We can process by steps: just send the PR with the minimal changes (will do the alpine version bump later) to make dovecot talk to admin without podop and we can take it from there

> Is there a place where we can find all the changes that need to be made for a successful migration to dovecot 2.4?

The idea is to get rid of podop completely if we can; migrating dovecot proxy in front is the obvious next step, then we can look into postfix

nextgens:
In fact I've already migrated dovecot proxy in front, see https://github.com/Mailu/Mailu/blob/master/core/nginx/dovecot/login.lua

firstred:
> It's not difficult, it just needs to be done :)

I'm glad you don't find it as difficult! I've been having a hard time finding the right documentation. Also the fact that you can't AI-slopify (vibe-code) some of the configuration makes me believe there aren't many dovecot 2.3 -> 2.4 config examples out there that the LLMs could have crawled. Dovecot 2.4 feels more like a new major version.

> Yes! We can process by steps: just send the PR with the minimal changes (will do the alpine version bump later) to make dovecot talk to admin without podop and we can take it from there

Will do!

> In fact I've already migrated dovecot proxy in front, see https://github.com/Mailu/Mailu/blob/master/core/nginx/dovecot/login.lua

Yes, I took that script as an example. Works a treat!

PR #3968, opened 2025-12-27
feature: Anonymous E-Mail aliases per API

lhw:
## What type of PR?

Feature

## Preamble

I know this is on the bigger side, if you'd prefer I split up the PR along some lines (i.e. changes to the original source code first) let me know. It's a nice to have feature for me and I think it fits well into the project.

## What does this PR do?

- Introduces anonymous/random alias generation and management across API and admin UI, plus the required database and authentication changes.
- The main idea is to have an automatic API for e.g. the Bitwarden client to generate aliases on a per-login basis.
- I selected the API that is used by [SimpleLogin](https://github.com/bitwarden/clients/blob/main/libs/tools/generator/core/src/integration/simple-login.ts) as its simple enough.
- Similar to the SimpleLogin interface users that own the aliases can disable and delete aliases.
- Aliases are, also similar to to SimpleLogin, generated either with a unique id or a combination of website+randomword style.
- I tried to be as secure as possible, while relying on already existing decorators, but I am not too familiar with the codebase yet, so I might have missed something.
- I added a lot of tests for the changes that were introduced. 

## Discussion points

- [x] I don't know where the filename for the migration comes from. I'll update accordingly.
- [x] I reused the authentication tokens for the API access of the new random alias endpoint. This required some changes, as the current authentication tokens are only used in conjunction with an E-Mail providing user information. The API would be token only. This might be an issue for you, in which case I would try to add API access tokens. But this seemed elegant enough without affecting the existing code too much (only added an index to identify the tokens).
- [x] The backfill of the existing tokens is not an automated process that runs for every token, but it's on a per-user basis. There might be a better solution for this.
- [ ] There are three different levels of access to the anonymous aliases (admin, managed access and global allow per domain).

### Screenshots

The bitwarden feature in question:
image

The anonymous email alias list from the user view:
Screenshot From 2025-12-27 14-36-42

Access denied without any domain access:
Screenshot From 2025-12-27 14-37-55

Domain manager interface to allow additional users on said domain:
Screenshot From 2025-12-27 14-38-30

Domain managers flag to allow domain for general use:
Screenshot From 2025-12-27 14-38-50

The list of aliases when viewed in the normal alias list:
Screenshot From 2025-12-27 14-39-52

### Related issue(s)
/

## Prerequisites


nextgens:
Thank you for your contribution; Let's have a chat about it on matrix if you can: https://app.element.io/#/room/#mailu-dev:matrix.org

lhw:
TODOs:
- [x] Allow for automatic domain selection if none is given
- [x] Delete aliases when destination is deactivated
- [x] Replace token with email:token and use the same process as nginx login with looping
- [x] Reverse the check order on the token. rate-limit first, then ip restrictions, then validity

lhw:
@nextgens 
You mentioned a filter option for the alias tables. But after getting jquery to work and finally finding some time i think the datatables search already works quite well
image
Did you have something else in mind?

core/admin/mailu/api/common.py
```diff
@@ -32,3 +34,56 @@ def decorated_function(*args, **kwds):
...
+            if flask_login.current_user.is_authenticated:
+                flask.g.user = flask_login.current_user
+                return func(*args, **kwds)
+            abort(401, 'A valid Authentication header is mandatory')
+        
+        if ':' not in auth:
+            utils.limiter.rate_limit_ip(client_ip)
+            abort(401, 'Invalid credentials format (expected email:token)')
+        
+        user_email, token = auth.split(':', 1)
```
ghostwheel42: I think "partition" is the method to use here:
```suggestion
        user_email, _, token = auth.partition(':')
        if not token:
            utils.limiter.rate_limit_ip(client_ip)
            abort(401, 'Invalid credentials format (expected email:token)')

```

core/admin/mailu/api/common.py
```diff
@@ -32,3 +34,56 @@ def decorated_function(*args, **kwds):
...
+        # Rate limit check first
+        if utils.limiter.should_rate_limit_ip(client_ip):
+            abort(429, 'Too many attempts from your IP (rate-limit)')
+        
+        auth = request.headers.get('Authentication')
+        if not auth:
+            if flask_login.current_user.is_authenticated:
+                flask.g.user = flask_login.current_user
+                return func(*args, **kwds)
+            abort(401, 'A valid Authentication header is mandatory')
```
ghostwheel42: Style: I would negate the if and swap the abort/return to "fail early":

```suggestion
            if not flask_login.current_user.is_authenticated:
                abort(401, 'A valid Authentication header is mandatory')
            flask.g.user = flask_login.current_user
            return func(*args, **kwds)
```

core/admin/mailu/api/common.py
```diff
@@ -32,3 +34,56 @@ def decorated_function(*args, **kwds):
...
+        # Try to get the user
+        try:
+            user = models.User.query.get(user_email) if '@' in user_email else None
+        except:
+            user = None
+        
+        if not user:
+            utils.limiter.rate_limit_ip(client_ip)
+            flask.current_app.logger.warn(f'Invalid user {user_email!r} from {client_ip}.')
+            abort(403, 'Invalid credentials')
```
ghostwheel42, line 80: What about logging the sql error like this:
```suggestion
        user = None
        cause = "not found"
        try:
            user = models.User.query.get(user_email)
        except sqlalchemy.exc.StatementError as exc:
            cause, _, _ = str(exc).partition('\n')
        
        if not user:
            utils.limiter.rate_limit_ip(client_ip)
            flask.current_app.logger.warn(f'Invalid user {user_email!r} from {client_ip}: {cause}')
            abort(403, 'Invalid credentials')
```

core/admin/mailu/ui/forms.py
```diff
@@ -17,16 +17,21 @@ def checkStrippable(form, field):
...
     """
 
     validator = re.compile(r'^.+@([^.@][^@]+)$', re.IGNORECASE)
 
     def iter_choices(self):
-        managed = [
-            str(email)
-            for email in flask_login.current_user.get_managed_emails()
-        ]
+        from .. import models
```
ghostwheel42: Please move the import to the top

core/admin/mailu/ui/forms.py
```diff
@@ -17,16 +17,21 @@ def checkStrippable(form, field):
...
+        
+        managed = []
+        for email in flask_login.current_user.get_managed_emails():
+            # Filter out anonymous aliases (those with owner_email set)
+            if isinstance(email, models.Alias) and email.owner_email is not None:
+                continue
+            managed.append(str(email))
```
ghostwheel42: I think this would be more "pythonic":
```suggestion
        managed = [
            str(email)
            for email in flask_login.current_user.get_managed_emails()
            if not (isinstance(email, models.Alias) and email.owner_email is not None)
        ]
```

nextgens:
bors try

nextgens:
@lhw can you fix the doc please?

lhw:
Anything else I should work on? For my use case it's feature complete for now.

core/admin/mailu/models.py
```diff
@@ -789,6 +797,22 @@ def resolve(cls, localpart, domain_name):
         return None
 
 
+@event.listens_for(db.session, 'before_commit')
+def disable_aliases_on_user_deactivation(session):
```
ghostwheel42: I wonder if disabling the owned addresses when the owner is disabled is the correct thing to do.
I would suggest to have the alias and user disabled flag to be completely separate and change the places where the "enabled" state of the alias is necessary from `not alias.disabled` to `not alias.disabled and not owner.disabled`. Do you think this would be possible?
Otherwise re-enabling a disabled owner would leave all aliases disabled.

ghostwheel42 (dismissed):
Hello and sorry for the delay. I haven't had much time for Mailu in the last few months.
Thanks for your time - automatic aliases are probably a good add-on.
Here are some comments. I did not yet test the changes, though, but I want to do that.

core/admin/mailu/api/common.py
lhw: Didnt know that existed. Added it.

core/admin/mailu/ui/forms.py
lhw: I actually made sure to avoid list comprehensions. I use them a lot in scripts but I thought with a bigger project the readability would suffer. But if it's the preferred style I am absolutely fine with that.

core/admin/mailu/models.py
lhw: I changed it so that the user disabled state is now a second query. Hence if a user is reactivated all the aliases states will be recovered as well.

ghostwheel42:
bors try

ghostwheel42:
bors try

ghostwheel42 (approved):
Okay, tests look fine, so let's merge it.

core/admin/mailu/api/simplelogin.py
```diff
@@ -0,0 +1,121 @@
...
+
+        if hostname and not note:
+            note = f'Website: {hostname}'
+
+        # Generate unique alias localpart
+        localpart = None
+        for _ in range(flask.current_app.config.get('ANONMAIL_MAX_RETRIES', 10)):
+            candidate = utils.generate_anonymous_alias_localpart(hostname=hostname)
+            email_candidate = f"{candidate}@{domain_name}"
+            if not models.Alias.query.filter_by(email=email_candidate).first() and not models.User.query.filter_by(email=email_candidate).first():
```
nextgens, line 86: This won't cover regexp-based aliases

core/admin/mailu/api/simplelogin.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+        
+        data = api.payload or {}
+        note = data.get('note')
+        hostname = request.args.get('hostname')
+
+        # Find all domains with anonmail access
+        accessible_domains = []
+        for d in models.Domain.query.all():
+            if (g.user.global_admin or models.has_domain_access(d.name, user=g.user) or
```
nextgens, line 68: maybe it would make sense to limit these to domains where anonmail is enabled?

core/admin/mailu/ui/views/aliases.py
```diff
@@ -64,3 +66,117 @@ def alias_delete(alias):
...
+
+
+@ui.route('/anonalias/list', methods=['GET'])
+@access.authenticated
+def anonalias_list():
+    user = flask_login.current_user
+    has_access = user.global_admin
+    if not has_access:
+        for d in models.Domain.query.all():
+            if models.has_domain_access(d.name, user=user) or (d.anonmail_enabled and user.domain and d.name == user.domain.name):
```
nextgens, line 78: I don't understand this access check. We are only displaying aliases created for **authenticated** users.

Either it's redundant or you should create another route that lets the username be customised (like we do elsewhere)

core/admin/mailu/ui/views/aliases.py
```diff
@@ -64,3 +66,117 @@ def alias_delete(alias):
...
+                has_access = True
+                break
+    
+    # Query user's anonymous aliases, standard aliases do not have an owner_email
+    aliases = models.Alias.query.filter_by(owner_email=user.email).all()
+    
+    return flask.render_template('alias/anonaliases.html', has_access=has_access, aliases=aliases)
+
+
+@ui.route('/anonalias/create', methods=['GET', 'POST'])
```
nextgens, line 88: same here

core/admin/mailu/ui/views/aliases.py
```diff
@@ -64,3 +66,117 @@ def alias_delete(alias):
...
+        hostname = form.hostname.data
+        note = form.note.data
+        destination = [user.email]
+        max_retries = app.config.get('ANONMAIL_MAX_RETRIES', 10)
+        
+        localpart = None
+        for _ in range(max_retries):
+            candidate = utils.generate_anonymous_alias_localpart(hostname=hostname)
+            email_candidate = f"{candidate}@{domain_name}"
+            if not models.Alias.query.filter_by(email=email_candidate).first() and not models.User.query.filter_by(email=email_candidate).first():
```
nextgens, line 117: This won't cover regexp-based aliases

nextgens (approved):
I have made a bunch of comments but approved nevertheless.

It can be fixed in an other PR.

core/admin/mailu/api/simplelogin.py
lhw: This covers multiple different "access" to the anonmail. As a manager you can whitelist the whole domain for anonmail or allow single users. But it could probably be moved into just the has_domain_access. All this was added one after the other.

core/admin/mailu/api/simplelogin.py
lhw: I'll check on that. That did not cross my mind that you could have a regex matching the same. I can just replace the first check with `Alias.resolve` if i see that correctly

core/admin/mailu/ui/views/aliases.py
lhw: Same as above i'll replace it with `Alias.resolve`

core/admin/mailu/ui/views/aliases.py
lhw: The whole check is a mess. true. the global admin being outside of the check itself as well. But it is just for the display in the form. For the error message. I'll check what is a good way to reduce the amount of calls here.

core/admin/mailu/ui/views/aliases.py
lhw: I think this one is valid as this is an access check to list domains in the form which are available to the user. not an authentication check.

core/admin/mailu/ui/views/aliases.py
nextgens: Elsewhere in the code we do that type of checks to enable admins to set/create/update/delete things on behalf of users. For that to work you need a new route.

For instance:
https://github.com/Mailu/Mailu/blob/master/core/admin/mailu/ui/views/tokens.py#L12

```python3
@ui.route('/token/list', methods=['GET', 'POST'], defaults={'user_email': None})
@ui.route('/token/list/<path:user_email>', methods=['GET'])
@access.owner(models.User, 'user_email')
def token_list(user_email):
```

lhw: Alright. Now not from my phone, i honestly mostly forgot about the PR. I think either I named the variable in such a way that it is easily misunderstood or i should have put a comment. The has_access only regulates if the user can see the "Create Alias" button, its not in itself an authentication check like the token_list decorator.
I simplified the check in #4023

Issue #1839, opened 2021-06-29
Use code lint / formating for python


parisni:
The python codebase would benefit from linters and formater. I suggest to add black and pylint/flake8 as additional CI breaker steps.

Diman0:
I had a look at black and it looks interesting. It would be handy to have such a tool that enforces PEP8 rules and rewrites all code to use the same style. I reckon this will improve the readability of the code. This could perhaps be run before the BORS try. I have no experience with pylint/flake8. Pylint looks more interesting when I compare the websites. It seems to have better documentation.

We could run this automatically before the bors try step. 
1). Black reformats the code and pushes the changes to the PR branch.
2). pylint/flake8 analyse the code.
3). If 1 and 2 succeeded, mergify will comment bors try. 
1/2 could trigger each time when a push is made on a PR.

parisni:
This sounds good. I should work on it during july.

decentral1se:
I've seen this tool called "pre-commit" used to group multiple linters into a single config and it is quite nice. A working example on another Python-based project is over in https://github.com/ansible-community/molecule/blob/master/.pre-commit-config.yaml. Then from a CI/CD and CLI perspective, you just run one command once and get all the tools reporting back to you.

parisni:
On Wed Jun 30, 2021 at 10:32 AM CEST, decentral1se wrote:
> I've seen this tool called "pre-commit" used to group multiple linters
> into a single config and it is quite nice


So https://pre-commit.com/ act as a pre-commit hook ? I am not a big fan
of pre-commit hooks. They make commit way longer and as a result tends
to commit less regularly. Also both black and pylint on this project would 
last few seconds.

parisni:
On June 30, 2021 8:32:05 AM UTC, decentral1se ***@***.***> wrote:
>I've seen this tool called "pre-commit" used to group multiple linters into a single config and it is quite nice. 

Wait, I will re-evaluate my position on this and try it out


Nicolas paris

Diman0:
If we implement this, we must be very careful that all irrelevant parts of the code are ignored by the linter.
See bugfix PR #1843 
@ghostwheel42 followed the advice of the linter he used to change sqlalchemy code which was already correct. The advice of the linter actually broke the code. All these areas must not be analyzed by the linter. We do not want it to break the code.

So If we include this, we must first check what areas must be excluded.

Diman0:
[...]

PR #3991, opened 2026-02-20
Add Ruff to CI

DjVinnii:
## What type of PR?

enhancement

## What does this PR do?

- Add Ruff to the CI to check (and only check) for formatting and linting. This will run before everything else so it ensures the fail fast principle
- Add Pre commit config to run Ruff before committing
- Ignore all current lint findings for now (separate PR's can be created so resolve the findings)

### Related issue(s)
-  #1839

## Prerequisites

### DIFF
--- a/docs/contributors/guidelines.rst
+++ b/docs/contributors/guidelines.rst
@@ -187,11 +187,19 @@ Coding
 Coding standards
 ````````````````
 
-All Python code should comply with PEP-8. We should review our code using
-pylint.
+All Python code should comply with PEP-8. To make sure all new code is
+compliant, the CI pipeline checks linting and formatting on all PRs.
+If the checks fail, the PR will not be merged until the issues are fixed.
+In order to make this easier as a developer, you can (optionally) run the
+linting and formatting checks locally before pushing your code by using
+pre-commit hooks. This can be done using `pre-commit`_ or the modern
+alternative `prek`_.
 
 We should comply with architectural recommendations from the Flask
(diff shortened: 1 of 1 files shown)


nextgens:
Ok, let's try it out.

bors try

DjVinnii:
> Ok, let's try it out.

Seems like it is working as intended, however a lot should be formatted to make the CI green. Do you want me to make a push for the formatting?

nextgens:
We are going in circles with this; There won't be a massive PR to "fix the formatting".

No, I'd like the linter to run exclusively on the piece of code that is in the PR: ie prevent non-formatted code to be introduced and do just that. Similarly the pre-commit hook should only concern itself with what is being changed in the commit, not the rest of the file.

DjVinnii:
Yeah, that makes sense. I updated the CI to get the diff between the base and head and filer on only python files. This also means that no linting exclusions should be present as a default to start with.

nextgens:
Great, what about the pre-commit?

bors try

DjVinnii:
I saw that the variables were empty strings, so I just replaced them with the actual Environment Variables and added a little bit of documentation. 

> Great, what about the pre-commit?

The pre-commit hooks can be optionally  run locally by a developer with a tool of choice (pre-commit or the more modern alternative prek). The CI makes sure that it is still checked if a developer does not use the pre-commit hooks. Also the pre-commit hooks only check the files being committed and the CI fully compares both branches.

PR #3993, opened 2026-02-26
fix: replace 17 bare except clauses with except Exception

haosenwang1018:
Replace bare `except:` clauses with `except Exception:` for PEP 8 compliance. Bare except catches SystemExit, KeyboardInterrupt, and GeneratorExit which is rarely intended.


haosenwang1018:
Thanks for reviewing. Closing this PR to avoid long-running open backlog; can re-open as smaller patches.

PR #4022, opened 2026-05-11
Fix PR-3864: Add bool "invisible" to model and set fetchmail option correctly

### LINKED ISSUE #4021: Database migration / model change missing when upgrading to current master
The newly introduced column "invisible" is not created when the database is upgraded which lead to fetchmail not working:

```
fetchmail-1        | Traceback (most recent call last):
fetchmail-1        |   File "/fetchmail.py", line 72, in run
fetchmail-1        |     options += " invisible" if fetch["invisible"] else ""
fetchmail-1        |                                ~~~~~^^^^^^^^^^^^^
fetchmail-1        | KeyError: 'invisible'
fetchmail-1        | Sleeping for 60 seconds
```

Before creating a new release we need to add that database migration and model change.

ghostwheel42:
## What type of PR?

bug-fix

## What does this PR do?

Add bool invisible to model and database. Without this fetchmail is broken.

### Related issue(s)

closes #4021
closes #3915

## Prerequisites


nextgens (approved):

nextgens (changes requested):
We're discussing it on #3915

There's a global fetchmail instance shared amongst users, if we want invisible mode support we should have two instances (one that is invisible and what that is not) or a simpler way of making it a global setting (env variable rather than per user pref)

ghostwheel42:
The current `fetchmail.py` uses a fetchmail instance per fetched remote mailbox.
So it is fine to use `--invisible` on the command line instead of an option in the rc file.

ghostwheel42:
bors try

nextgens (dismissed):
LGTM.

You're right, I was looking at the wrong branch/version of the code

ghostwheel42:
The database change in here is based on #3968, so this needs to be merged first.
After that I'll rebase this one.

ghostwheel42:
bors try

ghostwheel42:
Okay. rebase is done. I've also renamed the migration for the anon email db change to match the others.

nextgens (approved):

PR #4023, opened 2026-05-11
enhancement: checks for anonmail access and ensure no wildcard collisions

lhw:
## What type of PR?

Enhancement to resolve remaining comments from #3968

## What does this PR do?

- Simplifies the anonmail has_access checks (removed some duplicate checks for global admin
- Fixes the alias generation check to avoid collisions with wildcard aliases

Two Notes:
The `Alias.resolve` would make domains that have catch alls with `%@domain` essentially non-functional for anonmail generation, even if allowed as it would return an alias for every single call. Aliases are matched by length with explicit matches first and then by length of match if got this right:
```
.order_by(cls.wildcard, sqlalchemy.func.char_length(cls.localpart).desc())
```
Question is if that is what we want here. 

Second note. Right now the match with `Alias.resolve` requires the extra check against the alias db anyhow due to disabled aliases not being listed by `Alias.resolve` and causing a collision. An optional change would be to introduce a form of `Alias.resolve_any` or an optional parameter to `Alias.resolve(cls, localpart, domain_name, list_disabled=False)` e.g.

### Related issue(s)
- #3968 

## Prerequisites

### DIFF
--- a/core/admin/mailu/api/simplelogin.py
+++ b/core/admin/mailu/api/simplelogin.py
@@ -65,7 +65,7 @@ class RandomAlias(Resource):
         # Find all domains with anonmail access
         accessible_domains = []
         for d in models.Domain.query.all():
-            if (g.user.global_admin or models.has_domain_access(d.name, user=g.user) or 
+            if (models.has_domain_access(d.name, user=g.user) or
                 (d.anonmail_enabled and g.user.domain and d.name == g.user.domain.name)):
                 accessible_domains.append(d.name)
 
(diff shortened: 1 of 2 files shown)


ghostwheel42 (approved):
Let's not change Alias.resolve for now.
I'll approve this, but maybe adding some note in the docs about a domain with catch-all (or any other wildcards) would be appropriate.

nextgens (approved):
Maybe we should iterate over domains that have anonmail_enabled exclusively instead of all domains, no?

Do we really want global admins to have aliases created on domains where it's not enabled?

ghostwheel42:
> Do we really want global admins to have aliases created on domains where it's not enabled?

I'd say: No. We don't

ghostwheel42:
bors try

nextgens:
Let's get it merged and it can be done in another PR

bors r+

PR #4043, opened 2026-06-25
Quote vacation :from display name so auto-replies with special characters compile

### LINKED ISSUE #3983: Automatic reply doesn't work when "Displayed name" is an E-mail address
When a mail account has automatic replies enabled but the displayed name in that account is e.g. the mail itself, the replies will not work:

```
Error: sieve: default: line 32: specified :from address 'xxx@yyy <xxx@yyy>' is invalid for vacation action: address ends in invalid characters
Error: sieve: default: validation failed
Error: sieve: Failed to compile script `dict:proxy:/tmp/podop.socket:sieve;user=xxx@yyy'
Info: sieve: msgid=<1234>: stored mail into mailbox 'INBOX'
```

Probably missing some sort of escape?

maximilize:
## Problem

Automatic replies (vacation) fail to compile when the account's display name contains special characters such as `@` — e.g. when the display name is set to an email address (#3983):

```
sieve: ... specified :from address 'xxx@yyy <xxx@yyy>' is invalid for vacation action: address ends in invalid characters
```

## Cause

`core/admin/mailu/internal/templates/default.sieve` renders the vacation `:from` with the display name placed **bare** before the address:

```
:from "{{ user.displayed_name ... }} <{{ user.email ... }}>"   →   xxx@yyy <xxx@yyy>
```

An RFC 5322 display name that contains specials (`@`, `<`, `.`, …) must be a quoted-string, so the bare form is an invalid address and pigeonhole rejects the script.

## Fix

Wrap the display name in (escaped) quotes — a quoted-string is valid for any display name:

```
:from "\"{{ user.displayed_name ... }}\" <{{ user.email ... }}>"
```

- `John Doe` → `"John Doe" <john@d>` (still valid)
- `xxx@yyy` → `"xxx@yyy" <xxx@yyy>` (now valid; was invalid)

The existing internal-quote escaping is kept.

## Verification

Rendering `default.sieve` for a user whose display name is `xxx@yyy`:
- before: `:from "xxx@yyy <xxx@yyy>"` (invalid)
- after: `:from "\"xxx@yyy\" <xxx@yyy>"` (valid)

Fixes #3983.

### DIFF
--- a/core/admin/mailu/__init__.py
+++ b/core/admin/mailu/__init__.py
@@ -97,6 +97,15 @@ def create_app_from_config(config):
     def format_datetime(value):
         return utils.flask_babel.format_datetime(value) if value else ''
 
+    @app.template_filter()
+    def sieve_string(value):
+        # Escape a value for safe inclusion inside a Sieve quoted-string.
+        # RFC 5228 requires escaping '\' and '"'; we also escape '$' so a
+        # user-controlled value cannot inject Sieve variable expansion
+        # ('${...}', RFC 5229) once the "variables" extension is required.
+        # Backslash must be escaped first so the escapes we add aren't re-escaped.
+        return str(value).replace('\\', '\\\\').replace('"', '\\"').replace('$', '\\$')
+
     def ping():
         return ''
     app.route('/ping')(ping)
--- a/core/admin/mailu/internal/templates/default.sieve
+++ b/core/admin/mailu/internal/templates/default.sieve
@@ -31,6 +31,6 @@ if spamtest :percent :value "gt" :comparator "i;ascii-numeric" "{{ user.spam_thr
 
 {% if user.reply_active %}
 if not address :localpart :contains ["From","Reply-To"] ["noreply","no-reply"]{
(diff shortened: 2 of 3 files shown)


nextgens:
Thank you for your contribution!

This is unfortunately not enough: we need to escape ``\`` ``$`` and ``"`` from all fields to prevent sieve injection problems.

Would you mind adding a filter in lieu of the `` replace(`` too?

maximilize:
Done. Added a `sieve_string` Jinja filter that escapes `\`, `"` and `$`, and applied it to every user-controlled field in the vacation action (`displayed_name`, `email`, `reply_subject`, `reply_body`) in place of the inline `replace(...)`. Escaping `$` neutralises `${...}` expansion from the required `variables` extension; backslash is escaped first so the added escapes aren't re-escaped.

Verified by rendering `default.sieve` with hostile inputs — `ev"il`, `a\b`, `hi ${1} there`, and a breakout attempt `x"; discard; #` all stay inside the quoted string (the last renders as `… "x\"; discard; #";`), so no premature close and no injected action. The original #3983 case still renders `:from "\"xxx@yyy\" <xxx@yyy>"`.

nextgens:
bors try

ghostwheel42 (approved):

nextgens (approved):
LGTM

Thank you!

PR #4045, opened 2026-06-26
fix: preserve recipient-delimiter detail when resolving explicit aliases (#2988)

### LINKED ISSUE #2988: `lmtp_save_to_detail_mailbox` and aliases
### Environment

- [X] docker compose
- [ ] kubernetes
- [ ] docker swarm

### Version

- Version: `2.0`

## Description

Dovecot comes with the `lmtp_save_to_detail_mailbox` option, which directs incoming mails into the corresponding mailbox folder, if such folder exists (e.g. `user+test@domain.tld` would be saved into the `test` folder), otherwise into `INBOX`.

If enabled (via `lmtp_save_to_detail_mailbox` in `overrides/dovecot/dovecot.conf`), this works fine, however not in case of `alias+test@domain.tld` - in this case, the mail is delivered to the INBOX.

The `Delivered-To` header shows `user+test@domain.tld` in case the mail is addressed to `user+test@domain.tld`, but only `user@domain.tld` in case the mail is addressed to `alias+test@domain.tld`, so the detail part after the `recipient_delimiter` gets thrown away.

It seems that the detail part (e.g. `test`) gets thrown away (I could be wrong here):

`docker compose exec smtp curl http://admin:8080/internal/postfix/alias/mail+test@domain.xyz` returns `mail+test@domain.xyz`

`docker compose exec smtp curl http://admin:8080/internal/postfix/alias/alias+test@domain.xyz` returns `mail@domain.xyz`

Maybe the detail part should be included here:
https://github.com/Mailu/Mailu/blob/ab59c32296a26187e226b70dd08459b72746ace5/core/admin/mailu/models.py#L479-L480

...as it is here (implicitely):
https://github.com/Mailu/Mailu/blob/ab59c32296a26187e226b70dd08459b72746ace5/core/admin/mailu/models.py#L460-L467

## Replication Steps
1. Set `lmtp_save_to_detail_mailbox = yes` in `overrides/dovecot/dovecot.conf`
2. Create an IMAP folder named `test`
3. Confirm that `lmtp_save_to_detail_mailbox` is working by sending mail to any user with the respective local part (e.g. `yourusername+test@domain.tld` - the mail should end up in the IMAP folder `test`.
4. Send a mail to an alias of the user with the respective local part (e.g. `youralias+test@domain.tld`).

## Observed behaviour
The mail will end up in the `INBOX`.

## Expected behaviour
The mail should be saved into `test`.

maximilize:
Fixes #2988.

## Problem

With `lmtp_save_to_detail_mailbox`, mail to `alias+detail@domain` loses the `+detail` part when `alias` is an explicit alias. `Email.resolve_destination` strips the delimiter to find the alias, then returns the alias destination **without re-attaching the detail**. The user path (`user+detail@domain`) already preserved it; only aliases were affected.

## Why not just "remove the split"

The thread suggested dropping the delimiter split and letting Postfix `propagate_unmatched_extensions` do the rest. That regresses the catch-all case: Mailu models a catch-all as a wildcard alias, so `Alias.resolve('team+test')` already matches the wildcard — Postfix gets a hit on its first lookup and never retries the specific `team` alias. With a catch-all plus an explicit `team` alias, `team+test` then goes to the catch-all instead of `team`. The `test_catchall_does_not_swallow_explicit_alias_detail` test captures exactly this.

## Fix

Re-attach the stripped detail to the destination of an **explicit (non-wildcard)** alias — mirroring `propagate_unmatched_extensions` — and leave wildcard/catch-all resolution untouched:

- `team+test@dom` → `joe+test@dom` (was `joe@dom`)
- `random+x@dom` via catch-all → unchanged
- `user+detail@dom` → unchanged

`+15/-1` in `core/admin/mailu/models.py`, plus a `2988.bugfix` news fragment.

## Tests

`tests/anonmail/test_alias_detail.py` (5 cases): explicit-alias detail preserved, multiple destinations, catch-all not swallowing an explicit alias, user-detail unchanged, and the `/internal/postfix/alias/` endpoint. Red without the fix, green with it; the full `tests/anonmail` suite (26 tests) passes.

### DIFF
--- a/core/admin/mailu/models.py
+++ b/core/admin/mailu/models.py
@@ -506,6 +506,14 @@ class Email(object):
             domain_name = alternative.domain_name
         return (localpart, domain_name)
 
+    @staticmethod
+    def _append_detail(address, detail):
+        """ insert a recipient-delimiter detail into an address localpart """
+        if not detail or '@' not in address:
+            return address
+        localpart, domain_name = address.rsplit('@', 1)
+        return f'{localpart}{detail}@{domain_name}'
+
     @classmethod
     def resolve_destination(cls, localpart, domain_name, ignore_forward_keep=False):
         """ return destination for email address localpart@domain_name """
@@ -543,7 +551,13 @@ class Email(object):
                 return pure_alias.destination
 
         if stripped_alias := Alias.resolve(localpart_stripped, domain_name):
-            return stripped_alias.destination
+            if stripped_alias.wildcard:
+                return stripped_alias.destination
+            # Re-attach the recipient delimiter detail (e.g. '+tag') that was
+            # stripped for the lookup, so it is not lost for the recipient. This
+            # mirrors postfix' propagate_unmatched_extensions for explicit aliases.
+            detail = localpart[len(localpart_stripped):]
+            return [cls._append_detail(dst, detail) for dst in stripped_alias.destination]
 
         if pure_alias:
             return pure_alias.destination
--- a/tests/anonmail/test_alias_detail.py
+++ b/tests/anonmail/test_alias_detail.py
@@ -1,0 +1,69 @@
+"""Regression tests for #2988: recipient-delimiter detail must be preserved
+when an address resolves through an explicit (non-wildcard) alias."""
+import os
+
+from mailu import models
+
+
+def _base(app):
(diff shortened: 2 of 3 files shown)


nextgens (approved):
LGTM

ghostwheel42 (approved):

PR #3254, opened 2024-05-10
fix #3252

nextgens:
## What type of PR?

bug-fix

## What does this PR do?

Ensure that sending emails to invalid addresses with an '@' in localpart is permanently and not temporarily rejected.

This is the sister commit of 15569c62df2336faf3c3ffabb7190a3caff2983a

### Related issue(s)
- closes #3252

## Prerequisites
Before we can consider review and merge, please make sure the following list is done and checked.
If an entry in not applicable, you can check it or remove it from the list.

- [ ] In case of feature or enhancement: documentation updated accordingly
- [x] Unless it's docs or a minor change: add [changelog](https://mailu.io/master/contributors/workflow.html#changelog) entry file.

Issue #3910, opened 2025-08-25
RFC-compliant email addresses with non-standard characters cause API errors


StuHoward:
API operations on mail aliases fail when the address contains '#' (possibly other ASCII/UTF-8 characters too), despite these being valid according to the RFCs.  For example, an API request to delete the alias '`myalias#two@aliasdomain.com`':

```
{
"email":"real.address@gmail.com",
"proxyEmail":"myalias#two@aliasdomain.com",
"mailuId":"01925e3c-9bfd-7254-88cf-c8f6d3db8aaa@mydomain.com"
}
```

...produces the following error:

```
[2025-08-25 10:42:26,849] ERROR in app: Exception on /api/v1/alias/myalias [GET]
Traceback (most recent call last):
  File "/app/venv/lib/python3.11/site-packages/sqlalchemy/engine/base.py", line 1800, in _execute_context
    context = constructor(
              ^^^^^^^^^^^^
  File "/app/venv/lib/python3.11/site-packages/sqlalchemy/engine/default.py", line 1121, in _init_compiled
    param = {
            ^
  File "/app/venv/lib/python3.11/site-packages/sqlalchemy/engine/default.py", line 1122, in <dictcomp>
    key: processors[key](compiled_params[key])
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/venv/lib/python3.11/site-packages/sqlalchemy/sql/type_api.py", line 1660, in process
    return process_param(value, dialect)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/mailu/models.py", line 63, in process_bind_param
    raise ValueError('invalid email address (no "@")')
ValueError: invalid email address (no "@")

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/app/venv/lib/python3.11/site-packages/flask/app.py", line 1823, in full_dispatch_request
    rv = self.dispatch_request()
         ^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/venv/lib/python3.11/site-packages/flask/app.py", line 1799, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #4047, opened 2026-06-27
Fix 500 on internal Postfix lookup endpoints for addresses with multiple @ (#3252)

### LINKED ISSUE #3252: Double @ Sign In Sent Email Causing Rate Limiting
## Environment & Version

Docker Compose version v2.26.1

Mailu version 2.0

## Description

If a user attempts to send an email to an invalid email address containing two @ signs, for example to "john@1234@gmail.com", podop raises an exception and fails to lookup the domain part of the email. Postfix treats this as a temporary error, and advises the mail client to try again later. This repeats until the client is rate limited.

I suspect that changing line 443 in `core/admin/mailu/models.py` to split on the last '@' sign instead of the first may fix the issue, but I can't say for sure.

Note that '@' signs are allowed in the local part of an email address as long as they are within quotes. (Which is not the case here, but still.) https://en.wikipedia.org/wiki/Email_address#Local-part

## Replication Steps

Send an to email address with two '@' signs.

## Observed behaviour

Client rate limited.

## Expected behaviour

Fail gracefully.

## Logs
```
smtp-1  | ERROR:root:Error when processing request
smtp-1  | Traceback (most recent call last):
smtp-1  |   File "/app/venv/lib/python3.11/site-packages/podop/postfix.py", line 104, in process_request
smtp-1  |     result = await table.get(key)
smtp-1  |              ^^^^^^^^^^^^^^^^^^^^
smtp-1  |   File "/app/venv/lib/python3.11/site-packages/podop/table.py", line 35, in get
smtp-1  |     raise Exception(request.status)
smtp-1  | Exception: 500
smtp-1  | May 09 09:36:59 mail postfix/cleanup[123452]: warning: socketmap:unix:/tmp/podop.socket:recipientmap socketmap server temporary error: unknown error
smtp-1  | May 09 09:36:59 mail postfix/cleanup[123452]: warning: socketmap:unix:/tmp/podop.socket:recipientmap lookup error for ""john@1234"@gmail.com"
smtp-1  | May 09 09:36:59 mail postfix/cleanup[123452]: warning: 5E63228215BF: recipient_canonical_maps map lookup problem for "john@1234"@gmail.com -- message not accepted, try again later
smtp-1  | ERROR:root:Error when processing request
smtp-1  | Traceback (most recent call last):
smtp-1  |   File "/app/venv/lib/python3.11/site-packages/podop/postfix.py", line 104, in process_request
smtp-1  |     result = await table.get(key)
smtp-1  |              ^^^^^^^^^^^^^^^^^^^^
smtp-1  |   File "/app/venv/lib/python3.11/site-packages/podop/table.py", line 35, in get
smtp-1  |     raise Exception(request.status)
smtp-1  | Exception: 500
smtp-1  | May 09 09:36:59 mail postfix/cleanup[123452]: warning: socketmap:unix:/tmp/podop.socket:alias socketmap server temporary error: unknown error
smtp-1  | May 09 09:36:59 mail postfix/cleanup[123452]: warning: socketmap:unix:/tmp/podop.socket:alias lookup error for ""john@1234"@gmail.com"
smtp-1  | May 09 09:36:59 mail postfix/cleanup[123452]: warning: 5E63228215BF: virtual_alias_maps map lookup problem for john@1234@gmail.com -- message not accepted, try again later
```

maximilize:
## What

The internal Postfix lookup endpoints `/internal/postfix/alias`, `recipient/map`
and `sender/login` returned a 500 for an address with more than one `@` (or a
quoted local part): `resolve_destination` binds the key through `IdnaEmail`,
which raises `email local part must not contain "@"`, and `recipient/map` lets
`srslib` raise on the same input. Postfix treats a 500 as a temporary failure
and retries, which trips the sender rate limit — the behaviour reported in #3252.

`sender/map` already guarded against this. This applies the same guard to the
other three endpoints via a shared `_unsupported_address()` helper, so they
answer 404 (address not found) instead.

## Test

Adds a `tests/anonmail` regression test that hits all four endpoints with
`a@b@example.com` and `"a@b"@example.com` and asserts 404. Without the fix the
alias / recipient-map / sender-login endpoints 500 (reproduced); with it the
full `tests/anonmail` suite passes. This is the test coverage that was missing
when #3254 was set aside.

Fixes #3252

### DIFF
--- a/core/admin/mailu/internal/views/postfix.py
+++ b/core/admin/mailu/internal/views/postfix.py
@@ -10,10 +10,12 @@ import srslib
 
 def _unsupported_address(address):
     """ Return True for a lookup key Mailu cannot resolve: an address with
-    more than one ``@`` or a quoted local part. Resolving such a key would let
-    IdnaEmail/srslib raise and 500 the endpoint, which makes Postfix treat the
-    lookup as a temporary failure and retry — tripping the sender rate limit
-    (see #3252). These endpoints must answer 404 instead. """
+    more than one ``@`` or a quoted local part. Binding such a key in
+    ``resolve_destination`` lets ``IdnaEmail.process_bind_param`` raise (the
+    local part still contains an ``@``), which SQLAlchemy surfaces as a
+    ``StatementError`` and 500s the endpoint. Postfix then treats the lookup as
+    a temporary failure and retries — tripping the sender rate limit (#3252).
+    The resolve-path endpoints must answer 404 instead. """
     return address.count('@') > 1 or address.startswith('"')
 
 @internal.route("/postfix/dane/<domain_name>")
@@ -120,8 +122,6 @@ def postfix_recipient_map(recipient):
 
     This is meant for bounces to go back to the original sender.
     """
-    if _unsupported_address(recipient):
-        return flask.abort(404)
     srs = srslib.SRS(flask.current_app.srs_key)
     if srslib.SRS.is_srs_address(recipient):
         try:
@@ -137,8 +137,6 @@ def postfix_sender_map(sender):
 
     This is for bounces to come back the reverse path properly.
     """
-    if _unsupported_address(sender):
-        return flask.abort(404)
     srs = srslib.SRS(flask.current_app.srs_key)
     domain = flask.current_app.config["DOMAIN"]
     try:
--- a/core/admin/mailu/internal/views/postfix.py
+++ b/core/admin/mailu/internal/views/postfix.py
@@ -8,6 +8,14 @@ import re
 import sqlalchemy.exc
 import srslib
 
+def _unsupported_address(address):
+    """ Return True for a lookup key Mailu cannot resolve: an address with
+    more than one ``@`` or a quoted local part. Resolving such a key would let
+    IdnaEmail/srslib raise and 500 the endpoint, which makes Postfix treat the
+    lookup as a temporary failure and retry — tripping the sender rate limit
+    (see #3252). These endpoints must answer 404 instead. """
+    return address.count('@') > 1 or address.startswith('"')
+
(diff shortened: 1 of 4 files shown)


nextgens:
I think that the way forward is to upgrade srslib and drop our guards/simplify our code instead.

It's now fixed upstream: https://github.com/jichu4n/srslib/pull/3

maximilize:
Reworked accordingly: bumped srslib 0.1.4 → 0.1.5 (your jichu4n/srslib#3) and dropped the guard from the two SRS endpoints. On 0.1.5 `is_srs_address`/`forward` handle the extra `@` via `rsplit`, so `recipient/map` answers 404 (not an SRS address) and `sender/map` SRS-forwards an external sender as usual — neither 500s.

I kept `_unsupported_address` on `alias` + `sender/login`: those never touch srslib. They go through `resolve_destination` → `User.query.get(f'{localpart}@{domain_name}')`, and `resolve_domain` only `rsplit`s once, so the bound value still holds an `@` and hits `IdnaEmail.process_bind_param` → `ValueError: email local part must not contain "@"`, which SQLAlchemy wraps as `StatementError` → 500 (same class as #3910). The bump alone doesn't cover those two.

The regression test now characterizes all four endpoints under 0.1.5: `alias` / `recipient/map` / `sender/login` → 404, and `sender/map` → 404 for a local domain, 200 (SRS-forwarded) for an external one — none 500. One behavior change worth flagging: `sender/map` now returns a 200 SRS address for an external double-`@` sender instead of 404, i.e. it's treated like any other external sender. Full `tests/anonmail` suite passes on 0.1.5.

ghostwheel42 (approved):

nextgens (approved):

PR #4048, opened 2026-06-27
Fix domain list page hanging with many users by counting via aggregates (#3996)

### LINKED ISSUE #3996: Admin Domain List Page Hangs with Large User Table (~500k users)
# Admin Domain List Page Hangs with Large User Table (~500k users)

## Environment

- **Mailu version:** 2024.06 (Docker Compose)
- **Database:** PostgreSQL (external)
- **User table size:** ~500k rows
- **Domains:** ~20
- **Reverse proxy:** Nginx Proxy Manager in front of Mailu `front`

---

## Description

When opening the **Domain list** in the admin UI (`/admin/domain`), the request does not finish within **1–2 minutes**.  

The front (Nginx) logs **499 (client closed connection)** and the page never loads.

Other admin pages (e.g. **user settings**) and login work normally.

When the number of users was smaller, the domain list page loaded correctly.

---

## Steps to Reproduce

1. Deploy **Mailu 2024.06** with **PostgreSQL**.
2. Import a large number of users (e.g. **500k+**) across multiple domains (e.g. **~20 domains**).
3. Log in to the admin UI.
4. Open **Domain list** (click **Domains** in the sidebar).

Result:

- The request hangs.
- After the proxy timeout (e.g. **60s**) the client disconnects.
- front logs **499**.
- high cpu usage
Host:
```
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND                                                     
 479132 dhcpcd    20   0   12.4t 592368   3872 S  98.7  17.0  24:30.42 gunicorn                                                    
 192166 root      20   0  102.3g 222860  20784 S   5.3   6.4  49:59.21 node                                                        
   2130 root      20   0 2915380  56520  17936 S   0.3   1.6 174:17.07 dockerd    
```
admin container:
```
CONTAINER ID   NAME            CPU %     MEM USAGE / LIMIT     MEM %     NET I/O           BLOCK I/O         PIDS 
7eb33d94788a   mailu-admin-1   99.32%    551.4MiB / 3.317GiB   16.23%    9.86GB / 1.83GB   20.5GB / 21.1GB   5 
```
---

## Expected Behavior

The domain list page should load within a few seconds, or at least within a reasonable timeout (e.g. **<30s**).

Possible approaches for large datasets:

- Domains page load instantly

maximilize:
## What

The domain list page counted users and aliases per row with
`domain.users | count` / `domain.aliases | count`. Those are `lazy='select'`
backrefs, so Jinja's `count` filter loads each domain's **entire** user/alias
collection into memory on every page load. On an install with ~500k users this
materializes all of them per request and the page hangs (#3996).

`domain_list()` now computes the counts with one grouped `func.count()` query per
table and passes them to the template, which renders `user_counts.get(...)` /
`alias_counts.get(...)`. No per-domain collection is loaded. This is the
materialization, not a missing index — an index alone wouldn't avoid loading the
rows.

## Test

Adds a `tests/anonmail` regression test that renders the page for two populated
domains while capturing SQL, and asserts no per-domain
`... WHERE ? = user.domain_name` / `alias.domain_name` load happens (it does on
the old code) and that the counts render correctly. Full `tests/anonmail` suite
passes.

Fixes #3996


nextgens:
Please send us another PR (this one will be backported and anonmail only exists in master) to make the anonmail tests run on CI; I don't think they are run at the moment.

nextgens (dismissed):
LGTM!

Thank you for looking into it

nextgens:
Scratch the above, let's do it here... and we don't need to backport this. Those that require it can upgrade to master :)

maximilize:
Added a `unit tests` workflow (`.github/workflows/unittests.yml`) that runs `tests/anonmail` on every PR and on pushes to master. It's a standalone, lightweight job — checkout → setup-python 3.12 → install `requirements-prod.txt` + `tests/requirements.txt` + pytest → `pytest tests/anonmail` — kept separate from the docker-compose matrix in `build_test_deploy.yml`, since the anonmail suite is pure in-memory SQLite and needs no mail stack. `podop`/`socrate` are vendored, so they're put on `PYTHONPATH` rather than pip-installed.

I verified the exact steps in a clean `python:3.12-slim`: the suite passes, including this PR's #3996 regression test. If you'd rather fold it into `build_test_deploy.yml`, or run it inside the built admin image instead of a separate workflow, say the word and I'll move it.

ghostwheel42 (dismissed):

nextgens:
I'd like you to fold it and run it on the admin image instead (to ensure the python version and dependencies match)

maximilize:
Done — folded into `build_test_deploy.yml` as a dedicated `unit-tests` job and dropped the standalone `unittests.yml`. It runs the suite inside the freshly built admin image:

```
docker run --rm -v $GITHUB_WORKSPACE/tests:/app/tests:ro -w /app \
  ghcr.io/<org>/admin:<version>-build \
  sh -c "pip install --no-cache-dir pytest && python -m pytest tests/anonmail"
```

So they exercise the exact python and pinned deps that ship in the image; `deploy` now depends on it. Verified locally against `ghcr.io/mailu/admin` with this PR's code mounted in: `24 passed`. The `conftest.py` `sys.path` shim is a harmless no-op there — `mailu` already lives at `/app`, so the import resolves against the image.

nextgens:
bors try

nextgens (approved):
LGTM

ghostwheel42 (approved):

PR #4049, opened 2026-06-27
fetchmail: remove duplicate logging of fetch errors

### LINKED ISSUE #3970: Obsolete code in fetchmail.py?
Hi,

while working on a new revision of my fetchmail.py I came across 2 potential issues:

Is this obsolete code? I did not see any code in fetchmail's git that is putting out "No mail...". There was a driver.c file in earlier versions but not anymore.
https://github.com/Mailu/Mailu/blob/bc8fc85fa316ad60ddc2b47b6db89ce2d514198f/optional/fetchmail/fetchmail.py#L94

And I believe this will only output the error message twice in a specific case which to me looks fully redundant. The line mentioned above already catches this and could be reduced to simply `print(error_message)`
https://github.com/Mailu/Mailu/blob/bc8fc85fa316ad60ddc2b47b6db89ce2d514198f/optional/fetchmail/fetchmail.py#L98-L100

Your thoughts? Any intentions behind the code that I do not get?

maximilize:
Closes #3970.

The error handler in `fetchmail.py` printed the fetch output twice. The "messages seen" branch reprints `error_message`, but it can only fire on output the preceding check already printed: a `fetchmail: No mail …` line (the only case the first check skips) never contains `(seen `, so the second branch was always a duplicate. Removed it along with the now-unused `user_info`.

I kept the `No mail` check (the other half of #3970): current fetchmail still emits `No mail for %s` (verified in the `fetchmail` binary shipped by `apk add fetchmail`), so dropping that suppression would log a "no mail" line on every poll of an empty mailbox.

No behavior change beyond removing the duplicate log line — the status posted back to admin (`error_message.split('\n')[0]`) is unchanged.

### DIFF
--- a/optional/fetchmail/fetchmail.py
+++ b/optional/fetchmail/fetchmail.py
@@ -93,12 +93,6 @@ def run(debug):
                 # No mail is not an error
                 if not error_message.startswith("fetchmail: No mail"):
                     print(error_message)
-                user_info = "for %s at %s" % (fetch["user_email"], fetch["host"])
-                # Number of messages seen is not a error as well
-                if ("messages" in error_message and
-                        "(seen " in error_message and
-                        user_info in error_message):
-                    print(error_message)
             finally:
                 requests.post("http://{}:8080/internal/fetch/{}".format(os.environ['ADMIN_ADDRESS'],fetch['id']),
                     json=error_message.split('\n')[0]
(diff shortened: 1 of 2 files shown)


nextgens (approved):
LGTM

ghostwheel42 (approved):

PR #4050, opened 2026-06-28
Prevent a domain manager from taking over a higher-privileged account (#2685)

### LINKED ISSUE #2685: Domain manager can reset admin password
### Environment

- [x] docker-compose
- [ ] kubernetes
- [ ] docker swarm

### Version

- Version: `1.9`

## Description
Domain manager can change password on admin users

## Replication Steps
create a admin user on domain.tld
create a  domain manager on domain.tld
login as domain manager on domain.tld
try to change admin users password as domain manager

## Observed behaviour
admin password is reset and i can login with the new password

## Expected behaviour
domain manager is not allowed to change admin password because admin has higher priveleges than domain manager

maximilize:
## What

`access.domain_admin` (used by `user_edit`, `user_password`) and `access.owner`
(used by `token_create`) authorize purely on the **target's** domain —
`domain in current_user.get_managed_domains()` — with no rank check. So a
manager of a domain can edit, reset the password of, or mint an API token for a
**global admin** (or a **manager of another domain**) who merely owns a mailbox
in the managed domain, taking that privileged account over. Exactly as
@fastlorenzo described.

This adds a single `_can_manage_user(actor, target)` helper and calls it from
both decorators when the resource is a `User`: a non-global manager may only act
on a target that does not outrank them — the target must not be a global admin,
and every domain the target manages must also be managed by the actor. Global
admins and self-service are unaffected.

Scope note (re your comment): this gates **all** management of a higher-or-equal-rank
target (edit / password / token), not only password+token. The flip side is the
case you raised — a domain manager can then no longer flip the `disabled` bit on
a compromised global-admin (or other-domain-manager) account that lives in their
domain; that stays a global-admin action. `_can_manage_user` is the single place
to scope it, so narrowing this to only password+token (leaving edit/disable
open) is a one-line change if you prefer that trade-off.

## Test

Adds `tests/anonmail/test_domain_admin_privesc.py` (8 cases), exercising the
decorators via GET — the decorator runs before the view body, so 200 = allowed,
403 = denied:

- manager → reset / edit / create-token-for a global admin in the managed domain → **403** (was 200)
- manager → reset a manager of another domain → **403** (was 200)
- manager → reset / edit a plain subordinate → **200** (unchanged)
- global admin → reset a manager → **200**; user → create own token → **200**

Verified rot→green; the full `tests/anonmail` suite passes (31).

Fixes #2685

### DIFF
--- a/core/admin/mailu/ui/access.py
+++ b/core/admin/mailu/ui/access.py
@@ -46,6 +46,24 @@ def global_admin(args, kwargs):
     return flask_login.current_user.global_admin
 
 
+def _can_manage_user(actor, target):
+    """ Whether ``actor`` may manage the account of ``target`` (a User).
+
+    Global admins may manage anyone, and any user may manage themselves.
+    Otherwise a (domain) manager may only manage a target that does not
+    outrank them: the target must not be a global admin, and every domain the
+    target manages must also be managed by the actor. Without this, a manager
+    of a domain could reset the password of -- or mint an API token for -- a
+    global admin (or a manager of another domain) who merely owns a mailbox in
+    the managed domain, taking over that privileged account (#2685).
+    """
+    if actor.global_admin or target.email == actor.email:
+        return True
+    if target.global_admin:
+        return False
+    return set(target.get_managed_domains()) <= set(actor.get_managed_domains())
+
+
 @permissions_wrapper
 def domain_admin(args, kwargs, model, key):
     """ The view is only available to specific domain admins.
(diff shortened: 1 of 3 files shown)


nextgens (approved):
LGTM

It looks neat!

Thank you

nextgens:
@ghostwheel42 any opinion on whether we should backport this?

ghostwheel42 (approved):

ghostwheel42:
> @ghostwheel42 any opinion on whether we should backport this?

Hmm. It somehow was like it was by design, I think. 

I think upgrading deps and releasing the next version would be better.

PR #4051, opened 2026-06-28
Reject case-only-duplicate users/aliases instead of crashing (closes #2695, closes #2718)

### LINKED ISSUE #2695: Creating a duplicate of a user or alias with a capital letter returns: Status 500 Internal Server Error
### Environment

- [x] docker-compose
- [ ] kubernetes
- [ ] docker swarm

### Version

- Version: `master`

## Description
If you create a duplicate user and the existing user has a capital letter in the email address the server returns:
Status 500 Internal Server Error

Internal Server Error
The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.

instead of returning: Email is already used

This only happens if the existing user has a capital letter in their email address

The same thing also happens if you try to create a duplicate alias where the existing alias has a capital letter.
## Replication Steps
Create a user with a capital letter in the email address
try to create a duplicate of that user

This also happens with aliases
Create an alias with a capital letter in the email address
try to create a duplicate of that alias

## Observed behaviour
The server returns:
Status 500 Internal Server Error

Internal Server Error
The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.

instead of: Email is already used

## Expected behaviour
The server returns: Email is already used

### LINKED ISSUE #2718: Possible to create alias with same email as user or user with same email as alias if existing user or alias contains a capital letter.
### Environment

- [x] docker compose
- [ ] kubernetes
- [ ] docker swarm

### Version

- Version: `master`

## Description

It is possible to create a alias with a email that is already used by a user or create a user with a email that is already used by a alias if the existing alias or users email contain capital letters. Related to #2695

## Replication Steps

Create a alias with a email address that contains a capital letter.
Create a user with the same email address.

Create a user with a email address that contains a capital letter.
Create a alias with the same email address.

## Observed behaviour

You can create a alias or user with a email address that is already in use. Sending a email to the email address delivers the email to the user.

## Expected behaviour

Trying to create a user or alias with a email that already exists should return the error message: Email is already used

maximilize:
## Problem (closes #2695, closes #2718)

Creating a user or alias whose address differs from an existing one only by letter case (e.g. an existing `AbC@example.com` and a new `abc@example.com`) is not detected as a duplicate. Depending on the path it either silently creates a colliding entry (#2718) or crashes with an HTTP 500 (#2695).

## Root cause

Email local parts are case-insensitive, and the email primary key is stored lowercased (`IdnaEmail` / the `email` hybrid setter). But the `localpart` column is written **directly** by the UI create views via `form.populate_obj(...)` (`ui/views/users.py`, `ui/views/aliases.py`), and `Email._update_localpart` did not lowercase it — so a localpart created with capitals is stored mixed-case while its email PK is lowercased.

`Domain.has_email()` lowercases only its *input* and compares it against the stored, mixed-case `localpart`:

```python
localpart = localpart.lower()
for email in chain(self.users, self.aliases):
    if email.localpart == localpart:   # stored 'AbC' != 'abc'
        return True
```

So the duplicate gate misses and creation proceeds — colliding with the lowercased PK and raising `IntegrityError` → 500 (#2695), or creating a shadowing duplicate (#2718).

## Fix

In `core/admin/mailu/models.py`:
1. Lowercase the localpart at set time (`_update_localpart`, via `retval=True`) so it stays in sync with the lowercased email PK — a "localpart is always lowercase" invariant matching the email setter.
2. Compare case-insensitively in `has_email()` (`email.localpart.lower() == localpart`) so already-stored legacy mixed-case rows are matched too.

## How tested

New headless regression tests `tests/anonmail/test_2695_case_duplicate.py` and `tests/anonmail/test_2718_case_dupe.py` (existing pytest fixtures, in-memory SQLite, no network), covering the user and alias paths and the actual PK collision. Red→green confirmed (fails on master, passes with the fix); full `tests/anonmail` suite green (34 passed), no regressions. Towncrier fragments added (`2695.bugfix`, `2718.bugfix`).

### DIFF
--- a/core/admin/mailu/models.py
+++ b/core/admin/mailu/models.py
@@ -334,7 +334,7 @@ class Domain(Base):
         """ checks if localpart is configured for domain """
         localpart = localpart.lower()
         for email in chain(self.users, self.aliases):
-            if email.localpart == localpart:
+            if email.localpart.lower() == localpart:
                 return True
         return False
 
@@ -469,8 +469,12 @@ class Email(object):
 
     @staticmethod
     def _update_localpart(target, value, *_):
+        # email local parts are case-insensitive: keep them lowercased so the
+        # localpart stays in sync with the (lowercased) email primary key (#2695)
+        value = value.lower() if value else value
         if target.domain_name:
             target._email = f'{value}@{target.domain_name}'
+        return value
 
     @staticmethod
     def _update_domain_name(target, value, *_):
@@ -480,7 +484,7 @@ class Email(object):
     @classmethod
     def __declare_last__(cls):
         # gets called after mappings are completed
-        sqlalchemy.event.listen(cls.localpart, 'set', cls._update_localpart, propagate=True)
+        sqlalchemy.event.listen(cls.localpart, 'set', cls._update_localpart, propagate=True, retval=True)
         sqlalchemy.event.listen(cls.domain_name, 'set', cls._update_domain_name, propagate=True)
 
     def sendmail(self, subject, body):
--- a/core/admin/migrations/versions/9a5866105f5a_.py
+++ b/core/admin/migrations/versions/9a5866105f5a_.py
@@ -1,0 +1,173 @@
+""" Lowercase existing mixed-case localparts and e-mail addresses
+
+Revision ID: 9a5866105f5a
+Revises: fdff7f84d363
+Create Date: 2026-07-04 12:00:00.000000
+
+#2695 / #2718 let users and aliases be stored with a mixed-case localpart, so
+rows such as ``Foo@example.com`` and ``foo@example.com`` could coexist. The code
+change in this PR lowercases localparts on write; this migration lowercases the
+rows that already exist.
+
+It mirrors the 2018 migration ``5aeb5811408e`` ("Convert all domains and emails
+to lowercase") and extends it to the tables added since — ``alias.owner_email``
+and ``domain_access`` (Anonymous Email Service).
+
+Because the bug allowed case-only *duplicates*, a blind ``lower()`` would collide
+on the ``user`` / ``alias`` primary key and abort the upgrade half way. We
+therefore run a read-only pre-check first and abort — listing the offending
+addresses — so an administrator can rename or remove one of each pair before
+retrying. Auto-merging two distinct accounts is never safe, so that is
+deliberately left to a human.
+"""
+
+# revision identifiers, used by Alembic.
+revision = '9a5866105f5a'
+down_revision = 'fdff7f84d363'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# Constraint naming convention, identical to mailu.models.Base.metadata, so that
+# batch_alter_table can drop and recreate the foreign keys by name on SQLite and
+# PostgreSQL (same approach as migration 546b04c886f0).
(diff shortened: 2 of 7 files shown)


nextgens:
Why not making a migration to lowercase existing mixed case entries in the DB?

maximilize:
Good call — there's even precedent: `5aeb5811408e` ("Convert all domains and emails to lowercase", 2018) already lowercases localpart/email across user/alias/token/fetch/manager by dropping the FKs, updating, and restoring them.

The one wrinkle versus 2018: this bug let case-only *duplicates* coexist (#2718), so a blind `lower()` would hit a PK collision on exactly those rows and abort the upgrade. So I'd model the migration on `5aeb5811408e` (plus `DomainAccess`, which postdates it) but add a pre-pass that detects `lower(email)` collisions and aborts with the offending addresses listed — auto-merging two real accounts/aliases isn't safe, so that's left to the admin.

I'll add it to this PR on that basis.

nextgens:
@maximilize please do send the migration so that we can merge this

maximilize:
Done — pushed migration `9a5866105f5a` on top of the PR.

It mirrors `5aeb5811408e`, extended to the tables added since (`alias.owner_email` and `domain_access`): drop the FKs referencing `domain.name`/`user.email`, lowercase every stored name/address, then restore the FKs — using the `naming_convention` batch pattern from `546b04c886f0` so the constraint names resolve on SQLite and PostgreSQL.

As we discussed on #2718, a blind `lower()` collides on the `user`/`alias` primary key wherever case-only duplicates exist, so it runs a read-only pre-check first and aborts listing the offending pairs; merging two real accounts is left to the admin.

Tested headless on SQLite (`tests/anonmail/test_2695_lowercase_migration.py`): full lowercasing across every table with `PRAGMA foreign_key_check` clean, plus the collision-abort path. I haven't run it against a live PostgreSQL/MySQL instance.

nextgens:
bors try

nextgens:
Hmm, the tests don't pass anymore, see https://github.com/Mailu/Mailu/actions/runs/28703132262/job/85124466880
=========================== short test summary info ============================
FAILED tests/anonmail/test_2695_lowercase_migration.py::test_migration_lowercases_every_table
FAILED tests/anonmail/test_2695_lowercase_migration.py::test_migration_aborts_on_case_only_duplicate
================= 2 failed, 45 passed, 112 warnings in 50.04s ==================

maximilize:
Fixed — the failure was in the test, not the migration. It located the migration file via `parents[2]/core/admin/migrations/…`, which only resolves in the repo tree; in the built admin image the Dockerfile flattens `core/admin/` into `/app` (`COPY migrations/ ./migrations/`), so the file is at `/app/migrations/versions/` and the test hit `FileNotFoundError`. Now resolved relative to the `mailu` package, which sits next to `migrations/` in both layouts (`c6c52166`).

Verified against CI parity this time: ran the two tests inside `ghcr.io/mailu/admin:master` with the migration + tests mounted at the `/app` layout — both pass. Ready for another `bors try`.

nextgens:
bors try

nextgens (approved):
Thanks!

LGTM

ghostwheel42 (approved):

PR #4052, opened 2026-06-28
config-import: clear configuration through the ORM to avoid FK violations (closes #2715)

### LINKED ISSUE #2715: "flask mailu config-import" results in violation of foreign key constraint
### Environment

- [x] docker compose
- [ ] kubernetes
- [ ] docker swarm

### Version

- Version: `1.9`

## Description

When trying to import the same configuration I exported just before I get an error about violating a foreign key constraint on the "domain" table.

## Replication Steps

```
docker compose exec -T admin flask mailu config-export > tmp
docker compose exec -T admin flask mailu config-import -v < tmp
```

## Observed behaviour

```
Error: [IntegrityError] (psycopg2.errors.ForeignKeyViolation) update or delete on table "domain" violates foreign key constraint "user_domain_name_fkey" on table "user" DETAIL: Key (name)=(goeddel.net) is still referenced from table "user". [SQL: DELETE FROM domain] (Background on this error at: https://sqlalche.me/e/14/gkpj)
```

## Expected behaviour

I expect the import command to not execute any `DELETE FROM` query when I import stuff that's already in place. It better should create some kind of difference to the current configuration before applying it.

## Logs
There are no logs.

maximilize:
## Problem (#2715)

`flask mailu config-import` fails with a foreign-key constraint violation on FK-enforcing databases (e.g. PostgreSQL) while clearing the existing configuration before loading the new one.

## Root cause

`MailuConfig.clear()` (`core/admin/mailu/models.py`) deletes each table with a bulk `query(model).delete()`. Bulk deletes bypass ORM cascades and the many-to-many `manager` secondary-row cleanup, and are not ordered by FK dependency. So `DELETE FROM domain` runs while `manager` / alias / user rows still reference it → `FOREIGN KEY constraint failed`. SQLite doesn't enforce FKs by default, which is why this surfaces on PostgreSQL.

## Fix

Clear via the ORM: load each model's rows and `session.delete()` them, then `flush()`. The unit of work emits deletes in dependency order and cascades to dependent rows. config-import is not a hot path, so the extra load is negligible.

## How tested

New `tests/anonmail/test_2715_config_import_fk.py`. **Note on the repro:** the isolated headless harness can't spin up a real PostgreSQL, so the test enables SQLite FK enforcement (`PRAGMA foreign_keys=ON`) to mimic it — this reproduces the exact `FOREIGN KEY constraint failed` reported, failing on master and passing with the fix. A quick sanity check against a real PostgreSQL instance would be welcome. Full `tests/anonmail` suite green, no regressions. Towncrier fragment `2715.bugfix` added.

### DIFF
--- a/core/admin/mailu/models.py
+++ b/core/admin/mailu/models.py
@@ -1128,9 +1128,19 @@ class MailuConfig:
 
     def clear(self, models=None):
         """ remove complete configuration """
+        # Delete via the ORM (load + session.delete) instead of a bulk
+        # query(model).delete(). Bulk deletes neither honour FK ordering nor
+        # trigger ORM cascades / many-to-many (manager) secondary cleanup, so on
+        # a FK-enforcing backend (e.g. PostgreSQL) "DELETE FROM domain" fails
+        # while users/aliases/managers still reference it (#2715). Loading the
+        # rows lets the unit-of-work emit deletes in dependency order and
+        # cascade to dependent rows. The explicit flush keeps the tables
+        # physically empty before the new configuration is loaded.
         for model in self._models:
             if models is None or model in models:
-                db.session.query(model).delete()
+                for item in model.query.all():
+                    db.session.delete(item)
+        db.session.flush()
 
     def check(self):
         """ check for duplicate domain names """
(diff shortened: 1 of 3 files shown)


nextgens (approved):
LGTM

ghostwheel42 (approved):

Issue #3853, opened 2025-06-02
DKIM signing for outbound mail without domain registration in Mailu


blainvillem:
### **Context**
I'm using Mailu as an outbound SMTP relay only (SmartHost). In this setup:

> Mailboxes are not hosted on Mailu,
> Domains like a.com send mail through Mailu, but their actual mail services (IMAP/POP/etc.) are hosted elsewhere,
> Mailu is used only for outbound message delivery and DKIM signing.

### **Problem**
Currently, Mailu requires that a domain be registered in the admin interface in order to:

> Accept mail from that domain,
> Sign outgoing messages with DKIM.

This leads to serious delivery issues in my use case:
> When a user a@a.com sends a message to b@a.com, Mailu tries to deliver it locally, because it believes it "owns" the domain.
> But since a.com is not actually hosted in Mailu (mailboxes are elsewhere), delivery fails with "user not found".
> I do not want Mailu to accept or route local mail for these domains — it should simply act as a relay.

### **What I'm trying to achieve**
> I want Mailu to sign all outgoing messages with DKIM, for certain domains (like a.com),
> Without registering the domain in Mailu’s admin UI,
> And without having to define any mailboxes, users, or local routing for those domains.

### **Current limitation**
> Mailu’s DKIM feature appears tightly coupled to domain registration and local handling.
> There’s no clear way to sign outbound mail for a domain without making Mailu think it hosts that domain.
> This prevents Mailu from being used as a DKIM-signing smart relay.

### **Desired behavior**
I’d like to be able to:

> Manually configure DKIM keys for specific domains (via environment variables or mounted files),
> Instruct Mailu to sign outbound messages for those domains, even if the domain is not registered in the admin UI,
> Prevent Mailu from attempting local delivery of messages for those domains.

Issue #1332, opened 2020-01-29
Split mail delivery GSuite <--> Mailu


deze333:
Has anyone had success with **split email delivery** placing Mailu "behind" GSuite (or whatever) mail server?

Split delivery config well described in [this doc from iRedMail](https://gist.github.com/iredmail/fda90a4fd0fc4453f6a2c6c18638e047):

>  If you're hosting same mail domain on 2 servers, e.g. Google G-suite and iRedMail server, you can follow this tutorial to update iRedMail server settings to relay non-existing addresses (under the hosted mail domain) to G-suite. For example, mailbox `xyz@domain.com` exists on Gsuite but NOT on iRedMail, with this tutotial, you're able to relay email sent to `xyz@` to Gsuite.

My simplistic implementation looks like this:

**`main.cf`**

```ini
transport_maps = pcre:/overrides/transport_maps
smtpd_sender_login_maps = pcre:/overrides/sender_logins, ${podop}senderlogin

# Disable sender address matching to allow all, not just known addresses
smtpd_client_restrictions =
  ...
  # check_sender_access ${podop}senderaccess,
  ...
```

**`/overrides/transport_maps`**

```ini
# Relay any user that belongs to our domain
/^.+domain.com$/    smtp:[smtp-relay.gmail.com]:25  
```

**`/overrides/sender_logins`**

```ini
# Allow all users from our domain
 /^(.*)@domain.com$/   ${1}@domain.com
```
This config obviously creates a "too many hops" problem when sending from GSuite address to our mail server address, as our mail server simply relays that email back to GSuite.

Is there a way to tackle this at the config level, or this can only be solved via adding a podop service?

kaiyou:
This is in fact a duplicate of #350. We already have relay enabled for domains, simply using the relay feature.

At the time of the referenced issue it was difficult to implement because everything relied in database queries. Now that we have podop and API calls implemented, it should be doable to add a "fallback MX" field to domains, and use dynamic transport maps for the rest.

The issue with your config is that all email is forwarded back to gmail, not that of non existing local addresses. Implementing this behavior in Postfix directly might prove tricky.

kaiyou:
TL;DR this is tricky to do with postfix directly but should be okay using podop, any contribution to implement it is welcome.

deze333:
Thank you for the explanation! Some more clarification please, if you don't mind.

I've had a look at podop and it appears to be a mechanism for a basic "key-to-value" retrieval. 

[...]

PR #4055, opened 2026-07-04
Add Domain.outgoing_only for send-only domains

nextsux:
## What type of PR?

Feature

## What does this PR do?

Adds a per-domain "Outgoing only" flag. When set, Postfix does not resolve
the domain via `postfix_mailbox_domain`, so it is not included in
`mailbox_domains`. This lets Mailu act as an authenticated outbound relay
(with DKIM signing) for a domain whose incoming mail is handled by another
provider — Google Workspace, Microsoft 365, or another external MX — without
other Mailu-hosted domains bouncing mail addressed to it locally with a
spurious `User doesn't exist`.

Exposed in the admin UI (Domain create/edit form + list), the REST API v1
(`POST`/`PATCH /api/v1/domain`), and the YAML config import/export schema.
Includes docs (`docs/outgoing-only.rst`), an idempotent Alembic migration,
and pytest coverage for the internal Postfix endpoint (runs in the existing
admin unit-tests CI job).

### Related issue(s)
- Related to #3853 (partial overlap — this PR does not attempt env-var-based
  DKIM without domain registration; scope is just the local-delivery escape
  hatch)

## Prerequisites
- [x] Documentation updated (`docs/outgoing-only.rst`, `docs/index.rst`, `docs/cli.rst`)
- [x] Changelog entry added — currently `+outgoing_only.feature`; happy to rename to `<PR-number>.feature` on request

### DIFF
--- a/core/admin/mailu/api/v1/domain.py
+++ b/core/admin/mailu/api/v1/domain.py
@@ -16,6 +16,7 @@ domain_fields = api.model('Domain', {
     'max_aliases': fields.Integer(description='maximum number of aliases', min=-1, default=-1),
     'max_quota_bytes': fields.Integer(description='maximum quota for mailbox', min=0),
     'signup_enabled': fields.Boolean(description='allow signup'),
+    'outgoing_only': fields.Boolean(description='domain is used only for outgoing mail; skip local delivery'),
     'alternatives': fields.List(fields.String(attribute='name', description='FQDN'), example='["example.com"]'),
 })
 
@@ -25,6 +26,7 @@ domain_fields_update = api.model('DomainUpdate', {
     'max_aliases': fields.Integer(description='maximum number of aliases', min=-1, default=-1),
     'max_quota_bytes': fields.Integer(description='maximum quota for mailbox', min=0),
     'signup_enabled': fields.Boolean(description='allow signup'),
+    'outgoing_only': fields.Boolean(description='domain is used only for outgoing mail; skip local delivery'),
     'alternatives': fields.List(fields.String(attribute='name', description='FQDN'), example='["example.com"]'),
 })
 
@@ -36,6 +38,7 @@ domain_fields_get = api.model('DomainGet', {
     'max_aliases': fields.Integer(description='maximum number of aliases', min=-1, default=-1),
     'max_quota_bytes': fields.Integer(description='maximum quota for mailbox', min=0),
     'signup_enabled': fields.Boolean(description='allow signup'),
+    'outgoing_only': fields.Boolean(description='domain is used only for outgoing mail; skip local delivery'),
     'alternatives': fields.List(fields.String(attribute='name', description='FQDN'), example='["example.com"]'),
     'dns_autoconfig': fields.List(fields.String(description='DNS client auto-configuration entry')),
     'dns_mx': fields.String(Description='MX record for domain'),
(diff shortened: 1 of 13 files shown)


nextgens:
I am sorry but no. This is not a use-case we want to encourage.

#3854 already describes a workaround

nextsux:
I don't think RELAYNETS/sign_networks actually solves this, for the reason I gave: it means trusting an IP range without authentication. That's fine for a small static subnet, but my sender runs in AWS - the equivalent trusted range would be effectively all of AWS, which I'm not willing to whitelist, on Mailu or anywhere else.

To be clear on what this PR changes: authentication is untouched. Sending still requires a real Mailu mailbox and normal SMTP auth. The flag only stops Postfix from treating the domain as locally hosted when it isn't, so mail to that domain doesn't bounce with a false "user not found." It's opt-in per domain and off by default.

Could you clarify what the concern is - abuse/security risk, or more that this kind of split setup (mailbox elsewhere, Mailu as authenticated signing relay) isn't something you want to support at all regardless of implementation? If it's the latter, I'd still like to understand why, since the auth model doesn't change.

nextgens:
Our missing statement. We aim to provide a full-featured mail server, not an e-marketer (aka spammer) wet dream.

nextsux:
To be fair, RELAYNETS is the one with the real openness problem here, not my PR. My flag doesn't touch authentication - sending still requires valid Mailu credentials. RELAYNETS + sign_networks trusts an entire IP range with no authentication at all, which is a much bigger attack surface than "an authenticated user sends as a domain they're not registered as owning."

This also isn't a one-off ask. #1332 is the same split-delivery pattern (mailboxes on GSuite, Mailu handling the rest), and #3853/#3854 hit the same wall from the DKIM side - and in that thread the RELAYNETS workaround was reported as not actually applying DKIM signing. So this keeps coming up because there isn't a clean, authenticated way to do it.

I'll also say, honestly, I didn't expect a PR to get closed with a one-line "no" and no discussion - no questions about the design, no counter-argument, nothing. I put real effort into this (migration, tests, docs) and would have appreciated at least a chance to address whatever the concern actually is before it got shut down. That's not the kind of process I'd expect from a project like this.

Can you say concretely what risk you see in my PR, given auth is unchanged? If it's more a maintenance/scope call than a security one, I'd rather hear that directly.

(Also do you really think spammers would use mailu for sending spam? They can run postfix on their own)

nextgens:
This is a niche feature (it's been asked for what? maybe 3 times in the project's lifetime - 10y) that has no clear good use-case we actually want to support.

Can you articulate why you wouldn't want to use Google's infrastructure for sending if you already trust them with your data (your MUA will likely send them a copy of your sent emails)?
Why would you want a different authentication domain (your users having two separate sets of credentials with incompatible authentication protocols - google will require xoath2 that we don't support)?

nextsux:
Because for example I need to send a confirmation of contact form or order from a webpage to users that requested that. Because I need to send legal documents from eshop to customer? There are lot's of legitimate usecases that does not involve end user and his mail client. If SPF and DKIM are correct, it's absolutely legit way to send emails. Why not to use solution that's already working?

I agree this feature is not for everyone. But it's clear that it would help more ppl than only me.

nextgens:
All the usecases you have given about are machines sending transactional emails.

This is not what Mailu is meant for. Just like Google, we don't want the headache of dealing with that (nor the users having such features would bring).

Like you have identified, you can deploy your own postfix instance (or Mailu fork) and do what you want on your own infrastructure, this does not mean it is a good reason for your code to be up-streamed.

PR #4056, opened 2026-07-05
api: return proper HTTP status instead of 500 on malformed input

maximilize:
## What type of PR?

bug-fix

## What does this PR do?

Two admin API v1 endpoints returned HTTP 500 (an uncaught exception) on input that should produce a clean 4xx:

- `PATCH /api/v1/domain/<name>` for a non-existent domain built its 404 message from `data`, which is only assigned on the following line, so it raised `UnboundLocalError` → 500. It now uses the `domain` argument in the message and returns 404 as intended.
- `POST` / `PATCH /api/v1/user` parse `reply_startdate` / `reply_enddate` by hand; a value that is not a valid `YYYY-MM-DD` (e.g. `2022-02-30` or `notadate`) raised `ValueError` → 500. The dates are now parsed through a small helper that returns 400 on a malformed value.

Verified with a new headless regression test (`tests/anonmail/test_api_v1_malformed_input.py`): each malformed-input case returned 500 before and now returns 404/400; the full `tests/anonmail` suite passes (51 tests).

### Related issue(s)

## Prerequisites

### DIFF
--- a/core/admin/mailu/api/v1/domain.py
+++ b/core/admin/mailu/api/v1/domain.py
@@ -163,7 +163,7 @@ class Domain(Resource):
             return { 'code': 400, 'message': f'Domain {domain} is not a valid domain'}, 400
         domain_found = models.Domain.query.get(domain)
         if not domain_found:
-            return { 'code': 404, 'message': f'Domain {data["name"]} does not exist'}, 404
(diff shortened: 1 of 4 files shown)


nextgens (approved):
LGTM

ghostwheel42 (approved):

Issue #4053, opened 2026-07-01
Invalid email domain not handled properly and showing a 451


thohai:
Hey! When somebody using my mailu server (version 2024.06 - latest) tried to send an email to emaildomain containing an "_" (yes this is an invalid address).

Mailu responded with:
```
Transcript of session follows.

 Out: 220 srv.my.redacted ESMTP Postfix
 In:  EHLO srv.my.redacted
 Out: 250-srv.my.redacted
 Out: 250-SIZE 50000000
 Out: 250-VRFY
 Out: 250-ETRN
 Out: 250-AUTH PLAIN LOGIN
 Out: 250-ENHANCEDSTATUSCODES
 Out: 250-8BITMIME
 Out: 250-DSN
 Out: 250-SMTPUTF8
 Out: 250 CHUNKING
 In:  MAIL FROM: [<sender@my.redacted>](mailto:sender@my.redacted)
 Out: 250 2.1.0 Ok
 In:  RCPT TO: [<recipient@invalid_example.ex>](mailto:recipient@invalid_example.ex)
 Out: 250 2.1.5 Ok
 In:  DATA
 Out: 354 End data with <CR><LF>.<CR><LF>
 Out: 451 4.3.0 Error: queue file write error
 In:  QUIT
 Out: 221 2.0.0 Bye


For other details, see the local mail logfile
```

I would have expected a different error code or error message when the recipient email domain is invalid.

Either the email is rejected because of the special character that should not be there OR it shold be rejected because the domain is invalid - but it seems it failed earlier because it tries to write a filename with that character.

Is it maybe the case that the underscore is not properly escaped here?

PR #4057, opened 2026-07-05
internal: return a clean status instead of 500 on a malformed username

maximilize:
## What type of PR?

bug-fix

## What does this PR do?

Two internal endpoints returned HTTP 500 (an uncaught exception) on a client-supplied address that is not a valid stored e-mail, where a clean status was intended:

- `/internal/auth/basic` decoded the Basic credentials without guarding against a malformed header — invalid base64, or no `:` separator — raising `binascii.Error` / `ValueError`. The decode is now guarded and falls through to the generic 401. The analogous decode in `nginx.py` is already wrapped this way. (Reachable pre-auth as the `auth_request` for the WebDAV location when `WEBDAV != none`.)
- `/internal/dovecot/passdb`, `/dovecot/quota` and `/dovecot/sieve/data` looked the user up with an unguarded `User.query.get()`, so a username without `@` or with a non-IDNA domain raised a `StatementError`. They now go through a helper that catches it and returns 404, matching the sibling `/dovecot/userdb` endpoint which already does this.

Verified with a new headless regression test (`tests/anonmail/test_internal_malformed_input.py`): each malformed-input case returned 500 before and now returns 401/404; the full `tests/anonmail` suite passes (52 tests).

### Related issue(s)

## Prerequisites

### DIFF
--- a/core/admin/mailu/internal/views/auth.py
+++ b/core/admin/mailu/internal/views/auth.py
@@ -5,6 +5,7 @@ from flask import current_app as app
 import flask
 import flask_login
 import base64
+import binascii
 import sqlalchemy.exc
 import urllib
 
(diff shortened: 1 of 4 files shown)


nextgens:
Is that #4053 or something else?

maximilize:
Something else — this is independent of #4053; I found these by reading the code.

This PR hardens three admin-internal surfaces that raise an uncaught 500 on a malformed *username*: `/internal/auth/basic` (bad base64 / missing `:`) and the `/internal/dovecot/{passdb,quota,sieve}` lookups (a username that isn't a valid stored address — no `@`, or a non-IDNA domain — makes the `User.query.get` pk-bind raise a `StatementError`).

#4053 is a different path: outbound delivery to an invalid *recipient* domain (`invalid_example.ex`) ending in a Postfix `451 queue file write error`. That recipient isn't a local mailbox, so it never reaches the dovecot passdb/userdb lookups this PR touches — the shared trigger is "invalid domain", but the failing code is not the same. So #4053 stays open.

nextgens (approved):
LGTM

ghostwheel42 (approved):

PR #4058, opened 2026-07-05
config: coerce ANONMAIL_MAX_RETRIES to int

maximilize:
## What type of PR?

bug-fix

## What does this PR do?

`ANONMAIL_MAX_RETRIES` is used numerically — `range(...)` in the anonymous-alias creation path (`api/simplelogin.py`) and a comparison in the UI view (`ui/views/aliases.py`) — but, unlike the other numeric config values, it was never `int()`-coerced after being read from the environment. Overriding it via an environment variable therefore left it a string, so anonymous-alias creation raised `TypeError: 'str' object cannot be interpreted as an integer`. It is now coerced alongside the other numeric settings in `ConfigManager.init_app`.

Verified with a new headless regression test (`tests/anonmail/test_anonmail_max_retries_config.py`): with `ANONMAIL_MAX_RETRIES=20` set, the config value was the string `"20"` before and is now the integer `20`; the full `tests/anonmail` suite passes.

### Related issue(s)

## Prerequisites


nextgens (approved):
LGTM

ghostwheel42 (approved):

PR #4059, opened 2026-07-05
schemas: don't re-enable a disabled alias on merge-import

maximilize:
## What type of PR?

bug-fix

## What does this PR do?

`flask mailu config-import --update` (merge mode) is meant to leave any field the imported YAML omits unchanged — auto-generated schema fields never set a `load_default`, so an absent field keeps the value already stored. `AliasSchema.disabled` was the exception: it declared `load_default=False`, so importing an alias entry that omitted `disabled` injected `False` and silently **re-enabled an alias an admin had disabled** (e.g. an abusive anonymous alias), a data-integrity regression on every merge-import.

The fix drops the `load_default`. On create the model column default (`disabled = db.Column(..., nullable=False, default=False)`) still applies, and an explicit `disabled: true` / `disabled: false` in the YAML is still honoured.

Verified with a new headless test (`tests/anonmail/test_config_import_disabled_alias.py`): merge-import omitting `disabled` re-enabled the alias before and now keeps it disabled; an explicit toggle still works; a newly-imported alias still defaults to enabled. The full `tests/anonmail` suite passes (50 tests).

### Related issue(s)

## Prerequisites

### DIFF
--- a/core/admin/mailu/schemas.py
+++ b/core/admin/mailu/schemas.py
@@ -1220,7 +1220,9 @@ class AliasSchema(BaseSchema):
     note = fields.String(allow_none=True)
     hostname = fields.String(allow_none=True)
     owner_email = fields.String(allow_none=True, dump_only=True)
-    disabled = fields.Boolean(load_default=False, dump_default=False)
+    # No load_default: in merge mode (config-import --update) an omitted field must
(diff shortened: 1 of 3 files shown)


nextgens (approved):
LGTM

ghostwheel42 (approved):

PR #4060, opened 2026-07-05
ui: fix two admin form-flow bugs (pw-change redirect, create-user input reset)

maximilize:
## What type of PR?

bug-fix

## What does this PR do?

Fixes two small admin-UI form-flow bugs:

- **Forced password change always redirected to `WEB_ADMIN`.** On login, the intended post-change destination is stored under `session['redirect_to']`, but `pw_change()` read it back as `session['redir_to']` — the keys never matched, so the stored destination was dropped and the user always landed on the admin UI instead of the target they picked (e.g. webmail). Read the same key that is written. (No security impact: both values are trusted server config.)
- **Create User discarded the admin's input on a failed submit.** `form.process()` ran unconditionally after `validate_on_submit()`, re-initialising the form and dropping both the entered values and the validation errors, so an invalid submit showed an empty form with no feedback. It now only (re)applies the dynamic field defaults on the initial GET and keeps the input and errors on a failed submit.

Verified with new headless tests (`tests/anonmail/test_admin_ui_flow.py`): the pw-change redirect went to `WEB_ADMIN` before and now goes to the stored destination; a failed Create-User submit dropped the entered localpart before and now preserves it. The full `tests/anonmail` suite passes (49 tests).

### Related issue(s)

## Prerequisites


nextgens (approved):
LGTM

ghostwheel42 (approved):

PR #4064, opened 2026-07-09
Fix roundcube admin link for multi-segment WEB_WEBMAIL paths

### LINKED ISSUE #3164: Link to admin UI in roundcube broken for `WEB_WEBMAIL` path with more than two subdirectories
## Environment & Version
- Version: `master`

## Description
When `WEB_WEBMAIL` is set to a path with more than two subdirectories the link back to the mailu admin UI in roundcube is broken.
This is due to the link not point to "`$WEB_ADMIN`" but to "../..`$WEB_ADMIN`", see https://github.com/Mailu/Mailu/blob/9d0325fa1501c9e9febbbc7eaa5093cc6f8682c7/webmails/roundcube/config/config.inc.php#L60

With less than two subdirectories this does not matter since the superfluous `../` are simply ignored.

## Replication Steps
1. Set `WEB_WEBMAIL=/webmail/sub/directory`
2. Set `WEB_ADMIN=/admin`
3. Klick on mailu link in roundcube taskbar

### Observed behaviour
The browser navigates to `/webmail/admin`.

### Expected behaviour
The browser should navigate to `/admin`.

## Further background

The obvious solution would be to simply change  https://github.com/Mailu/Mailu/blob/9d0325fa1501c9e9febbbc7eaa5093cc6f8682c7/webmails/roundcube/config/config.inc.php#L60
to
~~~php
 $config['support_url'] = '{{ WEB_ADMIN }}'; 
~~~

Unfortunately this doesn't work because the URL would the be "corrected" by roundcube to be relative to the skin directory
* in [rcmail_output_html::file_callback()](https://github.com/roundcube/roundcubemail/blob/b3e5efa870b41a4fb4bc6d0e6fccca5e3fe0d2fb/program/include/rcmail_output_html.php#L1005-L1009)
* called by [rcmail_output_html::fix_paths()](https://github.com/roundcube/roundcubemail/blob/b3e5efa870b41a4fb4bc6d0e6fccca5e3fe0d2fb/program/include/rcmail_output_html.php#L992)
* called by [rcmail_output_html::_write()](https://github.com/roundcube/roundcubemail/blob/b3e5efa870b41a4fb4bc6d0e6fccca5e3fe0d2fb/program/include/rcmail_output_html.php#L2107)
* called by [rcmail_output_html::write()](https://github.com/roundcube/roundcubemail/blob/b3e5efa870b41a4fb4bc6d0e6fccca5e3fe0d2fb/program/include/rcmail_output_html.php#L695)

TowyTowy:
## What type of PR?
bug-fix

## What does this PR do?
The Mailu admin button in roundcube used a relative support_url with a fixed two-level climb (`../..{{ WEB_ADMIN }}`). roundcube's `fix_paths()` rewrites absolute hrefs to be skin-relative, so the link must stay relative and is resolved by the browser against the webmail page URL (`WEB_WEBMAIL/`). When WEB_WEBMAIL has three or more path segments (e.g. `/webmail/sub/directory`), two `../` are not enough to reach the web root, so the button pointed to `/webmail/admin` instead of `/admin`.

This emits one `../` per WEB_WEBMAIL path segment so the link reaches the web root at any depth, then appends WEB_ADMIN. It is behaviour-preserving for the default single-segment `/webmail`.

### Related issue(s)
- Fixes #3164

## Prerequisites
- [x] Unless it's docs or a minor change: add changelog entry file.

Validated by rendering the template across WEB_WEBMAIL depths and simulating the browser's relative-URL resolution: the old value reproduces the reported `/webmail/admin` bug, the new one resolves to `/admin` at every depth and is unchanged for the default `/webmail`.

### DIFF
--- a/towncrier/newsfragments/3164.bugfix
+++ b/towncrier/newsfragments/3164.bugfix
@@ -1,0 +1,1 @@
+Fix the link back to the Mailu admin UI in roundcube being broken when WEB_WEBMAIL is set to a path with more than two segments
--- a/webmails/roundcube/config/config.inc.php
+++ b/webmails/roundcube/config/config.inc.php
@@ -57,7 +57,9 @@ $config['managesieve_mbox_encoding'] = 'UTF8';
 // roundcube customization
 $config['product_name'] = 'Mailu Webmail';
 {%- if ADMIN and WEB_ADMIN %}
-$config['support_url'] = '../..{{ WEB_ADMIN }}';
+{# This link must stay relative: roundcube's fix_paths() rewrites absolute paths to be skin-relative. -#}
+{# One "../" per WEB_WEBMAIL path segment climbs back to the web root regardless of its depth (see #3164). -#}
+$config['support_url'] = '{{ "../" * (WEB_WEBMAIL.strip("/").split("/")|length) }}{{ WEB_ADMIN.strip("/") }}';
 {%- endif %}
 $config['plugins'] = array({{ PLUGINS }});
 


nextgens:
This is not something we want to support; You should run mailu on its own FQDN

Issue #2575, opened 2022-12-06
New feature - Support external Open ID Connect provider


fastlorenzo:
### Environment

- [ x] docker-compose
- [ x] kubernetes
- [ x] docker swarm

### Version

- Version: `master`

## Description

It would be great to be able to use external authentication systems using Open ID Connect (Keycloak in my case).
I stumbled upon a [fork that does that](https://github.com/heviat/Mailu-OpenID/). I've tried to reach out to the author to identify if they wanted to make a PR to the main repo, but as I still got no answer decided to go ahead and propose the changes here directly.
All credits goes to them for the implementation.

I've tested this for a couple of weeks now and everything looks stable. I've isolated the changes required in PR #2574.

## Replication Steps

N/A

## Observed behaviour

N/A

## Expected behaviour

N/A

## Logs

N/A

smolinari:
Commenting to be able to follow this PR, as I don't trust the :eyes: emoji. :grin: 

Scott

nextgens:
As discussed on the PR and #mailu-dev, the way forward for this is to document how to do it with a 3rd party tool through the #1972 functionality.

Whether https://github.com/oauth2-proxy/oauth2-proxy or something else is what ends up being documented is still open for discussion; so is extending what is already in Mailu so that attributes can be passed through by the IdP.

batistein:
highly appreciated feature! the use via oauth2 proxy sounds extremely cumbersome to me. We use this tool for kubernetes and I dislike it there as well. I would definitely favor having native support in a application. For many organizations this would be a deal breaker. 
I also tested the fork on the weekend and so far no issues, please see also: https://github.com/heviat/Mailu-OpenID/issues/1#issuecomment-1426220099

Both forks are currently working!

nextgens:
Comments like this make me wonder why I bother to contribute to OSS.

You have forks that do provide the features you want, use them.

Have you reported upstream the reasons why you dislike oauth2-proxy or are you just whining here?

"For many organizations", who are you talking about here? What deal are you talking about? Them using Mailu? Who cares and why should anyone?

migs35323:
nice! looking forward for this feature!

konradmoesch:
Seeing that this feature request has more than 40 upvotes, are there any plans on integrating native OICD support into mailu? 
I really like mailu, but this is the biggest missing feature for me. I'd really like to get native support (not via some external proxy) for OICD in mailu
[...]

Issue #3708, opened 2025-01-05
[RFC] SCIM support for auto-provisioning


nextgens:
We should look into [System for Cross-domain Identity Management](https://scim.cloud/) support for auto-provisioning.

I've seen many requests for "oAuth2" that are in fact requests for "centrally managing identities elsewhere". Well, guess what? Even if we enable account creation on first use (like we did with ``PROXY_AUTH_CREATE``) it will not handle their lifecycle (when accounts get disabled or renamed - off-boarding).

This has been briefly discussed on #mailu-dev and the current consensus of those that said they would use OIDC is that it's overkill. This issue is open to track interest, add a thumb up if you require it.

#783 #1984 #3675

nextgens:
SCIM

strugee:
Ah, gotcha. Honestly, having implemented a (basic) SCIM endpoint at $dayjob, this surprises me. It's not exactly a trivial protocol, but it isn't _that_ complex. OIDC AFAICT is a lot more complicated, but to be fair, I learned that OIDC supports provisioning _literally_ from this issue. So take that assessment with a large pile of salt.

I do not think it will be complicated to implement or maintain a SCIM API. Honestly I'd even consider doing it as a completely separately maintained bridge server, just to make maintenance and runtime operation simpler. All the server would need to do is translate SCIM API calls into calls to Mailu's existing REST API - you don't even need to store any state because you can have the SCIM bridge just pass through whatever authentication token it receives unchanged.

nextgens:
> I do not think it will be complicated to implement or maintain a SCIM API. 

I agree and this is why I've created this ticket.

IMHO we could just add new endpoints, we don't need a bridge server.

azmeuk:
FTR there are Python libs to handle the SCIM protocol both server and client sides.
https://github.com/python-scim

PR #4068, opened 2026-07-14
Add native OpenID Connect web login

bryce-d-greybeard:
## What type of PR?

Feature / enhancement

## What does this PR do?

Adds native OpenID Connect web login support for Mailu admin, including:

- `/sso/oidc/login` and `/sso/oidc/callback` routes
- configurable provider metadata, scopes, client credentials, redirect URI, and PKCE support
- ID-token validation against provider JWKS with configurable RSA algorithms (`RS256`, `RS384`, `RS512`)
- issuer, audience, `azp`, `exp`, `iat`, optional `nbf`, and nonce checks
- optional UserInfo claim enrichment with matching `sub` enforcement
- configurable claim mapping for email/name
- optional domain-restricted user auto-creation
- login form button, configuration docs, and Authentik integration notes

### Related issue(s)

- #2575

## Prerequisites

## Local validation

- `python -m pytest -vv --tb=long --full-trace tests/oidc_test.py tests/anonmail`  
  Result: `61 passed, 119 warnings`
- `python -m py_compile core/admin/mailu/sso/oidc.py core/admin/mailu/sso/views/base.py core/admin/mailu/configuration.py tests/oidc_test.py`
- `git diff --check`
- `docker buildx bake -f tests/build.hcl admin --progress=plain`  
  Result: built `mailu/admin:local` and `mailu/admin:latest`
- Built image smoke:
  - `python -m py_compile mailu/sso/oidc.py mailu/sso/views/base.py mailu/configuration.py`
  - `import mailu.sso.oidc`
- Docs-only follow-up: documented that Authentik providers should use an RSA signing key so discovery advertises `RS256`, matching Mailu's default `OIDC_JWT_ALGORITHMS=RS256`.


nextgens:
Thank you for your contribution but I am not sure we are ready to accept large patches from a first time contributor.

I've skimmed over what you are proposing; Which version of OpenID Connect is this claiming to implement?

I am not in favour of introducing new code in 2026 using RS256 / RSA; I'd also drop support for HS256 as that is almost always a bad idea. What about EdDSA or ES256 (with deterministic nonces) instead?

bryce-d-greybeard:
Thanks for the review.

This is intended to implement the OpenID Connect Core 1.0 authorization code flow for Mailu's browser/admin login only, with state, nonce, PKCE S256, ID-token signature validation against the provider JWKS, issuer/audience/time-claim checks, and optional UserInfo claim merge. It does not change SMTP/IMAP/POP3/submission/sieve authentication.

I pushed `2d5ed7b5` to address the signing-algorithm concern:

- added `EdDSA` / Ed25519 JWKS verification
- added `ES256` / P-256 JWKS verification, with JOSE raw ECDSA signature handling
- kept `RS256` accepted for provider compatibility, but no longer make it the only/default direction
- kept symmetric `HS*` unsupported/rejected; Mailu validates ID tokens from JWKS only
- updated docs to recommend EdDSA/ES256 where supported, with RS256 as compatibility fallback
- added focused tests for ES256, EdDSA, and HS256 rejection

Local validation before pushing:

- `python -m py_compile core/admin/mailu/sso/oidc.py tests/oidc_test.py`
- direct crypto smoke in `mailu/admin:local`: RS256, ES256, EdDSA verify; HS256 rejects
- `sudo docker buildx bake -f tests/build.hcl admin --progress=plain`

bryce-d-greybeard:
One clarification: this is not a solo/drive-by change. We are a small team of three working through this against a real Mailu deployment: Danny, Linus, and Bryce.

Our current deployment uses Authentik, and the PR as submitted is working there today for browser/admin login. That said, the concerns about signing algorithms are fair. We will make the suggested changes, prefer modern asymmetric algorithms such as EdDSA/ES256 where supported, keep symmetric `HS*` tokens rejected, and update the PR accordingly.

bryce-d-greybeard:
Here is a demo of it working against our real Painted Pony Mailu/AuthentiK test deployment. The wider site/deployment is still being finished, so I’d describe this as production-like testing rather than a polished public production case, but the OIDC browser/admin login flow is working against a real integration rather than a synthetic-only implementation.

We needed this functionality ourselves, so we thought it would be better to contribute it back to Mailu rather than keep it as a local-only patch.

https://github.com/user-attachments/assets/58d9db4e-7ad5-4394-8955-09d6034a7861

Lichtblitz:
I am in the process of researching how to get up and running with mailu and the OIDC flow currently implemented. The documentation seems to be targeting experts in the field. Having native OIDC support in mailu would be very welcome for those of us who are comfortable with OIDC but would like to avoid the complexity of additional moving parts in the form of more third party software in the mix.

timiti29:
Is this PR add possibility to create automatically mailbox when creating user un authentik for exemple ?

is it possible to have multiple mailboxes for one user (for exemple user1@domain1.com and user1@domain2.com) ? or needed to create multiple users in authentik ?

bryce-d-greybeard:
@Lichtblitz Thanks — that is exactly the use case this PR is intended to address. It adds OIDC as a built-in browser/admin login path, so an external authentication proxy is not required. The Authentik guide is one worked example, but your point about the general documentation being expert-oriented is fair; we can make the provider-neutral setup path clearer as part of this work. SMTP, IMAP, POP3, submission, and sieve authentication remain separate and continue to use Mailu-managed credentials or tokens.

bryce-d-greybeard:
@timiti29 Yes, with an important distinction: when `OIDC_CREATE_USER=true`, Mailu can create the mailbox automatically on the user’s first successful OIDC login, using the verified email address returned by Authentik. This is just-in-time provisioning; it does not react immediately when the user is created in Authentik. The Mailu domain must already exist, and `OIDC_ALLOWED_DOMAINS` can restrict which domains may be provisioned.

For provisioning at Authentik user creation—and for later updates, disabling, and offboarding—SCIM is the appropriate mechanism. Mailu tracks that separately in #3708. We have a separate SCIM implementation in progress and deliberately kept it out of this OIDC PR so the login and lifecycle-provisioning changes can be reviewed independently.

One OIDC login currently maps one returned email address to one Mailu user/mailbox. It does not map a single Authentik identity to several independent Mailu mailboxes. If several addresses should deliver to the same inbox, Mailu aliases are the appropriate setup. If they need to be separate mailboxes, they must be separate Mailu users, each with its own email mapping/login identity.

Also, this PR covers browser/admin login only; mail clients still authenticate to SMTP/IMAP using Mailu-managed credentials or application tokens.

PR #4072, opened 2026-07-19
internal: scope allow_spoofing to the sender's own domain (#2686)

### LINKED ISSUE #2686: Domain managers can spoof other domains
### Environment

- [x] docker-compose
- [ ] kubernetes
- [ ] docker swarm

### Version

- Version: `master`

## Description
Domain managers can give themself or another user on their domain "allow-spoofing" which allows them or the user to spoof mail addresses even on domains where the domain manager is not a manager.

## Replication Steps
add a second domain if you don't already have one
create a domain manager on domain1.tld
login to the domain manager you just created
give yourself or another user "allow-spoofing"
if you gave "allow-spoofing" to another user switch to that user
switch to webmail
you can now spoof addresses on domain2.tld

## Observed behaviour
domain managers can successfully give themself or another user "allow-spoofing" and they or the user can spoof addresses on domains where they are not domain managers.

## Expected behaviour
domain managers should not be able to allow themself or other users to spoof email addresses on domains where they are not domain managers.

maximilize:
## What type of PR?

bug-fix (security)

## What does this PR do?

`allow_spoofing` was consumed with no domain scoping in the `/postfix/sender/login/<sender>` lookup: it unioned in *every* user with `allow_spoofing=True` for *any* requested sender, making it an instance-wide "authenticate as any sender on any domain" capability. Since the flag is an ordinary `UserForm` field written by `user_create`/`user_edit` (guarded only by `@access.domain_admin`), a domain manager could enable it for themselves (or a user they create) and then send DKIM/SPF-authenticated mail spoofing domains they do not manage.

This scopes the lookup to the sender's own domain (`domain_name=domain_name`) — the same-domain direction you preferred over gating the field to global admins. `allow_spoofing` now means "may authenticate as other senders within my own domain".

Tested with a new anonmail regression test (`test_postfix_sender_login_spoofing_is_scoped_to_own_domain`): a spoofing user in `attacker.com` is still a valid login for `coworker@attacker.com`, but is no longer returned for `victim@victim.com` — red before the change, green after. The full `tests/anonmail` suite (55) stays green.

### Related issue(s)
- closes #2686

## Prerequisites

### DIFF
--- a/core/admin/mailu/internal/views/postfix.py
+++ b/core/admin/mailu/internal/views/postfix.py
@@ -159,7 +159,10 @@ def postfix_sender_login(sender):
     localpart = localpart[:next((i for i, ch in enumerate(localpart) if ch in flask.current_app.config.get('RECIPIENT_DELIMITER')), None)]
     destinations = set(models.Email.resolve_destination(localpart, domain_name, True) or [])
     destinations.update(wildcard_senders)
-    destinations.update(i[0] for i in models.User.query.filter_by(allow_spoofing=True).with_entities(models.User.email).all())
+    # allow_spoofing lets a user authenticate as other senders, but only within
+    # their own domain — scoping the lookup to the sender's domain prevents a
+    # domain manager from granting themselves cross-domain spoofing (#2686).
+    destinations.update(i[0] for i in models.User.query.filter_by(allow_spoofing=True, domain_name=domain_name).with_entities(models.User.email).all())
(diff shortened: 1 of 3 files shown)


nextgens (approved):
Thanks!

LGTM. We won't be backporting it as it's a behaviour change... But it will be in the next major release

ghostwheel42 (approved):

PR #4079, opened 2026-08-01
api: authenticate /alias/me with the caller's own credentials

### LINKED ISSUE #4061: API: /api/v1/alias/me endpoints return 500 for the API-token caller (flask.g.user never set)
## Environment & Version

- Found by reading the code at `master` (HEAD `8061788`); reproduced headless against the admin app, so it is not deployment-specific.
- Version: `master`

## Description

The anonymous-alias "me" endpoints — `GET /api/v1/alias/me`, `DELETE /api/v1/alias/me/<alias>`, `PATCH /api/v1/alias/me/<alias>` (`core/admin/mailu/api/v1/alias.py`) — return HTTP 500 for the normal API-token caller.

Each handler resolves the caller with:

```python
user_email = flask.g.user.email if hasattr(flask.g, 'user') else flask_login.current_user.email
```

but they are decorated with `@common.api_token_authorization`, which validates the global `API_TOKEN` and **never sets `flask.g.user`** (only `common.user_token_authorization` does). So with a `Bearer` token `flask.g.user` is unset, the fallback hits `flask_login.current_user` (an `AnonymousUserMixin` with no `.email`), and the request raises `AttributeError` → 500.

Their sibling `POST /api/alias/random/new` (`api/simplelogin.py`) — which shares the same `flask.g.user` model — is decorated with `@common.user_token_authorization`, which suggests the `/me` endpoints were meant to use that decorator too and got the admin-token one by copy-paste from the other `alias` endpoints.

## Replication Steps

`GET /api/v1/alias/me` with `Authorization: Bearer <API_TOKEN>` and no session cookie.

Reproduced headless (in-memory SQLite, no mailstack):

```python
def test_alias_me_with_bearer_token(app, client):
    with app.app_context():
        rv = client.get('/api/v1/alias/me',
                        headers={'Authorization': f'Bearer {app.config["API_TOKEN"]}'})
        assert rv.status_code == 500
```

## Observed behaviour

`AttributeError: 'AnonymousUserMixin' object has no attribute 'email'` → HTTP 500.

## Expected behaviour

A defined response for the intended authentication (the caller's own aliases, or a clean 401/403), not a 500.

## Question / proposed fix

The `flask.g.user` usage and the `/random/new` sibling both point to `@common.user_token_authorization` (per-user `Authentication: email:token`) as the intended auth for `/me`. Should these three endpoints use it? Since this changes an endpoint's authentication I'd rather confirm the intended auth first; I'll then open a PR switching the decorator with a headless regression test.

maximilize:
Closes #4061.

As requested in https://github.com/Mailu/Mailu/issues/4061#issuecomment-5107331847 — switches the three `/api/v1/alias/me*` endpoints to per-user authentication, with tests.

## What was wrong

`GET /alias/me`, `DELETE /alias/me/<alias>` and `PATCH /alias/me/<alias>` resolve the caller through `flask.g.user`, which only `common.user_token_authorization` sets. They were decorated with `common.api_token_authorization`, which validates the global `API_TOKEN` and never sets `flask.g.user`. The fallback then reached `flask_login.current_user` — an `AnonymousUserMixin` with no `.email` — so every call raised `AttributeError: 'AnonymousUserMixin' object has no attribute 'email'` → HTTP 500.

## The change

- The three handlers now use `common.user_token_authorization`, matching their sibling `POST /api/alias/random/new`, which shares the same `flask.g.user` model.
- Dropped the `if hasattr(flask.g, 'user') else flask_login.current_user.email` fallback. It is unreachable now: `user_token_authorization` sets `flask.g.user` on **both** paths — the `Authentication: email:token` header and the logged-in-session path (`flask_login.current_user` when already authenticated). Session-cookie callers keep working.
- Declared an `Authentication` security scheme on the v1 API so the Swagger page advertises the header these endpoints actually accept, instead of `Bearer`.

`/alias/domains/available` was left on the admin token — it never touches `flask.g.user` and was not affected.

## One existing test had to change — worth a look

`tests/anonmail/test_postfix.py::test_postfix_sees_generated_alias` disabled its alias by `PATCH /api/v1/alias/me/<alias>` with the **global API token**, commented "the /me endpoints require API token auth", and passed on master. That was an artifact of the test harness, not evidence the Bearer path worked:

The test wraps all of its requests in a single `with app.app_context():`. `flask.g` is bound to the app context, not the request, so the `flask.g.user` set by the earlier `POST /api/alias/random/new` (which *does* use `user_token_authorization`) **survived into the later PATCH**. The `hasattr` fallback then silently reused that leftover identity. I verified this directly against unmodified master:

```
BEFORE any request: hasattr(g,"user") = False
after user-token request: 201
AFTER request:  hasattr(g,"user") = True -> test1@example.com
```

In production every request gets a fresh app context, so `flask.g.user` is unset — which is exactly the reported 500. The same PATCH, issued as the first call in a fresh context with a Bearer token, raises the `AttributeError` on master.

That step now authenticates as the owning user. I mention it because it means the one place in the tree that appeared to exercise the Bearer path never actually did.

## Tests

New `tests/anonmail/test_4061_alias_me_auth.py` (13 cases), red before the fix and green after:

- all three endpoints work with `Authentication: email:token` (list / patch / delete);
- `/me` returns only the caller's own aliases, and patch/delete on another user's alias give 404 without modifying it;
- the reported symptom is gone: an admin Bearer token and an unauthenticated call now get 401 on all three endpoints rather than a 500;
- an invalid user token gives 403.

Full `tests/anonmail` suite: **73 passed** (60 on master + the 13 new), no other test touched. Run headless in a container against in-memory SQLite, no mailstack.

Newsfragment: `towncrier/newsfragments/4061.bugfix`.

## Note on compatibility

This changes an endpoint's accepted credentials, which is the point of the issue, so it is worth stating plainly: a caller that today sends the global API token to `/alias/me*` gets 401 instead of 500. Nothing in the tree calls these endpoints with a Bearer token (the only such caller was the test above), and no request that previously *succeeded* stops working — on master there was no success path for the admin token here.

### DIFF
--- a/core/admin/mailu/api/v1/__init__.py
+++ b/core/admin/mailu/api/v1/__init__.py
@@ -12,6 +12,12 @@ authorization = {
         'type': 'apiKey',
         'in': 'header',
         'name': 'Authorization'
+    },
+    # per-user credentials ('email:token'), used by the /alias/me endpoints
+    'Authentication': {
+        'type': 'apiKey',
+        'in': 'header',
+        'name': 'Authentication'
     }
 }
 
--- a/core/admin/mailu/api/v1/alias.py
+++ b/core/admin/mailu/api/v1/alias.py
@@ -1,5 +1,4 @@
 import flask
-import flask_login
 from flask_restx import Resource, fields, marshal
 from . import api, response_fields
 from .. import common
@@ -175,11 +174,11 @@ class AliasWithDest(Resource):
 class AnonAliases(Resource):
     @alias.doc('list_anonaliases')
     @alias.marshal_with(alias_fields, as_list=True, skip_none=True)
-    @alias.doc(security='Bearer')
-    @common.api_token_authorization
+    @alias.doc(security='Authentication')
+    @common.user_token_authorization
     def get(self):
         """List anonymous aliases created by the current user"""
-        user_email = flask.g.user.email if hasattr(flask.g, 'user') else flask_login.current_user.email
+        user_email = flask.g.user.email
         return models.Alias.query.filter_by(owner_email=user_email).all()
 
 
@@ -188,13 +187,13 @@ class AnonAlias(Resource):
     @alias.doc('delete_anonalias')
     @alias.response(200, 'Success', response_fields)
     @alias.response(400, 'Input validation exception', response_fields)
-    @alias.doc(security='Bearer')
-    @common.api_token_authorization
+    @alias.doc(security='Authentication')
+    @common.user_token_authorization
     def delete(self, alias):
         """Permanently delete the specified alias"""
         if not validators.email(alias):
             return {'code': 400, 'message': f'Provided alias (email address) {alias} is not a valid email address'}, 400
-        user_email = flask.g.user.email if hasattr(flask.g, 'user') else flask_login.current_user.email
+        user_email = flask.g.user.email
         alias_found = models.Alias.query.filter_by(email=alias, owner_email=user_email).first()
         if not alias_found:
             return {'code': 404, 'message': f'Alias {alias} cannot be found'}, 404
@@ -205,13 +204,14 @@ class AnonAlias(Resource):
     @alias.expect(alias_fields_update)
     @alias.doc('update_anonalias')
     @alias.response(200, 'Success', response_fields)
-    @common.api_token_authorization
+    @alias.doc(security='Authentication')
+    @common.user_token_authorization
     def patch(self, alias):
         """Modify an alias (toggle disabled, update note/hostname/destination)"""
         data = api.payload or {}
         if not validators.email(alias):
             return {'code': 400, 'message': f'Provided alias (email address) {alias} is not a valid email address'}, 400
-        user_email = flask.g.user.email if hasattr(flask.g, 'user') else flask_login.current_user.email
+        user_email = flask.g.user.email
         alias_found = models.Alias.query.filter_by(email=alias, owner_email=user_email).first()
         if not alias_found:
             return {'code': 404, 'message': f'Alias {alias} cannot be found'}, 404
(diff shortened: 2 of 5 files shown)


nextgens:
Thank you for the bugreport and PR! This turns out to be more involved than what I thought...

I've floated an alternative: rename `/api/alias/me` to `/api/alias/me/<email>` so that the target email is known (it's only an issue for the GET since the other operations have it already) - see https://test.mailu.io/api/v1/

We are discussing it on [#mailu-dev](https://app.element.io/?updated=1.12.14#/room/#mailu-dev:matrix.org), feel free to join in :)

I will let you know what ends up being decided in any case.

maximilize:
Two things about the handlers as they stand today, in case they save you a round-trip on that shape:

- `DELETE` and `PATCH /alias/me/<alias>` also read `flask.g.user`, not to identify the target but for the `owner_email=user_email` filter in the query (`core/admin/mailu/api/v1/alias.py:198` and `:215`). If those two stay on `api_token_authorization`, that filter still has no identity source and the `AttributeError` → 500 from #4061 remains. Either the filter goes (an admin token isn't scoped to an owner anyway) or the owner has to come from the path there too.
- The rename has no in-tree callers to migrate: `git grep alias/me` on master hits only `tests/anonmail/test_postfix.py:76`.

I'll rework the PR to whatever shape you land on — just say the word.

nextgens:
@maximilize Join us on [#mailu-dev:matrix.org](https://app.element.io/#/room/#mailu-dev:matrix.org) ; until we have a dev meeting those questions won't be answered and we will not make progress. I am failing at organising one.

maximilize:
I'm not on Matrix, I only follow the project here on GitHub, so I'll miss whatever gets discussed in the room.

No rush from my side though, the PR can sit as long as it needs to. If the meeting lands on `/api/alias/me/<email>`, just say so here and I'll rework it to that shape, including the `owner_email` filter that DELETE and PATCH still read `flask.g.user` for.

And if the meeting stays hard to organise, decide it on your own and I'll build whatever you pick.

PR #4084, opened 2026-08-17
ui: accept every valid localpart as an auto-forward destination

### LINKED ISSUE #4078: Auto-forward validation issue (~)
`~@domain.com` does not pass the validation.

### Steps to reproduce

1. Open User Settings page

    `/{server}/user/usersettings/{email}`
2. In "Auto-forward > Destination" enter `~@domain.com`

    Image
3. Save Settings

    Image

#### Expected behavior

Allow `~@domain.com` - it's allowed everywhere else in Mailu, and it's valid by email naming standards
(my email is `~@01eg.art` and I use it many years - I also have a Google account with it)

maximilize:
## What type of PR?

Bug fix.

## What does this PR do?

Fixes #4078: the *Auto-forward → Destination* field on the user settings page rejects addresses that Mailu itself considers valid, e.g. `~@example.com`.

### Root cause

`MultipleEmailAddressesVerify` (`core/admin/mailu/ui/forms.py:53`) validates the destination against a hand-written pattern whose localpart class is `[_a-z0-9\-\+]`. That is a strict subset of what the rest of the code accepts:

| path | localpart rule | `~@example.com` |
|---|---|---|
| create a user / alias (`UserForm`, `UserSignupForm`, `AliasForm`) | `LOCALPART_REGEX` (`forms.py:13`), RFC 5322 dot-atom | accepted |
| alias destinations in the UI (`DestinationField.validator`, `forms.py:27`) | `^.+@([^.@][^@]+)$` | accepted |
| forward destination via the API (`api/v1/user.py:123`, `:226`) | `validators.email(dest)` | accepted |
| **forward destination in the UI** (`MultipleEmailAddressesVerify`) | `[_a-z0-9\-\+]` | **rejected** |

So an address that Mailu can create, and that can be set as a forward destination through the API, cannot be entered in the web UI. 17 of the 19 special characters permitted by `LOCALPART_REGEX` are affected (`! # $ % & ' * / = ? ^ \` { | } ~`); only `+`, `-` and `_` pass today.

### Fix

Derive both patterns from one shared character class so the two cannot drift apart again:

```python
LOCALPART_CHARS = r'a-zA-Z0-9!#$%&\'*+/=?^_`{|}~-'
LOCALPART_REGEX = rf'^[{LOCALPART_CHARS}]+(?:\.[{LOCALPART_CHARS}]+)*$'
EMAIL_ADDRESS_REGEX = rf'[{LOCALPART_CHARS}]+(?:\.[{LOCALPART_CHARS}]+)*@(?:[a-z0-9-]+\.)*[a-z]+'
```

`LOCALPART_REGEX` keeps its exact previous value (asserted in the tests), so address creation is untouched. The domain part, the case-insensitive match and the comma-separated list handling are unchanged.

### Behaviour change, in full

I diffed the old and the new pattern over 33 inputs; exactly three results change, all of them widening:

- `~@example.com` → now accepted (the reported bug)
- `a!b@example.com` and the other atext specials → now accepted
- `one@a.com,two@b.c` → now accepted: the old pattern required a 2+ character TLD for the *second and later* addresses while allowing a 1 character TLD for the first one. Both now use the same rule.

Nothing that was accepted before is rejected now, and invalid input stays rejected (`notanemail`, `user@`, `@example.com`, `user@@example.com`, `a..b@example.com`, `one@a.com,,two@b.com`, …).

## Tests

`tests/anonmail/test_4078_forward_destination_localpart.py` drives the real `UserSettingsForm` through a request context:

- `~@example.com` is accepted (RED before this change)
- every special character `LOCALPART_REGEX` allows is accepted as a destination localpart, parametrized (17 of these fail before the change)
- the previously accepted forms still pass (dots, `+tag`, subdomains, mixed case, comma lists with and without spaces)
- invalid destinations are still rejected

Run in a throwaway container against this branch, offline: **93 passed** (`tests/anonmail`, 60 before this PR + 33 new). Without the fix: 18 failed.

A towncrier fragment `4078.bugfix` is included.

### DIFF
--- a/core/admin/mailu/ui/forms.py
+++ b/core/admin/mailu/ui/forms.py
@@ -10,7 +10,11 @@ import re
 import string
 import ipaddress
 
-LOCALPART_REGEX = r'^[a-zA-Z0-9!#$%&\'*+/=?^_`{|}~-]+(?:\.[a-zA-Z0-9!#$%&\'*+/=?^_`{|}~-]+)*$'
+# Characters a localpart may contain (RFC 5322 dot-atom), kept in a single place
+# so that the fields creating an address and the ones referencing one agree.
+LOCALPART_CHARS = r'a-zA-Z0-9!#$%&\'*+/=?^_`{|}~-'
+LOCALPART_REGEX = rf'^[{LOCALPART_CHARS}]+(?:\.[{LOCALPART_CHARS}]+)*$'
+EMAIL_ADDRESS_REGEX = rf'[{LOCALPART_CHARS}]+(?:\.[{LOCALPART_CHARS}]+)*@(?:[a-z0-9-]+\.)*[a-z]+'
 
 def checkStrippable(form, field):
     if field.data != field.data.strip(string.whitespace):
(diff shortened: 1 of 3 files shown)


nextgens:
bors try

nextgens:
bors try

nextgens (approved):
LGTM

ghostwheel42 (approved):

ghostwheel42 (approved):

